Issue #15, opened 2019-12-18
What compilers and platforms should be supported


certik:
Let's use this issue to figure out what compilers we have to support (which version and platform).

For my own usage I need at least:

* GFortran
* Intel Fortran
* NAG

In addition my colleagues use: PGI, Cray, IBM.

certik:
In addition I suggest we only use features that work with CMake.

milancurcic:
I agree that supporting and testing on as many compilers as we can is good for stdlib. I'm concerned that requiring support for several compilers may put a brake on the language features we can use for implementation. (example: do all compilers support finalization? I don't know)

Ensuring that non-free compilers build and run stdlib will require some custom CI+CD hacking with dedicated machines with these compilers. 

For start, can we require support for free compilers at first, run it through CI with free tools, and then work on scripts that members with access to Cray, IBM, and similar machines could run there and provide build reports?

certik:
We need to try and see. For the CI, we would simply use gfortran, and we would just test other compilers by hand. Later we can see if there is a way to automate it.

zbeekman:
One of the nice things about CMake is its introspection capabilities. Rather than maintaining a list of supported compilers & compiler versions you can test at configure time if a certain language feature is working and enable it or disable it as part of the build.

Additionally, there are ways to BYOL (bring your own license) for the commercial compilers.

certik:
I tested GFortran, Intel, NAG and PGI, and updated the issue description, either things work, or I created an issue. 

I do not have access to other compilers such as Cray or IBM. Does anyone have access to those?

scivision:
IBM arch is available via Travis-CI--I haven't tried to see if no-cost XL compiler is also there.
On CMake's Gitlab instance, there are a couple people who have contributed code to CMake for Cray Fortran:

* Chuck [Craynor](https://gitlab.kitware.com/cmake/cmake/merge_requests/2894)
* Willem [Deconinck](https://gitlab.kitware.com/cmake/cmake/merge_requests/3504)

scivision:
If stdlib brings in Fortran 2008 improvements, then the current releases of all commonly used Fortran compilers should be doable. Corner cases can be handled by build system detection.

* NAG 7.0 - OK Fortran 2008 as would be used by stdlib
* PGI 19.10 - OK Fortran 2008. PGI 19.4 was a lot buggier/missing syntax
* Gfortran >= 6  OK Fortran 2008
* Intel >= 16 (2016)
[...]

PR #30, opened 2019-12-21
Add initial GitHub CI

certik:
TODO:

* [x] Install gfortran on Linux
* [ ] Install gfortran on macOS
* [ ] Install gfortran on Windows

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,6 +21,7 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: Install GFortran
+      run: export PATH="/Users/runner/miniconda3/bin:$PATH"
       run: conda install -c conda-forge gfortran_osx-64
     - name: cmake
       run: cmake .
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,7 +23,9 @@ jobs:
     - name: Add conda to $PATH
       run: echo ::add-path::$CONDA/condabin
     - name: Install GFortran
-      run: sudo conda install -c conda-forge gfortran_osx-64
+      run: |
+          sudo conda install -c conda-forge gfortran_osx-64
+          which gfortran
     - name: cmake
       run: cmake .
     - name: make
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,6 +33,8 @@ jobs:
     runs-on: windows-latest
     steps:
     - uses: actions/checkout@v1
+    - name: Install GFortran
+      run: conda install -c conda-forge flang
     - name: cmake
       run: cmake -G "Visual Studio 16 2019" .
     - name: make
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,6 +20,8 @@ jobs:
     runs-on: macos-latest
     steps:
     - uses: actions/checkout@v1
+    - name: Install GFortran
+      run: conda install -c conda-forge gfortran_osx-64
     - name: cmake
       run: cmake .
     - name: make
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,17 +1,37 @@
 name: CI
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
-  build:
-
+  Linux:
     runs-on: ubuntu-latest
-
     steps:
     - uses: actions/checkout@v1
-    - name: Run a one-line script
-      run: echo Hello, world!
-    - name: Run a multi-line script
-      run: |
-        echo Add other actions to build,
-        echo test, and deploy your project.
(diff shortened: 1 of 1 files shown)


certik:
Cool, now the CI runs. (It fails, because gfortran is not installed.)

certik:
First, I requested to include gfortran by default at https://github.com/actions/virtual-environments/issues/202. In the meantime, we need to install it by hand.

certik:
Things install on Linux now and tests pass. Now we need to figure out how to install GFortran on macOS and Windows.

certik:
I guess we have to install our own Conda on macOS and Windows.

certik:
It's going to take some time to get this PR ready. In the meantime I contributed just the Linux build in #33, so that we have at least some kind of testing.

zbeekman:
@certik Do we have to use conda? I'd rather just use pip, pipenv, etc.

zbeekman:
I'd recommend not using conda, except maybe on windows... Is flang from conda F18, or "legacy" flang? I have a lot of experience with CI, so I may submit a competing PR, at least to get macOS up and running and some odds and ends for macOS and linux.

certik:
@zbeekman can you submit a PR getting Fortran working on macOS and Windows using Github Actions? That would be super helpful. We don't need to use Conda.

zbeekman:
I'm happy to submit a PR.. I'll start with mac and linux, windows will take some time for me to figure out... namely, which compiler to use and how to install it? (Cygwin? Msys2? Mingw? WSL?)

zbeekman:
(On windows I always use Intel, which is hard to do on CI)

certik:
Linux is already done. If you can do just macOS with gfortran, that would be super helpful. Then we'll go from there.

zbeekman:
Can I get us a newer gfortran on Linux? And a newer CMake? We should be using 3.14 at least for submodule support.

certik:
> Can I get us a newer gfortran on Linux? And a newer CMake? We should be using 3.14 at least for submodule support.

We have to extend our CI to test with newer and older gfortran, so yes (in addition). Regarding CMake, I would suggest to require the oldest possible version that still works. Right now we do not use submodules, so I would not require a newer CMake until we need some feature from it, and then we can discuss if it is worth dropping older version support. That's related to #15. For submodules we would need to investigate the compilers that we want to support if they actually support submodules.

Eitherway, I suggest you submit several PRs that are small, so that we can reach an agreement quickly and merge it (as opposed to a large PR with a complicated CI system). 

The most useful PR right now would be to get just macOS working. The PR can look like #33, but it installs gfortran on macOS instead.

The second most useful would be a PR for Windows. Probably the mingw-w64 from msys2 is the most "native" solution.

zbeekman:
Installing CMake binaries is so trivial I think it's worthwhile using more up-to-date versions. You can download statically linked, re-locatable binaries for all major OSes.

I'll start by opening a PR with multiple versions of GFortran for Linux and mac, then I'll look into windows...

zbeekman:
FYI, your loadtxt CMake doesn't work with out of source builds.

certik:
Yes, we need to fix the out of tree builds and test them on the CI

> Installing CMake binaries is so trivial I think it's worthwhile using more up-to-date versions. You can download statically linked, re-locatable binaries for all major OSes.

On many HPC systems they either do not have an internet connection, or if the do, they do not like you to be installing binaries from the internet. So being able to install using the preinstalled CMake and Fortran versions is a huge plus. In order to increase chances that things just work with default pre-installed versions, we should only require a newer CMake if we absolutely need the latest features. Besides submodules, which other feature do you find useful in newer CMake?

zbeekman:
On all the HPC systems I've ever run on grabbing a statically linked binary and putting it in my home directory has never been an issue. I check the GPG signature and cryptographic checksums and I'm off to the races.

There are a number of features that I like in newer CMake and I can't keep track of which version each was added in. For example, I think it's nice to declare a target and then add sources to it later. There are also bug fixes for `FindMPI` and in transient usage requirements... Using submodules is also pretty compelling, because it separates the interface from the implementation and speeds up builds.

If it's decided not use a suitably new CMake I probably won't be motivated to contribute. Why use CMake at all then? Why not just use Makefiles? We're already handy capped as we wait on compiler vendors to implement or fix new Fortran language features and I don't see a compelling reason to not use a suitably Fortran aware build system. If people are "just grabbing sources and putting them in their build trees" then it doesn't matter whether or not they're allowed to grab CMake binaries.

certik:
@zbeekman thank you for improving the CI build system.

You have a good point that since I think we'll probably decide to maintain manual Makefiles at #36, we might be able to get away with requiring the very lastest CMake, as for people who don't have the latest CMake, they can just use manual Makefiles. Let's move the discussion to #41.

I am closing this PR in favor of #39.

PR #31, opened 2019-12-22
Add initial GitHub CI

certik:
This is an empty CI using GitHub Actions. Once this is in master, we can send PRs modifying it to do what we need.


certik:
I am going to go ahead and merge this, so that I can send a PR with the actual CI.

Issue #11, opened 2019-12-16
Proposal for ascii


ivan-pi:
This module should include functions for character classification and conversion (lower, upper). I have prepared a basic implementation at https://github.com/ivan-pi/fortran-ascii.

The plan is to cover the same functionality as found in the C, C++, and D libraries:
* C : http://www.cplusplus.com/reference/cctype/
* C++: https://en.cppreference.com/w/c/string/byte
* D: https://dlang.org/phobos/std_ascii.html)

@Zbeekman has already opened an issue (see https://github.com/ivan-pi/fortran-ascii/issues/1) on dealing with different character kinds. The problem is that the ascii and iso_10646 character sets need not be supported by the compilers. Even if they are supported their bitwise representation might be different from the default kind. 

I realized while creating these functions, that agreeing upon a style guide #3 and documentation #4 early on would be helpful to improve future pull requests. Some agreement upon unit testing will also be necessary.

cc: @jacobwilliams

ivan-pi:
One of the simple points to discuss (connected to the style guide #3) is naming conventions. For example for the function that checks whether a character is a letter some of the possible names are:
* `is_alpha`
* `isalpha` (C/C++)
* `isAlpha` (D language, but in Fortran it's the same as the previous suggestion)

Personally, I like the underscored version (it matches the general verbosity of Fortran). 

I have also noticed that in C++, they have a separate set of the these function for "wide" strings (e.g. `iswalpha`, see https://en.cppreference.com/w/c/string/wide) which if I understand correctly are Unicode characters.

certik:
@ivan-pi let's move the naming conventions discussion into https://github.com/fortran-lang/stdlib/issues/3#issuecomment-566302138, where I just commented.

zbeekman:
To address character kinds, this is my comment from the ascii proto-type repo:

> If I understand correctly you are suggesting I create several copies of these function to operate on the following character kinds:

Well, not exactly, because, as you noted, they're not guaranteed to exist, and when they do exist, "DEFAULT" is often/usually the same kind as "ASCII", so you can't create overloaded functions with arguments that are "ascii" and "default". (In that case you'd have a duplicate interface.)

That's one nice thing about jin2for: It doesn't assume anything and interrogates the numeric kinds from the compiler to then generate the code. So if only one character kind is supported then your code will only have that one kind. I'll cross post this on the new issue you made.

ivan-pi:
[...]

PR #32, opened 2019-12-22
Add module for validating ASCII characters and upper/lower conversion

ivan-pi:
This PR addresses https://github.com/fortran-lang/stdlib/issues/11 (there are a few open question left there). 

Tested with both the gfortran and Intel Fortran compilers for the 'default' (ascii) character set.

The tests are essentially a port of those at https://github.com/dlang/phobos/blob/master/std/ascii.d (hopefully not a licensing issue?).

### DIFF
--- a/src/tests/ascii/test_ascii.f90
+++ b/src/tests/ascii/test_ascii.f90
@@ -1,7 +1,12 @@
 program test_ascii
 
     use stdlib_experimental_error, only: assert
-    use stdlib_experimental_ascii
+    use stdlib_experimental_ascii, only: lowercase, uppercase, digits, &
+        octal_digits, fullhex_digits, hex_digits, lowerhex_digits, &
+        whitespace, letters, is_alphanum, is_alpha, is_lower, is_upper, &
+        is_digit, is_octal_digit, is_hex_digit, is_white, is_blank, &
+        is_control, is_punctuation, is_graphical, is_printable, is_ascii, &
+        to_lower, to_upper, LF, TAB, NUL, DEL
 
     write(*,*) "Lowercase letters: ", lowercase
     write(*,*) "Uppercase letters: ", uppercase
@@ -34,6 +39,9 @@ program test_ascii
 
     call test_is_white_short
     call test_is_white_long
+
+    call test_is_blank_short
+    call test_is_blank_long
 
     call test_is_control_short
     call test_is_control_long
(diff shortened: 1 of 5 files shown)


certik:
Would you mind rebasing on top of the master to pick up tests?

src/tests/ascii/test_ascii.f90
```diff
@@ -0,0 +1,508 @@
+program test_ascii
+
+    use stdlib_experimental_error, only: assert
+    use stdlib_experimental_ascii
```
certik: Can you please explicitly import the symbols that are being used? I think we should follow that approach of "explicit imports", as in Python.

certik: (The reason I noticed this is that I came here to see what the public API actually is. As it is a bit hard to immediately see from the module itself, because there is not a single `public` line, but rather quite a few symbols are decorated with `public`.)

certik:
So I think this looks really nice. The public API seems nicely done, the naming convention I think is good.

Will the `is_*` functions work with utf8 strings?

The `to_upper` and `to_lower` will only work on the ascii parts of an utf8 string I assume.
Python calls these just `upper` and `lower` methods. In C, they are called `toupper` and `tolower`. Finally in C++, Boost has [to_upper](https://www.boost.org/doc/libs/1_41_0/doc/html/boost/algorithm/to_upper.html).

I think using `to_upper` is readable.

certik:
Python also has an [isupper](https://docs.python.org/3/library/stdtypes.html#str.isupper) method. Should we also implement `is_upper` and `is_lower`?

certik:
General question: should we maintain `stdlib_ascii` and `stdlib_unicode`, or should we simply have `stdlib_string`, and all methods would automatically work with `utf8`?

I think getting the unicode working (as in Python) will be some work, so we can start with `stdlib_ascii`, then in another PR we can see how many of these extend to unicode and create `stdlib_unicode` or merge them. Since this is all in experimental, we can do that.

Things like `to_upper` just need to be extended to support Unicode (utf8), see the Python documentation.

src/tests/ascii/test_ascii.f90
zbeekman: I think we should have `private` declared at the top of all modules, and below one or more lists of `public :: ...`

certik: @zbeekman, I agree, it makes it very easy to see what the public symbols are.

ivan-pi:
I think we should have separate modules for ASCII and Unicode characters. In fact using only the intrinsic Fortran character functions (achar and iachar) it is not possible to find say uppercase Slavic letters č, š, ž... I think it will be necessary to interface with C to achieve Unicode support. A second issue is that some preprocessing will be necessary, as not all Fortran compilers support the extended Unicode character set.

The current ascii module already contains `is_upper` and `is_lower` characters. At the moment all functions in the ascii module are limited to work on single characters. They are meant as support functions for a separate string module.

I will rebase and import explicitly the public functions for the test driver asap.

certik:
@ivan-pi yes, if it is not possible to merge unicode with ascii, then we need two modules.

certik:
The other thing is --- since you used https://github.com/dlang/phobos/blob/434429f273d0359744b6d3ba9db36d3bef1c7593/std/ascii.d as the original, we have to cite their license.

Overall this looks good to me. It would be nice to get some more reviews on this before we merge. @marshallward, @jacobwilliams, @milancurcic do you have any feedback on the API here?

certik:
@milancurcic thanks for the review.

@ivan-pi would you mind updating `whitechar` in `stdlib_experimental_io` to use your new functionality please?

dev-zero:
since I've been pointed here, this project might be interesting: https://github.com/lemire/fastvalidate-utf-8 althought I don't see how that could be implemented in Fortran given missing inline assembly.

certik:
@dev-zero thanks! I made a comment in https://github.com/fortran-lang/stdlib/issues/11#issuecomment-619108015.

Issue #4, opened 2019-12-14
Documentation


milancurcic:
Use this issue how to best document the stdlib. I put a placeholder Markdown file [here](https://github.com/fortran-lang/stdlib/blob/master/doc/API.md).

jvdp1:
Regarding the documention, is it planned to use some tools, like Doxygen or Ford to generate documention from the code?

certik:
I think we should write documentation as comments in the code and generate nice html documentation from it, that way the documentation and code is at one place and easier to maintain and keep in sync. So whatever tool allows us to do that, or if we have to write or improve some tools, I think we should do that.

zbeekman:
FORD is my preference unless someone knows a good way to do that with sphinx.

I think it can be worthwhile to have additional docs. I like hosting markdown based docs on RTD and API docs on gh-pages.

certik:
Down the road I would like to use LFortran to execute "doctests" just like in Python. Here is the issue for it: https://gitlab.com/lfortran/lfortran/issues/73

ivan-pi:
FORD also supports the option to include additional markdown based docs (http://d3denergetic.github.io/FIDASIM/ is my favourite example). 

Personally, I find the FORD generated documentation easier to navigate than doxygen. In case we decide to use FORD we should agree on the symbols to indicate documentation and whether we use preceding documentation strings or not.

I am not sure if FORD can be used together with jin2for. Does one only document the generic interface block in that case?

cc: @cmacmackin

cmacmackin:
@ivan-pi I've never used jin2for so I can't be certain. However, FORD does allow files to be preprocessed prior to it parsing for documentation, so that could work. You'd probably want to give a detailed explanation only in the generic blocks, but still document the arguments in each specific implementation.

marshallward:
MOM6 uses Doxygen-style comments, but these are converted into reST files which are then passed onto Sphinx.

An example Doxygen-marked file: https://github.com/NOAA-GFDL/MOM6/blob/dev/gfdl/src/parameterizations/vertical/MOM_vert_friction.F90

This uses the `!<`and `!!` syntax for marking up variables, subroutines, etc.

The Doxygen XML is then converted to reST by a local tool developed by a former visitor (itself forked from another project): https://github.com/angus-g/sphinxcontrib-autodoc_doxygen

although we are looking to migrate to this tool: https://github.com/vovkos/doxyrest

[...]

Issue #35, opened 2019-12-22
How to implement same procedures for different numeric kinds


milancurcic:
This question comes up in #34 and elsewhere. How to implement specific procedures that work on different kinds (`sp`, `dp`, `qp`, `int8`, `int16`, `int32`, `int64`) as well as characters, where the body of the procedure is the same (can be copy/pasted entirely without breaking it). Let's first just focus on this scenario, and we can consider more complex cases later.

I know of a few approaches:

1. Repeat the code, that is, implement all specific procedures explicitly. That's what I did in functional-fortran, see https://github.com/wavebitscientific/functional-fortran/blob/master/src/lib/mod_functional.f90. Repeating is fine if you do it once and forget about it. The upside is that you can see the specific code and it needs no extra tooling. The downside is combinatorial explosion if you have procedures that are to handle all combinations of types and kinds. Most procedures are rather simple (one or two arguments), and I ended up with > 3K lines of code for 23 generic procedures. Most work was in editing the argument types to specific procedures, and less work was in copy/pasting of the repeatable content. I don't recommend this approach for stdlib.

2. Approach 1 can be somewhat eased by explicitly typing out the interfaces, and using `#include 'procedure_body.inc'`, defined in a separate file. Then your procedure body collapses to one line. This reduces the total amount of code, but not so much the amount of work needed, as most work is in spelling out the interfaces. This approach still doesn't need extra tooling as a C preprocessor comes with all compilers that I'm aware of.

3. Use a custom preprocessor or templating tool. For example, a function that returns a set of an array:

```fortran
pure recursive function set(x) result(res)
  integer, intent(in) :: x(:) !! Input array
  integer, allocatable :: res(:)
  if(size(x) > 1)then
    res = [x(1), set(pack(x(2:), .not. x(2:) == x(1)))]
  else
    res = x
  endif
end function set
```

A template could look like this:

```fortran
pure recursive function set(x) result(res)
  {int*, real*}, intent(in) :: x(:) !! Input array
  {int*, real*}, allocatable :: res(:)
  ... ! body omitted for brevity
end function set
```
or similar, where the custom preprocessor would spit out specific procedures for all integer and real kinds. Some additional or alternative syntax would be needed if you wanted all combinations of type kinds between arguments.

[...]

PR #37, opened 2019-12-22
Single and quadruple precisions for load/savetxt

jvdp1:


src/stdlib_experimental_io.f90
```diff
@@ -1,28 +1,65 @@
...
 
 interface savetxt
     module procedure ssavetxt
     module procedure dsavetxt
+    module procedure qsavetxt
 end interface
 
 contains
 
+!PUBLIC
```
certik: What does this `!PUBLIC` do?

src/stdlib_experimental_io.f90
```diff
@@ -1,28 +1,65 @@
 module stdlib_experimental_io
-use iso_fortran_env, only: sp=>real32, dp=>real64
+use iso_fortran_env, only: sp=>real32, dp=>real64 ,qp=>real128
```
certik: Space after `,` and no space before?

src/stdlib_experimental_io.f90
```diff
@@ -1,28 +1,65 @@
...
     module procedure ssavetxt
     module procedure dsavetxt
+    module procedure qsavetxt
 end interface
 
 contains
 
+!PUBLIC
 subroutine sloadtxt(filename, d)
+! Loads a 2D array from a text file.
```
certik: For documentation, once we move to use templates as in #35, then the documentation will be written only once, and the templating engine will generate all these `sloadtxt` automatically with the same documentation. Until then, it's fine to copy it by hand.

src/stdlib_experimental_io.f90
```diff
@@ -50,34 +87,59 @@ subroutine dloadtxt(filename, d)
 !     11 12 13
 !     ...
 !
-character :: c
-integer :: s, ncol, nrow, ios, i
-logical :: lastwhite
-real(dp) :: r
+integer :: s
+integer ::nrow,ncol,i
```
certik: Space after `::` ?

src/stdlib_experimental_io.f90
```diff
@@ -50,34 +87,59 @@ subroutine dloadtxt(filename, d)
...
-lastwhite = .true.
-do
-   read(s, '(a)', advance='no', iostat=ios) c
-   if (ios /= 0) exit
-   if (lastwhite .and. .not. whitechar(c)) ncol = ncol + 1
-   lastwhite = whitechar(c)
-end do
-
-rewind(s)
+ncol=number_of_columns(s)
```
certik: Space around `=` ?

src/stdlib_experimental_io.f90
```diff
@@ -115,6 +195,70 @@ subroutine dsavetxt(filename, d)
...
+integer :: s, i
+open(newunit=s, file=filename, status="replace")
+do i = 1, size(d, 1)
+    write(s, *) d(i, :)
+end do
+close(s)
+end subroutine
+
+
+!PRIVATE
```
certik: I assume this does a similar thing as the `!PUBLIC` above.

src/stdlib_experimental_io.f90
jvdp1: I am used to sort the different procedures in groups of public and private procedures. This `!PUBLIC` (and `!PRIVATE` below) refer to these two groups. I can remove them.

src/stdlib_experimental_io.f90
jvdp1: Removed

src/stdlib_experimental_io.f90
jvdp1: Corrected

src/stdlib_experimental_io.f90
jvdp1: Corrected

src/stdlib_experimental_io.f90
jvdp1: Corrected

src/stdlib_experimental_io.f90
jvdp1: Now removed.

certik:
This is +1 from me to merge as is. The changes are straightforward, they do not change the public API and they all happen in experimental, so I am going to merge it.

certik:
@jvdp1 thanks for the contribution!

src/stdlib_experimental_io.f90
certik: I see. Once we fix #4, we'll consolidate all such things.

fiolj:
I've mentioned in the issue on quadrature that I'd like to have implementations for integrating complex functions. I think, that seeing as complex numbers are implemented in fortran, we could include them in the library. I've made the modifications to common.fypp, stdlib_experimental_io.fypp, and test_savetxt.f90 to add support for complex numbers. Is that desired? Should I open a pull request?

jvdp1:
@fiolj same comment as in [#62 comment](https://github.com/fortran-lang/stdlib/issues/62#issuecomment-581108893): I think it is a good idea to add `complex` numbers to the library. Because it is an addition to already existing API, I suggest to open a PR.
Could you also add them to `sdtlib_experimental_stat`, please? So the whole library would support `complex` numbers. The community can then further discuss its utility/wish around a PR.

fiolj:
Yes @jvdp1, I'll try this week to put in stats, probably not before tuesday or wednesday. When I was to do it I realized that I should start with `io`
I just put the pull request to `stdlib_experimental_io`, and modified optval, I'll put this also in stats.

Issue #3, opened 2019-12-14
Style Guide


milancurcic:
Use this issue to discuss the code style for the stdlib.

The most widely supported elements of style will eventually be merged into the [Style Guide for contributors](https://github.com/fortran-lang/stdlib/blob/master/STYLE_GUIDE.md).

certik:
@ivan-pi wrote in https://github.com/fortran-lang/stdlib/issues/11#issuecomment-566296346:

For example for the function that checks whether a character is a letter some of the possible names are:
* `is_alpha`
* `isalpha` (C/C++)
* `isAlpha` (D language, but in Fortran it's the same as the previous suggestion)

--------

I have discussed this particular issue with several people many years ago and we eventually converged towards the following compromise:

https://www.fortran90.org/src/best-practices.html#naming-convention

So in the above case, `isAlpha` would not be used, but both `is_alpha` and `isalpha` would be allowed, and in this case `isalpha` would be preferred (only two syllables). If the name was `is_char_alpha` vs `ischaralpha`, then the underscore version would be preferred because it is three syllables.

Let me know what you think.

ivan-pi:
`isalpha` contains three syllables. 

In general I agree with your suggested naming conventions and also the ones in [Fortran Best Practices](https://github.com/Fortran-FOSS-Programmers/Best_Practices). To stay close to the standard, I think it's best if we avoid CamelCase.

Specifically, for the ascii functions there are 11 functions starting with `is`, some are short (`is_alpha`), but some are long (`is_hex_digit`, `is_octal_digit`). I suppose it's easiest to just follow the [C library](https://en.cppreference.com/w/c/string/byte) and get rid of the underscores and shorten the names to 7-8 characters (`isalpha`, `iszdigit` - z is used in hexadecimal boz constants, `isodigit` - o is used in octal boz constants).

certik:
You are right, isalpha is 3 syllables.

I think the rules can be broken where it makes sense and we have a good reason. Otherwise if we have no reason then the rules provide a good consistent default.

milancurcic:
I'm on the same page here. I like all lowercase with underscores universally. For short variable and procedure names, dropping the underscore is OK I think. It's nice to have a universal rule to follow (always use underscores to separate words), but for some short names I think it's OK to make an exception. 

[...]

PR #42, opened 2019-12-22
Style guide (see #3)

zbeekman:


STYLE_GUIDE.md
```diff
@@ -1,3 +1,66 @@
...
+> Let robots handle your project’s pedantry and humans handle your project’s empathy.
+
+-- @mikemcquaid, [Homebrew] project leader<sup>[1]</sup>
+
+[1]: https://mikemcquaid.com/2018/06/05/robot-pedantry-human-empathy/
+[Homebrew]: https://brew.sh
+
+## File naming conventions
+
+- Source files should contain at most one `program` or `module`/`submodule`.
```
milancurcic:
```suggestion
- Source files should contain at most one `program`, `module`, or `submodule`.
```

STYLE_GUIDE.md
```diff
@@ -1,3 +1,66 @@
...
+- The filename should match the program or module name and have the file extension `.f90` or `.F90` if preprocessing is required.
```
milancurcic: As far as I understand, `.F90` for files to be preprocessed is only useful for the compiler to imply preprocessing. We can explicitly enable preprocessing with a flag (`-cpp`). Should we recommend just the `.f90` suffix?

STYLE_GUIDE.md
```diff
@@ -1,3 +1,66 @@
...
+
+## Indentation & whitespace
+
+By setting and following a convention for indentation and whitespace, code reviews and git-diffs can
+focus on the semantics of the proposed changes rather than style and formatting.
+
+* __TABs vs spaces__: __Spaces__: TABs should never be used for indentation
+* __Trailing whitespace__: All trailing whitespace must be removed
+* __End of File (EOF)__: The file must end in a single new-line character
+* __Line length__: __132__: The Fortran 90 (and later) standard mandates lines be no longer than 132 characters
```
milancurcic: This whole section seems overly technical, rather than part of the guide. I suggest we extract the actionable suggestions and list them in plain language, for example:

* The body of every Fortran construct should be indented by 4 spaces
* Line length should be limited to 80 characters, but must not exceed 132.

If trailing whitespaces and EOL and EOF stuff will be handled by CI, no need to mention it in the guide.

STYLE_GUIDE.md
```diff
@@ -1,3 +1,66 @@
...
+* __Indentation__: __Two (2) spaces__ for all control constructs, line continuations, etc.
+* __Line ending character__: __Native__: Git will convert line endings to `\n` (Unix style) on commit, and will checkout files with native line endings
+
+## Variable naming
```
milancurcic:
```suggestion
## Variable and procedure naming
```

STYLE_GUIDE.md
```diff
@@ -1,3 +1,66 @@
...
+* __End of File (EOF)__: The file must end in a single new-line character
+* __Line length__: __132__: The Fortran 90 (and later) standard mandates lines be no longer than 132 characters
+* __Indentation__: __Four (4) spaces__ for all control constructs, line continuations, etc.
+* __Line ending character__: __Native__: Git will convert line endings to `\n` (Unix style) on commit, and will checkout files with native line endings
+
+## Variable naming
+
+Variable names should be descriptive, and not artificially short.
+By default, where it makes sense to do so, variable names shall be made up of one or more full words separated by an underscore.
+For cases where a conventional & appropriate shortening of a word is used then the underscore may be omitted.
```
certik: I suggest to rework this section. 

Here is a suggestion (from https://www.fortran90.org/src/best-practices.html#naming-convention):

---

* Use lowercase for all Fortran constructs (`do`, `subroutine`, `module`, ...).
* Follow short mathematical notation for mathematical variables/functions (`Ylm`, `Gamma`, `gamma`, `Enl`, `Rnl`, ...).
* For other names use all lowercase: try to keep names to one or two syllables; if more are required, use underscores to clarify (`sortpair`, `whitechar`, `meshexp`, `numstrings`, `linspace`, `meshgrid`, `argsort`, `spline`, `spline_interp`, `spline_interpolate`, `stoperr`, `stop_error`, `meshexp_der`).

For example "spline interpolation" can be shortened to `spline_interpolation`, `spline_interpolate`, `spline_interp`, `spline`, but not to `splineint` ("int" could mean integration, integer, etc. --- too much ambiguity, even in the clear context of a computational code). This is in contrast to `get_argument()` where `getarg()` is perfectly clean and clear.

The above are general guidelines. In general, choosing the right name certainly depends on the word being truncated as to whether the first syllable is sufficient. Usually it is but clearly not always. Thus some thought should go into step "try to keep names to 2 syllables or less" since it can really affect the indicativeness and simplicity. Simple consistent naming rules are a real help in this regard – for both collaboration and for one’s own sanity when going back to some old code you haven’t seen in while.

---

Here is the source code so that you can copy & paste it:
```markdown
* Use lowercase for all Fortran constructs (`do`, `subroutine`, `module`, ...).
* Follow short mathematical notation for mathematical variables/functions (`Ylm`, `Gamma`, `gamma`, `Enl`, `Rnl`, ...).
* For other names use all lowercase: try to keep names to one or two syllables; if more are required, use underscores to clarify (`sortpair`, `whitechar`, `meshexp`, `numstrings`, `linspace`, `meshgrid`, `argsort`, `spline`, `spline_interp`, `spline_interpolate`, `stoperr`, `stop_error`, `meshexp_der`).

For example "spline interpolation" can be shortened to `spline_interpolation`, `spline_interpolate`, `spline_interp`, `spline`, but not to `splineint` ("int" could mean integration, integer, etc. --- too much ambiguity, even in the clear context of a computational code). This is in contrast to `get_argument()` where `getarg()` is perfectly clean and clear.

The above are general guidelines. In general, choosing the right name certainly depends on the word being truncated as to whether the first syllable is sufficient. Usually it is but clearly not always. Thus some thought should go into step "try to keep names to 2 syllables or less" since it can really affect the indicativeness and simplicity. Simple consistent naming rules are a real help in this regard – for both collaboration and for one’s own sanity when going back to some old code you haven’t seen in while.
```

STYLE_GUIDE.md
```diff
@@ -1,3 +1,66 @@
...
+
+Examples:
+
+__GOOD__:
+
+``` fortran
+logical :: has_failed
```
certik: Based on my above comment, both `has_failed` and `hasfailed` can be used, depending on the situation.

STYLE_GUIDE.md
```diff
@@ -1,3 +1,66 @@
...
+real function linspace(...)
+```
+
+__BAD__:
+
+``` fortran
+logical :: has_failed
+real function lin_space(...)
+```
```
milancurcic: (I can't seem to work out GitHub suggestion for multi-lines)

This whole thing can be condensed to:

* Variable names should be made up of one or more full words separated by an underscore, for example `has_failed` is preferred over `hasfailed`
* Where conventional and appropriate shortening of a word is used then the underscore may be omitted, for example `linspace` is preferred over `lin_space`

STYLE_GUIDE.md
```diff
@@ -1,3 +1,66 @@
...
+
+## Keyword case
+
+Fortran keywords should __*not*__ be UPPERCASE.
+Modern editors and IDEs can highlight Fortran syntax and UPPERCASE keywords.
```
certik: What do you mean by the sentence `Modern IDEs can highlight UPPERCASE keywords`? Are you trying to say `Modern IDEs can make lowercase keywords uppercase just for the screen (they stay lowercase in the file)`? Or are you trying to say `Modern IDEs can only highlight UPPERCASE keywords`? Or is it a typo and you meant `Modern IDEs can highlight lowercase keywords`.

certik:
Thanks for writing it down. I think this mostly captures the style guide that most people agree on. I left some comments.

STYLE_GUIDE.md
```diff
@@ -1,3 +1,66 @@
...
+```
+
+__BAD__:
+
+``` fortran
+logical :: has_failed
+real function lin_space(...)
+```
+
+## Keyword case
```
milancurcic: This section can be dropped if we recommend lowercase for all variable and procedure names, as well as Fortran keywords, as in suggestion above.

STYLE_GUIDE.md
```diff
@@ -1,3 +1,66 @@
...
+
+Fortran keywords should __*not*__ be UPPERCASE.
+Modern editors and IDEs can highlight Fortran syntax and UPPERCASE keywords.
+UPPERCASE keywords give Fortran source code the appearance of being antiquated.
+
+## End <scope> block closing statements
+
+Fortran allows certain block constructs or scopes to include the name of the program unit in the end statement.
+The convention adopted herein is to __NOT__ include procedure names, `module` names and `program` names in the `end` statement.
```
milancurcic: I disagree here. stating the name of the unit on `end` statements help readers see what exactly is being ended.

STYLE_GUIDE.md
certik: I think `has_failed` and `lin_space` is equivalent. They could both be used, but if it's just two syllables, it is perfectly fine to remove the `_` and just make it `linspace` and `hasfailed`. It all depends on the context if one should be preferred over the other.

In general, it's worth not to be typing `_` all the time, and it actually somtimes improves readability to remove `_` for two syllable names. If it's more than two syllables, then the `_` improves readability.

STYLE_GUIDE.md
```diff
@@ -1,3 +1,66 @@
...
 
-This document will describe the code style to use.
+> [A]utomate and codify as much as you possibly can while remembering that the human touch is still necessary to praise
+> and be kind to your contributors.
+> Let robots handle your project’s pedantry and humans handle your project’s empathy.
+
+-- @mikemcquaid, [Homebrew] project leader<sup>[1]</sup>
+
+[1]: https://mikemcquaid.com/2018/06/05/robot-pedantry-human-empathy/
+[Homebrew]: https://brew.sh
```
milancurcic: Personally not crazy about this quote.

STYLE_GUIDE.md
milancurcic: I see your point but I don't think they're quite equivalent. I see `linspace` is a programming portmanteau for "linear space". `linspace` makes more sense because `lin` is not a word on its own. `has_failed` however are two independent words, and like `is`, word `has` implies a logical state. I find `hasfailed` about as readable as the other one, but `has_*` has a semantic meaning implied in it (similar to `is_`, `get_`, `set_`, etc.).

Perhaps these are not the clearest example for the guide because the difference is nuanced.

STYLE_GUIDE.md
certik: Yeah, me neither.

I do want to automate formatting and everything else. But the quote doesn't quite get get excited.

STYLE_GUIDE.md
certik: I do it sometimes for longer subroutines. For shorter ones I usually don't, so I quite liked the suggestion. But we can remove it, not a big deal.

STYLE_GUIDE.md
certik: Yes, I agree that `is_alpha` as we discussed should be with `_` because `is` has a semantic meaning. In the same spirit, if that is the case with `has`, then it should be `has_failed`. (All I was saying is that I can imagine a context when `has` does not have a semantic meaning, in which case `hasfailed` works also.) So I wanted to keep our rules general enough to allow for both, based on the context.

STYLE_GUIDE.md
certik: Regarding 80 characters, I follow very strictly 80 character limit (and frequently have several terminals side by side on a screen, which the 80 characters allow), but not everybody does, and so I like the way you phrased it as `Line length should be limited to 80 characters and must not exceed 132.`, to recommend to try to keep it under 80, but if it can't be done, then one can go over. @zbeekman is that an acceptable compromise for you?

STYLE_GUIDE.md
zbeekman: There is a *defacto* convention for using `.F90` and letting the compiler do the preprocessing. I would rather stick to this established convention. It indicates to the reader what to expect and works more smoothly with compilers and build tools like CMake.

STYLE_GUIDE.md
zbeekman: @certik yes, I like the way I phrased the line length stuff to address line length. Keeping it less than 80 is ideal, but under 132---to provide an absolute limit somewhere, and conform to the standard---is a good compromise IMO. Especially if you have 4 spaces of indentation.

@milancurcic : Well, when you say:

> If trailing whitespaces and EOL and EOF stuff will be handled by CI, no need to mention it in the guide.

what are you envisioning?

If contributors aren't notified ahead of time about CI checks it could be frustrating to learn after you've committed code that you need to make stylistic changes. Or are you suggesting that we have CI automatically alter and commit changes to contributed code so that it would conform with the style?

STYLE_GUIDE.md
zbeekman: I think I should get rid of that `UPPERCASE`, it's not pertinent.

STYLE_GUIDE.md
zbeekman: Yeah, I started the PR mostly as a place to discuss particulars. I'm not hugely against adding the ending names other than the fact that for short blocks it's more typing in an already chatty language.

STYLE_GUIDE.md
zbeekman: Haha, OK, that's fine with me. The point was just to convey a need for human empathy, and to highlight the fact to other maintainers that our time is better spent on substantive discussions. Where people have strong opinions about, e.g., style, then it's nice to have a robot complaining about it instead of a maintainer. Contributors tend to not take offense or want to argue, and everyone else gets to spend more time doing productive work. If you want to remove it, that's fine by me. BTW, I would recommend reading the article linked, and also another one about maintaining open source and the user-contributor-maintainer funnel (like a sales funnel).

STYLE_GUIDE.md
zbeekman: Also, I should note, there are only so many things that can be programmatically enforced for Fortran; its grammar is fairly complex and there are only so many tools that can enforce it. Fortunately, I think *requiring*/*adding* matching named end constructs is doable with `findent`. My suggestion as it stands is not, as far as I can tell, enforceable.

STYLE_GUIDE.md
zbeekman: Here is the contributor funnel article, if you're interested: https://mikemcquaid.com/2018/08/14/the-open-source-contributor-funnel-why-people-dont-contribute-to-your-open-source-project/

And here is the article that the quote is from: https://mikemcquaid.com/2018/06/05/robot-pedantry-human-empathy/

I think both are definitely worth a read.

STYLE_GUIDE.md
zbeekman: I've done my best to simplify and cleanup this section using more plain language, and respond to your suggestions.

@milancurcic I've included some stuff about tabs and trailing white space because I think it's worthwhile to be up-front about it, and don't envision a CI system commiting changes on a users behalf.

STYLE_GUIDE.md
zbeekman: As a starting point I'm adopting @milancurcic's text/suggestion, but I'm happy to iterate on it more if someone can propose a better alternative that we agree upon.

zbeekman:
I've decided to not add the enforcement half of this work quite yet, and let this PR be a place to discuss what the style guide is, before any effort to enforce it is made. (After all, setting style conventions seems to be a necessary prerequisite to trying to enforce any.)

STYLE_GUIDE.md
```diff
@@ -15,23 +15,23 @@ This style guide is a living document and proposed changes may be adopted after
 
 ## File naming conventions
 
-* Source files should contain at most one `program`, `module`, or `submodule`.
-* The filename should match the program or module name and have the file extension `.f90` or `.F90` if preprocessing is required.
+* Source files should contain at most one `program`, `module`, or `submodule`
+* The filename should match the program or module name and have the file extension `.f90` or `.F90` if preprocessing is required
```
ivan-pi: The first bullet and last bullet points are more or less the same.

zbeekman: Good point. I was moving this from another section and didn't realize it was already covered here. Thanks!

STYLE_GUIDE.md
zbeekman: This was obsoleted and removed due to other suggestions, marking resolved.

STYLE_GUIDE.md
zbeekman: I followed @milancurcic's suggestions from below but I'm happy to iterate on this more too. I'm marking as resolved because its outdated and changed, but if you feel strongly that what is there now needs changing, please don't hesitate to say so.

STYLE_GUIDE.md
milancurcic: > Or are you suggesting that we have CI automatically alter and commit changes to contributed code so that it would conform with the style?

Sorry, indeed I assumed this would be done automatically via CI. I agree, if the developer is to worry about these things, we should state it in the style guide.

zbeekman: We can setup pre-commit hooks to automatically fix issues like this and do auto-formatting, but it's hard to ensure that people are always using them.

In CI, I don't think it makes sense (at least not right now) to have CI authoring commits on the users behalf.

zbeekman:
@milancurcic Just out of curiosity, do we want to codify/formalize how code reviews and PRs work? When is it OK to merge? By whom? This gets into a bigger issue: governance, but we can punt on that for a bit.

milancurcic:
Yes, we should discuss that in #5 and come up with some workflow to start with.

STYLE_GUIDE.md
```diff
@@ -1,3 +1,49 @@
 # Fortran stdlib Style Guide
 
-This document will describe the code style to use.
+Adopting a consistent style can improve code legibility through the choice of good naming conventions.
+In addition, style checks will be run during CI to flag any severe non-conformance.
+This allows code review discussions to focus on semantics and substance rather than pedantry.
+Consistent whitespace usage, and not poluting line endigs with trailing white space makes `git diff`s considerably more legible.
```
certik:
```suggestion
Consistent whitespace usage, and not polluting line endings with trailing white space makes `git diff`s considerably more legible.
```

certik:
+1 to merge

STYLE_GUIDE.md
zbeekman: Speaking of typos, can I add https://github.com/reviewdog/action-misspell as a github action? It may need a new/different token setup. It creates really nice PR comments and would have caught this.

PR #49, opened 2019-12-28
Place ASCII control characters in derived type 

ivan-pi:
As discussed in #11, I have moved the ascii control characters into a derived type. A single instance of this type (with the parameter attribute) is exposed to the public and can be accesed using
```Fortran
use stdlib_experimental_ascii, only: ascii_control_char 
write(*,*) ascii_control_char%TAB//"Hello"//ascii_control_char%LF
```

The ascii character validation and conversion procedures are now elemental and work on character arrays. One of the tests has been modified to demonstrate this by allocating character arrays filled with different subsets of characters and generating a true/false table:
```
          is_control
          | is_printable
          | | is_whitespace
          | | | is_blank
          | | | | is_graphical
          | | | | | is_punctuation
          | | | | | | is_alphanum
          | | | | | | | is_alpha
          | | | | | | | | is_upper
          | | | | | | | | | is_lower
          | | | | | | | | | | is_digit
 decimal  | | | | | | | | | | | is_hex_digit
 -------------------------------------------
     0-8  T F F F F F F F F F F F
       9  T F T T F F F F F F F F
   10-13  T F T F F F F F F F F F
   14-31  T F F F F F F F F F F F
      32  F T T T F F F F F F F F
   33-47  F T F F T T F F F F F F
   48-57  F T F F T F T F F F T T
   58-64  F T F F T T F F F F F F
   65-70  F T F F T F T T T F F T
   71-90  F T F F T F T T T F F F
   91-96  F T F F T T F F F F F F
  97-102  F T F F T F T T F T F T
 103-122  F T F F T F T T F T F F
 123-126  F T F F T T F F F F F F
     127  T F F F F F F F F F F F
```

Last, I have replaced the `whitechar` function in `stdlib_experimental_io` with the `is_blank` function from the ascii module.

### DIFF
--- a/src/stdlib_experimental_ascii.f90
+++ b/src/stdlib_experimental_ascii.f90
@@ -1,118 +1,135 @@
 module stdlib_experimental_ascii
 
-    implicit none
-    private
+implicit none
+private
 
-    ! Character validation functions
-    public :: is_alpha, is_alphanum
-    public :: is_digit, is_hex_digit, is_octal_digit
-    public :: is_control, is_white, is_blank
-    public :: is_ascii, is_punctuation
-    public :: is_graphical, is_printable
-    public :: is_lower, is_upper
+! Character validation functions
+public :: is_alpha, is_alphanum
+public :: is_digit, is_hex_digit, is_octal_digit
+public :: is_control, is_white, is_blank
+public :: is_ascii, is_punctuation
+public :: is_graphical, is_printable
+public :: is_lower, is_upper
 
-    ! Character conversion functions
-    public :: to_lower, to_upper
+! Character conversion functions
+public :: to_lower, to_upper
 
-    ! All control characters in the ASCII table (see www.asciitable.com).
-    character(len=1), public, parameter :: NUL = achar(z'00') !! Null
-    character(len=1), public, parameter :: SOH = achar(z'01') !! Start of heading
-    character(len=1), public, parameter :: STX = achar(z'02') !! Start of text
-    character(len=1), public, parameter :: ETX = achar(z'03') !! End of text
-    character(len=1), public, parameter :: EOT = achar(z'04') !! End of transmission
-    character(len=1), public, parameter :: ENQ = achar(z'05') !! Enquiry
-    character(len=1), public, parameter :: ACK = achar(z'06') !! Acknowledge
-    character(len=1), public, parameter :: BEL = achar(z'07') !! Bell
-    character(len=1), public, parameter :: BS  = achar(z'08') !! Backspace
-    character(len=1), public, parameter :: TAB = achar(z'09') !! Horizontal tab
-    character(len=1), public, parameter :: LF  = achar(z'0A') !! NL line feed, new line
-    character(len=1), public, parameter :: VT  = achar(z'0B') !! Vertical tab
-    character(len=1), public, parameter :: FF  = achar(z'0C') !! NP form feed, new page
-    character(len=1), public, parameter :: CR  = achar(z'0D') !! Carriage return
-    character(len=1), public, parameter :: SO  = achar(z'0E') !! Shift out
-    character(len=1), public, parameter :: SI  = achar(z'0F') !! Shift in
-    character(len=1), public, parameter :: DLE = achar(z'10') !! Data link escape
-    character(len=1), public, parameter :: DC1 = achar(z'11') !! Device control 1
-    character(len=1), public, parameter :: DC2 = achar(z'12') !! Device control 2
-    character(len=1), public, parameter :: DC3 = achar(z'13') !! Device control 3
-    character(len=1), public, parameter :: DC4 = achar(z'14') !! Device control 4
-    character(len=1), public, parameter :: NAK = achar(z'15') !! Negative acknowledge
-    character(len=1), public, parameter :: SYN = achar(z'16') !! Synchronous idle
-    character(len=1), public, parameter :: ETB = achar(z'17') !! End of transmission block
-    character(len=1), public, parameter :: CAN = achar(z'18') !! Cancel
-    character(len=1), public, parameter :: EM  = achar(z'19') !! End of medium
-    character(len=1), public, parameter :: SUB = achar(z'1A') !! Substitute
-    character(len=1), public, parameter :: ESC = achar(z'1B') !! Escape
-    character(len=1), public, parameter :: FS  = achar(z'1C') !! File separator
-    character(len=1), public, parameter :: GS  = achar(z'1D') !! Group separator
-    character(len=1), public, parameter :: RS  = achar(z'1E') !! Record separator
-    character(len=1), public, parameter :: US  = achar(z'1F') !! Unit separator
-    character(len=1), public, parameter :: DEL = achar(z'7F') !! Delete
+! Ascii control characters
+public :: ascii_control_char
 
-    ! Constant character sequences
-    character(len=*), public, parameter :: fullhex_digits = "0123456789ABCDEFabcdef" !! 0 .. 9A .. Fa .. f
-    character(len=*), public, parameter :: hex_digits = fullhex_digits(1:16) !! 0 .. 9A .. F
-    character(len=*), public, parameter :: lowerhex_digits = "0123456789abcdef" !! 0 .. 9a .. f
-    character(len=*), public, parameter :: digits = hex_digits(1:10) !! 0 .. 9
-    character(len=*), public, parameter :: octal_digits = digits(1:8) !! 0 .. 7
-    character(len=*), public, parameter :: letters = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz" !! A .. Za .. z
-    character(len=*), public, parameter :: uppercase = letters(1:26) !! A .. Z
-    character(len=*), public, parameter :: lowercase = letters(27:) !! a .. z
-    character(len=*), public, parameter :: whitespace = " "//TAB//VT//CR//LF//FF !! ASCII _whitespace
+! Constant character sequences
+public :: fullhex_digits, hex_digits, lowerhex_digits, digits, octal_digits
+public :: letters, uppercase, lowercase, whitespace
+
+
+! All control characters in the ASCII table (see www.asciitable.com).
+type :: ascii_control_char_t
+    character(len=1) :: NUL = achar(z'00') !! Null
+    character(len=1) :: SOH = achar(z'01') !! Start of heading
+    character(len=1) :: STX = achar(z'02') !! Start of text
+    character(len=1) :: ETX = achar(z'03') !! End of text
+    character(len=1) :: EOT = achar(z'04') !! End of transmission
+    character(len=1) :: ENQ = achar(z'05') !! Enquiry
+    character(len=1) :: ACK = achar(z'06') !! Acknowledge
+    character(len=1) :: BEL = achar(z'07') !! Bell
+    character(len=1) :: BS  = achar(z'08') !! Backspace
+    character(len=1) :: TAB = achar(z'09') !! Horizontal tab
+    character(len=1) :: LF  = achar(z'0A') !! NL line feed, new line
+    character(len=1) :: VT  = achar(z'0B') !! Vertical tab
+    character(len=1) :: FF  = achar(z'0C') !! NP form feed, new page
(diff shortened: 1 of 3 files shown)


ivan-pi:
I tried to squash two commits. I am not sure what went wrong with the MacOS tests as the ubuntu tests run successfully.

certik:
I think this is fine. Generally it's better not to use derived types in a library because it forces all applications to also use them, even if they don't want to. But in this case it seems it is ok, it's used more like a namesoace.

milancurcic:
Looks good to me. @certik In what scenario would an application not want to use a derived type?

certik:
It doesn't quite apply in this PR, but a good example is a sparse CSR matrix, represented by three arrays Ap, Aj,  Ax. We will provide operations on it, such as matmul. The lowest level public API should simply accept the three arrays as input arguments, not use a derived type CSRMatrix. The reason for that is that as an application I would like to have a choice what data structures I use. If we only accept CSRMatrix as an argument then the application is forced to use it and what if this is some big production code that already uses its own derived type for CSR matrix, perhaps with a few more members such as name, or some other application specific metadata? Then the application is forced to create CSRMatrix derived type and copy the arrays there, or it is forced to redo its data structures. Not optimal. If on the other hand we provide an API that only accepts the three arrays directly, then the application simply passes the arrays in from its internal data structure directly.

Stdlib can still optionally provide a higher level API that uses the CSRMatrix derived type.

So the answer is: as an application I would almost always like to use a derived type for a CSR matrix, but on my own terms. Not forced by a library like stdlib.

CSR matrix is an example where most people including me would agree a derived type simplifies code and generally is appropriate for an end application. Other use cases, such as saveppm and creating an Image_t derived type, are much worse. The best is to leave applications the freedom to decide what derived types to use and when; and in stdlib to provide the actual functionality to operate on a CSR matrix or PPM images.

gronki:
I agree with @certik that placing character constants in a derived type is
a bad design decision. Derived types are not designed to store constants.
They are not namespaces. They should be not used as namespaces. If somebody
cares about pollution of name space, they can use only operator which was
designed exactly for that. I think we should use things as are intended and
not poorly implement ideas from other languages. Or the result will be a
Frankenstein.

niedz., 29 gru 2019, 16:07 użytkownik Ondřej Čertík <
notifications@github.com> napisał:

ivan-pi:
Thank you @certik and @gronki for explaining your opinions. It is good we discuss this usage case as it sets a precedence and will also influence future design choices. 

If we do not want the control character constants to polute the `stdlib_experimental_ascii` namespace, aside from the current PR which wraps them in a derived type (only exposing a single instance publically), we could put them in a separate module `stdlib_experimental_ascii_constants` along with the other constant character sequences (letters, digits).

We could also just give them longer names, e.g. `ascii_control_char_tab` instead of the current `tab`, and thereby hopefully prevent name clashes.

@gronki Do you offer any other suggestions?

jvdp1:
> If we do not want the control character constants to polute the `stdlib_experimental_ascii` namespace, aside from the current PR which wraps them in a derived type (only exposing a single instance publically), we could put them in a separate module `stdlib_experimental_ascii_constants` along with the other constant character sequences (letters, digits).

I agree with the comments. And this @ivan-pi 's solution is a solution I often use in my programs. If a user doesn't need constants, he doesn't need to call the module. While not ideal, it would also allow the use to simply write
`use stdlib_experimental_ascii`
without a need to specify what it just wants.

>We could also just give them longer names, e.g. ascii_control_char_tab instead of the current tab, and thereby hopefully prevent name clashes

Maybe the name of the module could be used in front of the variable, e.g., for `stdlib_experimental_ascii_constants`, `tab` would become `stdlib_ascii_constants_tab` (where I remove the `experimental` since it would be eventually removed).
I am afraid that something like `ascii_control_char_xxx` is still too generic. Also, using the names of the module in the names of the variables may help to find its origin in complex libraries.

certik:
Let's put them in a separate module then? In that case their names can stay as is, cannot they?

There is an open issue to make modules work as namespaces here:

https://github.com/j3-fortran/fortran_proposals/issues/1

Which would solve a lot of these issues. I do not recommend people to "use some_module" without the "only" part. Just like in Python.

zbeekman:
@ivan-pi I suspect the failures are just "cloud" issues. I'm not a member of the fortran-lang org, however, so I cannot re-run the CI test to confirm this. Given that the failures won't expand their logs, this is most likely the cause of the issue.

milancurcic:
> It doesn't quite apply in this PR, but a good example is a sparse CSR matrix, represented by three arrays Ap, Aj, Ax. We will provide operations on it, such as matmul. The lowest level public API should simply accept the three arrays as input arguments, not use a derived type CSRMatrix. 

I agree, you're talking in the context of providing procedural/functional vs. object-oriented APIs. Here it's meant exactly to emulate a namespace.

milancurcic:
@gronki Can you elaborate why you think it's a bad design choice? What are the specific caveats or downsides? I don't see them right now.

No matter if we adopt it here or not, I'd bet that it'd come up again as we keep working on stdlib. It'd be good for all of us to understand what are the downsides. I don't care if it was originally intended for a purpose or not. What I care about is what is the problem we're trying to solve and does the proposed method solve it. 

Different question is whether this is a problem at all. That's why I asked about this in #11.

milancurcic:
> Let's put them in a separate module then? In that case their names can stay as is, cannot they?

I think this solves only the scenario of user doing `use stdlib_experimental_ascii`, to prevent populating the global namespace with 33 short-named constants. The same user would do `use stdlib_experimental_ascii_constants` to get the control characters, which is convenient but against our recommended use of modules (`use ..., only ...:`). The downside is small complexity overhead from having a separate module. I don't recommend this approach.

Alternatives that remain are:

* Prefix control constants with `ascii_`. This would solve the problem of polluting a namespace in the event of `use stdlib_experimental_ascii`.
* Leave as is (reject this PR) and reconsider only in the event of users complaining.

I like both latter options better than adding the constants module.

certik:
I restarted the checks, but I don't know if it did anything... GitHub actions might not be as production ready as I was hoping. 

We can move to Azure pipelines, which can also check all three platforms (Linux, macOS and Windows) in one framework.

Otherwise we can do Travis + AppVeyor.

zbeekman:
The conflict may need to be resolved too. I'll try reproducing CI locally on macOS to see if there's anything obvious.

milancurcic:
Let's keep this ball rolling. How do we want to proceed?

Options:

* Emulate namespace with derived type (this PR). Supported by @ivan-pi and myself, cautiously supported by @certik; @gronki against (if you can provide some concrete downsides or caveats it would help!)
* Prefix control constants with `ascii_`. This would prevent polluting the global namespace with 33 short constants, but doesn't make it any easier to import;
* Mark this as "not a problem" (reject this PR)

@jvdp1 @marshallward @jacobwilliams do you mind chiming in?

milancurcic:
Also, @zbeekman what do you think?

certik:
By priority, my first preference is to simply reject the PR. My personal opinion is that we should not use derived type for everything to try to emulate namespaces. Rather, let's work on fixing this issue: https://github.com/j3-fortran/fortran_proposals/issues/1 and then this issue: https://github.com/j3-fortran/fortran_proposals/issues/86. With those fixed, one could access it by `std%ascii%tab`. I recommend against doing `use stdlib_ascii`, but rather always `use, stdlib_ascii, only: tab`. So there is no polluting.

My second preference is to put them in a separate module, that will not be imported by default when you `use stdlib_ascii`, so it won't pollute the namespace.

My last preference is a derived type. But these particular types will not be used that often, so if the majority wants a derived type, then I am fine with that.

More pragmatically, this is in the experimental module, so we can always revert this change later.

jvdp1:
My preferences would be:

1. to move all ASCII control characters AND constant character sequences in another module (with, e.g., `ascii_`, in front of them).
2. to put  all ASCII control characters AND constant character sequences in a public derive type.
3. to let the module as it was, BUT still adding `ascii_` (or similar) in front of the control characters and others. There will be always someone who will do `use stdlib_ascii` without `only`, and this will most likely lead to problems because of the existence of some `ff`, `lf`, `bel`,...

My feelind is that we should find, at least, something to avoid polluting namespaces by variables as `ff`, `lf`, `bel`,...

ivan-pi:
> This may be controversial, and I'm not necessarily saying this is `the right approach`™️: What if everything was wrapped up in an object including the procedures as TBPs/methods?

If I understand correctly you mean something along the lines of:
```Fortran
type :: ascii_tools_t
contains
... procedures for character conversion, etc.
end type
type(ascii_tools_t) :: ascii_tools = ascii_tools_t()
```
The user would then import a single instance of this derived type and just call the TPB/methods, completely avoiding the hassle of listing ten different functions:
```Fortran
use std_ascii, only: at => ascii_tools
print *, at%is_upper('A') ! prints T
```
It is kind of like a swiss-army knife. You only take out the tool you need. I would like to see what the others thoughts are on this one, however I having the feeling it is somehow too radically different to what most users are accustomed too. 

--- 

Given the replies from @certik and @jvdp1 we are now at 50/50 for or against using a derived type to emulate a namespace. While it would be easier to judge with some real world usage examples in the end it might turn out more comfortable to move the constants to a different module, e.g. `std_ascii_constants`. Then someone who just needs the tab character and carriage return can easily do:
```
use stdlib_ascii_constants, only: ascii_TAB, ascii_CR
```
and someone who wants all control characters will just leave the `only` part out. You can always wrap the constants into a derived type in the client code (I am here reversing the argument of @zbeekman).

Under the solution with a separate module, would the sequences for letters, digits, whitespace, and punctuation stay in the current module or would they move to the new one?

Could we use this issue to further motivate the development of namespaces over at [j3-fortran](https://github.com/j3-fortran/fortran_proposals/issues/1)? With the namespace syntax, we could have:
```
use, namespace :: ctr => ascii_constants
print *, ctr%TAB//"Hello"//ctr%CR//ctr%LF
```

certik:
> Could we use this issue to further motivate the development of namespaces over at [j3-fortran](https://github.com/j3-fortran/fortran_proposals/issues/1)? With the namespace syntax, we could have:

Yes!! Please do. Comment at the issue. Then go to https://github.com/j3-fortran/fortran_proposals/issues/122, and put this high in your priority list. :)

milancurcic:
I'm not adding a separate module just for these constants. It only solves half of the problem and introduces complexity cost. 

Considering we don't have majority agreement on way forward, I agree we should shelf this away (close PR) as @certik suggested.

certik:
Ok. Let's close this one for now.

This does not mean that we are saying "no" forever. It's just that we can't reach a solid agreement on this right now, and let's use our energy on things where we can reach an agreement. Once we get more users, we can revisit this.

marshallward:
We have successfully used a class-like type in Australia's main climate model's library, [libaccessom2](https://github.com/COSIMA/libaccessom2).  (All credit to the author, @nichannah, who designed and programmed this class, and would probably be very interested in the efforts going on here.)

The definition is based on `procedure` statements:

https://github.com/COSIMA/libaccessom2/blob/9fda758e017b5d1c55e7f39797da351317fb9390/libcouple/src/accessom2.F90#L19-L103

The overhead to load it is very low:

https://github.com/mom-ocean/MOM5/blob/50dc61e9d77c181e6ad0047925fafdcead64d87a/src/accessom_coupler/ocean_solo.F90#L111

Here is the `accessom2` class in action:

https://github.com/mom-ocean/MOM5/blob/50dc61e9d77c181e6ad0047925fafdcead64d87a/src/accessom_coupler/ocean_solo.F90#L228-L257

I am a big fan of this approach.  It worked well on our supercomputer, which was your typical hornet's nest of old and new libraries, so it is hopefully reasonably robust these days.

I could see a design approach like using `procedure` adapted to the ASCII methods here.

And big apologies for taking so long to reply, I'm slowly getting caught up on all the activities going on here.

epagone:
As a user, I would prefer the OO implementation along the lines suggested by @zbeekman and @marshallward . As soon as the standard is fixed (as suggested by @certik), then the interface can be simplified or changed. My 2 cent.

milancurcic:
> We have successfully used a class-like type in Australia's main climate model's library, libaccessom2. (All credit to the author, @nichannah, who designed and programmed this class, and would probably be very interested in the efforts going on here.)

Always nice to stumble upon datetime-fortran used in the wild :)

PR #50, opened 2019-12-29
set the cmake build to out-of-source

pdebuyl:
Change the github workflow for the cmake build.

I open the PR to trigger the workflow for the branch. The manual makefiles are not touched.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -52,10 +52,10 @@ jobs:
       run: brew install gcc@${GCC_V} || brew upgrade gcc@${GCC_V} || true
 
     - name: Configure with CMake
-      run: cmake -Wdev -S . -B . # We should build out of source but loadtxt needs fixing
+      run: cmake -Wdev -S . -B build # We should build out of source but loadtxt needs fixing
 
     - name: Build and compile
-      run: cmake --build . || cmake --build . --verbose --parallel 1
+      run: cmake --build build || cmake --build build --verbose --parallel 1
 
     - name: test
-      run: cmake --build . --target test
+      run: cmake --build build --target test
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -52,7 +52,7 @@ jobs:
       run: brew install gcc@${GCC_V} || brew upgrade gcc@${GCC_V} || true
 
     - name: Configure with CMake
-      run: cmake -Wdev -S . -B build # We should build out of source but loadtxt needs fixing
+      run: cmake -Wdev -S . -B build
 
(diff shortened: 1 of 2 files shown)


certik:
I suggest we test both in tree and out of tree. 

The out of tree can be the default to test various compilers, etc. But on each platform we should have at least one test for in tree builds to ensure they still work.

zbeekman:
@certik 
> I suggest we test both in tree and out of tree.

Out of curiosity, what is the motivation for maintaining in-tree builds?

Out-of-source builds are easier to cleanup after, and don't risk clobbering WIP changes. If we truly want to support both then I suppose we should test in source builds too, but I don't see a strong case for insisting on supporting both. In addition, out-of-source builds are more likely to fail from a misconfigured build system than in-source builds are.

I don't care that much about whether or not we want to support in source builds, but we don't want a combinatorial explosion of our build matrix if we can avoid it.

.github/workflows/main.yml
```diff
@@ -52,10 +52,10 @@ jobs:
       run: brew install gcc@${GCC_V} || brew upgrade gcc@${GCC_V} || true
 
     - name: Configure with CMake
-      run: cmake -Wdev -S . -B . # We should build out of source but loadtxt needs fixing
+      run: cmake -Wdev -S . -B build # We should build out of source but loadtxt needs fixing
```
zbeekman:
```suggestion
      run: cmake -Wdev -S . -B build
```

zbeekman: Actually, this comment should be deleted if the issue is fixed and being tested.

pdebuyl:
I removed the comment. I didn't see that one could accept "suggested commits" online and did it via the CLI.

Regarding in-tree builds, I don't use them in my projects, I don't know how useful they are.

I proposed the cmake patch because I tested the build on my computer with the "idiomatic"

    mkdir build
    cd build
    cmake ..
    make

Maybe we need test users, with different coding habits.

zbeekman:
It seems @certik is one such "test user" as he reported elsewhere that he typically does in-tree CMake builds... Given the goal of supporting Makefiles and CMake, I would advocate for in-source builds being "unsupported". Interested parties can still try to keep them working, but I don't want to explode our CI test matrix anymore than 3 oses and 3 versions of Fortran compilers, unless there is a really good reason.

jacobwilliams:
I don't think we should encourage in-source builds. :)

certik:
This PR is ready to go in, so I am going to merge it.

certik:
I submitted #55 for in-tree CI tests.

PR #51, opened 2019-12-30
improve cmake build

### LINKED ISSUE #47: CMake build in place or copy test files
Without the datafiles, the test would fail on my machine when running an out of source build.

The branch here https://github.com/pdebuyl/stdlib/tree/cmake_copy_dat_files make a copy of the data files. I preferred to open an issue since I don't know if a PR would be of any use here.

scivision:
* use CMAKE_Fortran_MODULE_DIRECTORY to avoid introspecting directory structure
* don't build in-source, use test WORKING_DIRECTORY instead
* use modern add_test() syntax to avoid introspecting project directory structure
* use open(..., action=) to help indicate / safeguard intended file use
* don't have to accommodate CMake <= 2.4 circa 2006

closes #47

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -52,7 +52,7 @@ jobs:
       run: brew install gcc@${GCC_V} || brew upgrade gcc@${GCC_V} || true
 
     - name: Configure with CMake
-      run: cmake -Wdev -S . -B build
+      run: cmake -Wdev -DCMAKE_BUILD_TYPE=Release -S . -B build
 
     - name: Build and compile
       run: cmake --build build || cmake --build build --verbose --parallel 1
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,9 +1,8 @@
-cmake_minimum_required(VERSION 3.5.0 FATAL_ERROR)
-
-enable_language(Fortran)
-
-project(stdlib)
-
+cmake_minimum_required(VERSION 3.5.0)
+project(stdlib Fortran)
 enable_testing()
+
+# this avoids stdlib and projects using stdlib from having to introspect stdlib's directory structure
+set(CMAKE_Fortran_MODULE_DIRECTORY ${CMAKE_BINARY_DIR})
 
 add_subdirectory(src)
--- a/src/stdlib_experimental_io.f90
+++ b/src/stdlib_experimental_io.f90
@@ -46,7 +46,7 @@ real(sp), allocatable, intent(out) :: d(:,:)
 integer :: s
 integer :: nrow,ncol,i
 
-open(newunit=s, file=filename, status="old")
+open(newunit=s, file=filename, status="old", action="read")
 
 ! determine number of columns
 ncol = number_of_columns(s)
@@ -89,7 +89,7 @@ real(dp), allocatable, intent(out) :: d(:,:)
 integer :: s
 integer :: nrow,ncol,i
 
-open(newunit=s, file=filename, status="old")
+open(newunit=s, file=filename, status="old", action="read")
 
 ! determine number of columns
 ncol = number_of_columns(s)
@@ -132,7 +132,7 @@ real(qp), allocatable, intent(out) :: d(:,:)
 integer :: s
 integer :: nrow,ncol,i
 
-open(newunit=s, file=filename, status="old")
+open(newunit=s, file=filename, status="old", action="read")
 
 ! determine number of columns
 ncol = number_of_columns(s)
@@ -164,7 +164,7 @@ real(sp), intent(in) :: d(:,:)           ! The 2D array to save
 ! call savetxt("log.txt", data)
 
 integer :: s, i
-open(newunit=s, file=filename, status="replace")
+open(newunit=s, file=filename, status="replace", action="write")
 do i = 1, size(d, 1)
     write(s, *) d(i, :)
 end do
@@ -187,7 +187,7 @@ real(dp), intent(in) :: d(:,:)           ! The 2D array to save
 ! call savetxt("log.txt", data)
 
 integer :: s, i
-open(newunit=s, file=filename, status="replace")
+open(newunit=s, file=filename, status="replace", action="write")
 do i = 1, size(d, 1)
     write(s, *) d(i, :)
 end do
@@ -210,7 +210,7 @@ real(qp), intent(in) :: d(:,:)           ! The 2D array to save
 ! call savetxt("log.txt", data)
 
 integer :: s, i
-open(newunit=s, file=filename, status="replace")
+open(newunit=s, file=filename, status="replace", action="write")
 do i = 1, size(d, 1)
     write(s, *) d(i, :)
 end do
@@ -243,9 +243,9 @@ integer function number_of_rows_numeric(s)
  ! determine number or rows
  integer,intent(in)::s
  integer :: ios
- 
+
  real::r
- 
+
  rewind(s)
  number_of_rows_numeric = 0
  do
@@ -253,7 +253,7 @@ integer function number_of_rows_numeric(s)
     if (ios /= 0) exit
     number_of_rows_numeric = number_of_rows_numeric + 1
  end do
- 
+
  rewind(s)
 
 end function
--- a/src/tests/ascii/CMakeLists.txt
+++ b/src/tests/ascii/CMakeLists.txt
@@ -1,5 +1,3 @@
-project(ascii)
-
 add_executable(test_ascii test_ascii.f90)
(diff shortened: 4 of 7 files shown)


certik:
Thanks a lot. I like the spirit of these changes.

 One issue with the working directory is that the tests also create files. So they will pollute the working directory. That I think is not expected when you use a build directory that ctest will create files outside of it.

scivision:
Thanks, I changed that test to accept an input directory argument

.github/workflows/main.yml
```diff
@@ -52,10 +52,10 @@ jobs:
       run: brew install gcc@${GCC_V} || brew upgrade gcc@${GCC_V} || true
 
     - name: Configure with CMake
-      run: cmake -Wdev -S . -B . # We should build out of source but loadtxt needs fixing
+      run: cmake -Wdev -DCMAKE_BUILD_TYPE=Release -B build
```
certik: Don't you want to keep the `-S .` option? That's where the root directory is.

src/stdlib_experimental_io.f90
```diff
@@ -46,7 +46,7 @@ subroutine sloadtxt(filename, d)
 integer :: s
 integer :: nrow,ncol,i
 
-open(newunit=s, file=filename, status="old")
+open(newunit=s, file=filename, status="old", action="read")
```
certik: Is the `action` argument required to be used in `open`? I had to look it up what it means. The standard says "If this specifier is omitted, the default value is processor dependent.", which seems to imply that we have to use it, because if the processor dependent default value happens to be `read`, then we cannot write to the file. This seems to be a pretty bad design of the Fortran standard, similar to `status`, which you have to specify also, otherwise it is processor dependent.

src/tests/ascii/CMakeLists.txt
```diff
@@ -1,9 +1,4 @@
-include_directories(${PROJECT_BINARY_DIR}/src)
```
certik: Thanks for removing it, this was a bad design that slipped through in our review of #23 and #32 (which just copied it).

src/tests/loadtxt/test_savetxt.f90
```diff
@@ -4,57 +4,77 @@ program test_loadtxt
...
 
-    subroutine test_sp()
+    function get_outpath() result(outpath)
+    integer :: ierr
+    character(256) :: argv
+    character(:), allocatable :: outpath
+
+    call get_command_argument(1, argv, status=ierr)
+    if (ierr==0) then
+    outpath = trim(argv)
```
certik: Indent this line.

src/tests/loadtxt/test_savetxt.f90
```diff
@@ -4,57 +4,77 @@ program test_loadtxt
...
+    else
+    outpath = '.'
```
certik: Indent this line.

src/tests/loadtxt/test_savetxt.f90
certik:
```suggestion
        outpath = trim(argv)
```

src/tests/loadtxt/test_savetxt.f90
certik:
```suggestion
        outpath = '.'
```

certik:
@scivision please allow people with push access to modify your source branch (it's a check box somewhere). Then I can commit my suggestions above and merge this. Tons of other PRs depend on this.

.github/workflows/main.yml
scivision: `cmake -S` is a new feature for CMake 3.13. Since it seems the trend is to have CMake 3.14 as a minimum, it's fine to keep -S .

src/stdlib_experimental_io.f90
zbeekman: Agreed about the shortcoming of this in the standard. Either way it's good practice to

1. Indicate intent
2. Prevent accidental writing when you only should be reading

src/tests/ascii/CMakeLists.txt
```diff
@@ -1,9 +1,4 @@
...
-
-project(ascii)
-
 add_executable(test_ascii test_ascii.f90)
 target_link_libraries(test_ascii fortran_stdlib)
 
-add_test(test_ascii ${PROJECT_BINARY_DIR}/test_ascii)
-
+add_test(NAME ASCII COMMAND $<TARGET_FILE:test_ascii>)
```
zbeekman: 👍

src/tests/ascii/CMakeLists.txt
zbeekman: Yeah, this command should never be used.

src/stdlib_experimental_io.f90
scivision: `action=` is not a required parameter, but I think it's "safer" to use it to handle edge cases like use on network drives without write permission or other weird permissions. And just to help avoid writing a file if a handle variable is used inadvertently. I have not had any issues using `action=` everywhere.

src/tests/ascii/CMakeLists.txt
scivision: old CMake syntax is painful in every way, yes

certik:
I added in, and then approved it. Don't know why, because it should not be used.

CMakeLists.txt
```diff
@@ -1,9 +1,8 @@
...
-enable_language(Fortran)
-
-project(stdlib)
-
+cmake_minimum_required(VERSION 3.5.0)
+project(stdlib Fortran)
 enable_testing()
 
+# this avoids stdlib and projects using stdlib from having to introspect stdlib's directory structure
+set(CMAKE_Fortran_MODULE_DIRECTORY ${CMAKE_BINARY_DIR})
```
zbeekman: @scivision This is a pretty big hammer. For now, I'm fine with this, but it comes with its own complexities and issues. Eventually it would be nice to install a package-config (`.pc`) file and a [CMake package configuration file](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#creating-a-package-configuration-file) that could help consumers use the software if it's installed somewhere.

The drawbacks of this design choice are that:

1. All client code will assume modules will always be in the same place: at the root of the build directory
2. Modules with the same name will clobber each other (as was the case for the bug I found and fixed in your CMake with submodules repository)
3. Any testing & infrastructure modules that we may want to keep private to this project now get mixed in with modules intended to be consumed by client code.

Overall I think this move makes sense right now, but that a more nuanced approach is warranted in the future.

src/tests/loadtxt/CMakeLists.txt
```diff
@@ -1,15 +1,10 @@
...
 
-add_test(test_loadtxt ${PROJECT_BINARY_DIR}/test_loadtxt)
-add_test(test_savetxt ${PROJECT_BINARY_DIR}/test_savetxt)
-
-file(COPY array1.dat  array2.dat  array3.dat  array4.dat
-     DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
+add_test(NAME load_text COMMAND $<TARGET_FILE:test_loadtxt> ${CMAKE_CURRENT_BINARY_DIR}
+  WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+add_test(NAME save_text COMMAND $<TARGET_FILE:test_savetxt> ${CMAKE_CURRENT_BINARY_DIR}
+  WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
```
zbeekman: The alternative to this is to use test fixtures to move files around. In addition, care should be taken to prevent issues when testing in parallel if multiple tests read and/or modify the same files, or if one test consumes a file created or modified by another test.

zbeekman: (But, just to clarify, this approach makes sense here, as far as I can tell from a very cursory look at the code in the tests.)

zbeekman:
Looks like this is ready to :ship:

src/tests/loadtxt/CMakeLists.txt
scivision: yes I think so too, the test is rudimentary at the moment and doesn't yet need these protections.

CMakeLists.txt
scivision: yes more typically I set Fortran_Module_Directory as a property on each and every target, perhaps with foreach() loop if lots of targets. Just trying to level up from `include_directories()` until things get further along.

certik: What problems does setting `CMAKE_Fortran_MODULE_DIRECTORY` fix? I see the drawbacks, that @zbeekman described, but I do not see the advantages.

PR #52, opened 2019-12-30
Fortran 77 print instead of write(*,*)

scivision:
I didn't see this discussed previously--

```fortran
write(*,*)
```
is for printing to console. This makes it harder to recursively search files for where console output is given with special formatting etc. 

I would kindly suggest to consider Fortran 77 standard `print`, which is identical at the [assembly code level](https://www.scivision.dev/print-vs-write-fortran/).

### DIFF
--- a/src/tests/ascii/test_ascii.f90
+++ b/src/tests/ascii/test_ascii.f90
@@ -8,13 +8,13 @@ program test_ascii
         is_control, is_punctuation, is_graphical, is_printable, is_ascii, &
         to_lower, to_upper, LF, TAB, NUL, DEL
 
-    write(*,*) "Lowercase letters: ", lowercase
-    write(*,*) "Uppercase letters: ", uppercase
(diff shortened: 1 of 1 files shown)


jacobwilliams:
Is `write(*,*)` really nonstandard? I never knew that.

scivision:
yea I was wrong there, I was thinking in Fortran 2003 terms. In any case, I think the issue is over being able to find where non-stdout text is / is not being output, I have found it convenient to use print, which is identical from the compiler's perspective.

jacobwilliams:
how about using `output_unit`?

certik:
I also only use `print`, never `write(*,*)`. +1 to merge.

certik:
@jacobwilliams are you thinking of something like this:
```fortran
...
write (output_unit, *) ...
```
? That seems more complicated than just:
```
print *, ...
```

PR #54, opened 2019-12-30
add system module

scivision:
There are a number of capabilities it would be useful to bring from
cstdlib and STL. This is an initial demonstration, replacing
the non-cross-compiler sleep() with a standard implementation that
works across compilers and operating systems, with millisecond integer
input.
For example, Intel compilers can hang at runtime if `call sleep()` is used without `use ifport`. This PR technique doesn't suffer from that issue across operating systems and compilers.

There are a number of other useful functions that can be implemented like this, including resolving absolute filenames and expanding ~ to user home path

Adding CI for Windows brought an interesting corner case to light--runtime segfaults on Windows. I don't see these on my Windows Gfortran PC. This is a common type of corner case, tricky/shaky tests using uncommon features where it isn't feasible to whitelist/blacklist systems. A future PR would do a "check_fortran_source_run()" on a snippet of code to ensure the program can run instead of nuisance failure.

### DIFF
--- a/src/stdlib_experimental_system.F90
+++ b/src/stdlib_experimental_system.F90
@@ -1,0 +1,43 @@
+module stdlib_experimental_system
+use, intrinsic :: iso_c_binding, only : c_int, c_long
+implicit none
+private
+public :: sleep
+
+interface
+#ifdef _WIN32
+subroutine winsleep(dwMilliseconds) bind (C, name='Sleep')
+!! void Sleep(DWORD dwMilliseconds)
+!! https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-sleep
+import c_long
+integer(c_long), value, intent(in) :: dwMilliseconds
+end subroutine winsleep
+#else
+integer(c_int) function usleep(usec) bind (C)
+!! int usleep(useconds_t usec);
+!! https://linux.die.net/man/3/usleep
+import c_int
+integer(c_int), value, intent(in) :: usec
+end function usleep
+#endif
+end interface
+
+contains
+
+subroutine sleep(millisec)
+integer, intent(in) :: millisec
+integer(c_int) :: ierr
+
+#ifdef _WIN32
+!! PGI Windows, Ifort Windows, ....
+call winsleep(int(millisec, c_long))
+#else
+!! Linux, Unix, MacOS, MSYS2, ...
+ierr = usleep(int(millisec * 1000, c_int))
+if (ierr/=0) error stop 'problem with usleep() system call'
+#endif
+
+
+end subroutine sleep
+
+end module stdlib_experimental_system
--- a/src/tests/system/CMakeLists.txt
+++ b/src/tests/system/CMakeLists.txt
@@ -1,5 +1,5 @@
 add_executable(test_sleep test_sleep.f90)
 target_link_libraries(test_sleep fortran_stdlib)
 
-add_test(NAME Sleep COMMAND $<TARGET_FILE:test_sleep> 650)
-set_tests_properties(Sleep PROPERTIES TIMEOUT 5)
+add_test(NAME Sleep COMMAND $<TARGET_FILE:test_sleep> 350)
+set_tests_properties(Sleep PROPERTIES TIMEOUT 1)
--- a/src/tests/system/CMakeLists.txt
+++ b/src/tests/system/CMakeLists.txt
@@ -1,0 +1,5 @@
+add_executable(test_sleep test_sleep.f90)
+target_link_libraries(test_sleep fortran_stdlib)
+
+add_test(NAME Sleep COMMAND $<TARGET_FILE:test_sleep> 650)
+set_tests_properties(Sleep PROPERTIES TIMEOUT 5)
--- a/src/tests/system/test_sleep.f90
+++ b/src/tests/system/test_sleep.f90
@@ -1,0 +1,18 @@
+program test_sleep
+
+use stdlib_experimental_system, only : sleep
+
+implicit none
+
+integer :: ierr, millisec
+character(8) :: argv
+
+millisec = 780
+call get_command_argument(1, argv, status=ierr)
+if (ierr==0) read(argv,*) millisec
+
+if (millisec<0) millisec=0
+
+call sleep(millisec)
+
+end program
--- a/src/tests/system/test_sleep.f90
+++ b/src/tests/system/test_sleep.f90
@@ -1,11 +1,15 @@
 program test_sleep
-
+use, intrinsic :: iso_fortran_env, only : int64, real64
 use stdlib_experimental_system, only : sleep
 
 implicit none
 
 integer :: ierr, millisec
 character(8) :: argv
+integer(int64) :: tic, toc, trate
+real(real64) :: t_ms
+
+call system_clock(count_rate=trate)
 
 millisec = 780
 call get_command_argument(1, argv, status=ierr)
@@ -13,6 +17,17 @@ if (ierr==0) read(argv,*) millisec
 
 if (millisec<0) millisec=0
 
+call system_clock(count=tic)
 call sleep(millisec)
+call system_clock(count=toc)
+
+t_ms = (toc-tic) * 1000._real64 / trate
+
+if (millisec > 0) then
+  if (t_ms < 0.5 * millisec) error stop 'actual sleep time was too short'
+  if (t_ms > 2 * millisec) error stop 'actual sleep time was too long'
+endif
+
+print '(A,F8.3)', 'OK: test_sleep: slept for (ms): ',t_ms
 
 end program
--- a/README.md
+++ b/README.md
@@ -1,5 +1,9 @@
 # Fortran Standard Library
 
+[![Actions Status](https://github.com/fortran-lang/stdlib/workflows/CI/badge.svg)](https://github.com/fortran-lang/stdlib/actions)
+[![Actions Status](https://github.com/fortran-lang/stdlib/workflows/CI_windows/badge.svg)](https://github.com/fortran-lang/stdlib/actions)
+
+
 ## Goals and Motivation
 
 The Fortran Standard, as published by the ISO (https://wg5-fortran.org/), does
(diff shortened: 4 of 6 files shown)


certik:
Thanks @scivision for pushing this. Can you create another PR where you get the Windows build working? I am working reviewing #51 now, let's that one in first.

I like what you are trying to do, and let's get it polished an in.

scivision:
Yes basically the CMake needs #51 to get working where I could put the necessary Windows stuff if. For these Windows tests, it will require CMake >= 3.14 to do the workaround.

certik:
@scivision in that case we can depend on CMake 3.14, which I know @zbeekman would be very happy about. :)

zbeekman:
> @scivision in that case we can depend on CMake 3.14, which I know @zbeekman would be very happy about. :)

I told you there be bugs...  all's well that ends well.

.github/workflows/ci_windows.yml
```diff
@@ -0,0 +1,37 @@
...
+
+    - name: CTest
+      run: ctest --parallel -V
+      working-directory: build
+
+    - uses: actions/upload-artifact@v1
+      if: failure()
+      with:
+        name: WindowsCMakeTestlog
+        path: build/Testing/Temporary/LastTest.log
```
zbeekman: I can consolidate this with the build matrix once this is merged so that it's not a separate workflow. I'll have to look into the windows runners some more. If it is possible to test multiple GFortrans that would be nice but any windows tests are a win in my book.

src/tests/system/test_sleep.f90
```diff
@@ -0,0 +1,16 @@
+program sleepy
+
+use stdlib_experimental_system, only : sleep_millisec
```
certik: Add `implicit none`.

src/stdlib_experimental_system.F90
```diff
@@ -0,0 +1,37 @@
...
+end subroutine system_sleep
+#else
+subroutine system_sleep(us) bind (C, name='usleep')
+import c_int
+integer(c_int), value, intent(in) :: us
+end subroutine system_sleep
+#endif
+end interface
+
+public :: sleep_millisec
```
certik: I would move this up in the module.

Is the `sleep_millisec` a good name? Perhaps `sleep_ms` is better? Or should this be just called `sleep`? How does C++ and Python handle this and what name do they use?

src/tests/system/test_sleep.f90
```diff
@@ -0,0 +1,16 @@
+program sleepy
```
milancurcic:
```suggestion
program test_sleep
```

src/stdlib_experimental_system.F90
scivision: The issue is generating compiler warnings because of their own `sleep()` implementations. I haven't looked into if they can be disabled. This may be an issue other such functions in this module would encounter.

src/stdlib_experimental_system.F90
```diff
@@ -0,0 +1,37 @@
...
+
+contains
+
+subroutine sleep_millisec(t)
+integer, intent(in) :: t
+integer(c_int) :: tsys
+
+tsys = int(t, c_int)
```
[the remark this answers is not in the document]
scivision: the issue is potential compiler warnings about non-explicit cast. I didn't see if that was an issue here, just a general lint habit I've gotten into.

zbeekman:
> I would strongly suggest to split the unrelated and approved parts of this PR into separate PRs, so that we can merge them:

I'll have a go at this in a moment.

certik:
@scivision most of your other unrelated improvements in this PR have been merged into master. Do you want to submit a clean PR with just the system module?

src/stdlib_experimental_system.F90
```diff
@@ -0,0 +1,44 @@
...
+integer(c_int) function usleep(usec) bind (C)
+!! int usleep(useconds_t usec);
+!! https://linux.die.net/man/3/usleep
+import c_int
+integer(c_int), value, intent(in) :: usec
+end function usleep
+#endif
+end interface
+
+public :: sleep
```
certik: I think it's clearer to put this line right after `private`. So that people can see the public API right away.

src/tests/system/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+add_executable(test_sleep test_sleep.f90)
+target_link_libraries(test_sleep fortran_stdlib)
+
+add_test(NAME Sleep COMMAND $<TARGET_FILE:test_sleep> 1234)
```
certik: This sleeps for over 1s, so it makes all the tests hang to wait for this one to finish. See the timings:
```
$ ctest 
Test project /home/ondrej/repos/stdlib
      Start  1: AlwaysSkip
      Start  2: AlwaysFail
      Start  3: ASCII
      Start  4: loadtxt
 1/11 Test  #1: AlwaysSkip .......................***Skipped   0.01 sec
 2/11 Test  #2: AlwaysFail .......................   Passed    0.01 sec
 3/11 Test  #3: ASCII ............................   Passed    0.01 sec
 4/11 Test  #4: loadtxt ..........................   Passed    0.01 sec
      Start  5: savetxt
      Start  6: loadtxt_qp
      Start  7: savetxt_qp
      Start  8: open
 5/11 Test  #8: open .............................   Passed    0.00 sec
 6/11 Test  #7: savetxt_qp .......................   Passed    0.00 sec
 7/11 Test  #6: loadtxt_qp .......................   Passed    0.00 sec
 8/11 Test  #5: savetxt ..........................   Passed    0.00 sec
      Start  9: parse_mode
      Start 10: OPTVAL
      Start 11: Sleep
 9/11 Test #10: OPTVAL ...........................   Passed    0.10 sec
10/11 Test  #9: parse_mode .......................   Passed    0.10 sec
11/11 Test #11: Sleep ............................   Passed    1.24 sec

100% tests passed, 0 tests failed out of 11

Label Time Summary:
quadruple_precision    =   0.00 sec*proc (2 tests)

Total Test time (real) =   1.26 sec

	  1 - AlwaysSkip (Skipped)
```
I would lower the sleeping amount.

src/tests/system/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 add_executable(test_sleep test_sleep.f90)
 target_link_libraries(test_sleep fortran_stdlib)
 
-add_test(NAME Sleep COMMAND $<TARGET_FILE:test_sleep> 1234)
+add_test(NAME Sleep COMMAND $<TARGET_FILE:test_sleep> 650)
```
certik: Why not to reduce this to 10ms? 0.65s is still very much noticeable.

scivision: If a mistake is made the test might just pass through without sleeping at all. Especially with the test run in parallel it shouldn't really add much time. so the test is set up to be a finite sleep time so that we can look back and see that yes indeed it actually was sleeping.

certik:
@scivision I would like to merge your `system` module, but when you bundle unrelated CI changes, I don't want to just merge it, because some people might disagree with the CI changes, or want to further discuss them.

Do you think it would be please possible to not bundle unrelated changes in a single PR? 

Let's split this PR, so that we can merge the `system` module, which I think is a great start and exactly as you say, it would allow us to use this approach for other things.

src/tests/system/CMakeLists.txt
certik: I feel that's a very fragile way of testing sleeping, if it requires manual "eyeballing". And somebody can adjust these numbers in the future also. 

Isn't there some way of using [ctest_sleep](https://cmake.org/cmake/help/latest/command/ctest_sleep.html) to test this robustly with a shorter timeout?

scivision: I think sleep() once it works as-is, would only be likely to break on systems where there are POSIX implementation issues. To enforce a minimum time in a platform agnostic way might be easiest done within Fortran by checking CPU_TIME. Trying to push these time durations too low can run into issues whether due to slow CPU or OS limitations like WIndows. Where the procedure setup/teardown can be longer than sleep()

scivision: So how I have seen non-standard/proprietary sleep() fail is to hang forever (ifort) or not sleep at all (implementation mistakes)

certik: Ok. I am fine with the current solution for now.

scivision: 504f9b3 should address the auto-test of time.

certik: Thanks, I think that looks good.

src/stdlib_experimental_system.F90
```diff
@@ -0,0 +1,43 @@
...
+
+subroutine sleep(millisec)
+integer, intent(in) :: millisec
+integer(c_int) :: ierr
+
+#ifdef _WIN32
+!! PGI Windows, Ifort Windows, ....
+call winsleep(int(millisec, c_long))
+#else
+!! Linux, Unix, MacOS, MSYS2, ...
```
zbeekman: @scivision does MSYS2 not define `_WIN32`? I thought it did but I only ever use Intel on Windows.

scivision: MSYS2 only defines `_WIN32` with flag `-mwin32`

certik:
Let's not worry about formatting too much in individual PRs --- rather let's figure out how to do this automatically for all source files.

certik:
@zbeekman thanks for the review!

@milancurcic can you please review this also?

For new functionality, I want several independent reviews, to prevent putting things in that might not be well designed or too bloated.

certik:
@milancurcic can you please review this also? 

@scivision sorry for the delay.

urbanjost:
Are we discussing something like an interface to C system routines like often proposed for POSIX interfaces?  I use routines like that a lot but cooked my own (not sure how many platforms they would work on, but worked on all the ones I needed) in 
[M_system](https://github.com/urbanjost/M_system).  A lot of compilers supported some variant of a "POSIX" interface, but in different ways, which made it highly non-portable. Never quite sure why this did not become part of the standard, especially when "POSIX" was a hot topic.  The only one I remember doing a full PXF interface was Cray.

nncarlson:
The NAG compiler comes with a collection of intrinsic `f90_unix_*` modules that implement a bunch of POSIX stuff. They should be examples to consider if we wanted to head in that direction with stdlib. The documentation for them is [here](https://www.nag.com/nag-compiler#tab-357).

urbanjost:
Almost the same list as I have; which is a good sign it can be done relatively generically.  Cray, Intel, and gfortran all have a lot of the same functionality as extensions, but all as proprietary extensions so not very portable. Probably others too. None are OOPs, which is interesting.

PR #55, opened 2019-12-30
Test in-tree builds

certik:
This adds one in-tree build test for each platform, to ensure our CMake
build system works both out-of-tree as well as in-tree.

In terms of costs, it seems to add 1s on one Linux build, and 3s on one macOS build.


certik:
Thanks for the review!

PR #57, opened 2019-12-31
Let's use 4 spaces instead of 2

certik:
I think most people agreed to use 4 spaces instead of just 2.

### DIFF
--- a/STYLE_GUIDE.md
+++ b/STYLE_GUIDE.md
@@ -29,7 +29,7 @@ This style guide is a living document and proposed changes may be adopted after 
 By setting and following a convention for indentation and whitespace, code reviews and git-diffs can
 focus on the semantics of the proposed changes rather than style and formatting.
 
-* The body of every Fortran construct should be indented by __two (2) spaces__
+* The body of every Fortran construct should be indented by __two (4) spaces__
 * Line length *should be limited to 80 characters* and __must not exceed 132__
 * Please do not use <kbd>Tab</kbd> characters for indentation
 * Please remove trailing white space before committing code


zbeekman:
@certik you need to change [`.editorconfig`] too!

[`.editorconfig`]: https://github.com/fortran-lang/stdlib/blob/master/.editorconfig

I list the reasons why I changed it back to 2 below, but if you want 4 that's fine by me.

## Preferences (from #3):

- @zbeekman: 2
- @milancurcic 2
- @jvdp1 1 or 2
- @longb don't care
- @gronki don't care
- @certik 4
- @ivan-pi 4
- @marshallward 4

__3__ preferences for 2 spaces and __3__ for 4 spaces.
So, as you can see there is effectively a tie.

## Further considerations from #3

My further understanding from #3 was that a consensus was reached by the majority that:

- module, submodule and program units should be indented
- procedure, type and interface bodies should be indented
- usual control constructs should be indented
- should strive to stay within 80 chars, with a hard limit of 132

Given the concerns about line length when indenting procedure bodies etc. and the effective tie in preference for 2 vs 4 combined that 2 is the default for `findent` I thought it prudent to switch to 2. However, I may have misinterpreted or overstated the consensus reached on the other bullet points above.

We should pick a convention and finalize this decision so people can sink their teeth into writing some code.

## Setting up editorconfig for per-project style/indentation

Also, a note for @certik and other vim users: If you install the [vim editorconfig plugin](https://github.com/editorconfig/editorconfig-vim#readme) you won't have to mess with your global preferences. It will/should use the indentation specified in  [`.editorconfig`] when working on files for stdlib, and then your default when not. [Many/most editors respect `.editorconfig` out of the box](https://editorconfig.org/#download), and plugins for others, like [Emacs](https://github.com/editorconfig/editorconfig-emacs#readme) are easily installed.

gronki:
Ugh.

wt., 31 gru 2019, 18:40 użytkownik zbeekman <notifications@github.com>
napisał:

certik:
The survey is too small and not representative. Can you please list here the number of spaces used by the top 30 or more projects listed at:

https://github.com/fortran-lang/stdlib/wiki/List-of-popular-open-source-Fortran-projects

? That would be a lot more representative what the community at large uses.

milancurcic:
Considering the development here, I change my preference to "don't care". :)

I'd also approve of allowing both 2 and 4 spaces, not mixed together within a source file.

jacobwilliams:
I vote for 4 spaces! :)

zbeekman:
I was only responding to your comment at the top of the PR and providing a rationale and evidence in support of my past decision:
> I think most people agreed to use 4 spaces instead of just 2.

I'm fine with either. Listing the number of spaces used by the top Fortran projects is not a good use of my time.

The biggest one, OpenBLAS, appears to use 3 spaces FWIW.

A final note before I completely abstain from further comment on indentation: Just because a library is big and popular doesn't mean we want to do what they did; tradition for tradition's sake alone isn't a compelling reason to do something! (If it were we'd all be writing F77 still.)

Ah, thank goodness, @jacobwilliams broke the tie!!! Haha. Yes, please god let us use 4 spaces, just make sure that `.editorconfig` is updated too.

milancurcic:
> Ugh.

@gronki Please, please if you don't have anything constructive to add to the thread, just don't post. This is a good example of the kind of communication that isn't helpful. Please read [Code of Conduct](https://github.com/fortran-lang/stdlib/blob/master/CODE_OF_CONDUCT.md) if you haven't yet.

jvdp1:
Let's change my preference from `1 or 2` to `don't care`. I can easily adapt Vim.

septcolor:
Just FWIW, here are some indentation used by the standard libraries of some other languages (which I think may be more or less related to numerical computing).

4 spaces:

- Julia
https://github.com/JuliaLang/julia/blob/master/base/array.jl
- Rust
https://github.com/rust-lang/rust/blob/master/src/libcore/array/iter.rs
- D
https://github.com/dlang/phobos/blob/master/std/array.d
- Python
(I haven't checked this but probably)

2 spaces:

- Scala
https://github.com/scala/scala/blob/v2.13.1/src/library/scala/math/Numeric.scala
- Chapel
https://github.com/chapel-lang/chapel/blob/master/modules/standard/Math.chpl
- Nim
https://github.com/nim-lang/Nim/blob/devel/lib/system.nim

(I haven't checked other languages like JS, C#, Java, etc). So the choice seems to vary. I personally use 4 spaces for distinguishing block structures more clearly, but also feel that codes with 2 spaces are more compact. (I used 3 spaces for Fortran77, which was the default of Emacs at that time...)

gronki:
> > Ugh.
> 
> @gronki Please, please if you don't have anything constructive to add to the thread, just don't post. This is a good example of the kind of communication that isn't helpful. Please read [Code of Conduct](https://github.com/fortran-lang/stdlib/blob/master/CODE_OF_CONDUCT.md) if you haven't yet.

I know the Code of Conduct, what I basically meant is that it doesn't matter in my opinion considering the amount of decisions that have to be made that do matter. If you find my comments not constructive I think it's time from my side to move away to where I can actually help.

zbeekman:
I'll open a PR to change back to 4 spaces, and fix `.editorconfig` to reflect this while I'm at it.

milancurcic:
Thanks. I was addressing this specific comment and not your other posts, many of them helpful. Clean and constructive communication is important.

certik:
I am away from my computer. Please close this PR I am retracting it. Let's spend our time constructively.

Let's submit a PR removing the requirement for 2 spaces. Or change it to 2 or 4 spaces.

Let's not lose any more time with this.

We have bigger fish to fry.

zbeekman:
Honestly I think everyone is fine with 4 at this point. I'll change the text to 2 or 4 if you prefer, but I'd rather pick one (4) that can be automated in hooks and checked in PRs.

scivision:
normally I prefer 2 spaces for Fortran projects, where the procedures are often longer than ideal. Since this is a fresh project, there may be more consciousness of procedure cyclomatic complexity. Thus 4 spaces would be less likely to result in excessive indentation->too much line continuation.

certik:
@zbeekman and I talked about this on the phone. We can either choose now, or postpone this decision for later. We both prefer to decide now, because that allows us to setup automatic formatting at the CI, and move on. So we both agreed to make a choice now, see how it works for us, and we can revisit later if needed.

Then the question is whether to choose 2 or 4 spaces today. We agreed to go with 4.

certik:
I created this PR via GitHub and it pushed a branch into the main repository. So I just created #60 from my own fork. I am going to close this PR in favor of #60.

PR #58, opened 2019-12-31
ci: add windows workflow

scivision:
It can be advantageous to have distinct workflows on a per-operating system basis,
to quickly see the scope of a failure.

This CI will fail until the quad-precision code is made conditional on a build system test (CMake >= 3.14 `check_fortran_source_runs()`) to make quad-precision optional in library and tests.
In the library, this can be done via Fortran `submodule`. In the test, it can be done via preprocessor `#ifdef usereal128` or similar introduced via CMake as a result of configure checks

### DIFF
--- a/.github/workflows/ci_windows.yml
+++ b/.github/workflows/ci_windows.yml
@@ -1,0 +1,37 @@
+name: CI_windows
+
+on: [push, pull_request]
+
+env:
+  CI: "ON"
+
+jobs:
+  Build:
+    runs-on: windows-latest
+    strategy:
+      fail-fast: false
+
+    steps:
+    - uses: actions/checkout@v1
+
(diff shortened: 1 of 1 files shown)


certik:
Thanks! You can upgrade cmake, or do whatever needs to be done to make all tests pass.

certik:
I am trying to trouble shoot this in #61. Is this a compiler bug?

scivision:
I think it is a compiler bug it's not uncommon to find such things with rarer configurations like quad precision.

certik:
I see. Why don't we disable quad precision for now. I think having Windows tests is essential. Then we can re-enable quad precision in future PRs on platforms where it works.

certik:
This was merged as part of #61. Thanks @scivision!

Issue #25, opened 2019-12-19
How should stdlib handle single, double, quadruple precision types


certik:
When we declare a real variable in our API, we have to use some variable to represent the real kind. Available options for single, double and quadruple precision that people have used in codes:

1. `real(sp)`, `real(dp)`, `real(qp)`
2. `real(real32)`, `real(real64)`, `real(real128)`
3. `real(r32)`, `real(r64)`, `real(r128)`
4. `real(float32)`, `real(float64)`, `real(float128)`
5. `real(f32)`, `real(f64)`, `real(f128)`
6. `real(wp)` (for default real)
7. `real(r4)`, `real(r8)`, `real(r16)`

I will keep appending to this list if we find some code that uses different names.

Those variables must be defined somewhere. I will use the case 1. below, for other cases we simply substitute different names. The available options:

a. `stdlib_types` module that provides `sp`, `dp`, `qp` for single, double and quadruple precision as follows:
```fortran
integer, parameter :: sp=kind(0.), &             ! single precision
                      dp=kind(0.d0), &           ! double precision
                      qp=selected_real_kind(32)  ! quadruple precision
```
The main idea behind this option is that there is a module in `stdlib` that provides the types and every other module uses it. The proposal #13 is similar to it. There are several options how the types are defined inside the module: one can define `sp` and `dp` using `selected_real_kind` also. Another alternative is to define `sp`, `dp` and `qp` using iso_fortran_env as in b. to a. The module `stdlib_types`  can be called differently also.

b. `use iso_fortran_env, only: sp=>real32, dp=>real64, qp=>real128` (if the case 2. above is used, then one does not need to rename, so it simplifies to just `use iso_fortran_env, only: real32, real64, real128`). Unlike a., this option does not introduce a new module in stdlib. One simply uses `iso_fortran_env` everywhere directly.

c. `use iso_c_binding, only: sp=>c_float, dp=>c_double, qp=>c_float128`. Unlike a., this option does not introduce a new module in stdlib.

I will keep this list updated if more options become available.

certik:
My own preference that I have been using is 1. together with a.

certik:
[...]

Issue #13, opened 2019-12-17
Proposal for a common data model module to set default KINDS


rweed:
The following (sorry about the length) is a module that sets defaults for real and integer KIND parameters (what I call a data model) but allows users to define there own or change the defaults using preprocessor defines. I think something like this along with some module procedures for checking things like storage size and interrogating machine parameters ala R1MACH from Linpack etc. should be a basic part of the library. This would allow users to build different versions with different combinations of integer and real default types. I prefer this approach over trying to provide a different routine for each possible combination of integer and real types. Again, thats a personal preference but I prefer having 4 different versions of a library over trying to to provide four different versions of a subroutine/function and providing a generic interface. Also, I think part of the coding standard for the library should forbid real(8), real*8, implicit double precision etc.

*** dataModel.F90 ***
```fortran
Module dataModel

! Define default data model precisions (kinds) for a project
! The default data model is assumed to be 32 bit integers and
! 64 bit reals

! Define intrinsic KIND parameters for Modern Fortran programs. 
! If we have a Fortran 2008 compliant version of ISO_FORTRAN_ENV, we will 
! use it; if not we will create our own versions of the equivalent Fortran 
! integer and real kinds using SELECT_INT_KIND and SELECT_REAL_KIND. We
! also define a "machine zero" function for both 32 and 64 bit Real values 

  Use ISO_FORTRAN_ENV

#ifdef HAVE_USER_DATA_MODEL
  USE USERDATAMODEL, ONLY: USER_DEFAULT_INT, USER_DEFAULT_REAL
#endif

  Implicit NONE 

#ifdef NO_F2008_ENV

! Define Integer and Real intrinsic KINDS with same names as in Fortran 2008
! ISO_FORTRAN_ENV
   
  Integer, Parameter :: INT8    = SELECTED_INT_KIND(2) 
  Integer, Parameter :: INT16   = SELECTED_INT_KIND(4) 
  Integer, Parameter :: INT32   = SELECTED_INT_KIND(9)
  Integer, Parameter :: INT64   = SELECTED_INT_KIND(18)
  Integer, Parameter :: REAL32  = SELECTED_REAL_KIND(P=6,  R=37)
  Integer, Parameter :: REAL64  = SELECTED_REAL_KIND(P=15, R=307)

#endif

#ifdef NO_REAL128
!  Integer, Parameter :: REAL128 = REAL64
#else
#ifdef NO_F2008_ENV
[...]

PR #63, opened 2020-01-02
Add stdlib_experimental_kinds.f90 and use it

certik:
This implements the latest discussion at #25.

### DIFF
--- a/src/stdlib_experimental_kinds.f90
+++ b/src/stdlib_experimental_kinds.f90
@@ -3,7 +3,7 @@ module stdlib_experimental_kinds
 !use iso_fortran_env, only: sp=>real32, dp=>real64, qp=>real128
 !use iso_fortran_env, only: int32, int64, int128
 use iso_c_binding, only: sp=>c_float, dp=>c_double, qp=>c_float128
-use iso_c_binding, only: int32=>c_int32_t, int64=>c_int64_t, int128=>c_int128_t
+use iso_c_binding, only: int8=>c_int8_t, int16=>c_int16_t, int32=>c_int32_t, int64=>c_int64_t
 implicit none
 private
 public sp, dp, qp, int32, int64, int128
--- a/src/stdlib_experimental_kinds.f90
+++ b/src/stdlib_experimental_kinds.f90
@@ -6,5 +6,5 @@ use iso_c_binding, only: sp=>c_float, dp=>c_double, qp=>c_float128
 use iso_c_binding, only: int8=>c_int8_t, int16=>c_int16_t, int32=>c_int32_t, int64=>c_int64_t
 implicit none
 private
-public sp, dp, qp, int32, int64, int128
+public sp, dp, qp, int8, int16, int32, int64
 end module
--- a/src/stdlib_experimental_kinds.f90
+++ b/src/stdlib_experimental_kinds.f90
@@ -1,9 +1,9 @@
 module stdlib_experimental_kinds
-! Instead of iso_fortran_env, we use iso_c_binding, to be compatible with C
-!use iso_fortran_env, only: sp=>real32, dp=>real64, qp=>real128
-!use iso_fortran_env, only: int32, int64, int128
-use iso_c_binding, only: sp=>c_float, dp=>c_double, qp=>c_float128
-use iso_c_binding, only: int8=>c_int8_t, int16=>c_int16_t, int32=>c_int32_t, int64=>c_int64_t
+use iso_fortran_env, only: sp=>real32, dp=>real64, qp=>real128
+use iso_fortran_env, only: int8, int16, int32, int64
+! If we decide later to use iso_fortran_env instead of iso_fortran_env:
+!use iso_c_binding, only: sp=>c_float, dp=>c_double, qp=>c_float128
+!use iso_c_binding, only: int8=>c_int8_t, int16=>c_int16_t, int32=>c_int32_t, int64=>c_int64_t
 implicit none
 private
 public sp, dp, qp, int8, int16, int32, int64
--- a/src/stdlib_experimental_kinds.f90
+++ b/src/stdlib_experimental_kinds.f90
@@ -1,0 +1,10 @@
+module stdlib_experimental_kinds
+! Instead of iso_fortran_env, we use iso_c_binding, to be compatible with C
+!use iso_fortran_env, only: sp=>real32, dp=>real64, qp=>real128
+!use iso_fortran_env, only: int32, int64, int128
+use iso_c_binding, only: sp=>c_float, dp=>c_double, qp=>c_float128
+use iso_c_binding, only: int32=>c_int32_t, int64=>c_int64_t, int128=>c_int128_t
+implicit none
+private
+public sp, dp, qp, int32, int64, int128
+end module
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -3,6 +3,7 @@ set(SRC
     stdlib_experimental_io.f90
     stdlib_experimental_error.f90
     stdlib_experimental_optval.f90
+    stdlib_experimental_kinds.f90
 )
 
 add_library(fortran_stdlib ${SRC})
--- a/src/Makefile.manual
+++ b/src/Makefile.manual
@@ -2,6 +2,7 @@ SRC = stdlib_experimental_ascii.f90 \
       stdlib_experimental_error.f90 \
       stdlib_experimental_io.f90 \
       stdlib_experimental_optval.f90 \
+      stdlib_experimental_kinds.f90 \
       f18estop.f90
 
 LIB = libstdlib.a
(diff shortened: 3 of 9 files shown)


src/stdlib_experimental_kinds.f90
```diff
@@ -0,0 +1,10 @@
+module stdlib_experimental_kinds
+! Instead of iso_fortran_env, we use iso_c_binding, to be compatible with C
+!use iso_fortran_env, only: sp=>real32, dp=>real64, qp=>real128
+!use iso_fortran_env, only: int32, int64, int128
+use iso_c_binding, only: sp=>c_float, dp=>c_double, qp=>c_float128
+use iso_c_binding, only: int32=>c_int32_t, int64=>c_int64_t, int128=>c_int128_t
```
jvdp1: Is `int128` supported by all compilers?

zbeekman: That's a good question. If C has standardized `int128` I would image that it is. But I need to consult the standard for more insight.

jvdp1: I would add:
```
use iso_c_binding, only: int8=>c_int8_t, int16=>c_int16_t, int32=>c_int32_t, int64=>c_int64_t
```

src/stdlib_experimental_kinds.f90
```diff
@@ -0,0 +1,10 @@
...
+implicit none
+private
+public sp, dp, qp, int32, int64, int128
```
jvdp1:
```
public sp, dp, qp, int8, int16, int32, int64
```

src/stdlib_experimental_kinds.f90
jvdp1: At least I don't find `c_int128_t` in `iso_c_binding or `int128` in `iso_fortran_env`

src/stdlib_experimental_kinds.f90
certik: @jvdp1 go ahead and make the suggestion in GitHub. (Select the line and click "make suggestion"). Then I can just click "commit".

src/stdlib_experimental_kinds.f90
certik: But it compiled. It might be only supported by gfortran?

jvdp1:
```suggestion
use iso_c_binding, only: int8=>c_int8_t, int16=>c_int16_t, int32=>c_int32_t, int64=>c_int64_t
```

src/stdlib_experimental_kinds.f90
jvdp1:
```suggestion
public sp, dp, qp, int8, int16, int32, int64
```

src/stdlib_experimental_kinds.f90
zbeekman: The standard says that it will compile, but be set to a special value when not supported (missing) if my memory is correct. So compilation doesn't guarantee existence of the kind. (I think the special value is 0 or a negative integer.)

zbeekman:
Looks like the Makefile needs updating.

certik:
@marshallward I think you were initially against having a `kinds` module, so I would like your feedback on this. This PR is based on the discussion in #25.

zbeekman:
> @marshallward I think you were initially against having a `kinds` module, so I would like your feedback on this. This PR is based on the discussion in #25.

I think @jacobwilliams expressed some reservation in #13 too, but I might be reading between the lines too much.

src/tests/loadtxt/test_savetxt.f90
```diff
@@ -1,5 +1,5 @@
 program test_loadtxt
```
jvdp1: Just spotted this small one:
```suggestion
program test_savetxt
```

certik:
I rebased on top of the latest master and resolved all conflicts.

certik:
I rebased again. The required Makefile changes are now minimal.

@marshallward, @jacobwilliams just a reminder that this waits for you to review it, as you might have some objections to this.

marshallward:
@certik Sorry, missed this request back when I was still travelling, I will get caught up on this tonight and give my feedback.

src/stdlib_experimental_kinds.f90
marshallward: It looks like C99 has no guaranteed 128-bit integer (`long long` is "at least 64 bits" and `intN_t` is only guaranteed up to 64).

Looking at my own include files (libc 2.30), I don't see anything for 128-bit integers except a few places under an `__ILP32__` preprocessor flag, which must have been unset when my platform-specific includes were preprocessed for my system.

marshallward: Also, I think one does need to be careful with quad precision.  On its own, it does not have a clear meaning.  There are (at least) three active definitions:

https://gcc.gnu.org/onlinedocs/gcc/Floating-Types.html

For example, on my machine, `double` and `long double` both point to `_Float64`.  My libc also defines a `_Float64x` which as best I can tell corresponds to the x86 float80 format.

marshallward: BTW I am looking at a copy of the 2018 standard now (Table 18.2 in section 18.3.1), and I don't see any reference to `c_int128_t` in there?  Only up to `c_int64_t`.  Also no reference to `c_float128`.

marshallward:
I might revise my review and suggest that the 128 formats be dropped until they can be investigated robustly at build time (e.g. CMake).

certik:
I personally prefer to import the kinds from iso_fortran_env, but it seems most others preferred C. But now with the kinds module, this is easy to change.

Regarding dp, sp and qp, it's not set in stone, we are still in experimental. I plan to do a thorough review of Fortran projects out there to get a better idea what the community uses.

Regarding the quadruple precision, I think there is only one type: https://en.m.wikipedia.org/wiki/Quadruple-precision_floating-point_format

Besides double precision there is also 80 bit. I think what you meant to say is that the C float128 might not always map to quadruple precision and might sometimes map to 80bit? If that's the case, then we should use the iso_fortran_env

certik:
Given that they currently work in master, I would suggest to keep quadruple precision, but create an issue with a todo list of things that we need to review and possibly change in experimental, before considering to move to main.

marshallward:
@certik yes, that's a better way to put it.  There is a "float128", but it's often not what one gets when one requests a quad precision number.  (Maybe my brief experience on IBM Powers is showing?)

I agree that `iso_fortran_env` might be safer in this case, assuming that the Fortran compiler does the heavy lifting here.

marshallward:
Also, given that this to be a reference library, are we sure about using `c_int128_t` and `c_float128` when they are not defined in the standard?

Appending a bit of GFortran preprocessing might be OK, though.

certik:
@zbeekman, @scivision you were for using `iso_c_binding` here. Do you have any comments to @marshallward's review?

I think `iso_fortran_env` is probably the safest. But if we can't get an agreement on this, then let's revisit this later.

As a compromise, we can stay with `iso_fortran_env`, as that's what's already in master. 

@marshallward would you be against introducing the `stdlib_kinds` module, even if we use `iso_fortran_env` inside?

marshallward:
I don't have any objection to a `stdlib_kinds` module, and pointing to `iso_fortran_env` is probably the better option, since it would presumably(?) resolve the platform-specific issues.  (It also helps that `real128` is actually in the standard!)

I apologize if it seemed like I was against the idea of providing such a module, I think it is very valuable!  I was only a bit resistant to the names (as usual), and the handling of quad types.

In the future I could see it being extended it to include kinds for more explicit types, such as `float128`, `ibm128`, and `float80`.

This situation could re-emerge with the many new architectures coming into the market and the newer formats like `float16` and `decimalN` becoming more popular.  There may be value in providing kind IDs for these explicit data types.

certik:
@marshallward thanks for the clarification. So it seems we are all in agreement about having `stdlib_kinds` module. So I modified the PR to keep using `iso_fortran_env` as in master, but now it is in the module. Then let's create another PR or an issue where we can discuss if we want to change it to use `iso_c_binding`. I actually do not care at all about this, now when we have it at just one place in a `stdlib_kinds` module.

Regarding the names of `sp`, `dp`, `qp`, this PR does not change anything either. I plan to do a wide survey soon of the open source Fortran codes and gather some data on this. 

Anyway, is anybody against merging this PR as is?

scivision:
I think it's fine to use iso_fortran_env and when/if issues arise with C interoperability, fix it then.

certik:
Ok then.  I am going to merge this, as we seem to be in agreement or not opposed to this latest version of the PR. Now we can easily change how the constants are defined in one module.

As for the naming, I created #85, so that we do not forget.

zbeekman:
Yes, given the points raised, I think the best thing for now is to use `iso_fortran_env` thanks for the careful discussion and consideration everyone, especially @marshallward!

Issue #40, opened 2019-12-22
Pretty printing of matrices (and multidimensional arrays)


certik:
Currently the standard Fortran's `print *, A` prints a 2D array `A` as a 1D list of numbers. Rather, I would like `stdlib` to have a function `print_array` (we can discuss a better name) that would print the array as NumPy:
```
>>> numpy.arange(10000).reshape(250,40)
array([[   0,    1,    2, ...,   37,   38,   39],
       [  40,   41,   42, ...,   77,   78,   79],
       [  80,   81,   82, ...,  117,  118,  119],
       ..., 
       [9880, 9881, 9882, ..., 9917, 9918, 9919],
       [9920, 9921, 9922, ..., 9957, 9958, 9959],
       [9960, 9961, 9962, ..., 9997, 9998, 9999]])
```
or Julia:
```
julia> B = [1 2; 3 4; 5 6; 7 8; 9 10]
5×2 Array{Int64,2}:
 1   2
 3   4
 5   6
 7   8
 9  10
```
Julia can also use nice unicode characters for ... and vertical ... if the array is too large.

Then we should use this function at https://github.com/fortran-lang/stdlib/blob/ae5591f5b6d9f1df644c6246faf099056276bf0a/src/tests/loadtxt/test_loadtxt.f90#L21 and other places.

Then compilers can perhaps optionally use such `print_array` as default in the Fortran's language `print` statement.

jvdp1:
My preference is for a format like Julia 's one. It would be also a similar format as savetxt.

ivan-pi:
One prior art - Algorithm 892: DISPMODULE, a Fortran 95 module for pretty-printing matrices
https://dl.acm.org/citation.cfm?id=1486531

The code can be downloaded from netlib: http://netlib.org/toms/892.zip

jvdp1:
> One prior art - Algorithm 892: DISPMODULE, a Fortran 95 module for pretty-printing matrices
> https://dl.acm.org/citation.cfm?id=1486531
> 
> The code can be downloaded from netlib: http://netlib.org/toms/892.zip

Thank you for mentioning this librabry. It seems to be quite complete and flexible. Could we use it (with some modernisations) in `stdlib` (license?)?

ivan-pi:
Code published in TOMS is under the [ACM Software License Agreement](https://www.acm.org/publications/policies/software-copyright-notice) which allows usage for non-commercial purposes. I think that it is not compatible with what we want to achieve here.

[...]

PR #65, opened 2020-01-02
Split the loadtxt qp tests and skip them on Win

certik:


certik:
The `print_array` is duplicated a lot, but that will eventually be fixed by #40.

src/tests/loadtxt/test_loadtxt.f90
```diff
@@ -1,11 +1,10 @@
 program test_loadtxt
-use iso_fortran_env, only: sp=>real32, dp=>real64 ,qp=>real128
+use iso_fortran_env, only: sp=>real32, dp=>real64, qp=>real128
```
jvdp1:
```suggestion
use iso_fortran_env, only: sp=>real32, dp=>real64!, qp=>real128
```
I would comment `qp=>real128` to be in line with the rest of the program

src/tests/loadtxt/test_loadtxt_qp.f90
```diff
@@ -0,0 +1,38 @@
+program test_loadtxt_qp
+use iso_fortran_env, only: sp=>real32, dp=>real64, qp=>real128
```
jvdp1: To avoid confusion:

```suggestion
use iso_fortran_env, only: qp=>real128
```

src/tests/loadtxt/test_loadtxt_qp.f90
```diff
@@ -0,0 +1,38 @@
...
+
+ select type(a)
+  type is(real(sp))
+   do i = 1, size(a, 1)
+    print *, a(i, :)
+   end do
+  type is(real(dp))
+   do i = 1, size(a, 1)
+    print *, a(i, :)
+   end do
```
jvdp1: I would deleted `type is (real(sp))` and `type is(real(dp))` to avoid confusion with the name of the program.

src/tests/loadtxt/test_savetxt_qp.f90
```diff
@@ -0,0 +1,45 @@
+program test_loadtxt
```
jvdp1:
```suggestion
program test_loadtxt_qp
```

certik: Actually, it should be `test_savetxt_qp`! I need to fix it.

src/tests/loadtxt/test_loadtxt_qp.f90
certik: Done in 84975608064f4fdaa1160ee79bc45508b4d771e9.

src/tests/loadtxt/test_savetxt_qp.f90
certik: Fixed in 84975608064f4fdaa1160ee79bc45508b4d771e9.

certik:
> It seems that `test_save.f90` has not been modified for `qp`.

I think it is now. Let me know.

certik:
Thanks!

PR #70, opened 2020-01-02
Refactor the manual Makefiles

certik:
* Compile the quadruple precision loadtxt
* Implement "make test" to run tests and run them at the CI
* Create a static libstdlib.a library and use it in tests
* Use more modern syntax and simplify
* Pass in the FC/FCFLAGS variables automatically
* Consolidate stdlib tests targets

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -71,3 +71,5 @@ jobs:
       if: contains(matrix.os, 'ubuntu') && contains(matrix.gcc_v, '9')
       run: |
         make -f Makefile.manual
+        make -f Makefile.manual test
+        make -f Makefile.manual clean
--- a/Makefile.manual
+++ b/Makefile.manual
@@ -1,17 +1,21 @@
 # Fortran stdlib Makefile
 
 FC = gfortran
-FCFLAGS=-O0
+FCFLAGS = -Wall -Wextra -Wimplicit-interface -fPIC -g -fcheck=all
 
-.PHONY: all clean
+export FC
+export FCFLAGS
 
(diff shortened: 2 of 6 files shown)


certik:
Further simplifications can be made how the tests are handled to reduce code duplication in individual test directories (`loadtxt` and `ascii`), but we can do that later. The current design in this PR greatly simplifies the necessary changes to manual Makefiles required as part of #63 and #53.

certik:
@gronki, as another proponent of manual Makefiles (per your comments at https://github.com/j3-fortran/fortran_proposals/issues/104#issuecomment-563873795), I was wondering if you could please review this PR. I tried to simplify the Makefiles a lot.

certik:
Thanks @milancurcic for the review. I am rebasing on top of the latest master.

gronki:
@certik I only would suggest using ``FFLAGS`` for Fortran flags. I know there was an argument that ``FCFLAGS`` corresponds to f90+, but I think we should assume there is only one Fortran, and ``FFLAGS`` is clearly stated in [gnu make manual](https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html) as Fortran flags and followed by most Linux operating systems packages (such as rpmbuild that I use). I also think that ``.SUFFIXES`` is obsolete and not needed with new syntax using ``%``. The rest is perfect!

certik:
@gronki thanks for the review! I fixed both things in 0d80b89e4ecf31e89bb1fa039e8c7e47e51ebb96 and 340051548c74ba9d78175722cdd8ca640a65e0a4.

I am going to merge it now, and we can further improve upon this later.

Issue #62, opened 2020-01-02
Facilitate default values of optional arguments


nshaffer:
An annoyance with optional arguments is handling their default values. This usually looks something like:

    function mylog(x, base) result(y)
        real, intent(in) :: x
        real, intent(in), optional :: base
        real :: y

        real :: base_
        
        base_ = 10.0
        if (present(base)) base_ = base
        
        y = log(x)/log(base_)
    end function mylog

I propose to introduce a module that exports a generic function `default` which will in many cases allow for the elimination of the local copy of the optional argument. The above example could be rewritten

    function mylog(x, base) result(y)
        use default_values, only: default
        real, intent(in) :: x
        real, intent(in), optional :: base
        real :: y
        
        y = log(x)/log(default(10.0, base))
    end function mylog

This is a convenience, but it's incredibly handy. The module is very simple to write. See, e.g., [this CLF post by Beliavsky](https://groups.google.com/d/msg/comp.lang.fortran/OzSX0GC1pWU/GAlkcjddCQAJ), where I first learned of this trick. I put this in all my serious codes, as it removes a lot of the tedium and error potential with optional arguments.

I can easily spin up a illustrative PR that can be fleshed out once we've decided how we're going to automate generic interfaces.

certik:
I like this a lot. I would flip the order of arguments:

default(base, 10.0)

Sort of like Python's dictionary method get. Regarding the name, here are a few options:

default
argget
optarg
arg_default
opt_default

I like the name default the most so far.

marshallward:
I do like this, but the more frustrating issue for me is that I often need to define two variables in the same scope with identical purpose, in this case `base_` and `base`.

This function would let you replace `base_` with `default(base, 10.0)`, and is very effective if `base` only appears once or twice.  But if you may have to do this many times then it may not actually be what you want.

I feel constrained because the optional input is the one which becomes part of the API, and therefore should be the most readable.  But it is often the internal variable which will appear the most throughout the function, which can make the function less readable.

But this is still an improvement for lots of cases.  I agree with @certik that `default` is too general of a name to use here.

nshaffer:
The semantics of optional arguments make the "natural" order a little tricky. If we want the fallback value to come second, callers will have to explicitly make it a keyword arg. That is, consider the two implementations:

    default_first(to, x) result(y)
        ! Return first arg if second is not present
        !     default(1, 2) == 2
        !     default(1) == 1 ("x" not present, so we get the fallback value)
[...]

PR #73, opened 2020-01-03
Preliminary implementation of default values

nshaffer:
In support of #62
 
- New module "default_m" which exports the generic function name "default"
- New entries in src/CMakeLists.txt and src/Makefile.manual

I've implemented a generic interface for `real(sp)`, `real(dp)`, `real(qp)`, `int16`, `int32`, `int64`, `logical`, and `character`. I figure this is enough for a first pass. Complex types and discriminating between ASCII and UCS characters can come later. Probably no need to do specific logical kinds.

What's missing right now is tests. I don't know how to use CTest, but I would like to learn if anyone can link me a decent tutorial.


src/stdlib_experimental_default.f90
```diff
@@ -0,0 +1,142 @@
+module default_m
```
jvdp1:
```suggestion
module stdlib_experimental_default
```

src/stdlib_experimental_default.f90
```diff
@@ -0,0 +1,142 @@
...
+  !! to subprograms. If `x` is an `optional` parameter of a
+  !! subprogram, then the expression `default(x, y)` inside that
+  !! subprogram evaluates to `x` if it is present, otherwise `y`.
+  !!
+  !! It is an error to call `default` with a single actual argument.
+  !!
+  !! For additional clarity, `default` be called with keyword argument
+  !! for the fallback value, e.g., `default(x, to=1.0)`.
+  !!
+  use iso_fortran_env, only: sp => real32, dp => real64, qp => real128, int16, int32, int64
```
jvdp1: At some points, it should use `stdlib_experimental_kinds` proposed in #63 . But I would say it is fine for now.

src/stdlib_experimental_default.f90
```diff
@@ -0,0 +1,142 @@
...
+     module procedure default_dp
+     module procedure default_qp
+     module procedure default_int16
+     module procedure default_int32
+     module procedure default_int64
+     module procedure default_logical
+     module procedure default_character
+     ! TODO: complex kinds
+     ! TODO: differentiate ascii & ucs char kinds
+  end interface default
```
jvdp1: These lines show again the importance to find a solution to #35 soon, to avoid to go one more time across all files when #35 is solved.

src/stdlib_experimental_default.f90
```diff
@@ -0,0 +1,142 @@
...
+
+  private
+  public :: default
+  
+
+  interface default
+     module procedure default_sp
+     module procedure default_dp
+     module procedure default_qp
+     module procedure default_int16
```
jvdp1: `int8` should be added too.

src/stdlib_experimental_default.f90
```diff
@@ -0,0 +1,142 @@
...
+    character(len=:), allocatable :: y
+
+    if (present(x)) then
+       y = x
+    else
+       y = to
+    end if
+  end function default_character  
+  
+end module default_m
```
jvdp1:
```suggestion
end module stdlib_experimental_default
```

certik:
Thanks @nshaffer. I resolved the conflict with master and committed to your branch. Make sure you pull it before adding more commits.

Also, I recommend people to submit PRs from a different branch than their master, so that they can use their master to track the upstream master (to rebase, merge, etc.).

certik:
Regarding the test, try to copy let's say the `test_ascii` test and its CMake setup. It should be straightforward. If you have any questions, let us know.

src/stdlib_experimental_default.f90
nshaffer: Done, cheers.

nshaffer:
Thanks all for the comments. Seems like folks in #62 prefer the name "optval" so I converted to that everywhere and renamed the fallback argument "default", so that you can write either `optval(x, default=1.0)` if you want to be explicit. Also wrote tests for all currently implemented types.

certik:
Thank you! This is great. I'll review later tonight or tomorrow.

certik:
Thanks @nshaffer for the work!

certik:
Here is the first usage of `optval`: https://github.com/fortran-lang/stdlib/pull/71/commits/a8416a10a1b2db974dfa6192fe486c8d16b24e8c

nshaffer:
Cheers! 

It occurs to me that `optval` should be `pure elemental` for the intrinsic types. Shall I tack that on here or open a new PR? (I've added a dev branch to my fork now, if that makes any difference.)

certik:
Yes, send another PR please against master, from a new branch.

PR #74, opened 2020-01-03
Further Makefile improvements

certik:
This further reduces the number of lines to maintain in Makefiles. The tests' Makefiles are now very short. In general we now simply list all the source files, similar to CMake.


certik:
@gronki, here is further simplification of the Makefiles. Let me know what you think.

certik:
Thanks for the review!

PR #77, opened 2020-01-04
Modification of the function parse_mode in stdlib_experimental_io

jvdp1:
I modified the function `parse_mode` in `stdlib_experimental_io.f90` to be more flexible.
I also added a new test file `test_parse_mode.f90` and removed associated tests from `test_open` (in case we would use a public `parse_mode` only for tests).
I added several cases for testing `parse_mode`.

### DIFF
--- a/src/stdlib_experimental_io.f90
+++ b/src/stdlib_experimental_io.f90
@@ -356,26 +356,31 @@ end function
 character(3) function parse_mode(mode) result(mode_)
 character(*), intent(in) :: mode
 
+integer::i
+character(:),allocatable::a
+
 mode_ = 'r t'
+
 if (len_trim(mode) == 0) return
-mode_(1:1) = mode(1:1)
+a=trim(adjustl(mode))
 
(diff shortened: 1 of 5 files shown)


certik:
+1 from me to merge

certik:
@jvdp1 thanks, it looks great! The only additional improvement that I can think of is if something calls it like this `open(filename, "rr")`. In Python this gives an error if "r" and other modes are used more than once. Also some other combinations are probably not allowed, like "rw", or "rtb". But as it is now, it's already very useful.

jvdp1:
> @jvdp1 thanks, it looks great! The only additional improvement that I can think of is if something calls it like this `open(filename, "rr")`. In Python this gives an error if "r" and other modes are used more than once. Also some other combinations are probably not allowed, like "rw", or "rtb". But as it is now, it's already very useful.

Things like "rr" is not a problem, while things like "rw" are a problem. I will have a look to modify `parse_mode` to give an error when these cases are provided.

PR #78, opened 2020-01-05
CMake: introduce ADDTEST macro

certik:
This simplifies the CMake code a lot.

@zbeekman let me know if this goes in the direction that you prefer with CMake.


scivision:
That seems like a nice macro, particularly since so many tests have that pattern

certik:
Thanks for the review!

zbeekman:
@certik nice one! Yes, this is exactly the sort of thing I would have done myself. I'm not 100% sure about having it as a macro instead of a function, but we can iterate on that idea later. Great starting point!

certik:
@zbeekman go ahead and create a function. I didn't even know you can create functions in CMake.

PR #79, opened 2020-01-05
add missing implicit none, set compiler flags to force implicit none

scivision:
also set minimum CMake version to 3.14 per prior discussion and add example check requiring CMake 3.14: real128

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,12 +1,25 @@
-cmake_minimum_required(VERSION 3.5.0)
+cmake_minimum_required(VERSION 3.14.0)
 project(stdlib Fortran)
 enable_testing()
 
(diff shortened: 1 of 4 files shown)


certik:
I've never used `-fimplicit-none`. Should that be recommended to be used by default for all modern Fortran projects?

scivision:
yes `-fimplicit-none` was available at least back to Gfortran 4.1 (didn't try looking further back) and I don't think there are any drawbacks.

certik:
@scivision cool. This is relevant to: https://github.com/j3-fortran/fortran_proposals/issues/90.

certik:
Thanks for the improvements!

zbeekman:
Thanks all!

PR #80, opened 2020-01-05
use "select case" for clarity/brevity

scivision:
using `select case` here is clearer since the condition doesn't need to include spaces, and briefer

also add "pure" where suitable

### DIFF
--- a/src/stdlib_experimental_io.f90
+++ b/src/stdlib_experimental_io.f90
@@ -300,51 +300,53 @@ character(:),allocatable :: action_, position_, status_, access_, form_
 
 mode_ = parse_mode(optval(mode, ""))
 
-if (mode_(1:2) == 'r ') then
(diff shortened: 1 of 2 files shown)


src/stdlib_experimental_io.f90
```diff
@@ -300,51 +300,53 @@ integer function open(filename, mode) result(u)
 
 mode_ = parse_mode(optval(mode, ""))
 
-if (mode_(1:2) == 'r ') then
+select case (mode_(1:2))
```
jvdp1: It looks cleaner with `select case`.
However, I usually avoid to use `select case` with `character` since I got memory issues (leaks) with Intel Fortran compiler. Any of you observed such issues?

certik: If there are leaks, then we'll revert it.

PR #82, opened 2020-01-05
Addition of checks and use of stdlib_experimental_ascii

jvdp1:
Additions/changes:
- some checks in `parse_mode` to return errors for modes such as `rr`, `rw`, `r++t`, `rbt`. (Note: I couldn't add these tests in CMake, or it should be one per file)
- the function `open` can return the `iostat` if requested
- (@certik) since #49 has been closed, `whitechar` is now replaced by `is_blank`

@scivision : Because I added checks in the function `parse_mode`, it uses again `if` statements instead of `select case`. Good for you?

### DIFF
--- a/src/stdlib_experimental_io.f90
+++ b/src/stdlib_experimental_io.f90
@@ -287,7 +287,7 @@ character(*), intent(in), optional :: mode
 integer, intent(out), optional :: io
 
 integer :: io_
-character(3):: mode_
+character(3) :: mode_
 character(:),allocatable :: action_, position_, status_, access_, form_
 
 
--- a/src/stdlib_experimental_io.f90
+++ b/src/stdlib_experimental_io.f90
@@ -232,16 +232,16 @@ integer function number_of_columns(s)
 
  integer :: ios
  character :: c
- logical :: lastwhite
+ logical :: lastblank
 
  rewind(s)
  number_of_columns = 0
- lastwhite = .true.
+ lastblank = .true.
  do
     read(s, '(a)', advance='no', iostat=ios) c
     if (ios /= 0) exit
-    if (lastwhite .and. .not. is_blank(c)) number_of_columns = number_of_columns + 1
-    lastwhite = is_blank(c)
+    if (lastblank .and. .not. is_blank(c)) number_of_columns = number_of_columns + 1
+    lastblank = is_blank(c)
  end do
  rewind(s)
 
--- a/src/stdlib_experimental_io.f90
+++ b/src/stdlib_experimental_io.f90
@@ -266,7 +266,7 @@ integer function number_of_rows_numeric(s)
 
 end function
 
-integer function open(filename, mode, io) result(u)
+integer function open(filename, mode, iostat) result(u)
 ! Open a file
 !
 ! To open a file to read:
@@ -284,7 +284,7 @@ integer function open(filename, mode, io) result(u)
 
 character(*), intent(in) :: filename
 character(*), intent(in), optional :: mode
-integer, intent(out), optional :: io
+integer, intent(out), optional :: iostat
 
 integer :: io_
 character(3) :: mode_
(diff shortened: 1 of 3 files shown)


scivision:
yes I think it's fine/necessary to use `if` here with the more complex conditions now present

src/stdlib_experimental_io.f90
```diff
@@ -239,8 +240,8 @@ integer function number_of_columns(s)
  do
     read(s, '(a)', advance='no', iostat=ios) c
     if (ios /= 0) exit
-    if (lastwhite .and. .not. whitechar(c)) number_of_columns = number_of_columns + 1
-    lastwhite = whitechar(c)
+    if (lastwhite .and. .not. is_blank(c)) number_of_columns = number_of_columns + 1
+    lastwhite = is_blank(c)
```
ivan-pi: Would it be perhaps more meaningful to rename `lastwhite` as `lastblank`?

jvdp1: > Would it be perhaps more meaningful to rename `lastwhite` as `lastblank`?

@ivan-pi Thanks for the suggestion. I just pushed your proposition.

src/stdlib_experimental_io.f90
```diff
@@ -265,17 +266,7 @@ integer function number_of_rows_numeric(s)
...
-character, intent(in) :: char
-if (iachar(char) == 32 .or. iachar(char) == 9) then
-    whitechar = .true.
-else
-    whitechar = .false.
-end if
-end function
-
-integer function open(filename, mode) result(u)
+integer function open(filename, mode, io) result(u)
```
certik: Should the `io` be called `iostat`?

src/stdlib_experimental_io.f90
```diff
@@ -351,34 +344,44 @@ integer function open(filename, mode) result(u)
 open(newunit=u, file=filename, &
      action = action_, position = position_, status = status_, &
      access = access_, form = form_, &
-     iostat = io)
+     iostat = io_)
+
+if(present(io))io=io_
```
certik: If it is not present, then we now need to give an error. For the reason, I would probably let the compiler generate the proper error message, and write this like this:
```fortran
if (present(io)) then
    open(newunit=u, file=filename, &
        action = action_, position = position_, status = status_, &
        access = access_, form = form_, &
        iostat = io)
else
    open(newunit=u, file=filename, &
        action = action_, position = position_, status = status_, &
        access = access_, form = form_)
end if
```

src/tests/io/test_open.f90
```diff
@@ -38,21 +47,25 @@ program test_open
 filename = get_outpath() // "/io_open.stream"
 
 ! Test mode "w"
-u = open(filename, "wb")
+u = open(filename, "wb", io)
+call assert(io == 0)
```
certik: I would leave these tests without `io` present, thus letting the built-in `open` to generate an error.

The reason for that is that this is how mostly the `open` function will be used. So that's what needs to be tested the most.

In addition, let's add tests where `io` is also tested (both zero and non-zero). These tests (at least initially) do not have to be so thorough.

certik:
Thank you @jvdp1 for doing this, it looks really good. I left some comments regarding the API and testing.

src/stdlib_experimental_io.f90
jvdp1: I usually don't like to use official names. However, in this case using `iostat` makes sense.

jvdp1:
> Thank you @jvdp1 for doing this, it looks really good. I left some comments regarding the API and testing.

Thanks for the review.

certik:
In your last commit, in the tests I think you forgot to add io in the 3rd test from last. (I am on my phone, so I can't comment there directly.)

jvdp1:
> In your last commit, in the tests I think you forgot to add io in the 3rd test from last. (I am on my phone, so I can't comment there directly.)

Resolved!

certik:
Perfect! I think it looks good now. +1 to merge. Go ahead and merge it if you want.

PR #83, opened 2020-01-05
Make real128 optional

scivision:
this same concept is readily extended to other types. Based on the current architecture, this seemed to be the most expedient way to make a kind optional.

The only thing lacking is the Makefiles--how would you prefer to handle having some files .F90 and some .f90. One way would be to just make *all* the files .F90 as there is no harm in preprocessing every file, and eventually all the files may have preprocessing anyway.


certik:
I merged the kinds module, I didn't see you sent this. It will have to be rebased, sorry about that.

The fact that .f90 are not automatically preprocessed unfortunately means that we probably have to change all file extensions to .F90.

However, if there is a way to avoid the preprocessor, I think we should. In particular I think this PR is needed to make it work with Flang, correct?

What does Flang segfault on exactly? Can the qp be imported in the use statement? But it segfaults when it is used?

scivision:
Both PGI and Flang are broken in multiple ways for Fortran 2003. It might be best to just wait for them to fix themselves. However if there's any other non-universal kind we wish to add, it would currently be done perhaps like this, unless we refactored. E.g. if someone wants `real16` or something. 

So while this PR may be unnecessary as is, something like this would need to be done to accomodate kinds not available in all systems. Or a refactor.

scivision:
So this PR fixes only part of the things wrong with Flang and PGI. The rest are due to improper implementation of Fortran 2003 in Flang/PGI.

Issue #88, opened 2020-01-06
test_savetxt_qp fails with ifort


pdebuyl:
Using "ifort (IFORT) 18.0.1 20171018" on linux x86_64, test_savetxt_qp fails

The routine qsavetxt writes one number per line, which is allowed by the star format.

I propose a fix in https://github.com/pdebuyl/stdlib/tree/qsavetxt_format_string , tested with said ifort and with gfortran 6.4.0 and 8.3.0.

I have not opened a PR yet, as I don't know if ifort compatibility is considered necessary at this point. I would support keeping stdlib working with several compilers early on.

jvdp1:
Is this problem related to #81? Maybe you could implement the same strategy as in #81 to avoid `qp` with Intel compilers and open a new PR.

However, I think your fix could be also useful for very large arrays (to avoid multiple lines for one row of an array; I didn't test it yet). However, this implies a fixed format (e.g., 'f40.34`), and the same should be done for `sp` and `dp`. 
Another option would be to discuss the API of `loadtxt` and `savetxt` to allow the user to specify its own format. See #16 ( [comment](https://github.com/fortran-lang/stdlib/issues/16#issuecomment-568024782) ) and #14 for some discussions.

pdebuyl:
It is orthogonal to #81 I believe. ifort supports a qp type, it is only the write operation that does not output a "ncol x nrow" format.

Issue #16 is of course related but consider extensions of loadtxt and savetxt. Here, I just fix the currrent version :-)

I write a format string to avoid relying on a fixed format. The current format string can hold up to 999999 for the number of columns. At some point users should use HDF5 or another solution anyway.

certik:
Can you please submit a PR? Compatibility with ifort is essential.

scivision:
This would be addressed by https://github.com/fortran-lang/stdlib/pull/93

pdebuyl:
@scivision ifort supports qp, it is only the * formatting that fails to generate colummar data files.

pdebuyl:
It is fixed actually, so I close the issue.

certik:
This was fixed by #89.

PR #89, opened 2020-01-06
Use explicit formatting in qsavetxt

pdebuyl:
ifort does not save 2d arrays of "real(kind=qp)" in a rectangular fashion with "*" formatting.


certik:
Fixes #88.

PR #90, opened 2020-01-06
Use access = "stream" by default

certik:
See the discussion at #86.

### DIFF
--- a/src/stdlib_experimental_io.f90
+++ b/src/stdlib_experimental_io.f90
@@ -335,14 +335,14 @@ end select
 
 select case (mode_(3:3))
 case('t')
-    access_='sequential'
     form_='formatted'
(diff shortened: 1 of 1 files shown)


src/stdlib_experimental_io.f90
```diff
@@ -335,15 +335,15 @@ integer function open(filename, mode, iostat) result(u)
 
 select case (mode_(3:3))
 case('t')
-    access_='sequential'
     form_='formatted'
 case('b', 's')
```
jvdp1:
```suggestion
case('b')
```

certik:
(One builder is stuck, but it should pass, so I merged it.)

zbeekman:
LGTM! I was going to comment on this or suggest this.

PR #91, opened 2020-01-06
Addition of access in the API of open function

jvdp1:
Here is a proposition for allowing different accesses in the API of `open`.
The default access is `stream` for both formatted and unformatted files.
This is based on discussions in #86 and in #14, and needs further discussion.

### DIFF
--- a/src/stdlib_experimental_io.f90
+++ b/src/stdlib_experimental_io.f90
@@ -347,8 +347,8 @@ access_ = 'stream'
 
 if (present(access)) then
     select case (trim(adjustl(access)))
-    case('direct')
-        access_ = 'direct'
+!    case('direct')
+!        access_ = 'direct'
     case('sequential')
         access_ = 'sequential'
     case('stream')
--- a/src/stdlib_experimental_io.f90
+++ b/src/stdlib_experimental_io.f90
@@ -269,7 +269,7 @@ integer function number_of_rows_numeric(s)
 
 end function
 
-integer function open(filename, mode, iostat) result(u)
+integer function open(filename, mode, iostat, access) result(u)
 ! Open a file
 !
 ! To open a file to read:
@@ -287,6 +287,7 @@ integer function open(filename, mode, iostat) result(u)
 
 character(*), intent(in) :: filename
 character(*), intent(in), optional :: mode
+character(*), intent(in), optional :: access
 integer, intent(out), optional :: iostat
 
 integer :: io_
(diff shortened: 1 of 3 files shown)


jvdp1:
Thinking a bit: we might not be able to support `direct` access with our `open` function because the `open` statement must contain the `recl` argument, in addition to `access = direct`. 

So the `direct` access must most likely be deleted from this PR. Or we need an additional argument, but then our simple `open` function is becoming complex.

milancurcic:
> So the direct access must most likely be deleted from this PR. Or we need an additional argument, but then our simple open function is becoming complex.

Yep, otherwise soon we'd have the same old `open`, but now as a function. I don't think it's what we're looking for :)

milancurcic:
So again, please please let's discuss API changes and additions in #14. It's very easy to just approve  quick and dirty PRs because they look good on the surface, but over time the API just get piled up on. Let's discuss and plan. What do we want to accomplish, why, and how.

jvdp1:
> Yep, otherwise soon we'd have the same old open, but now as a function. I don't think it's what we're looking for :)
> 
> So again, please please let's discuss API changes and additions in #14. It's very easy to just approve quick and dirty PRs because they look good on the surface, but over time the API just get piled up on. Let's discuss and plan. What do we want to accomplish, why, and how.

I agree to both comments. If it is to get something like the `open` statement, then this function is not needed. IMHO the current `open` function can be tested and used as is.

The API needs more discussion if we want to extend it. It is also why I submitted this PR as a draft. So we can based our discussion on some codes.

certik:
I agree we need a plan. All I am saying is that the quality of discussion is much higher if we can discuss an actual code change, like this one. So thanks @jvdp1 for submitting a PR with this.

I would not do direct for the reasons discussed above.

In that light, perhaps it's not worth doing "sequential" at all. The main motivation would be that current codes could start using our `open` for more things. But as documented [here](https://github.com/fortran-lang/stdlib/wiki/Usage-of-%22open%22), some of the codes use "direct". So it seems that such usages would just use the old built-in `open`. And stdlib's `open` would just use `stream`.

zbeekman:
I do think that it can be easier and more natural to discuss technical details over actual code like a PR, but I haven't been keeping up with the details.

I am a strong proponent for using `stream` access wherever possible to be more interoperable, create smaller files, etc. If my memory serves me correctly, `stream` isn't mutually exclusive with `direct` access; A file written using one method may be read with the other subject to a few constraints and assumptions. Neither one have the pesky record length indicators in them, and direct access just indexes record*record_length bytes into the file. (I've spent a lot of time with a hex editor examining Fortran binary IO, but this was a while ago and I'm not 100% sure if the standard guarantees that compilers can't make up their own approach for some of this stuff.)

milancurcic:
> and direct access just indexes record*record_length bytes into the file

Except that it's not bytes on all compilers. With ifort you have to say `-assume byterecl`, see example here: https://gist.github.com/milancurcic/6590d74fbbd3e765832b823474623a27. Standard says it's system dependent.

But yes, your main point is, and I agree, we can read unformatted files written with `direct` access, using `stream` access just fine.

PR #92, opened 2020-01-06
ci ctest enhancements

scivision:
* broken tests sometimes hang forever, and that's CTest's default, so make CTest have a finite timeout by default
* don't rely on shell-specific `||` to catch test failures, use a platform-agnostic `if:`
* invoke `ctest` directly to allow easier option access

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -7,6 +7,7 @@ env:
   CMAKE_BUILD_PARALLEL_LEVEL: "2" # 2 cores on each GHA VM, enable parallel builds
   CTEST_OUTPUT_ON_FAILURE: "ON" # This way we don't need a flag to ctest
   CTEST_PARALLEL_LEVEL: "2"
+  CTEST_TIME_TIMEOUT: "5"  # some failures hang forever
   HOMEBREW_NO_ANALYTICS: "ON" # Make Homebrew installation a little quicker
   HOMEBREW_NO_AUTO_UPDATE: "ON"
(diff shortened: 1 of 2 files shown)


zbeekman:
I think this is a sufficiently straightforward/uncontroversial and necessary PR, so I'm going to merge this.

As an aside, @certik I took a look at windows CI and integrating it into the main CI build matrix, but it's not immediately clear to me what the best way to get multiple/different versions of GFortran installed is. It looks like the windows CI has mingw64 by default which ships a GCC/GFortran. It also has the Chocolatey package manager, which can provide MSYS2, which gives a route to install multiple different GCCs on windows. However, I'm concerned that installing MSYS2, and then GCC may have quite a bit of overhead associated with it, and may also not map particularly well into the build matrix. What do you @certik and @scivision think? Should we test multiple GCCs on Windows? Should we experiment with MSYS2? Or should we leave things as they are now?

scivision:
I tried MSYS2 Github Action a couple months ago and it didn't work, it was in development. Haven't tried since.
I think a lot of Homebrew/Chocolately users would be just using the latest release? 
CI for Meson itself just uses one MSYS and Cygwin version, while testing many versions of Visual Studio.
Are most of the users with old GCC versions on Linux PC/HPC?

My thought at this time would be to check across GCC >= 6 stable versions on Linux, but just the default latest version on MacOS and Windows. Unless/until specific issues are noted.

certik:
I agree with @scivision to only test the latest GFortran on Windows. (And even on macOS.) 

@zbeekman in general, I think we can do anything as long as the tests run quick (as they do now).

PR #94, opened 2020-01-06
Document workflow based on the discussion in #5

certik:


nncarlson:
I think item 5 is an absolute must, but I frankly think that item 4 should be accompanied by at least a draft of the specification. That is the API after all.

certik:
@nncarlson I like your idea that already at step 4. there should be a draft of the specification.

What we can do is to write the specification for `open` to show how it could be done, and then require it for each new functionality before it gets merged.

milancurcic:
Thanks for opening this PR, and I agree we should require a spec at step 4. I skimmed through and like what I see so far, but will give detailed read-through and suggestions later tonight.

certik:
I updated the document to include the specification requirement already in step 3.

certik:
Mostly agree. Instead of stabilization, why not to call it standardization? And experimental vs standard.

milancurcic:
I think "standard" may be confusing, especially if we use the word "Standardize". It sounds a bit too close to "become part of the Fortran Standard". But that's not what we mean because we hope that some of the functions eventually transition from stdlib to the Standard. That would indeed be to "Standardize".

Now, if we mean "standard" as "accepted as part of the standard library", then it implies that experimental is somehow non-standard, or not part of stdlib, which I think it is.

I think it's just inconvenient jargon that there is both a standard and stdlib but they mean different things. What we're really doing here is transition from what everybody else calls "development" -> "stable".

milancurcic:
I think "Standardize" should be Step 6, 5-10 years in the future :)

certik:
How would you describe our "stable" section of stdlib?
I think  we do more than just a regular stable branch from other projects.

certik:
For example we will require specification document, which other projects do not.

nncarlson:
How about "release"? It's been in experimental and now blessed for broad use.

milancurcic:
I'd describe a feature in stable or main as:

* Has spent some time in experimental
* Has API that is unlikely to change in the future
* Has a specification document
* After getting traction is external projects, is a candidate for proposal to Fortran standard committee

milancurcic:
I like release also.

certik:
I am going to go ahead and merge this, as it is good enough, so that we can move on. Let's submit further PRs if somebody has any ideas for improvement.

PR #96, opened 2020-01-07
Make optval pure or pure elemental where possible

nshaffer:
Exactly what is says on the tin. 
- real, integer, and logical versions of `optval` are now `pure elemental`
- character version is now `pure` (no elemental b/c the return value is allocatable)
- new tests for arrays of real, integer, and logical 

(sorry in advance for the whitespace diffs, forgot to M-x delete-trailing-whitespace the first time around)


certik:
I resolved the conflicts with master, it turns out it was already made `pure` in master, but not `elemental`.

Issue #108, opened 2020-01-13
Workarounds for NAG


certik:
For NAG 6.2:
1. Does not seem to support submodules.
```diff
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 72b3d25..6a11a43 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -9,12 +9,6 @@ set(SRC
 
 add_library(fortran_stdlib ${SRC})
 
-if(f18errorstop)
-  target_sources(fortran_stdlib PRIVATE f18estop.f90)
-else()
-  target_sources(fortran_stdlib PRIVATE f08estop.f90)
-endif()
-
 add_subdirectory(tests)
 
 install(TARGETS fortran_stdlib
diff --git a/src/stdlib_experimental_error.f90 b/src/stdlib_experimental_error.f90
index 3d932d6..b35083b 100644
--- a/src/stdlib_experimental_error.f90
+++ b/src/stdlib_experimental_error.f90
@@ -3,16 +3,46 @@ use, intrinsic :: iso_fortran_env, only: stderr=>error_unit
 implicit none
 private
 
-interface ! f{08,18}estop.f90
-module subroutine error_stop(msg, code)
+public :: assert, error_stop
+
+contains
+
+subroutine error_stop(msg, code)
 character(*), intent(in) :: msg
 integer, intent(in), optional :: code
-end subroutine error_stop
-end interface
 
-public :: assert, error_stop
+! Aborts the program with nonzero exit code
+! this is a fallback for Fortran 2008 error stop (e.g. Intel 19.1/2020 compiler)
+!
+! The "stop <character>" statement generally has return code 0.
+! To allow non-zero return code termination with character message,
+! error_stop() uses the statement "error stop", which by default
+! has exit code 1 and prints the message to stderr.
+! An optional integer return code "code" may be specified.
+!
+! Example
+! -------
+!
+! call error_stop("Invalid argument")
 
-contains
+write(stderr,*) msg
+
+if(present(code)) then
+  select case (code)
+  case (1)
+    error stop 1
+  case (2)
+    error stop 2
+  case (77)
+    error stop 77
+  case default
+    write(stderr,*) 'ERROR: code ',code,' was specified.'
+    error stop
+  end select
+else
+  error stop
+endif
+end subroutine
 
 subroutine assert(condition, code)
 ! If condition == .false., it aborts the program.
```

[...]

PR #109, opened 2020-01-13
Update CMakeLists handling of .mod files

nncarlson:
This resolves the issue @certik noted in #108 with the NAG compiler not finding the .mod files. It's unclear to me why things worked with other compilers, as this change is not NAG-specific at all.

This also handles installation of the .mod files (or whatever the compiler creates) which was missing before. One question is where the .mod files should be installed. This installs them into `include`, but it is arguably better to install them alongside the library file in the `lib` directory.

I've tested this with Intel and NAG (temporarily using the workarounds from #108).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,11 +2,6 @@ cmake_minimum_required(VERSION 3.14.0)
 project(stdlib Fortran)
 enable_testing()
 
-# this avoids stdlib and projects using stdlib from having to introspect stdlib's directory structure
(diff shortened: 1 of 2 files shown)


certik:
@scivision can you please also review this? You were the one who implemented `CMAKE_Fortran_MODULE_DIRECTORY` as part of #51.

scivision:
This is a nice programmatic way to do this, using the CMake generators.

certik:
Thanks for the review @scivision and thanks for the fix @nncarlson.

PR #120, opened 2020-01-22
stdlib_experimental_io: addition of a spec and extension of loadtxt and savetxt to support integers

jvdp1:
* Addition of a spec (`stdlib_experimental_io.md`) for the procedures in `stdlib_experimental_io.f90` (`loadtxt`, `savetxt`, and `open`). This spec is different than previously because it also describes subroutines.
* Extension of `loadtxt` and `savetxt` to support `integer`.
* The file `stdlib_experimental_io.f90` was generated from the file `stdlib_experimental_io.fypp` using `fypp`.

### DIFF
--- a/src/stdlib_experimental_io.md
+++ b/src/stdlib_experimental_io.md
@@ -57,16 +57,16 @@ Returns the unit number of a file opened to read, to write, or to read and write
 
 | Character | Meaning |
 | --------- | ------- |
-| `r` | open for reading (default) |
-| `w` | open for writing, truncating the file first |
-| `x` | open for exclusive creation, failing if the file already exists |
-| `a` | open for writing, appending to the end of the file if it exists |
-| `b` | binary mode |
-| `t` | text mode (default) |
-| `+` | open for updating (reading and writing) |
+| '`r`' | open for reading (default) |
+| '`w`' | open for writing, truncating the file first |
+| '`x`' | open for exclusive creation, failing if the file already exists |
+| '`a`' | open for writing, appending to the end of the file if it exists |
+| '`+`' | open for updating (reading and writing) |
+| '`b`' | binary mode |
+| '`t`' | text mode (default) |
 
 
-The default mode is `rt` (i.e. open for reading a text file).
+The default `mode` is '`rt`' (i.e. open for reading a text file). The `mode` may include one of the four different methods for opening a file (i.e., '`r`', '`w`', '`x`', and '`a`'). These four methods can be associated with the character '`+`' to open the file for updating. In addition, it can be specified if the file should be handled as a binary file ('`b`') or a text file ('`t`').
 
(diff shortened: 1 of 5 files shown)


src/stdlib_experimental_io.md
```diff
@@ -0,0 +1,120 @@
...
+
+| Character | Meaning |
+| --------- | ------- |
+| `r` | open for reading (default) |
+| `w` | open for writing, truncating the file first |
+| `x` | open for exclusive creation, failing if the file already exists |
+| `a` | open for writing, appending to the end of the file if it exists |
+| `b` | binary mode |
+| `t` | text mode (default) |
+| `+` | open for updating (reading and writing) |
```
ivan-pi: Maybe an extra sentence would be helpful to indicate which of these can be used together? I had to look at the code, to know I can select one letter r, w, x, a; one from b,t; and optionally the +. But perhaps this is not yet so critical as long as we are in the experimental stage.

jvdp1: Thank you. I think it is important that we get a correct spec at this stage already. Otherwise it will lead to other PRs.

Would the following sentence be enough:

> `mode` may include one character to define which mode you want to open the file in (i.e., `r`, `w`, `x`, and `a`) and one character to specify if the file should be handled as a binary or a text file (i.e., `b`, or `t`). `mode` may optionally include a third character `+` to open the file for updating. The default `mode` is `rt` (i.e. open for reading a text file).

ivan-pi: I will reply later, however before I forget I think the characters in the table should be enclosed with quotation marks (single or double?) so they get rendered in Markdown as a character literals?
| `r` | open for reading (default) |
| `'r'` | open for reading (default) | <-- like this

The same follow for the modes mentioned in the text, e.g. The default `mode` is `'rt'`.

jvdp1: @ivan-pi I modified the spec following your comments. Please let me know if it is now clear.

milancurcic: Quotes should be inside the inline code block, not outside. So, `'r'`, not '`r`' as it is now.

jvdp1: > Quotes should be inside the inline code block, not outside. So, `'r'`, not '`r`' as it is now.

Good. And modified.

ivan-pi: Compared to the Python documentation (e.g. [open](https://docs.python.org/3/library/functions.html#open)), the current explanation is relatively short and does not cover precisely what are the meanings of binary and text format and how they interact with the platform. 

But I do not want to hinder progress, so good to go from my side too. 👍

certik:
It's a start. We can improve upon it with subsequent PRs. Thanks for the review and merging!

PR #123, opened 2020-01-23
Do not use pip to install cmake

certik:
Fixes #122.

Issue #121, opened 2020-01-23
Proposal for test_condition


milancurcic:
Evolved from discussion in #116.

## Summary

This is a proposal to add a subroutine `test_condition` to `stdlib_experimental_error` that will:

* Test the value of a logical condition like `assert` in `stdlib_experimental_error` does;
* Allow printing a custom error message to stderr;
* Optionally, allow passing an exit code like with `assert`;
* Optionally, only print the message without stopping the program, i.e. warn only;

This is useful for testing both programmer and user errors. It can be used regardless of how the program is built (debug or release mode).

It could also be used as a basic building block toward higher-level testing or exception handling proposed in #95.

## Description

`test_condition` tests an input logical condition. If the condition evaluates to `.false.`, the default behavior is to stop the program and print the error message. Optionally, the user can pass an exit code to stop the program with. Optionally, the user can choose to not stop the program, but only print the warning.

## API

```fortran
subroutine test_condition(condition, msg, code, warn)
  logical, intent(in) :: condition
  character(*), intent(in), optional :: msg
  integer, intent(in), optional :: code
  logical, intent(in), optional :: warn
```

## Arguments

* `condition`: Logical condition to test
* `msg`: (optional) Character string to print to stderr. Default
* `code`: (optional) Integer exit code to set when stopping the program. (**what should be the default value? Current implementation of assert doesn't set any default code as far as I can tell.**)
* `warn`: (optional) Logical. If `.false.` (default), `test_condition` will stop the program if `condition` is also false. If `.true.`, it will print the error message to the screen and resume.

## Example

```fortran
use stdlib_experimental_error, only: test_condition
...
call test_condition(n > 0, 'n > 0 failed', warn=.true.)
```

## Implementation

```fortran
subroutine test_condition(condition, msg, code, warn)
  logical, intent(in) :: condition
  character(*), intent(in), optional :: msg
  integer, intent(in), optional :: code
  logical, intent(in), optional :: warn
  if (.not. condition) then
    if (optval(warn, .false.)) then
      write(srderr,*) optval(msg, "Test failed.")
    else
      call error_stop(optval(msg, "Test failed."), code)
    end if
  end if
end subroutine test_condition
```

[...]

Issue #76, opened 2020-01-04
Message for errors inside stdlib?


jvdp1:
In `stdlib_experimental_error.f90`, there is the following example:
`call error_stop("Invalid argument")`
A similar case can be found in `stdlib_experimental_io.f90`.

When running a large program, messages such as "Invalid argument" are quite useless. Should we discuss and agree on a good way to mention error messages, e.g.,

`ERROR (_name_of_the_function_): Invalid argument (_argument_)`

zbeekman:
The best way I've ever found to get good error messages and be able to do so even when using elemental or pure procedures is to define an `error_stack_t` class with a bunch of `pure` methods for pushing, popping, signalling (raising?) and then non-`pure` methods to catch and handle the errors.

I opened #95 to get more feedback. It could be used for the purpose described in this issue.

certik:
A minimal improvement is to at least list the name of the function in the error message as @jvdp1 suggested. Ideally the compiler would be able to optionally generate a nice stacktrace when the program ends with a non-zero exit code.

PR #124, opened 2020-01-23
addition of stdlib_experimental_stats mean

jvdp1:
follow-up of #119 and after merging #123  (See @certik [comments](https://github.com/fortran-lang/stdlib/pull/119#issuecomment-577918918)).

In addition to #119, CMakefiles and Makefiles.manual were modified to support `fypp`. Therefore, the auto-generated `.f90` files are not present in this PR.

**Issue:** I could not modify the CI for Windows properly. I would need help there.


certik:
This makes it pass on Windows: https://github.com/jvdp1/stdlib/pull/5

certik:
Now there is another problem. The CI suggests the file `stdlib_experimental_stats_mean.f90` takes over 4.5 minutes to compile (https://github.com/jvdp1/stdlib/pull/5/checks?check_run_id=406216313):
```
Thu, 23 Jan 2020 23:43:13 GMT [ 28%] Building Fortran object src/CMakeFiles/fortran_stdlib.dir/stdlib_experimental_stats_mean.f90.o
Thu, 23 Jan 2020 23:47:30 GMT [ 31%] Linking Fortran static library libfortran_stdlib.a
```
Which in my opinion is a big problem. Although perhaps this is only in Release mode.

jvdp1:
@certik it takes also only a few seconds to compile on my desktop.

jvdp1:
@certik : Here are the times on my desktop:
When the files includes arrays with up to 15 dimensions:
```
$ gfortran --version
GNU Fortran (GCC) 9.2.1 20190827 (Red Hat 9.2.1-1)
...
$ time make -j
[  5%] Generating stdlib_experimental_stats_mean.f90
[  5%] Generating stdlib_experimental_stats.f90
.....
[ 97%] Linking Fortran executable test_mean
[100%] Linking Fortran executable test_mean_f03
[100%] Built target test_mean
[100%] Built target test_mean_f03

real	0m21,621s
user	0m23,165s
sys	0m1,692s
````
When all loops for real arrays are replaced by `sum(x)` and `sum(x, dim)`:
```
$ time make -j
[  5%] Generating stdlib_experimental_stats.f90
...
[100%] Built target test_mean_f03

real	0m16,390s
user	0m18,122s
sys	0m1,502s
```
See [Actions of branch _stat_mean_dev_2](https://github.com/jvdp1/stdlib/tree/stat_mean_dev_2) for details with this branch. The tests with GFortran 9 took still >4 minutes.

When the files includes arrays with up to 7 dimensions (with the same compiler):
```
$ time make -j
[  5%] Generating stdlib_experimental_stats_mean.f90
[  5%] Generating stdlib_experimental_stats.f90
....
[ 97%] Linking Fortran executable test_ascii
[100%] Linking Fortran executable test_mean
[100%] Built target test_ascii
[100%] Built target test_mean

real	0m3,312s
user	0m4,973s
sys	0m1,250s
```

So,
* With CMake, I am far from the times we saw on Github. (Note: when using the Makefile.manual, GFortran 9 took over 2 minutes to compile the version with up to 15 dimensions)
* Replacing all loops for real arrays does not help much.
* The difference between GFortran 7 and Gfortran (8 or 9) can partially be explained by the fact that GFortran supports only arrays up to 7 dimensions, while GFortran 8 and 9 support arrays with dimentions up to 15.
* It is strange that GFortran 9 requires >2 times the time of GFortran 8 since they are based on the same code.

**Comment by jvdp1** [comment]
@certik @zbeekman 

In the branch [stat_mean_dev_3](https://github.com/jvdp1/stdlib/tree/stat_mean_dev_3), I replaced all loops by using `sum(x)`, or `sum(real(x,dp))`. Suprisingly to me, `sum(real(x,dp))` does not create a double-precision temporary array of the same dimensions as `x`. It was the main reason I used loops instead of `sum`. Since this reason is not valid (at least with GFortran 9; please explain me why ;)), I propose to merge this branch "stat_mean_dev_3" into this PR.

Regarding compile times of stat_mean_dev_3 on my desktop (GFortran 9):
```
$ time make -j
[  5%] Generating stdlib_experimental_stats_mean.f90
[  5%] Generating stdlib_experimental_stats.f90
...
[100%] Built target test_mean_f03

real	0m10,240s
user	0m11,773s
sys	0m1,729s
```
So there is a speedup of 2 (on my desktop) in comparison to this PR that includes all loops for real and integer arrays.

From [Github Actions](https://github.com/jvdp1/stdlib/runs/406875275), Build and compile times are 6s for GFortran7, 1m 15s for GFortran 8, and 1m 51s for GFortran 9. These times are better than the initial 4 minutes, but I can't still understand why GFortran 8 and 9 need more time on CI than on my own desktop.

**Comment by certik** [comment]
The other concerning thing is if a single function such as `mean` adds 4min to our build, then once we have say 50 such functions, we are talking of a build time over 3h. That does not scale well.

As a practical work around, we can have a CMake option for development, which would restrict dimensions to 3. That would allow us to develop with a reasonable compile time. 

@milancurcic what do you think we should do here?

**Comment by milancurcic** [comment]
More generally, can we have a build parameter to set the maximum rank to build? This would have a low default values for debug and release modes, and user could override it manually if they needed higher ranks? Something along the lines of:

```
cmake .. -DCMAKE_MAXIMUM_RANK=8 # overrides default maximum rank
```

**Comment by jvdp1** [comment]
> As a practical work around, we can have a CMake option for development, which would restrict dimensions to 3.

For GFortran 7, the number of dimensions restricticted to 7, and the compile time is 6-7 seconds on Github Actions. Would 7 dimensions be an option for you @certik? 
Or is it possible that CMake recognizes that the tests are running on Github Actions, and that it limits the number of dimensions to 4? (Since it is a preprocessed loop starting at 3, going to 4 would be good). 
When a user would compile the library, CMake would see that it is not a Github Action, and will compile `stdlib` accordingly to its environment (i.e., up to 7 or 15 dimensions; which is quite quick on my desktop).

**Comment by certik** [comment]
Yes, I am fine with any number of dimensions (including 7) as long as it is fast. If 7 can be made fast, then let's do it. Otherwise let's do less.

**Comment by jvdp1** [comment]
@certik @milancurcic : I implemented the proposed idea. So, for the CI, the maximum dimension for auto-generation of the `.f90` files is limited to 5 (by using cmake -DCMAKE_MAXIMUM_RANK=5). If the flag CMAKE_MAXIMUM_RANK is not defined, then CMake will test if the compiler supports 7 or 15 dimensions.

Issue: Something seems broken for [CI on Linux](https://github.com/fortran-lang/stdlib/pull/124/checks?check_run_id=408114214) when installing GFortran: 

> E: Failed to fetch https://packages.microsoft.com/repos/microsoft-ubuntu-bionic-prod/dists/bionic/main/binary-amd64/Packages.bz2  File has unexpected size (89974 != 89668). Mirror sync in progress? [IP: 40.76.35.62 443]
>    Hashes of expected file:
>     - Filesize:89668 [weak]
>     - SHA512:239b3775157309c1f1ff14624d2d0044c5a0c57f6bf9a431f628de1c9f19c91e107c4bc78e1d12bce44121e5866fdf67328e6482151b6edd3456f8c541dc5739
>     - SHA256:d737bf9c5418556ef337f43f0124a3db6e79c9f7209deb140ea87644169168b1
>     - SHA1:de9e09d12abdb967a62954611af31655b3a8b5d9 [weak]
>     - MD5Sum:9e9321f39e46dccc97d928ec9bffc87a [weak]
>    Release file created at: Thu, 23 Jan 2020 17:58:24 +0000
> E: Some index files failed to download. They have been ignored, or old ones used instead.
> ##[error]Process completed with exit code 100.

Any idea how to solve this?

**Comment by certik** [comment]
Thanks for implementing it. Yes, the Linux image sometimes has this problem and last time I think it took about a day to resolve. I am hoping it will be faster this time.

**Comment by jvdp1** [comment]
> Thanks for implementing it. Yes, the Linux image sometimes has this problem and last time I think it took about a day to resolve. I am hoping it will be faster this time.

OK. We will see when it will resolve. At least for `macos` and GFortran 9 the compile time reduced from >3 minutes to 7s. I guess we can expect similar times for Linux.

**Comment by jvdp1** [comment]
I limited the number of dimensions to 4 (with `cmake .. -DCMAKE_MAXIMUM_RANK=4`). All cheks passed, and compile times are <10s.
I think that all comments in #119 and in this PR were considered. Any other comments after an additional review?

File: src/stdlib_experimental_stats_mean.fypp [hunk 1]
```diff
@@ -0,0 +1,178 @@
...
+                  merge(size(x,${rank-1}$),size(x,${rank}$),mask = ${rank-1}$ < dim ) )
+
+    select case(dim)
+#:for fi in range(1,rank+1)
+     case(${fi}$)
+       res=sum(x, ${fi}$) / real(size(x, ${fi}$), ${k1}$)
+#:endfor
+     case default
+       call error_stop("ERROR (mean): wrong dimension")
+    end select
```
ivan-pi: A minor nitpick related to the discussion in https://github.com/fortran-lang/stdlib/issues/76:
- the error message would be more informative if it were to say something like "wrong dimension, obtained dim=_dim_ is larger than rank of the array x"

We need to wait for the numeric to string conversion functions to achieve this easily.

jvdp1: I agree with you. Ideally it should be check by the compiler (like for `sum` and other intrinsic functions).
I will keep in mind for a next improvement. It could be also related to #121 .

PR #130, opened 2020-01-29
Mean 1d

aradi:
* Substantially reduced code & number of fypp-constructs in `mean()`.
* Extended `mean()` to handle reduction from 1D-array to scalar (similar to `sum()`)
* Introduced tolerances when testing floating point results (note checking with `== 0.0` is highly unreliable and may give false positives)
* Reformatted test file according to style guide.

Commits in order of relevance, feel free to cherry pick if you don't like all of them.

### DIFF
--- a/src/tests/stats/test_mean.f90
+++ b/src/tests/stats/test_mean.f90
@@ -5,6 +5,8 @@ use stdlib_experimental_io, only: loadtxt
 use stdlib_experimental_stats, only: mean
 implicit none
 
+real(sp) :: s1(3) = [1.0_sp, 2.0_sp, 3.0_sp]
+
 real(sp), allocatable :: s(:, :)
 real(dp), allocatable :: d(:, :)
 
@@ -18,6 +20,10 @@ call loadtxt("array3.dat", s)
 call assert( mean(s) - sum(s)/real(size(s), sp) == 0.0_sp)
 call assert( sum( abs( mean(s,1) - sum(s,1)/real(size(s,1), sp) )) == 0.0_sp)
 call assert( sum( abs( mean(s,2) - sum(s,2)/real(size(s,2), sp) )) == 0.0_sp)
+
+! check reduction of rank one array to scalar
+call assert(mean(s1) - sum(s1) / real(size(s1), sp) == 0.0_dp)
+call assert(mean(s1, dim=1) - sum(s1, dim=1) / real(size(s1, dim=1), sp) == 0.0_dp)
 
 
 !dp
--- a/src/tests/stats/test_mean.f90
+++ b/src/tests/stats/test_mean.f90
@@ -5,6 +5,9 @@ use stdlib_experimental_io, only: loadtxt
 use stdlib_experimental_stats, only: mean
 implicit none
 
+real(sp), parameter :: sptol = 1.2e-06_sp
+real(dp), parameter :: dptol = 2.2e-15_dp
+
 real(sp) :: s1(3) = [1.0_sp, 2.0_sp, 3.0_sp]
 
 real(sp), allocatable :: s(:, :)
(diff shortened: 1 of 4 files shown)


src/tests/stats/test_mean.f90
```diff
@@ -1,75 +1,90 @@
...
+  use stdlib_experimental_error, only: assert
+  use stdlib_experimental_kinds, only: sp, dp, int32, int64
+  use stdlib_experimental_io, only: loadtxt
+  use stdlib_experimental_stats, only: mean
+  implicit none
 
-real(sp), allocatable :: s(:, :)
-real(dp), allocatable :: d(:, :)
+  real(sp), parameter :: sptol = 1.2e-06_sp
+  real(dp), parameter :: dptol = 2.2e-15_dp
```
jvdp1: Minor comment: why not use the function`epsilon()`?

src/tests/stats/test_mean.f90
```diff
@@ -1,75 +1,90 @@
...
-call loadtxt("array3.dat", d)
+  !dp
+  call loadtxt("array3.dat", d)
 
-call assert( mean(int(d, int32)) - sum(real(int(d, int32),dp))/real(size(d), dp) == 0.0_dp)
-call assert( sum(abs( mean(int(d, int32),1) - sum(real(int(d, int32),dp),1)/real(size(d,1), dp) )) == 0.0_dp)
-call assert( sum(abs( mean(int(d, int32),2) - sum(real(int(d, int32),dp),2)/real(size(d,2), dp) )) == 0.0_dp)
+  call assert( abs(mean(d) - sum(d)/real(size(d), dp)) < dptol)
+  call assert( sum( abs( mean(d,1) - sum(d,1)/real(size(d,1), dp) )) < dptol)
+  call assert( sum( abs( mean(d,2) - sum(d,2)/real(size(d,2), dp) )) < dptol)
```
jvdp1: I used the `==` on purpose since it is the same formula (e.g., `(sum(d)/real(size(d), sp))`) that are used for the function `mean`. 
However, I am also fine with these expressions.

src/tests/stats/test_mean.f90
aradi: Yes, we could use that. But then some power of epsilon (e.g. ˙epsilon(1.0_dp)**(7.0_dp / 8.0_dp)`) as epsilon itself is too strict (if the difference of two numbers of the order of 1.0 is less than epsilon, than by the definition of epsilon it can not be represented anyway...)

src/tests/stats/test_mean.f90
aradi: Despite being the same formula, there is no warranty for the exact bitwise equivalence at the end. If the first one is calculated in 80-bit float registers and rounded at the end, while the other does every summation with 64 bit floats, the results would be different.

src/tests/stats/test_mean.f90
jvdp1: Thank you. Good as it is.

src/tests/stats/test_mean.f90
jvdp1: I did not think to that possibility. Thank you for the explanations.

aradi:
Well, the styleguide says `the body of every Fortran construct should be indented by two (4) spaces`, so I have chosen the one I prefer more :laughing: .

Anyway, I have taken out the reformatting commit, it really does not belong to here, I agree. I'll wait with reformatting changes until the style guide becomes non-ambigous. :wink:

certik:
@aradi ha, that was a mistake, thanks for spotting it. I fixed it in #131.

certik:
@aradi the PR looks great, thank you!

PR #60, opened 2019-12-31
Use 4 spaces

certik:

PR #131, opened 2020-01-29
Fix a typo ("four" instead of "two")

certik:
This is an omission that was supposed to be part of #60, but we missed
it. This PR fixes it.


certik:
I am going to merge it, as it was an obvious typo that we forgot to do in the previous PR.

Issue #128, opened 2020-01-27
Addition of a mask to stdlib_experimental_stats function `mean`


jvdp1:
# Current

(Related to #113)

The current spec for `stdlib` `mean` is:

### Description

Returns the mean of all the elements of `array`, or of the elements of `array` along dimension `dim` if provided.

### Syntax

`result = mean(array)`

`result = mean(array, dim)`

### Arguments

`array`: Shall be an array of type `integer`, or `real`.

`dim`: Shall be a scalar of type `integer` with a value in the range from 1 to n, where n is the rank of `array`.

### Return value

If `array` is of type `real`, the result is of the same type as `array`.
If `array` is of type `integer`, the result is of type `double precision`.

If `dim` is absent, a scalar with the mean of all elements in `array` is returned. Otherwise, an array of rank n-1, where n equals the rank of `array`, and a shape similar to that of `array` with dimension `dim` dropped is returned.

##

# Proposal

I would like to propose to add `mask` into the API of `mean` as follows (similar to the intrinsic `sum`):

### Description

Returns the mean of all the elements of `array`, or of the elements of `array` along dimension `dim` if provided, and if the corresponding element in `mask` is `true`.

### Syntax

`result = mean(array [, mask])`

`result = mean(array, dim [, mask])`

### Arguments

`array`: Shall be an array of type `integer`, or `real`.

`dim`: Shall be a scalar of type `integer` with a value in the range from 1 to n, where n is the rank of `array`.

`mask` (optional): Shall be of type `logical` and either be a scalar or an array of the same shape as `array`.

### Return value

If `array` is of type `real`, the result is of the same type as `array`.
If `array` is of type `integer`, the result is of type `double precision`.

If `dim` is absent, a scalar with the mean of all elements in `array` is returned. Otherwise, an array of rank n-1, where n equals the rank of `array`, and a shape similar to that of `array` with dimension `dim` dropped is returned.

If `mask` is specified, the result is the mean of all elements of `array` corresponding to `true` elements of `mask`. If every element of `mask` is `false` , the result is zero.

# 

This definition of `mask` for `mean` agrees with the definition of `mask` of the intrinsic `sum` in the [Fortran Standard draft](http://isotc.iso.org/livelink/livelink?func=ll&objId=19442438&objAction=Open&viewType=1):

`p. 429`:
>MASK (optional) shall be of type logical and shall be conformable with ARRAY.

and

`p. 9`:
>3.36
>conformable
[...]

PR #132, opened 2020-01-29
mask_mean_dev: addition of a mask in mean API

jvdp1:
* Addition of `mask` in the API of `mean`, as discussed in #128. `NaN` returns are based on `ieee_arithmetic`;
* Update of the spec;
* Correction of a small typo/error introduced in a previous PR (an `integer` array was not converted to `real(dp)` before doing the `sum` (were our tests not severe enough for these scenarions?).

@aradi : I can't request you as reviewer (the system doesn't find you). Could you have also a look, please?

### DIFF
--- a/src/tests/stats/test_mean.f90
+++ b/src/tests/stats/test_mean.f90
@@ -37,6 +37,17 @@ call assert( sum( abs( mean(d,1) - sum(d,1)/real(size(d,1), dp) )) < dptol)
 call assert( sum( abs( mean(d,2) - sum(d,2)/real(size(d,2), dp) )) < dptol)
 
 
+! check mask = .false.
+
+call assert( isnan(mean(d, .false.)))
+call assert( any(isnan(mean(d, 1, .false.))))
+call assert( any(isnan(mean(d, 2, .false.))))
+
+! check mask of the same shape as input
+call assert( abs(mean(d, d > 0._dp) - sum(d, d > 0._dp)/real(count(d > 0._dp), dp)) < dptol)
+call assert( sum(abs(mean(d, 1, d > 0._dp) - sum(d, 1, d > 0._dp)/real(count(d > 0._dp, 1), dp))) < dptol)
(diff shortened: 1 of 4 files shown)


src/tests/stats/test_mean.f90
```diff
@@ -67,7 +78,7 @@ program test_mean
 
 !dp rank 4
 allocate(d4(size(d,1),size(d,2),3,9))
-d4 = 1.
+d4 = -1._dp
```
certik: Your change is correct, but even better is just:
```suggestion
d4 = -1
```
Then the precision is figured out automatically.

src/tests/stats/test_mean.f90
```diff
@@ -79,4 +90,20 @@ program test_mean
...
+
+call assert( isnan(mean(d4, .false.)))
+call assert( any(isnan(mean(d4, 1, .false.))))
+call assert( any(isnan(mean(d4, 2, .false.))))
+call assert( any(isnan(mean(d4, 3, .false.))))
+call assert( any(isnan(mean(d4, 4, .false.))))
+
+
+! check mask of the same shape as input
+call assert( abs(mean(d4, d4 > 0._dp) - sum(d4, d4 > 0._dp)/real(count(d4 > 0._dp), dp)) < dptol)
```
certik: Your code is correct, but one can simplify it as:
```suggestion
call assert( abs(mean(d4, d4 > 0) - sum(d4, d4 > 0)/real(count(d4 > 0), dp)) < dptol)
```
So I usually like it that way better. But that's up to you.

aradi:
@jvdp1 Sure. Actually, now it seems to be possible to select me as reviewer. I'll try to have a detailed look at it tomorrow.

src/tests/stats/test_mean.f90
jvdp1: modified!

src/tests/stats/test_mean.f90
jvdp1: modified

jvdp1:
Thank you @aradi for the review.

@milancurcic could  you have a last look and merge it if you are happy with the proposed changes?

milancurcic:
Great work, thanks @jvdp1

PR #140, opened 2020-02-04
Added support and test for mean of complex arrays

fiolj:
Added support and tests for complex types. 
I changed the names of the specific functions, following the convention used in optval.

### DIFF
--- a/src/stdlib_experimental_stats.fypp
+++ b/src/stdlib_experimental_stats.fypp
@@ -1,8 +1,6 @@
 #:include "common.fypp"
-
 #:set RANKS = range(1, MAXRANK + 1)
-
-
+#:set RC_KINDS_TYPES = REAL_KINDS_TYPES + CMPLX_KINDS_TYPES
 module stdlib_experimental_stats
   use stdlib_experimental_kinds, only: sp, dp, qp, &
       int8, int16, int32, int64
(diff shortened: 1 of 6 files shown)


src/stdlib_experimental_stats.fypp
```diff
@@ -12,92 +10,113 @@ module stdlib_experimental_stats
   public :: mean
 
   interface mean
-
-    #:for k1, t1 in REAL_KINDS_TYPES
+#:def name(Rank, Type, Kind)
+$:"mean_{0}_all_{1}{2}_{1}{2}".format(Rank, Type[0], Kind)
+#:enddef
```
jvdp1: I like idea to use a preprocessing macro for the names of functions, especially if we will other functions with the same API (e.g., `variance`, `median`,...)

So, to make these macros more general, would it be possible to pass the name of the function, e.g.,:
`#:def name(Varname, Rank, Type, Kind)`
where `Varname` would be equal to `mean`, `variance`, `median`,....?

Also, would it be possible to add these macros in `common.fypp`, since these will be used for other functions?

fiolj: Yes, I'll do. I started that way but because the original routine names were complex, I decided for hard-code the names in order not to change too much the code. I'll amend and send a new request.

Also @jvdp1, could we make the functions pure? (I've not used much of submodules before, so I don't know if there is restrictions there)

jvdp1: Functions in the submodule can be pure. However, it seems that the minimum required CMake does not support that (I am not sure if the last version of CMake supports it either; @zbeekma @certik @milancurcic ?).
Also the checks on the dimensions inside the functions may prohibit `pure` for some of them.

fiolj:
Ok, @jvdp1  @aradi What do you think of using a macro for names like the following?

```
#:def rname(gname, rank, type, kind, suffix='')
  $:"{0}_{1}_{2}{3}_{2}{3}".format(gname, rank, type[0], kind) if suffix == '' else "{0}_{1}_{2}{3}_{4}".format(gname, rank, type[0], kind, suffix)
#:enddef
```

Assuming that t1='real', k1='sp', rank=1,
`${rname("mean_all",rank, t1, k1)}$`
will give:
 `mean_all_1_rsp_rsp`

src/tests/stats/test_mean_f03.f90
```diff
@@ -5,34 +5,51 @@ program test_mean
 use stdlib_experimental_stats, only: mean
 implicit none
 
+real(dp), parameter :: dptol = 2.2e-15_dp
```
certik: This number seems to be `10 * epsilon(1._dp)`. Maybe you can use that instead.

certik: (Or you can keep it and we'll fix it in another PR.)

src/stdlib_experimental_stats_mean.fypp
```diff
@@ -12,28 +10,29 @@ submodule (stdlib_experimental_stats) stdlib_experimental_stats_mean
...
-      module function mean_${rank}$_all_${k1}$_${k1}$(x, mask) result(res)
+      #:set RName = rname("mean_all",rank, t1, k1)
+      module function ${RName}$ (x, mask) result(res)
         ${t1}$, intent(in) :: x${ranksuffix(rank)}$
         logical, intent(in), optional :: mask
         ${t1}$ :: res
 
         if (.not.optval(mask, .true.)) then
-          res = ieee_value(res, ieee_quiet_nan)
+          res = ieee_value(real(res, kind=${k1}$), ieee_quiet_nan)
```
jvdp1: `ieee_value` returns a NaN of the same type as `res`. So it should be OK without the conversion to a `real`, right? Or do I miss something?

src/tests/stats/test_mean_f03.f90
certik: I just sent a PR for this: #142.

fiolj:
I was getting an error for complex numbers. I am not sure but it seems
that it is not defined for complex numbers.

src/tests/stats/test_mean_f03.f90
fiolj: I did not see your message. I corrected it both in test_mean_f03.f90 and test_mean.f90

certik: No problem. Let's first get an agreement on this in #142, to ensure nobody is against it.

jvdp1:
> I was getting an error for complex numbers. I am not sure but it seems that it is not defined for complex numbers. 

It seems that you are right: it is not defined for `complex` (at least in [gfortran](https://github.com/gcc-mirror/gcc/blob/master/libgfortran/ieee/ieee_arithmetic.F90))

So it can be merged from my side.

certik:
I resolved conflicts and tests seem to pass. We got two approvals, so I am going to merge it now.

PR #141, opened 2020-02-04
Add spec for optval

nshaffer:
I based this on the structure of the spec for `mean`. Things I'm not 100% sure of

- Is an "implemented" section necessary for this case?
- Do we want language-defined terms capitalized or formatted some other way, e.g., PRESENT vs present vs `present`?
- Is my mixed usage of "default" and "fallback" a problem?

### DIFF
--- a/src/stdlib_experimental_optval.md
+++ b/src/stdlib_experimental_optval.md
@@ -10,7 +10,7 @@
 
 Returns `x` if it is PRESENT, otherwise `default`. 
 
-This function is intended to be called in a proceudre with one or more OPTIONAL arguments, in order to conveniently fall back to a default value if an OPTIONAL argument is not PRESENT.
+This function is intended to be called in a procedure with one or more OPTIONAL arguments, in order to conveniently fall back to a default value if an OPTIONAL argument is not PRESENT.
 
 ### Syntax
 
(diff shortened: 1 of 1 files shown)


jvdp1:
Thank you.

> I based this on the structure of the spec for `mean`. Things I'm not 100% sure of
> 
> * Is an "implemented" section necessary for this case?

I suggest to keep it; to follow the same format as other specs.

> * Do we want language-defined terms capitalized or formatted some other way, e.g., PRESENT vs present vs `present`?

I think it is @milancurcic who mentioned that, for the `mean` spec, he preferred e.g., `present` over PRESENT. I tried to follow that rule for the other specs.

> * Is my mixed usage of "default" and "fallback" a problem?

I am ok with it.

milancurcic:
Alternatively, to avoid confusion regarding what "present" means, we could write:

> Returns `x` if `present(x)` returns `.true.`, otherwise `default`.

instead of

> Returns `x` if it is present, otherwise `default`.

nshaffer:
Thanks for the feedback, all. I've changed styling to match Milan's suggestions, which seem reasonable to me. I've left in the Implemented section. I suspect if we write tools that parse the specs or something, it would be useful to have them all follow a consistent structure.

jvdp1:
I will merge this PR since most of us approved it.

certik:
Thanks!

Issue #137, opened 2020-02-02
Proposal for variance and centering functions


jvdp1:
Based on discussions in #113, #3, #128, I would like to propose the following addition to `stdlib_experimental_stats`:

# 

### Description

Returns the variance of all the elements of `array`, or of the elements of `array` along dimension `dim` if provided, and if the corresponding element in `mask` is `true`.

The variance is defined as the best unbiased estimator and is computed as:

```
 var(x) = 1/(n-1) sum_i (array(i) - mean(array))^2
```

### Syntax

`result = var(array [, mask])`

`result = var(array, dim [, mask])`

### Arguments

`array`: Shall be an array of type `integer`, or `real`.

`dim`: Shall be a scalar of type `integer` with a value in the range from 1 to n, where n is the rank of `array`.

`mask` (optional): Shall be of type `logical` and either by a scalar or an array of the same shape as `array`.

### Return value

If `array` is of type `real`, the result is of the same type as `array`.
If `array` is of type `integer`, the result is of type `double precision`.

If `dim` is absent, a scalar with the variance of all elements in `array` is returned. Otherwise, an array of rank n-1, where n equals the rank of `array`, and a shape similar to that of `ar
ray` with dimension `dim` dropped is returned.

If `mask` is specified, the result is the variance of all elements of `array` corresponding to `true` elements of `mask`. If every element of `mask` is `false`, the result is IEEE `NaN`.

### Example

```fortran
program demo_mean
    use stdlib_experimental_stats, only: var
    implicit none
    real :: x(1:6) = [ 1., 2., 3., 4., 5., 6. ]
    print *, var(x)                            !returns __TOBECOMPLETED__
    print *, var( reshape(x, [ 2, 3 ] ))       !returns __TOBECOMPLETED__ 
    print *, var( reshape(x, [ 2, 3 ] ), 1)    !returns [__TOBECOMPLETED__]
    print *, var( reshape(x, [ 2, 3 ] ), 1,&
                  reshape(x, [ 2, 3 ] ) > 3.)  !returns [__TOBECOMPLETED__]
end program demo_mean
```
# 

To be discussed (not exhaustive):
* Based on discussions in #3, I suggest to first implement a two-pass algorithm. Other algorithms can be implemented later,  as proposed in #134. Allowing `dim` and `mask` in the API will not lead to a function as simple as in [#3 comment](https://github.com/fortran-lang/stdlib/issues/3#issuecomment-580493032).

[...]

Issue #114, opened 2020-01-19
Returning allocatable array from function


leonfoks:
I was going to ask this in #113 but didn't want to derail that thread.
This is more of a personal question... since I've actually done very little with returning an allocatable array from a function.  I may have flawed logic, so please correct me if i'm wrong.
I've had compile issues in the past returning allocatable from functions (cant remember all the details but perhaps others have had these too)  and I "intuitively" thought there might be a slow down if those functions need calling multiple times.

Ill use the snippet from #113
```Fortran
function mean(mat, dim) result(res)
    real(sp), intent(in) :: mat(:,:)
    integer, intent(in), optional :: dim !dim = 1 or dim = 2 
    real(sp), allocatable :: res(:)
end function
```

If I have a large array, and I call this multiple times, does it reallocate memory for that variable every time? Or is there some cleverness that happens?

e.g. 1D result is already allocated, function allocates its own result, on return those results are copied to pre-allocated array.

```Fortran
real(sp), allocatable :: x(:), A(:, :)

! A = full of numbers
allocate(x(shape(A)(2)))

x(:) = mean(A, dim=1)
```
Is it known what happens specifically, memory wise? Is it compiler dependent?

I'm asking because it will be slower if allocations keep happening, we might want to be careful and always make subroutines available where preallocated memory can be accessed without any implicit allocations and then copies. Am I worrying about this too much haha.

certik:
This is really important.

I don't think we should use allocatable at all in the lowest level API, unless the size of the output is truly not known ahead of time, like in loadtxt.

Also, in the past I have found that if a function returns an array (non allocatable), it can be slower than an equivalent subroutine with intent(out) array. Apparently there is an extra copy happening. This is something where it feels Fortran compilers must do better. I feel ideally there shouldn't be an overhead of using functions over subroutines, but unfortunately it seems there is.

leonfoks:
Phew, I was hoping I wasn't crazy haha

nncarlson:
I don't believe there is any performance difference between an automatically-sized array valued function, for example
```fortran
function foo(array) result(res)
real, intent(in) :: array(:,:)
real :: res(size(array,dim=2))
```
versus one whose result is allocatable. In either case the size of the result is determined at run time. So there is absolutely no reason to avoid allocatable results *if one is determined to use a function.*

[...]

PR #144, opened 2020-02-06
Addition of variance function in stdlib_experimental_stats

jvdp1:
Based on #137 

With this PR, I propose to add a function for computing the variance of elements in arrays using the same API as `stdlib::mean`. The used algorithm is a two-pass algorithm (as discussed in #3).
Based on #3 and #114, I avoided to use the function `mean` (and to create a new function `center` for doing `x - mean`), to avoid loss in performance.

* How can we avoid `select case` statement:
```fortran
        select case(dim)
          #:for fi in range(1, rank+1)
          case(${fi}$)
            n = real(size(x, dim), ${k1}$)
            mean = sum(x, dim) / n
            do i = 1, size(x, dim)
              res = res + (x${rankindice(':', 'i', rank, fi )}$ - mean)**2
            end do
          #:endfor
          case default
            call error_stop("ERROR (mean): wrong dimension")
        end select
```

and

```fortran
        select case(dim)
          #:for fi in range(1, rank+1)
          case(${fi}$)
            n = real(count(mask, dim), ${k1}$)
            mean = sum(x, dim, mask) / n
            do i = 1, size(x, dim)
              res = res + merge( (x${rankindice(':', 'i', rank, fi )}$ - mean)**2,&
              #:if t1[0] == 'r'
                                  0._${k1}$,&
              #:else
                                  cmplx(0._${k1}$, 0._${k1}$, ${k1}$),&
              #:endif
                                  mask${rankindice(':', 'i', rank, fi)}$)
            end do
          #:endfor
          case default
            call error_stop("ERROR (mean): wrong dimension")
        end select
```
I probably miss something that should be obvious!

Another issue is the compilation time needed with the Makefiles in the CI! 

Note: Each new statistical function in `stdlib_stats` will potentially includes 600 additional functions. It really illustrates the issue of having no templates.


jvdp1:
@fiolj Could you check the implementation for `complex` numbers? Currently there is no tests implemented, to limit the number of tests (but I can implement some tests in a latter commit/PR).

fiolj:
I"ll do,  but I can't before this night or tomorrow morning

fiolj:
I found it really good overall, but I think the variance for complex arrays is always a real number
([wikipedia link](https://en.wikipedia.org/wiki/Complex_random_variable#Variance_and_pseudo-variance)). 
Thus, for instance in var_all, we should replace:
@@ -16,7 +16,7 @@ contains
       module function ${RName}$(x, mask) result(res)
         ${t1}$, intent(in) :: x${ranksuffix(rank)}$
         logical, intent(in), optional :: mask
-        ${t1}$ :: res
+        real(${k1}$) :: res

and
@@ -28,7 +28,7 @@ contains
         n = real(size(x, kind = int64), ${k1}$)
         mean = sum(x) / n

-        res = sum((x - mean)**2) / (n - 1._${k1}$)
+        res = sum(abs(x - mean)**2) / (n - 1._${k1}$)
 
and in rname("var",rank, t1, k1):
-        ${t1}$ :: res${reduced_shape('x', rank, 'dim')}$
+        real(${k1}$) :: res${reduced_shape('x', rank, 'dim')}$

and later:
-              res = res + (x${rankindice(':', 'i', rank, fi )}$ - mean)**2
+             res = res + abs(x${rankindice(':', 'i', rank, fi )}$ - mean)**2

Also, if `res` is real, we don't need the conversion to real to return ieee_nan
-          res = ieee_value(real(res, kind=${k1}$), ieee_quiet_nan)

Coincidentally, I was thinking that my solution to the ieee for complex was unnecessarily complex.
The above line: `res = ieee_value(1._${k1}$, ieee_quiet_nan)` should work also when res is a complex variable.

Finally, Tomorrow I can look into it in more detail, but I was thinking that may be, even for real numbers we don't have to match the kind of the input. Indeed if the input is a large quantity of real(sp) it may be desirable that the variance (and mean) be calculated in double precision. I don't know if I am missing something about availability of double precision for some machines but I would think that nowadays that should not be an issue

jvdp1:
Thanks @fiolj for your review. 
I will look tomorrow on your propositions for the variance of `complex`.
Below are some other comments.

> Coincidentally, I was thinking that my solution to the ieee for complex was unnecessarily complex.
> The above line: `res = ieee_value(1._${k1}$, ieee_quiet_nan)` should work also when res is a complex variable.

This is indeed a good idea, much simpler than the previous implementation. I will implement it here and for `mean` too.

> Finally, Tomorrow I can look into it in more detail, but I was thinking that may be, even for real numbers we don't have to match the kind of the input. 

The idea was to get the same API as `sum`. So the result is of the same type as the real array, as in `sum`. This API also avoids internal conversion from, e.g., `sp` to `dp`, before calling `sum` (since it would have little sense to do it after `sum` operations IMO).
Also, having all results in `dp`, will not reduce the number of generated functions.

>Indeed if the input is a large quantity of real(sp) it may be desirable that the variance (and mean) be calculated in double precision. 

I think it is the responsability of the user to check that (as it is already the case with `sum`).

jvdp1:
From @fiolj  comments, I push commits for:

* simplifying the `ieee_value` call (as proposed by @fiolj)
* implementing a corrected computation of variance of `complex`
* implementing tests for `complex` (I tested them against Octave)

Regarding the implementation, I did e.g.,
```fortran
      #:if t1[0] == 'r'
          res = sum((x - mean)**2) / (n - 1._${k1}$)
        #:else
          res = sum(abs(x - mean)**2) / (n - 1._${k1}$)
        #:endif
```
to avoid the `abs` function in the `real` `var` function (it is there useless and give some penalties to efficiency)

@fiolj  could you confirm it was what you suggested, please?

fiolj:
Thanks @jvdp1, that was exactly what I suggested.
I think it looks good

fiolj:
> This PR is fine. Thanks.
> 
> In the long term, I wonder if fypp can be improved so that there is not as much repetition. There are still 8 different blocks to declare the signature of the function (real/integer and `(x, mask)`/`(x, dim, mask)` and `mask` = scalar/array, all combinations 2x2x2=8). And even more long term, the Fortran language itself should be improved so that fypp is not needed.

I agree with @certik here. I was thinking along the same lines. A possible solution for real/int interfaces does not seem to difficult. Something like (for instance for the function mean):
```

    #:for k1, t1 in RCI_KINDS_TYPES
      #:for rank in RANKS
        #:set RName = rname("mean_all",rank, t1, k1)
        #:set ret_type = 'real(dp)' if t1[0] == 'i' else t1
        module function ${RName}$ (x, mask) result(res)
          ${t1}$, intent(in) :: x${ranksuffix(rank)}$
          logical, intent(in), optional :: mask
          ${ret_type}$ :: res
        end function ${RName}$
      #:endfor
    #:endfor

```
which is valid for all types (real, complex, int). For some simple functions we could write the implementation in some way that would be mostly the same also.

For function`var`, we could use something similar

```
    #:for k1, t1 in RCI_KINDS_TYPES
      #:for rank in RANKS
        #:set RName = rname("var_all",rank, t1, k1)
        #:set ret_type = 'real(dp)' if t1[0] == 'i' else "real({})".format(k1)
        module function ${RName}$(x, mask) result(res)
          ${t1}$, intent(in) :: x${ranksuffix(rank)}$
          logical, intent(in), optional :: mask
          ${ret_type}$ :: res
        end function ${RName}$
      #:endfor
    #:endfor

```

This would cut the code to a half, not solving the remaining four factor.

jvdp1:
Thank you for your review.

> For function`var`, we could use something similar
> 
> ```
>     #:for k1, t1 in RCI_KINDS_TYPES
>       #:for rank in RANKS
>         #:set RName = rname("var_all",rank, t1, k1)
>         #:set ret_type = 'real(dp)' if t1[0] == 'i' else "real({})".format(k1)
>         module function ${RName}$(x, mask) result(res)
>           ${t1}$, intent(in) :: x${ranksuffix(rank)}$
>           logical, intent(in), optional :: mask
>           ${ret_type}$ :: res
>         end function ${RName}$
>       #:endfor
>     #:endfor
> ```
> 
> This would cut the code to a half, not solving the remaining four factor.

That will reduce the number of blocks, but it might give a `fypp` file that is difficult to read and to follow due to too many conditional statements (if/set; especially for the implementation where `fypp` conditions (if) will be needed, as already used for `complex`). We could probably implement some functions in `common.fypp`, but I am still afraid for the complexity of the code.

For now, I would suggest that we merge this PR with the master. Then we can open a PR to (try to) reduce the number of blocks in `var` and `mean` (but unfortunately, it will not reduce the number of generated functions).

fiolj:
El 9/2/20 a las 10:47, Jeremie Vandenplas escribió:
> For now, I would suggest that we merge this PR with the master. Then we
> can open a PR to (try to) reduce the number of blocks in |var| and
> |mean| (but unfortunately, it will not reduce the number of generated
> functions).

Yes, agree. It is not a completely satisfactory solution yet, and we
should try to come out with something better. I put it forward mainly to
start thinking about it.

src/common.fypp
```diff
@@ -113,4 +113,33 @@ ${prefix + joinstr.join([line.strip() for line in txt.split("\n")]) + suffix}$
...
+#!   origrank (int): Rank of the original variable
+#!   dim (int): Index of the used expression varname1
+#!
+#! Returns:
+#!   Shape expression enclosed in braces, except for the index dim.
+#!
+#! E.g., (:, :, :, i, :, :)
+#!
+
+#:def rankindice(varname, varname1, origrank, dim)
```
aradi: We should probably rename this macro to have a more descriptive name.  If it is only used to select subarrays by reducing the dimension, we could have:
```
#:def select_subarray(origrank, selectors)
  #:assert origrank > 0
  #:set seldict = dict(selectors)
  #:call join_lines(joinstr=", ", prefix="(", suffix=")")
    #:for i in range(1, origrank + 1)
      $:seldict.get(i, ":")
    #:endfor
  #:endcall
#:enddef
```
and use it as
```
#!  -> x(:, i, :)
x${select_subarray(3, [(2, 'i')])}$
```
It could also be used, if we need to reduce more than one rank, e.g.
```
#!  -> x(:, :, i, j)
x${select_subarray(4, [(3, 'i'), (4, 'j')])}$
```

Also the description should be clarified a bit.

src/common.fypp
```diff
@@ -113,4 +113,33 @@ ${prefix + joinstr.join([line.strip() for line in txt.split("\n")]) + suffix}$
...
+  #:assert origrank > 0
+  #:if origrank > 0
```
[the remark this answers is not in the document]
jvdp1: Thank you for your review. I changed for your proposition in the other comment.

src/common.fypp
jvdp1: Implemented as suggested. The proposed macro is more general and better fit to its aim.
Could you have another review, please?

src/stdlib_experimental_stats_var.fypp
```diff
@@ -0,0 +1,265 @@
...
+      #:set RName = rname("var_all",rank, t1, k1)
+      module function ${RName}$(x, mask) result(res)
+        ${t1}$, intent(in) :: x${ranksuffix(rank)}$
+        logical, intent(in), optional :: mask
+        real(${k1}$) :: res
+
+        real(${k1}$) :: n
+        ${t1}$ :: mean
+
+        if (.not.optval(mask, .true.)) then
```
nshaffer: This is a weird idiom to me. Here, I'd prefer the more obvious
```fortran
if (present(mask)) then
    if (mask .eqv. .false.) then
```
But this is a matter of style rather than substance.

src/stdlib_experimental_stats.md
```diff
@@ -47,3 +48,55 @@ program demo_mean
...
+`array`: Shall be an array of type `integer`, `real`, or `complex`.
+
+`dim`: Shall be a scalar of type `integer` with a value in the range from 1 to n, where n is the rank of `array`.
+
+`mask` (optional): Shall be of type `logical` and either by a scalar or an array of the same shape as `array`.
+
+### Return value
+
+If `array` is of type `real` or `complex`, the result is of the same type as `array`.
+If `array` is of type `integer`, the result is of type `double precision`.
```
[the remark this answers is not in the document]
jvdp1: @nshaffer Thank you for your review.

I used `double precision` because they are declared as `dp`. But I agree they are actually `real(real64)`. The issue with using `real64` in the spec is that if the definition of `dp` 
 in `stdlib_experimental_kinds` changes (there has been already discussions on that), then we will need to modify the spec too.

Would it be better to write "..... the result is of type `dp`."?

src/stdlib_experimental_stats_var.fypp
jvdp1: Hopefully none of both options will be needed in a future standard.

milancurcic:
I just revisited this PR. It looks like everybody approved it. We can revisit any outstanding minor issues in a later PR. Merging, thanks @jvdp1!.

Issue #112, opened 2020-01-18
Proposal for quadrature


nshaffer:
## Overview

Let's get numerical integration into stdlib. As a baseline, here's what I'd want in a general-purpose integration module.

- Interfaces to QUADPACK routines, both simple wrappers and a more modern interface a la SciPy's `scipy.integrate.quad`. 
- Routines to integrate arrays as if they are tabulated function values. At a minimum, we'd want trapezoid rule and Simpson's rule. 
- Some niche or less well-known rules that are still useful. Examples might be endpoint-corrected discrete Fourier transforms (see Numerical Recipes Sec. 13.9) and double-exponential rules.

The idea would be to provide easy-to-use and reasonably general numerical integration routines that cover 95% of what people need. Some things which are out of scope in my opinion:

- ODE integration
- Monte Carlo integration
- Basis representations of abstract integral operators

I envision having two modules: one being a light f90 wrapper for QUADPACK which expert users can use, the other being the main module with a nicer interface to QUADPACK's functionality, as well as array-integrating routines.

## Integrating functions

I've already written wrappers and tests for "non-expert" QUADPACK routines ([link](https://github.com/nshaffer/modern_quadpack)). All these do is create explicit interfaces and give names to magic numbers, e.g., error codes and integrand weight selectors. It will need a little refactoring to port to stdlib, but I am pretty confident that my module `quadpack_interface` uses all the right types, kinds, and intents to agree with the F77 routines. No need to duplicate that effort here.

From a modern programmer's perspective, QUADPACK has some unattractive qualities:

- Explicitly passed workspace arrays with sizes requirements that are documented but not enforced in code. These mattered in the days of tight memory constraints, but nowadays, we can relieve the caller of this burden.
- Out-parameters that are not usually useful should be made optional.
- Technical-detail in-parameters should be made optional with sensible default values.
- The integrand function must be a scalar function of one variable. This can make it clumsy to integrate functions with additional parameters, and doing multiple integrals by nested QUADPACK calls gets cumbersome.
- Routines for weighted integrands can be greatly simplified and clarified. For instance, `qaws` implements weight functions that remove algebraic/logarithmic singularities at the integration endpoints. The type of singularity is controlled by an integer argument, whose value affects the meaning of other arguments. 

[...]

PR #146, opened 2020-02-06
First steps toward quadrature

nshaffer:
Here is a starting point for the proposed quadrature module (#112)

- `trapz`, `trapz_weights`
  - Simple API, implementation, and tests for 1-d real arrays only
  - TODO: complex arrays (straightforward for 1-d)
  - TODO: multi-dimensional arrays (still some API questions to resolve)

- `simps`, `simps_weights`
  - API for 1-d real arrays
  - TODO: Implementation & tests
  - TODO: complex arrays
  - TODO: multi-dimensional arrays

I did not get as far as I hoped in the past week, put I want to get some baseline code up for others to work off of. 

Some points I'm hoping for some specific feedback on:
- Edge cases. Do we like the proposed handling of small arrays?
- The `even` argument of `simps` and `simps_weights`. Does this seem like a reasonable thing to do? Does the API make sense?


nshaffer:
@certik Hm, I'm of two minds about that. Sometimes you're testing that a function return a specific value, and it is incorrect behavior if it differs even by one bit. That's the case in `optval` and for my edge-case tests for `trapz`. But I do like the ideal of warning-free compilation. I've gone ahead and changed all the real equality tests here to tolerance checks against `epsilon`.

src/stdlib_experimental_quadrature_trapz.fypp
```diff
@@ -0,0 +1,86 @@
...
+        real(${KIND}$), dimension(size(x)) :: w
+
+        integer :: i
+        integer :: n
+
+        n = size(x)
+
+        select case (n)
+        case (0)
+            ! no action needed
```
jvdp1: I agree that "no action" is needed. But should the function not return somethin for `w` (e.g. NaN, a huge number, 0,...)? Currently `w` is undefined.

The other functions return 0 for `case (0)`.

nshaffer: For `trapz`, zero-size inputs behave the same as summing a zero-size array, producing zero. For `trapz_weights`, you want the result to behave such that `sum(w*x)` also produces zero. 

There are also practical challenges to returning anything besides a zero-size array. It's OK that it's value is undefined. That's just how zero-size arrays go.

Issue #149, opened 2020-02-19
What kind of variance should `var()` return?


milancurcic:
Playing with a toy example today, I was surprised to see the result (I guess I should have read the specification first, LOL). 

Program:

```fortran
use stdlib_experimental_stats, only: mean, var
real :: a(5) = [1, 2, 3, 4 ,5]
print *, var(a), mean((a - mean(a))**2)
end
```

I expected two identical numbers. Result:

```
   2.50000000       2.00000000    
```

Then I looked at the code for var() and saw that we're making an average by dividing the sum by (N - 1), rather than N. 

Then I looked at this issue and the spec, and it's all good: It says that the variance is defined in such way so that we're dividing by (N - 1). The code works as advertised.

But then I wondered why N - 1 and not N, and did some Google searching and found that there are all kinds of variances out there and that this particular flavor is called "population variance", or as described here the best unbiased estimator. Dividing by just N seems to be called just "variance". 

How we define this not only affects the numerical result, but also in some cases the behavior of the program: variance of a scalar is 0, but population variance of a scalar is a NaN. Some discussion [here](https://stats.stackexchange.com/questions/4068/how-should-one-define-the-sample-variance-for-scalar-input). 

NumPy's `np.var()` for example defines it as a simple variance (divide by N), but you can optionally set a "delta degrees of freedom", so that `np.var(x, ddof=1)` corresponds to the population variance.

I'm not a statistician but a wave physicist. I expect my variance to divide by N, and NumPy as served me well so far. 

**Question**: Should we consider adding optional "delta degrees of freedom" to make both statisticians and physicists happy?

certik:
I would do what NumPy does, and divide by `N`. Then provide an option, just like NumPy, to allow the `N-1` convention.

jvdp1:
This [Wikipedia page](https://en.wikipedia.org/wiki/Variance#Population_variance_and_sample_variance) gives a nice explanation about the differences sample variance vs population variance.

In Octave, [var](https://octave.sourceforge.io/nan/function/var.html) uses `N-1` as default (`opt = 0`), and `N` as an option (`opt  = 1`)., So it is divided by `N - 1 + opt`.

In [NumPy var](https://docs.scipy.org/doc/numpy/reference/generated/numpy.var.html), it is divided by `N - opt`, with `opt = 0` as default.

Both R and [Julia](https://docs.julialang.org/en/v1/stdlib/Statistics/) use `N-1` as default denominator.

[...]

PR #151, opened 2020-02-22
Addition of corrected in API of var (following #149)

jvdp1:
This PR is following the discussion in #149 .

### New syntax
`result = var(array [, mask[, corrected]])`
`result = var(array, dim [,mask[, corrected]])`

1. `corrected` is as last argument because `mask` is not always `optional` in the implementation (while `corrected` is always `optional`).

2. `corrected` is used as follows, e.g.:
```fortran
res = res / (n - merge(1._dp, 0._dp, optval(corrected, .true.)))
```
When `mask` is an array, the additional condition `.and. n > 0._dp` is needed to ensure that the result is `NaN` when all elements are `.false.` and `corrected == .true.`, as specified in the spec.
If `corrected == .true.` and `n==0`, then we got something like `res = res / -1`, which may lead to a result equal to `0`, e.g., in the following case if the additional condition is not added:

```fortran
 res = sum((x - mean)**2, mask) / (n -&
                merge(1._${k1}$, 0._${k1}$, (optval(corrected, .true.)) .and. n > 0._${k1}$))
```

Tests have been adapted to consider this scenario.

I also added some tests for `corrected==.false.`. However, it is now only for `real(sp)`. I can add more tests later in this PR when we will agree on its implementation.


jvdp1:
Thank you @milancurcic  for your comments. I modified the code following them; a bad habit to lose.

Issue #95, opened 2020-01-06
Error stack class/object


zbeekman:
I would like to implement an error stack class/object. This will be helpful for addressing #76 (message for errors inside of stdlib). The reasons why this is useful are that:

1. You can raise/signal errors deep within the call stack, even in `pure` (or maybe `elemental` too, I'll have to think more carefully about that) procedures, and then handle them at a higher level
2. You can maintain a call stack with the object and optionally include the file name, procedure name, and line number along with an error message and error type/severity
3. IO for warnings, errors, statuses, etc. can be performed once outside performance critical and/or pure/elemental procedures
4. You could hook up some logging mechanism too, so that the file IO happens on return from `pure` procedures

Perhaps this should be part of #72 (standard assert and macros) or the existing [`stdlib_experimental_error.f90`](https://github.com/fortran-lang/stdlib/blob/master/src/stdlib_experimental_error.f90) or perhaps it should be its own module.

I'm waiting, in part, for #69 (string handling routines) to get to a point where there's some consensus, to take a crack at implementing that before proposing an implementation for this. String handling will be important for the manipulations required.

## Sketch interface prototype

``` Fortran
pure subroutine push(this, file_name, line, procedure_name)
    !! caller calls this on line before call to callee
    !! `file_name` can be set to a compile-time constant via CMake definition
    !! or can use `__FILE__` macro, but this may overrun the line length
    !! `line` can be passed `__LINE__` macro.
    class(error_stack_t), intent(inout) :: this
    character(len=*), intent(in) :: file_name
    integer, intent(in) :: line
    character(len=*), intent(in), optional :: procedure_name
end subroutine push
```

I like to let CMake ad a per-source-file definition, something like `THIS_FILE` that holds the relative path to the file in the source directory. It could fallback to `__FILE__` when using manual makefiles. The benefit over `__FILE__` is two fold:

1. Relative paths are shorter than absolute paths and less likely to overrun the line length limit, and CMake can check the length and insert a newline w/ continuation if needed
2. Embedding absolute paths prevents builds from being reproducible and can make package managers thing that the software is not relocatable when it actually is.

Also, one or more arguments could be made optional. It would be wonderful if there was a way to query the name of the current scope in Fortran and if the programmer didn't need to make two separate calls, one to the callee and one push info onto the error stack. You could add file name, line number and error_stack arguments to all procedures and then the callee could push the data onto the stack as an alternative to the caller doing it.

Feedback, advice and suggestions welcome here.

[...]

PR #153, opened 2020-03-01
Addition of a function to compute the k-th order central moment of an array

jvdp1:
Related to #113

In this __draft pull request__ I propose the implementation of a function for the k-th  order central moment of an array.

## Other languages
[Matlab moment](https://www.mathworks.com/help/stats/moment.html)
[SciPy moment](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.moment.html)
[Julia moment](https://juliastats.org/StatsBase.jl/v0.21/scalarstats.html#StatsBase.moment)
[R moment](https://www.rdocumentation.org/packages/e1071/versions/1.7-3/topics/moment)
[Octave moment](https://octave.sourceforge.io/octave/function/moment.html)

Note: The functions of R and Octave can also return absolute central moment, raw moment, and absolute raw moment.

### Description

Returns the _k_-th order central moment of all the elements of `array`, or of the elements of `array` along dimension `dim` if provided, and if the corresponding element in `mask` is `true`.

The _k_-th order central moment is defined as :

```
 moment(array) = 1/n sum_i (array(i) - mean(array))^k
```

where n is the number of elements.

### Syntax

`result = moment(array, order [, mask])`

`result = moment(array, order, dim [, mask])`

### Arguments

`array`: Shall be an array of type `integer`, `real`, or `complex`.

`order`: Shall be an scalar of type `integer`.

`dim`: Shall be a scalar of type `integer` with a value in the range from 1 to n, where n is the rank of `array`.

`mask` (optional): Shall be of type `logical` and either by a scalar or an array of the same shape as `array`.

### Return value

If `array` is of type `real` or `complex`, the result is of the same type as `array`.
If `array` is of type `integer`, the result is of type `real(dp)`.

If `dim` is absent, a scalar with the _k_-th central moment of all elements in `array` is returned. Otherwise, an array of rank n-1, where n equals the rank of `array`, and a shape similar t
o that of `array` with dimension `dim` dropped is returned.

If `mask` is specified, the result is the _k_-th central moment of all elements of `array` corresponding to `true` elements of `mask`. If every element of `mask` is `false`, the result is IE
EE `NaN`.

### Example

```fortran
program demo_moment
    use stdlib_experimental_stats, only: moment
    implicit none
    real :: x(1:6) = [ 1., 2., 3., 4., 5., 6. ]
    print *, moment(x, 2)                            !returns 2.9167
    print *, moment( reshape(x, [ 2, 3 ] ), 2)       !returns 2.9167
    print *, moment( reshape(x, [ 2, 3 ] ), 2, 1)    !returns [0.25, 0.25, 0.25]
    print *, moment( reshape(x, [ 2, 3 ] ), 2, 1,&
                     reshape(x, [ 2, 3 ] ) > 3.)     !returns [NaN, 0., 0.25]
end program demo_moment
```

## Questions for discussion

1. Agreement on API?
2. Agreement on implementation (e.g. for complex numbers)?
3. Should we support (absolute)(raw/central) moments in this function? E.g. through optional logicals `absolute` and `raw`? (see example in [this branch](https://github.com/jvdp1/stdlib/tree/optional_moment_dev) )

### DIFF
--- a/src/stdlib_experimental_stats_moment.fypp
+++ b/src/stdlib_experimental_stats_moment.fypp
@@ -1,0 +1,261 @@
+#:include "common.fypp"
+#:set RANKS = range(1, MAXRANK + 1)
+#:set RC_KINDS_TYPES = REAL_KINDS_TYPES + CMPLX_KINDS_TYPES
+submodule (stdlib_experimental_stats) stdlib_experimental_stats_moment
+
+  use, intrinsic:: ieee_arithmetic, only: ieee_value, ieee_quiet_nan
+  use stdlib_experimental_error, only: error_stop
+  use stdlib_experimental_optval, only: optval
+  implicit none
+
+contains
+
+  #:for k1, t1 in RC_KINDS_TYPES
+    #:for rank in RANKS
+      #:set RName = rname("moment_all",rank, t1, k1)
+      module function ${RName}$(x, order, mask) result(res)
+        ${t1}$, intent(in) :: x${ranksuffix(rank)}$
+        integer, intent(in) :: order
+        logical, intent(in), optional :: mask
+        ${t1}$ :: res
+
+        real(${k1}$) :: n
+        ${t1}$ :: mean
+
+        if (.not.optval(mask, .true.)) then
+          res = ieee_value(1._${k1}$, ieee_quiet_nan)
+          return
+        end if
+
+        n = size(x, kind = int64)
+        mean = sum(x) / n
+
+        res = sum((x - mean)**order) / n
+
+      end function ${RName}$
+    #:endfor
+  #:endfor
+
+
+  #:for k1, t1 in INT_KINDS_TYPES
+    #:for rank in RANKS
+      #:set RName = rname("moment_all",rank, t1, k1, 'dp')
+      module function ${RName}$(x, order, mask) result(res)
+        ${t1}$, intent(in) :: x${ranksuffix(rank)}$
+        integer, intent(in) :: order
+        logical, intent(in), optional :: mask
+        real(dp) :: res
+
+        real(dp) :: n, mean
+
+        if (.not.optval(mask, .true.)) then
+          res = ieee_value(1._dp, ieee_quiet_nan)
+          return
+        end if
+
+        n = size(x, kind = int64)
+        mean = sum(real(x, dp)) / n
+
+        res = sum((real(x, dp) - mean)**order) / n
+
+      end function ${RName}$
+    #:endfor
+  #:endfor
+
+
+  #:for k1, t1 in RC_KINDS_TYPES
+    #:for rank in RANKS
+      #:set RName = rname("moment",rank, t1, k1)
+      module function ${RName}$(x, order, dim, mask) result(res)
+        ${t1}$, intent(in) :: x${ranksuffix(rank)}$
+        integer, intent(in) :: order
+        integer, intent(in) :: dim
+        logical, intent(in), optional :: mask
+        ${t1}$ :: res${reduced_shape('x', rank, 'dim')}$
+
+        integer :: i
+        real(${k1}$) :: n
+        ${t1}$ :: mean${reduced_shape('x', rank, 'dim')}$
+
+        if (.not.optval(mask, .true.)) then
+          res = ieee_value(1._${k1}$, ieee_quiet_nan)
+          return
+        end if
+
+        res = 0
+        select case(dim)
+          #:for fi in range(1, rank+1)
+          case(${fi}$)
+            n = size(x, dim)
+            mean = sum(x, dim) / n
+            do i = 1, size(x, dim)
+              res = res + (x${select_subarray(rank, [(fi, 'i')])}$ - mean)**order
+            end do
+          #:endfor
+          case default
+            call error_stop("ERROR (mean): wrong dimension")
+        end select
+        res = res / n
+
+      end function ${RName}$
+    #:endfor
+  #:endfor
+
+
+  #:for k1, t1 in INT_KINDS_TYPES
+    #:for rank in RANKS
+      #:set RName = rname("moment",rank, t1, k1, 'dp')
+      module function ${RName}$(x, order, dim, mask) result(res)
+        ${t1}$, intent(in) :: x${ranksuffix(rank)}$
+        integer, intent(in) :: order
+        integer, intent(in) :: dim
+        logical, intent(in), optional :: mask
+        real(dp) :: res${reduced_shape('x', rank, 'dim')}$
+
+        integer :: i
+        real(dp) :: n
+        real(dp) :: mean${reduced_shape('x', rank, 'dim')}$
+
+        if (.not.optval(mask, .true.)) then
+          res = ieee_value(1._dp, ieee_quiet_nan)
+          return
+        end if
+
+        res = 0
+        select case(dim)
+          #:for fi in range(1, rank+1)
+          case(${fi}$)
+            n = size(x, dim)
+            mean = sum(real(x, dp), dim) / n
+            do i = 1, size(x, dim)
+              res = res + (real(x${select_subarray(rank, [(fi, 'i')])}$, dp) - mean)**order
+            end do
+          #:endfor
+          case default
+            call error_stop("ERROR (mean): wrong dimension")
+        end select
+        res = res / n
+
+      end function ${RName}$
+    #:endfor
+  #:endfor
+
+
+  #:for k1, t1 in RC_KINDS_TYPES
+    #:for rank in RANKS
+      #:set RName = rname("moment_mask_all",rank, t1, k1)
+      module function ${RName}$(x, order, mask) result(res)
+        ${t1}$, intent(in) :: x${ranksuffix(rank)}$
+        integer, intent(in) :: order
+        logical, intent(in) :: mask${ranksuffix(rank)}$
+        ${t1}$ :: res
+
+        real(${k1}$) :: n
+        ${t1}$ :: mean
+
+        n = count(mask, kind = int64)
+        mean = sum(x, mask) / n
+
+        res = sum((x - mean)**order, mask) / n
+
+      end function ${RName}$
+    #:endfor
+  #:endfor
+
+
+  #:for k1, t1 in INT_KINDS_TYPES
+    #:for rank in RANKS
+      #:set RName = rname("moment_mask_all",rank, t1, k1, 'dp')
+      module function ${RName}$(x, order, mask) result(res)
+        ${t1}$, intent(in) :: x${ranksuffix(rank)}$
+        integer, intent(in) :: order
+        logical, intent(in) :: mask${ranksuffix(rank)}$
+        real(dp) :: res
+
+        real(dp) :: n, mean
+
+        n = count(mask, kind = int64)
+        mean = sum(real(x, dp), mask) / n
+
+        res = sum((real(x, dp) - mean)**order, mask) / n
+
+      end function ${RName}$
+    #:endfor
+  #:endfor
+
+
+  #:for k1, t1 in RC_KINDS_TYPES
+    #:for rank in RANKS
+      #:set RName = rname("moment_mask",rank, t1, k1)
+      module function ${RName}$(x, order, dim, mask) result(res)
+        ${t1}$, intent(in) :: x${ranksuffix(rank)}$
+        integer, intent(in) :: order
+        integer, intent(in) :: dim
+        logical, intent(in) :: mask${ranksuffix(rank)}$
+        ${t1}$ :: res${reduced_shape('x', rank, 'dim')}$
+
+        integer :: i
+        real(${k1}$) :: n${reduced_shape('x', rank, 'dim')}$
+        ${t1}$ :: mean${reduced_shape('x', rank, 'dim')}$
+
+        res = 0
+        select case(dim)
+          #:for fi in range(1, rank+1)
+          case(${fi}$)
+            n = count(mask, dim)
+            mean = sum(x, dim, mask) / n
+            do i = 1, size(x, dim)
+              res = res + merge( (x${select_subarray(rank, [(fi, 'i')])}$ - mean)**order,&
+                #:if t1[0] == 'r'
+                  0._${k1}$,&
+                #:else
+                  cmplx(0,0,kind=${k1}$),&
+                #:endif
+                  mask${select_subarray(rank, [(fi, 'i')])}$)
+            end do
+          #:endfor
+          case default
+            call error_stop("ERROR (mean): wrong dimension")
+        end select
+        res = res / n
+
+      end function ${RName}$
+    #:endfor
+  #:endfor
+
+
+  #:for k1, t1 in INT_KINDS_TYPES
+    #:for rank in RANKS
+      #:set RName = rname("moment_mask",rank, t1, k1, 'dp')
+      module function ${RName}$(x, order, dim, mask) result(res)
+        ${t1}$, intent(in) :: x${ranksuffix(rank)}$
+        integer, intent(in) :: order
+        integer, intent(in) :: dim
+        logical, intent(in) :: mask${ranksuffix(rank)}$
+        real(dp) :: res${reduced_shape('x', rank, 'dim')}$
+
+        integer :: i
+        real(dp) :: n${reduced_shape('x', rank, 'dim')}$
+        real(dp) :: mean${reduced_shape('x', rank, 'dim')}$
+
+        res = 0
+        select case(dim)
+          #:for fi in range(1, rank+1)
+          case(${fi}$)
+            n = count(mask, dim)
+            mean = sum(real(x, dp), dim, mask) / n
+            do i = 1, size(x, dim)
+              res = res + merge((real(x${select_subarray(rank, [(fi, 'i')])}$, dp) - mean)**order,&
+                                  0._dp, mask${select_subarray(rank, [(fi, 'i')])}$)
+            end do
+          #:endfor
+          case default
+            call error_stop("ERROR (mean): wrong dimension")
+        end select
+        res = res / n
+
+      end function ${RName}$
+    #:endfor
+  #:endfor
+
+end submodule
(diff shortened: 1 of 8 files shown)


jvdp1:
I just implemented (absolute) central/raw moments in [this branch](https://github.com/jvdp1/stdlib/tree/optional_moment_dev). Personally I only use central moments, which is the default returned value for most packages I checked (that sometimes supports only the central moment).

Any thoughts @ivan-pi @aradi @fiolj @leonfoks?

ivan-pi:
I noticed the [Scipy version](https://github.com/scipy/scipy/blob/v1.4.1/scipy/stats/stats.py#L890-L970) uses [exponentiation by squares](https://eli.thegreenplace.net/2009/03/21/efficient-integer-exponentiation-algorithms) for efficiency. Do you think in Fortran we can simply trust the compiler to do the right thing?

Second thing I am wondering is what is the sensitivity of this function to underflow/overflow? The reason I am asking is because I recall reading [a paper](https://dl.acm.org/doi/abs/10.1145/3134441?download=true) by Hanson and Hopkin concerning some subtleties of the intrinsic Fortran `norm2`.

Concerning the API, the SciPy version allows to pass a list of moments orders, e.g.
```
real :: x(1:5) = [ 1., 2., 3., 4., 5. ]
print *, moment(x, [1, 2]) ! prints 0 2
```      
Since I have not used this function before, I don't know whether this is particularly useful or not? It sure would complicate the allocation of a LHS variable...

ivan-pi:
> 3. Should we support (absolute)(raw/central) moments in this function? E.g. through optional logicals `absolute` and `raw`? (see example in [this branch](https://github.com/jvdp1/stdlib/tree/optional_moment_dev) )

Concerning this point, I find the Octave solution of passing a character ("c", "a", "ac", "r", "ar") to an optional `type` dummy variable easier than two logicals:
```
real :: x(1:5) = [ 1., 2., 3., 4., 5. ]
print *, moment(x, 2, type='ar')
print *, moment(x, 2, absolute=.true.,raw=.true.)
```

jvdp1:
Thank you @ivan-pi for the comments.

> I noticed the [Scipy version](https://github.com/scipy/scipy/blob/v1.4.1/scipy/stats/stats.py#L890-L970) uses [exponentiation by squares](https://eli.thegreenplace.net/2009/03/21/efficient-integer-exponentiation-algorithms) for efficiency. Do you think in Fortran we can simply trust the compiler to do the right thing?

I hope so. At least I always trust it, while never check it. If it is not the case, it could a nice procedure for `stdlib`, e.g., `power(x, order)` that would perform something similar to Scipy.
 
> Second thing I am wondering is what is the sensitivity of this function to underflow/overflow? The reason I am asking is because I recall reading [a paper](https://dl.acm.org/doi/abs/10.1145/3134441?download=true) by Hanson and Hopkin concerning some subtleties of the intrinsic Fortran `norm2`.

Good point. The function mainly relies on intrinsic functions (e.g., `sum`).

I think that this point is applicable to all procedures in `stdlib` (e.g., `mean`, `var`, `trapz`,....). A possible solution (already proposed in another post; I don't remember which one) would be to have a standard implementation of the function (e.g., `moment`), and followed by other optimized/specific implementations of the same function (e.g., `moment_largearray`????)

> Concerning the API, the SciPy version allows to pass a list of moments orders, e.g.
> 
> ```
> real :: x(1:5) = [ 1., 2., 3., 4., 5. ]
> print *, moment(x, [1, 2]) ! prints 0 2
> ```
> 
> Since I have not used this function before, I don't know whether this is particularly useful or not? It sure would complicate the allocation of a LHS variable...

I am not in favor of such a behavior (the user can call several times the same function, with different order). The main issue would be the dimension of the result (what would happen if only one value is given? Is the result a scalar (`res`) or a vector (`res(1)`)? However, I may miss something.

jvdp1:
> > 1. Should we support (absolute)(raw/central) moments in this function? E.g. through optional logicals `absolute` and `raw`? (see example in [this branch](https://github.com/jvdp1/stdlib/tree/optional_moment_dev) )
> 
> Concerning this point, I find the Octave solution of passing a character ("c", "a", "ac", "r", "ar") to an optional `type` dummy variable easier than two logicals:
> 
> ```
> real :: x(1:5) = [ 1., 2., 3., 4., 5. ]
> print *, moment(x, 2, type='ar')
> print *, moment(x, 2, absolute=.true.,raw=.true.)
> ```
Nice suggestion. I agree that using characters may simplify the API. 

Here is my reasoning to propose logicals and not characters: 
* with logicals (`absolute` and `raw`), only 4 combinations are possible. With characters (e.g., `a`, `c`, and `r` as in Octave), many combinations are possible (there are even 5 combinations in Octave for 4 combinations allowed). This implies that the function will have 1) to check if the provided combination is allowed (e.g., `ar` and `ra` would be allowed and are the same, while `cr` or `arc` are not allowed) (so, 5 cases with characters, instead of 4 with logicals), and 2) to report an error message that the combination is incorrect. The latter may limit the implementation of `pure` functions. Using characters and its additional checks may also impact performance.

* it follows the API of `variance` that uses a `logical` to allow the Bessel 's correction of not. 

Again, I also find that the API with characters is nicer, but have some (unfounded?) concerns about performance.

ivan-pi:
Thank you for explaining these points. As long as the API does not change, we can always modify for performance/accuracy later, or as you suggest add specialized versions. 

I agree with your concern about the dimension of the result. MATLAB does not have it either, so I suppose there is no need really. The current behavior is also fairly easy to remember, for the case of computing the moment along a single axis, the user simply needs to remember, the target array is one rank smaller than the original array:
```
real :: array(:,:,:,:)
real, allocatable :: mom2(:,:,:)
mom2 = moment(array,order=2,dim=1)
```
For array slices, or chained calls to moments it  becomes more tricky (I don't know whether there are such use cases).

The purpose of this function is primarily statistics, right? I have used [central moments](https://en.wikipedia.org/wiki/Image_moment#Central_moments) before for image analysis purposes (e.g. to locate the centroid or area of an object), but I think a different function would be necessary for this.

ivan-pi:
> Nice suggestion. I agree that using characters may simplify the API.
> 
> Here is my reasoning to propose logicals and not characters:
> 
>   * with logicals (`absolute` and `raw`), only 4 combinations are possible. With characters (e.g., `a`, `c`, and `r` as in Octave), many combinations are possible (there are even 5 combinations in Octave for 4 combinations allowed). This implies that the function will have 1) to check if the provided combination is allowed (e.g., `ar` and `ra` would be allowed and are the same, while `cr` or `arc` are not allowed) (so, 5 cases with characters, instead of 4 with logicals), and 2) to report an error message that the combination is incorrect. The latter may limit the implementation of `pure` functions. Using characters and its additional checks may also impact performance.
> 
>   * it follows the API of `variance` that uses a `logical` to allow the Bessel 's correction of not.
> 
> 
> Again, I also find that the API with characters is nicer, but have some (unfounded?) concerns about performance.

I think your concerns about performance are unjustified. If you check the routines in other languages they performs checks for NaN, dimensions, etc., but they are still more popular than Fortran (simply because they provide these functions). LAPACK also uses characters in the API to specify the form of the system of equations (transpose, no transpose...), but they return the `info` argument to inform about incorrect usage.  Plus, the routines already stop if the dimension exceeds the rank of the array, meaning they can not be pure (at least until Fortran introduces some assert facility that could be used in pure routines). 

I do agree with your concern about the multiple character combinations. I suppose renaming the `absolute` dummy variable to the shorter `abs` is not possible, because it would clash with the intrinsic function? (Is it possible to rename intrinsic functions?) Introducing two new functions for `rawmoment` and `absmoment` is less appealing than the current proposition.

jvdp1:
> The purpose of this function is primarily statistics, right? I have used [central moments](https://en.wikipedia.org/wiki/Image_moment#Central_moments) before for image analysis purposes (e.g. to locate the centroid or area of an object), but I think a different function would be necessary for this.

Indeed, it is primarily statistics. I think the function you mentioned is an extension to 2D.

>I think your concerns about performance are unjustified. If you check the routines in other languages they performs checks for NaN, dimensions, etc., but they are still more popular than Fortran (simply because they provide these functions).

You convinced me. I will start an implementation with characters, instead of logicals. Then we can further discuss pros and cons based on the code.

However, it could be easier to first merge this PR (if there is a consensus about it), and then submit another PR with the different moments.

> LAPACK also uses characters in the API to specify the form of the system of equations (transpose, no transpose...), but they return the info argument to inform about incorrect usage. Plus, the routines already stop if the dimension exceeds the rank of the array, meaning they can not be pure (at least until Fortran introduces some assert facility that could be used in pure routines).

#95 could be usefull in these cases. 
On the other hand, we could also implement an optional argument `info` in the different functions (`mean`, `var`,...). However, such an argument would be useless for some implementations, and the API of these functions would not be the smae/similar to the API of intrinsic functions (like `sum`).

> Introducing two new functions for rawmoment and absmoment is less appealing than the current proposition.

I agree.

ivan-pi:
Looking at the source code closer, the error message for "wrong dimension" currently refers to the `mean` function. 

---

> However, it could be easier to first merge this PR (if there is a consensus about it), and then submit another PR with the different moments.

As the raw and absolute do not break backward compatibility of API I think this might be easier too. It would be nice to get an opinion from the remaining developers concerning their preference between logicals vs characters for option passing.

jvdp1:
> After taking a closer look at the source code, the error message for "wrong dimension" currently refers to the `mean` function.

Thank you for reviewing the code. It is now corrected as suggested.

>  It would be nice to get an opinion from the remaining developers concerning their preference between logicals vs characters for option passing.

I agree! It needs more discussions.

fiolj:
Sorry to barging in late to the discussion. I do not have strong feelings about the function since I do not usually use it.

  I do not want to disrupt the implementation when it already has a lot of work in it but there are two points that are not clear.

  First, probably the simplest is that I do not understand what happens when a scalar mask is provided. I though that perhaps for functions as `sum` it is used to avoid an extra `if` statement.

  Secondly, when started to ponder about the interface with logical vs character flags I went to look to other implementations and started to think on the usefulness of the options.

  Because is not my area of expertise, I only was able to think of 


  ## Some Applications ##

  - Statistics (centered)
  - System of particles
	- Center of mass: need weights (mass)
	- Kinetic energy: need weights (sqrt(mass))
	- Internal kinetic energy: need weigths (sqrt(mass)) and centered on v_cm (different from mean)
    - Moments of inertia (also centered in places other than the mean)
	

For these reasons I would favor (slightly) an API more similar to Julia in that the center be given as a float point number (or rather a number of the same type than the array).

	`result = moment(array, order [, center] [, mask])`
	`result = moment(array, order, dim [, center] [, mask])`
	
Thus, we only would have a remaining flag for absolute, which (because of my ignorance) I do not know how used/needed is.

Because of the advanced state of the current pull request, I'd be perfectly happy to leave it as it is now, and if there is any interest keep the discussion for a future next iteration.

jvdp1:
>Sorry to barging in late to the discussion. I do not have strong feelings about the function since I do not usually use it.

>I do not want to disrupt the implementation when it already has a lot of work in it but there are two points that are not clear.

Thank you for your comments @fiolj . There is no problem! I mainly implemented it to facilitate discussion.

> First, probably the simplest is that I do not understand what happens when a scalar mask is provided. I though that perhaps for functions as sum it is used to avoid an extra if statement.

Personally I do not understand the usefullness of a scalar logical, because it will always return `NaN` when it is `.false.`. I probably miss something there. So it was introduced in `mean`, `var` and `moment` to have the same API as the intrinsic function `sum`, which accept a scalar logical, and to be consistent with the standard.

Regarding the implementation, it could be indeed passed to the intrinsic function `sum`, instead of having the additional condition. I implemented it that way to avoid unnecessary computations when `mask = .false.`.

>Secondly, when started to ponder about the interface with logical vs character flags I went to look to other implementations and started to think on the usefulness of the options.
...
>For these reasons I would favor (slightly) an API more similar to Julia in that the center be given as a float point number (or rather a number of the same type than the array).
```
result = moment(array, order [, center] [, mask])
result = moment(array, order, dim [, center] [, mask])
```

This is a nice idea. `center` should be a scalar when `dim` is not provided, and an array of rank-1 when `dim` is provided (Julia does not allow to specify a `dim`). Not providing `center` would lead to the raw moment, right? Using the function `mean`, something like that could be written:
```fortran
result = moment(x, 2) ! raw moment
result = moment(x, 2, center = mean(x))  ! central moment
result = moment(x, 2, dim = 3, center = mean(x, 3, mask=(x >3)), mask = (x >3)) !central moment
```
Your proposed API adds some flexibility to the function `moment` and might interest a broader audience, So I would be in favor of such an API.
For completness, I suggest to add a flag for absolute moments.
So the API could be:
```
result = moment(array, order [, center] [,absolute] [, mask])
result = moment(array, order, dim [, center] [,absolute] [, mask])
```

What are your opinions about this API? @fiolj @ivan-pi @certik @milancurcic 

Re: weights, the aim was to implement a function for the [moment in terms of mathematics/statistics](https://en.wikipedia.org/wiki/Moment_(mathematics)). However, if the addition of an argument can extend its used in physics, I would agree.

src/stdlib_experimental_stats_moment.fypp
```diff
@@ -0,0 +1,265 @@
...
+
+        if (.not.optval(mask, .true.)) then
+          res = ieee_value(1._${k1}$, ieee_quiet_nan)
+          return
+        end if
+
+        n = size(x, kind = int64)
+        mean = sum(x) / n
+
+        res = sum((x - mean)**order) / n
```
milancurcic: Why not

```fortran
res = mean((x - mean(x))**order)
```

here? Is there a concern about the overhead from calling `mean()`?

jvdp1: It was indeed to avoid the overhead of calling `mean` (especially when `mean` is an array).

jvdp1:
> Sorry for the late review. This is excellent, thank you. I especially appreciate the hyperlinked TOC in the spec file.
> 
> I only had one question about the implementation. Otherwise, good to go.

Thank you @milancurcic for the review. Any thoughs about a different API:
```fortran
result = moment(array, order [, center] [,absolute] [, mask])
result = moment(array, order, dim [, center] [,absolute] [, mask])
```
where `center` would replace the computation of the `mean`, and `absolute` would be an optional logical for computing absolute moments?

fiolj:
I've started some simple timing tests with gfortran on the overhead of using mean(), I do not have conclusive results yet, but noticed a few (minor) points in the code. Some of these will change with the new API implementation, but I mention them anyway.

`n` is defined real, so we could do here the same that `mean()` does and convert explicitly the size to real:

```
@@ -53,7 +53,7 @@ contains
           return
         end if
 
-        n = size(x, kind = int64)
+        n = real(size(x, kind = int64), dp)
         mean = sum(real(x, dp)) / n

         res = sum((real(x, dp) - mean)**order) / n

``` 
Also, we could put all calculations under the right case (although it perhaps may become quite irrelevant with the new API, unless we decide to use "center moment" as a default)

```
@@ -82,11 +82,13 @@ contains
           return
         end if
 
-        n = size(x, dim)
-        mean = sum(x, dim) / n
-
-        res = 0
         select case(dim)
+
+          n = real(size(x, kind = int64), ${k1}$)
+          mean = sum(x, dim) / n
+
+          res = 0
+
           #:for fi in range(1, rank+1)
           case(${fi}$)
             do i = 1, size(x, dim)

```

jvdp1:
>   n = size(x, dim)
> -        mean = sum(x, dim) / n
> -
> -        res = 0
>          select case(dim)
> +
> +          n = real(size(x, kind = int64), ${k1}$)
> +          mean = sum(x, dim) / n
> +



> I've started some simple timing tests with gfortran on the overhead of using mean(), I do not have conclusive results yet, but noticed a few (minor) points in the code. Some of these will change with the new API implementation, but I mention them anyway.
> 
> `n` is defined real, so we could do here the same that `mean()` does and convert explicitly the size to real:
> 
> ```
> @@ -53,7 +53,7 @@ contains
>            return
>          end if
>  
> -        n = size(x, kind = int64)
> +        n = real(size(x, kind = int64), dp)
>          mean = sum(real(x, dp)) / n
> 
>          res = sum((real(x, dp) - mean)**order) / n

Funny. I implemented it in the way you proposed for `var` (see [comment](https://github.com/fortran-lang/stdlib/pull/151#pullrequestreview-363102731) ), but @milancurcic suggested to avoid explicit type casts, mainly to improve readability. The compiler should do the job.

milancurcic:
The API with optional `absolute` and `center` arguments seems reasonable to me. I am not the target audience for these though; I only ever calculate moments where `center == mean(x)`. So I can't speak for the usefulness of it.

I suggest to:

1. Merge the basic implementation (this PR) into master;
2. Open a PR for optional `center` and `absolute` (perhaps even separate PRs if they're independent);

Small and specific PRs will allow for a focused discussion.

certik:
I agree with @milancurcic about how to move forward. I am not a target audience either for this, so I do not have a strong opinion either.

jvdp1:
@milancurcic @certik Thank you for your advices. I will merge this PR.
I will later submit other PRs for optional `center` and `absolute` as discussed in this PR with @fiolj and @ivan-pi .

@fiolj I would be interested by your results regarding the overhead of calling `mean()` inside another function of `stdlib`, especially when the result is a large array (see #114 for a discussion on that topic). These results might help us on how to implement stat functions `skewness()` and `kurtosis()` because both need multiple computations of `mean`.

fiolj:
> @fiolj I would be interested by your results regarding the overhead of calling `mean()` inside another function of `stdlib`, especially when the result is a large array (see #114 for a discussion on that topic). These results might help us on how to implement stat functions `skewness()` and `kurtosis()` because both need multiple computations of `mean`.

A (naive) comparison of times for moment is in:

https://gist.github.com/fiolj/7ebb1611e200d1fcd248b5866f851b76

Strangely enough, I am not finding differences between using `mean()` and the intrinsic `sum()`.
I've just copied the implementation of mean and moment for a few (1D, 2D, 5D, real(dp)) cases and timed them with `cpu_time()`.
Please, somebody check that these are valid tests of speed, I may be missing something.

jvdp1:
Thank you @fiolj for these time comparisons.
I added a loop to repeat 100 times the computations in the last part. No difference either.
There are maybe a bit naive and we should probably use some tools to get better numbers. But at least it seems to show that the overhead of calling the function`mean()` is very small. Good to know. 
Thank you for the tests.

PR #155, opened 2020-03-09
small_corrections: small issues detected when compiled with ifort17.0.4

jvdp1:
Some small issues detected when compiled with ifort 17.0.4

### DIFF
--- a/src/tests/stats/test_var.f90
+++ b/src/tests/stats/test_var.f90
@@ -154,7 +154,7 @@ program test_var
     call assert( all( abs( var(d, 1, d < 11) -&
                  [20._dp / 3._dp, 20._dp / 3._dp, 0.5_dp]) < dptol))
     call assert( all( abs( var(d, 2, d < 11) -&
-                 [19.0_dp, 43._dp / 3._dp, 0.5 _dp, 0.5_dp]) < dptol))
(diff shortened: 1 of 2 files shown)


src/tests/stats/test_var.f90
```diff
@@ -154,7 +154,7 @@ program test_var
     call assert( all( abs( var(d, 1, d < 11) -&
                  [20._dp / 3._dp, 20._dp / 3._dp, 0.5_dp]) < dptol))
     call assert( all( abs( var(d, 2, d < 11) -&
-                 [19.0_dp, 43._dp / 3._dp, 0.5 _dp, 0.5_dp]) < dptol))
+                 [19.0_dp, 43._dp / 3._dp, 0.5_dp, 0.5_dp]) < dptol))
```
certik: How is it that this even compilers with gfortran?

jvdp1: I have no idea. I tested different versions of gfortran and it compiled fine with all of them.

certik: (I wonder if gfortran makes the space in `0.5 _dp` optional.)

jvdp1:
Good. I will merge it.

PR #157, opened 2020-03-13
Update CMAKE files

leonfoks:
I've updated the CMAKE build procedure to a similar setup that I have in Coretran.

Parsed out compiler flag options into different files for easier reading
Added DEBUG and RELEASE mode with DEBUG default.
Moved test dir to /bin
Moved .mod files to /include
Moved built libraries (.o .a .dylib etc) to /lib
Added OpenMP finder (we might not need this thought)
Added new directories to gitignore.

CMake is a pain for people who are not used to it. 
So I print extra information to the screen explaining the options chosen,
the flags being used etc. and how to change these options.

Added more information when using CMAKE to help users when compiling.

TODO: Add a compile flags file for PGI
TODO: Add a compile flags file for CRAY? I can do this.
TODO: Add information about how to change CMAKE_MAXIMUM_RANK - Easy

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,17 +1,34 @@
 cmake_minimum_required(VERSION 3.14.0)
+
 project(stdlib Fortran)
 enable_testing()
 
-include(${CMAKE_SOURCE_DIR}/cmake/stdlib.cmake)
+# Add our local modules to the module path
+if(NOT CMAKE_MODULE_PATH)
+  set(CMAKE_MODULE_PATH)
+endif()
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
+
+# Include the preprocessor options for stdlib
+include(${CMAKE_MODULE_PATH}/stdlib.cmake)
+
+# GNUInstallDirs is used to install libraries into correct locations
+# on all platforms.
+include(GNUInstallDirs)
+
+# Setup the Fortran cmake environment
+# Sets up the flags for compiling shared libraries or static libraries, and returns the libType parameter.
+include("${CMAKE_MODULE_PATH}/FortranEnvironment.cmake")
+
 
 # --- compiler options
-if(CMAKE_Fortran_COMPILER_ID STREQUAL GNU)
-  add_compile_options(-fimplicit-none)
-elseif(CMAKE_Fortran_COMPILER_ID STREQUAL Intel)
-  add_compile_options(-warn declarations)
-elseif(CMAKE_Fortran_COMPILER_ID STREQUAL PGI)
-  add_compile_options(-Mdclchk)
-endif()
+#if(CMAKE_Fortran_COMPILER_ID STREQUAL GNU)
+#  add_compile_options(-fimplicit-none)
+#elseif(CMAKE_Fortran_COMPILER_ID STREQUAL Intel)
+#  add_compile_options(-warn declarations)
+#elseif(CMAKE_Fortran_COMPILER_ID STREQUAL PGI)
+#  add_compile_options(-Mdclchk)
+#endif()
 
 # --- compiler feature checks
 include(CheckFortranSourceCompiles)
@@ -27,7 +44,8 @@ endif()
 # --- find preprocessor
 find_program(FYPP fypp)
 if(NOT FYPP)
-  message(FATAL_ERROR "Preprocessor fypp not found!")
+  message(FATAL_ERROR "Preprocessor fypp not found!\n"
+	              "fypp is a python package and can be installed via, 'pip install fypp'")
 endif()
 
 add_subdirectory(src)
--- a/cmake/FortranEnvironment.cmake
+++ b/cmake/FortranEnvironment.cmake
@@ -1,0 +1,118 @@
+# ================================
+# CMAKE Script for setting up the fortran compiling and linking flags for different operating systems and compilers
+
+# Call this cmake file from your project's CMakeLists.txt using 'include(/path/to/FortranEnvironment.cmake)'
+
+# ++++++++++++++++++++++++++++++++
+
+
+enable_language(Fortran)
+
+option (BUILD_SHARED_LIBS "Shared or static libraries"  ON)
+
+# Check if linux
+if(UNIX AND NOT APPLE)
+  set(LINUX TRUE)
+endif()
+
+
+# ================================
+# Default the Build type to RELEASE
+# ================================
+# Make sure the build type is uppercase
+string(TOUPPER "${CMAKE_BUILD_TYPE}" BT)
+
+if(BT STREQUAL "RELEASE")
+  set(CMAKE_BUILD_TYPE RELEASE CACHE STRING
+    "Choose the type of build, options are DEBUG, or RELEASE."
+    FORCE)
+elseif(BT STREQUAL "DEBUG")
+  set(CMAKE_BUILD_TYPE DEBUG CACHE STRING
+    "Choose the type of build, options are DEBUG, or RELEASE."
+    FORCE)
+elseif(NOT BT)
+  set(CMAKE_BUILD_TYPE DEBUG CACHE STRING
+    "Choose the type of build, options are DEBUG, or RELEASE."
+    FORCE)
+  message(STATUS "CMAKE_BUILD_TYPE not given, defaulting to DEBUG.")
+else()
+  message(FATAL_ERROR "CMAKE_BUILD_TYPE not valid, choices are DEBUG, or RELEASE.")
+endif(BT STREQUAL "RELEASE")
+
+# ++++++++++++++++++++++++++++++++
+
+# ================================
+# Find the openmp package and add compiler-specific flags
+# ================================
+find_package(OpenMP)
+if(OPENMP_FOUND)
+  set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} ${OpenMP_Fortran_FLAGS}")
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${OpenMP_EXE_LINKER_FLAGS}")
+endif()
+# ++++++++++++++++++++++++++++++++
+
+# ================================
+# Set gfortran compile flags
+# ================================
+message("\n")
+if(CMAKE_Fortran_COMPILER_ID MATCHES "GNU")
+  include(${CMAKE_MODULE_PATH}/gfortran_flags.cmake)
+
+# ================================
+# Set INTEL compile flags
+# ================================
+elseif(CMAKE_Fortran_COMPILER_ID MATCHES "Intel")
+  include(${CMAKE_MODULE_PATH}/intel_flags.cmake)
+
+elseif(CMAKE_Fortran_COMILER_ID MATCHES "PGI")
+  message(FATAL_ERROR "Need to define flags for PGI")
+
+endif()
+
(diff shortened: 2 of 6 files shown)


leonfoks:
Obviously, this is all subject to change. We can discuss paths for mod files etc.

jvdp1:
> CMake is a pain for people who are not used to it.
> So I print extra information to the screen explaining the options chosen,
> the flags being used etc. and how to change these options.
> 
> Added more information when using CMAKE to help users when compiling.

I can't jugde the changes, but this one is useful. Thanks!

CMakeLists.txt
```diff
@@ -1,17 +1,34 @@
...
+
 project(stdlib Fortran)
 enable_testing()
 
-include(${CMAKE_SOURCE_DIR}/cmake/stdlib.cmake)
+# Add our local modules to the module path
+if(NOT CMAKE_MODULE_PATH)
+  set(CMAKE_MODULE_PATH)
+endif()
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
```
scivision, line 10: I think line 10 should move to line 8 as
```cmake
set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
```

CMakeLists.txt
```diff
@@ -1,17 +1,34 @@
...
+# Include the preprocessor options for stdlib
+include(${CMAKE_MODULE_PATH}/stdlib.cmake)
+
+# GNUInstallDirs is used to install libraries into correct locations
+# on all platforms.
+include(GNUInstallDirs)
+
+# Setup the Fortran cmake environment
+# Sets up the flags for compiling shared libraries or static libraries, and returns the libType parameter.
+include("${CMAKE_MODULE_PATH}/FortranEnvironment.cmake")
```
scivision, line 21: no need for `"` quotes here

cmake/FortranEnvironment.cmake
```diff
@@ -0,0 +1,118 @@
...
+# CMAKE Script for setting up the fortran compiling and linking flags for different operating systems and compilers
+
+# Call this cmake file from your project's CMakeLists.txt using 'include(/path/to/FortranEnvironment.cmake)'
+
+# ++++++++++++++++++++++++++++++++
+
+
+enable_language(Fortran)
+
+option (BUILD_SHARED_LIBS "Shared or static libraries"  ON)
```
scivision, line 11: Default shared libs don't work in general for compilers like Intel on Windows

cmake/FortranEnvironment.cmake
```diff
@@ -0,0 +1,118 @@
...
+endif(BT STREQUAL "RELEASE")
+
+# ++++++++++++++++++++++++++++++++
+
+# ================================
+# Find the openmp package and add compiler-specific flags
+# ================================
+find_package(OpenMP)
+if(OPENMP_FOUND)
+  set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} ${OpenMP_Fortran_FLAGS}")
```
scivision, line 49: these two lines should be like
```cmake
string(APPEND CMAKE_Fortran_FLAGS " ${OpenMP_Fortran_FLAGS}")
```

cmake/FortranEnvironment.cmake
```diff
@@ -0,0 +1,118 @@
...
+elseif(CMAKE_Fortran_COMILER_ID MATCHES "PGI")
+  message(FATAL_ERROR "Need to define flags for PGI")
+
+endif()
+
+# ================================
+# Set the output directories for compiled libraries and module files.
+# Paths are relative to the build folder where you call cmake
+# ================================
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/../lib) # Static library location
```
scivision, line 76: To make it clear these lines are creating side effects (creating artifacts outside of the build directory) would it be possible to do like:
```cmake
set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${PROJECT_SOURCE_DIR}/lib)
```

cmake/FortranEnvironment.cmake
```diff
@@ -0,0 +1,118 @@
...
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/../lib) # Shared library location
+# Place module files in specific include folder
+set(CMAKE_Fortran_MODULE_DIRECTORY ${PROJECT_BINARY_DIR}/../include)
+# Place executables in bin
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/../bin)
+INCLUDE_DIRECTORIES(${CMAKE_Fortran_MODULE_DIRECTORY})
```
scivision, line 82: early in the Fortran stdlib development it was noted, and in general for modern CMake, we typically like to use per-target `target_include_directories(foo PRIVATE ...)`  Unless this was decided to be changed for this project

cmake/gfortran_flags.cmake
```diff
@@ -0,0 +1,88 @@
+# ================================
+# Set gfortran compile flags
+# ================================
+message(STATUS "Getting gfortran flags")
+
+# Set flags for all build types
+set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -std=f2008ts -cpp -ffree-line-length-none -fall-intrinsics")
```
scivision, line 7: `-fstd=f2008ts` is a [deprecated flag](https://gcc.gnu.org/onlinedocs/gfortran/Fortran-Dialect-Options.html). In general I've found there are bugs in Gfortran that makes these standard flags problematic (e.g. Gfortran 9.3.0 has -fstd bugs that make lots of spurious warnings). 

Is it better to not use -fstd except if the builder specifies them on the CMake command line?

cmake/gfortran_flags.cmake
```diff
@@ -0,0 +1,88 @@
...
+
+# ================================
+# Set GFORTRAN flags for a SHARED library
+# ================================
+if(BUILD_SHARED_LIBS)
+  # Add any shared library related stuff here
+  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -shared -fpic")
```
scivision, line 14: I think it's preferable to use [CMake POSITION_INDEPENDENT_CODE](https://cmake.org/cmake/help/latest/prop_tgt/POSITION_INDEPENDENT_CODE.html#prop_tgt:POSITION_INDEPENDENT_CODE)

cmake/gfortran_flags.cmake
```diff
@@ -0,0 +1,88 @@
...
+
+  if(LINUX OR APPLE)
```
scivision, line 16: `if(UNIX)`

cmake/gfortran_flags.cmake
```diff
@@ -0,0 +1,88 @@
...
+  # Static build options
+  if(APPLE)
+    # gcc on OS X defaults to the dynamic quadmath library instead of the static
+    # NOTE: LIBRARY_PATH environment variable must be properly defined for the
+    #   current compiler or find_library(quadmath) will return the static
+    #   system version of libquadmath.a
+    set(CMAKE_FIND_LIBRARY_SUFFIXES ".a")
+    find_library(LIB_QUADMATH quadmath)
+
+    if ("${LIB_QUADMATH}" STREQUAL "LIB_QUADMATH-NOTFOUND")
```
scivision, line 51: `if(NOT LIB_QUADMATH)`

cmake/gfortran_flags.cmake
```diff
@@ -0,0 +1,88 @@
...
+  elseif(WIN32)
+    set(CMAKE_FIND_LIBRARY_SUFFIXES ".lib")
+    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -static-libgfortran -static-libgcc")
+  elseif(${LINUX})
+    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -static -static-libgfortran -static-libgcc -lgfortran -lgcc")
+  endif()
+endif()
+
+set(CMAKE_Fortran_FLAGS_RELEASE "${CMAKE_Fortran_FLAGS_RELEASE} -O3 -funroll-all-loops -finline-functions")
+set(CMAKE_Fortran_FLAGS_DEBUG "${CMAKE_Fortran_FLAGS_DEBUG} -O0 -fbacktrace -fbounds-check -Waliasing -Wampersand -Wconversion -Wsurprising -Wc-binding-type -Wintrinsics-std -Wtabs -Wintrinsic-shadow -Wline-truncation -Wtarget-lifetime -Wreal-q-constant")
```
scivision, line 84: `string(APPEND CMAKE_Fortran_FLAGS_DEBUG " ....")`

cmake/intel_flags.cmake
```diff
@@ -0,0 +1,33 @@
+
+message(STATUS "Getting ifort flags")
+
+if(WIN32)
+  set(CMAKE_Fortran_FLAGS_RELEASE "${CMAKE_Fortran_FLAGS_RELEASE} -nologo -fpp -O3 -heap-arrays1024 -QaxCORE-AVX2,CORE-AVX-I,AVX,SSE4.2,SSSE3 -Qipo -fp:fast=2 -Qdiag-disable:remark -Qmkl")
```
scivision, line 5: Normally CMake handles these flags fine specifically for Intel compiler, including on Windows. What about older CPUs?  I typically use for Intel compiler `-march=native` or `/arch:native` instead of all these flags.

Note: For Gfortran, `-march=native` does not work in general, `-mtune=native` should be used instead.

certik:
The CI fails on Windows, so that has to be fixed. 

It adds a lot more complexity to the CMake, hundreds of new lines to maintain. I feel we should just let cmake do its thing, and keep the number of lines that we have to write and maintain to a minimum.

And ultimately, we are hoping to switch to the Fortran Package Manager for stdlib once it is ready, and `fpm` will be able to automatically generate the CMake files.

https://github.com/fortran-lang/fpm/issues/69

cmake/FortranEnvironment.cmake
aradi: Very good point. I think, it is generally not a nice idea to create anything outside of the build folder (not even in the source folder), unless the user explicitly asks for it. So, why not something like
```
set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/_install/lib)
```
and the user can override it, in case it should be put somewhere else.

milancurcic:
@leonfoks How would you like to proceed with this PR? Can you make it work on Windows? There are still quite a few unresolved suggestions.

Also, can you reduce the added code to a minimum, as per @certik's suggestion?

I wish I could provide more feedback here, but CMake is really not in my wheelhouse.

leonfoks:
@milancurcic yes i can get this working on windows i think.  sorry for the delay, Paternity leave!

milancurcic:
@leonfoks Great to hear and congrats (I have a 4-month old next to me as I type this)! No rush.

awvwgk:
@leonfoks, what is the current status of this patch? The CMake files got a couple of modifications from my work of making it reusable in subprojects, which are currently conflicting with this patch. To move this pull request forward it would require a rebase against the current default branch first.

milancurcic:
I will close this for now, @leonfoks please re-open when you're ready to continue working on it.

PR #158, opened 2020-03-14
Addition of TOCs in Markdown files

jvdp1:
Some specs already contain multiple functions. So I thought it could be good to add TOCs for clarity.
Of course, it can be discussed/modified/removed, as usual.

### DIFF
--- a/src/stdlib_experimental_io.md
+++ b/src/stdlib_experimental_io.md
@@ -1,7 +1,5 @@
 # IO
 
-## Implemented
-
 * [`loadtxt` - load a 2D array from a text file](#loadtxt---load-a-2d-array-from-a-text-file)
 * [`open` - open a file](#open---open-a-file)
 * [`savetxt` - save a 2D array into a text file](#savetxt---save-a-2d-array-into-a-text-file)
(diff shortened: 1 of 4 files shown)


nshaffer:
Nothing obviously amiss, except the vim plugin magic comment thing. 

Links for each routine are obviously useful and good to have. Subsection links (Description, etc.) don't seem useful, though.

jvdp1:
> Nothing obviously amiss, except the vim plugin magic comment thing.
> 
> Links for each routine are obviously useful and good to have. Subsection links (Description, etc.) don't seem useful, though.

Thanks @nshaffer for the review. I will remove Vim comments and subsection links (Their are indeed all the same).

certik:
Looks good to me.

jvdp1:
> Looks great! My only suggestion is to remove the "Implemented" heading from the top of each spec file--it serves no purpose, and in some files it's incorrect (for example, in quadrature it lists some functions that aren't implemented).

Thank you @milancurcic. Suggestion applied.

jvdp1:
Ok. I will merge. Thank you for your comments.

Issue #159, opened 2020-03-18
Build with nagfor(7.0) is fail because of "isnan"


sakamoti:
There are many `isnan` functions are used in test_mean.f90, test_moment.f90,  test_var.f90 and  test_varn.f90 files. But `isnan` is not a fortran standard. So, NAG fortran compiler (nagfor) fail to build these files. What about using `ieee_is_nan` instead of `isnan`?

jvdp1:
Thank you for testing `stdlib`.
`ieee_is_nan` could be used indeed. `ieee_arithmetic` is already used to generate `NaN` in `mean()`, `moment()`, and `var()` functions.

If you have time, could you modify the tests and submit a PR, please?

There has been several discussions around `NaN` and its implementation (e.g., #147). A more general solution is needed. Meanwhile, using `ieee_is_nan` should be fine, and a better solution than using compiler-dependent `is_nan`.

sakamoti:
Thank you for merging pull request #160 . Close this issue.

PR #160, opened 2020-03-19
fix build failure with nagfor because of "isnan" (issue #159)

sakamoti:
`ieee_arithmetic` module is added  and `isnan` is replaced to `ieee_is_nan`.
This fixes the issue https://github.com/fortran-lang/stdlib/issues/159.


certik:
Thank you for the PR!

PR #161, opened 2020-03-22
Addition of optional center in API of moment()

jvdp1:
As discussed in #153 
Related to #113 

Changed the API of `moment()` from
```
result = moment(array, order [, mask])
result = moment(array, order, dim [, mask])
```
to
```
result = moment(array, order [, center [, mask]])
result = moment(array, order, dim [, center[, mask]])
```

The default returned value before changes was central moments. Now it is raw moments, and central moments if `center` is provided.

Note: if `dim` is provided, `center` must be an array. Otherwise, it must be a scalar.

Comments are welcome, @fiolj @ivan-pi @certik @milancurcic


ivan-pi:
I appreciate that the current iteration also supports other centers. Is there any particular reason to prefer raw vs central moments as the default option? In Julia [StatsBase.jl](https://juliastats.org/StatsBase.jl/v0.21/scalarstats.html#StatsBase.moment) the default are central moments. The MATLAB [moment](https://de.mathworks.com/help/stats/moment.html) is also central.
```
! raw is default (currently)
raw = moment(array, order)
central = moment(array, order,center=mean(array))

! vs

! if central were default
central = moment(array, order)
raw = moment(array, order,center=0.0)
```
I am not sure which is more convenient. If central was default, then, for the moment along a specific dimension  of a multidimensional array it would be necessary to have a `zeros` function:
```
! raw is default (currently)
raw = moment(array, order,dim=2)
central = moment(array, order,dim=2,center=mean(array,dim=2))

! if central is default
central = moment(array, order,dim=2)
raw = moment(array, order,center=zeros(size(array,1),...)) !  center a rank > 1 array of zeros
```
For the moment of a multi-dimensional array along a specific dimension the currently proposed version seems easier to use, it is however opposite from MATLAB and Julia. With central as default on the other hand, inferring the correct size of the zeros array for raw moments is error-prone.

Thoughts?

jvdp1:
Thanks @ivan-pi for the comments. I though about this issue (default: raw or central moment) and wanted to have some discussion about that. I am glad you started it.

I choose 'raw moment' as default such that `center` can be optional. However, I agree that other languages (Julia, Matlab, but also Scipy, Octave,R) have 'central momen' as default. Julia does not provide raw moments explicitely. So it would make sense that `stdlib` provides a central moment as default.

### Possible solution:
A way to implement easily raw moments would be to allow a scalar `center` __when__ `dim` is provided, in addition to the current implementation (i.e., an array for `center` when `dim` is mentioned). (Or should we also allow array `center` when `dim` is not provided? With which rank?)
With such an implemenation, a `zero()` function is not needed.

So, `center` would be still optional. But the default returned value is 'central moment'. 
If `center` is provided, then the return value is  a moment about `center`. A raw moment could be computed by providing `center = 0`. This behaviour is similar to the one of Julia.

The API would remain the same. Only the default returned value would change.

ivan-pi:
For consistency with other programming languages used in scientific computing I would prefer to see central moments as default. That way it is easier for users to swap between languages.

So for the solution you are suggesting, the interface would be the following
```
! existing API
result = moment(array, order [, center [, mask]])  ! center is a scalar, default is `mean(array)`
result = moment(array, order, dim [, center[, mask]]) ! center is an array, default is `mean(array,dim)`

! possible solution
result = moment(array, order, dim, center [, mask]) 
    ! center is a scalar, raw moments with `center = 0`
```
Is this correct? Would the type of `center` match the input `array`? I think it makes sense to allow integer arrays to have a real center. For real arrays presumably, it will be necessary to specify the precision as `center = 0.0_sp`, otherwise the compiler will complain. 

The remaining option
```
result = moment(array, order, center [, mask]) ! center is an array, dim is not provided
```
does not make sense to me in the context of statistical moments. It is in contradiction with the first version above, where the rank of the array and the location of values in the array carry no significance in the moment calculation. It would however make sense for things such as image moments, in which case the `center` would probably be a vector of size equal to the rank of the array. Currently, I don't think these different use cases are compatible.

jvdp1:
Following the discussion with @ivan-pi 

1) Current default result is now central moment (instead of raw moment)
2) the  optional `center` must be a scalar if `dim` is not provided, OR can be a scalar or an array if `dim` is provided.
3) `center` must be of the same type as `res` (so for integer arrays, `center` must be of type `real(dp)`).

So, this can be done to compute 2-th raw moment:
```fortran
res = moment( x, 2, center = 0.)
res = moment( x, 2, dim = 1, center = 0.)
res = moment( x, 2, dim = 1, center = zero)   !where zero is an array of the same size as res
```

Tests have been implemented for both options.

jvdp1:
ISSUE:
The CI failted.
The mentioned error:
```
Note: The following floating-point exceptions are signalling: IEEE_INVALID_FLAG IEEE_OVERFLOW_FLAG IEEE_DENORMAL
```
Has anyone any hint how to solve that? All the tests passed on my computer?

certik:
I restarted the builds, they are failing. Probably due to the conflicts.

certik:
@jvdp1 I got an email where you were asking about some CI failures due to floating point exceptions, but I can't find the comment here. If there is still an issue, let me know, I can have a look at that.

fiolj:
Sorry for the delay. It looks very good now. I also would prefer center for compatibility with other languages/environment. Thanks for all the work!

milancurcic:
With #163 now in master, I think this needs to be rebased to latest master, and replace calls to `assert` with calls to `check`. I used sed for this, for example:

```
sed -i 's/assert/check/gi' test_moment.f90
sed -i 's/assert/check/gi' test_rawmoment.f90
```

jvdp1:
> @jvdp1 I got an email where you were asking about some CI failures due to floating point exceptions, but I can't find the comment here. If there is still an issue, let me know, I can have a look at that.

@certik thank you for proposing help. I solved this issue and deleted the comment (sorry, I should have update it now).

I will rebase the branche. It should be fine then.

jvdp1:
@certik The tests for the manual Makefiles failed, probably because they take too much time. Removing all the flags for the debug options leads to pass the tests [here](https://github.com/jvdp1/stdlib/runs/534278203?check_suite_focus=true).

What could we do for this issue?

fiolj:
> @certik The tests for the manual Makefiles failed, probably because they take too much time. Removing all the flags for the debug options leads to pass the tests [here](https://github.com/jvdp1/stdlib/runs/534278203?check_suite_focus=true).
> 
> What could we do for this issue?

I think you are right, the problem is the same that has been insinuating for a time now. Compilation times are getting too long due to rank-aware routines. It also happens with `cmake` system but my guess is that with Makefiles are slightly longer (at least for this case) and is crossing some time threshold in the CI system.  
I've done a simple test of building times in my machine, obtaining:
```
real	4m15.740s
user	4m2.226s
sys	0m7.115s
```
and then, setting MAXRANK manually to 7:

```
#! Ranks to be generated when templates are created
#:set MAXRANK = 7
#:if not defined('MAXRANK')
  #:if VERSION90
    #:set MAXRANK = 7
  #:else
    #:set MAXRANK = 15
  #:endif
#:endif
```
I've got for a new complete build:
```
real	0m42.008s
user	0m38.078s
sys	0m3.121s
```
**Note:** I've had to manually disable (comment out) some lines of test_mean_f03.f90 for it to compile.

Probably we should discuss the compiling times in a new issue, because it will get worse quickly when we add new routines that are made to work with all different ranks.

jvdp1:
> I think you are right, the problem is the same that has been insinuating for a time now. Compilation times are getting too long due to rank-aware routines. It also happens with `cmake` system but my guess is that with Makefiles are slightly longer (at least for this case) and is crossing some time threshold in the CI system.

Indeed, the problem also appeared with CMake, and it is why we limited the number of ranks to 4 for the CI.
We should probably do the same for the manual Makefiles. I will have a look at this option.

jvdp1:
Seeing the report for the CI, it seems that the flag FYPPFLAGS (to limit the number of ranks to 4 for the CI) is not applied when using the Makefile.manual.  I have no idea how to check it.
@certik @milancurcic Any ideas?

milancurcic:
I merged #164 into master and restarted the builds here. But I forgot that we need to rebase to master here. @jvdp1 Can you do that?

jvdp1:
I rebased this PR to master and all checks have passed,

milancurcic:
Thank you, great work! Merging into master.

Issue #162, opened 2020-03-23
Proposal to use Testing Framework


everythingfunctional:
Rather than do ad-hoc testing of the standard library, we should adopt and use a dedicated testing framework. We'll use this issue to discuss possible candidates, their pros and cons, and what requirement we might have for choosing one.

everythingfunctional:
I'm going to firstly suggest my library, [vegetables](https://gitlab.com/everythingfunctional/vegetables).

Cons:
* It comes with some additional dependencies

Pros:
* As I'm also working on fpm, and the long term goal is for us to start using it, I will likely port it to fpm, alleviating the above con

I'm open to suggestions about others though.

certik:
I think we need a testing framework. We should integrate it with `fpm`, so that things just work. Dependencies will then not be a problem.

In fact, I think Cargo allows you to use any testing framework you want. One simply specifies it in `[dev-dependencies]`. So we should design `fpm` in the same way. That way people can use any testing framework they like friction free.

So we can use `vegetables`. The issue I have is with the API. See:

https://github.com/fortran-lang/stdlib/issues/121#issuecomment-577858926

I don't like `assertEquals`. I think Fortran's conventions are to use something like `assert_equal` or `assert_eq`.

@everythingfunctional would you be willing to consider using such names, assuming most people would agree?

milancurcic:
Of all the testing frameworks I looked at, I like vegetables the best, minus the camelCase but that's a nit pick :).

What I'm still confused about is where do you write actual tests. Brad [wrote here](https://github.com/fortran-lang/stdlib/issues/121#issuecomment-602771405) that testing should be done from the outside. Concretely, what does that mean for stdlib?

1. Tests are written in a dedicated project stdlib-tests which has vegetables as a dependency?
2. Tests are written in fpm which has vegetables as a dependency?
3. Tests are written in vegetables?

Of the three, only the 1st makes some sense to me. Of course, the most sensible solution to me is to write tests in stdlib, and then the framework is a dependency there (if pure Fortran or C, not a biggie). But I got the impression that that's not what you're trying to do.

certik:
I propose to do what Rust / Cargo does. Have a `tests` directory, and each file there gets compiled to a binary. That file internally can import many other modules and use a test framework to execute tests.

[...]

PR #163, opened 2020-03-24
121 replace assert with check

milancurcic:
This PR implements the `check` subroutine, proposed to replace `assert` for internal testing. Specifically:

* Implements `check` as proposed [here](https://github.com/fortran-lang/stdlib/issues/121#issuecomment-602261670)
* Replaces existing calls to `assert` with calls to `check`
* Removes the definition of `assert` from `stdlib_experimental_error`

This PR doesn't cover making all `check` calls to print meaningful messages on failure, and to execute them with `warn=.true.`, to allow multiple test failures in a single run. These will be covered in a future PR.

This internal testing is meant as temporary until we can do it with `fpm test` as proposed in #162. For now, we consider this as a "good enough" solution that will help us move forward with development while [fpm](https://github.com/fortran-lang/fpm) is in the works.

Finally, there is also Brad's proposed implementation [here](https://github.com/fortran-lang/stdlib/issues/121#issuecomment-602632789) which we shouldn't ignore. If there is some support from the community for it, let's discuss and consider it.


src/stdlib_experimental_error.f90
```diff
@@ -1,34 +1,63 @@
...
+    !   * the message.
+    !
+    ! Arguments
+    ! ---------
+
+    logical, intent(in) :: condition
+    character(*), intent(in), optional :: msg
+    integer, intent(in), optional :: code
+    logical, intent(in), optional :: warn
+    character(*), parameter :: msg_default = 'Test failed.'
```
ivan-pi: Wouldn't it make more sense if the default message is `'Check failed.' just to be consistent with the subroutine name?

src/stdlib_experimental_error.f90
```diff
@@ -1,34 +1,63 @@
...
+
+    ! Examples
+    ! --------
+    !
+    ! ! Stops the program with exit code 1 and prints 'Test failed.'
+    ! call check(a == 5)
```
ivan-pi: There is an implicit assumption in the examples that `a` is not five. It remains to be discussed whether such examples in the docstrings should be self-contained.

milancurcic:
@ivan-pi You're right. The latest commit fixes that.

milancurcic:
Thanks, all. @everythingfunctional, let's do a separate issue and PR for your proposal when you're ready. Merging.

jvdp1:
@milancurcic Do we want that users use both `check` and `error_stop` outside `stdlib`? OR are these only aimed for `stdlib`?
If users can use them outside `stdlib`, then I think it would be good to have some specs (like for other functions)  (I can give a start based on the comments in the code). Personally, I could use both outside `stdlib` ;)

milancurcic:
@jvdp1 I think so. Sorry, I totally forgot about the spec. That should've been part of the original PR. Please go ahead with a draft spec if you don't mind. Thank you!

Issue #167, opened 2020-03-30
set order of attributes


JHenneberg:
While reading the style guide ```STYLE_GUIDE.md``` I noticed that the order of the attributes is not specified.

Going through the source files the order seems:
1. ```allocatable```
2. ```dimension```
3. ```intent(intent-spec)```
4. ```optional```

Although it is not clear if ```allocatable``` should be after oder before ```dimension```

Missing attributes I can think of right now are:
* ```target```, ```pointer```
* ```parameter```
* ```public```, ```private```, ```protected```
* ```asynchronous```
* ```save```
* ```external```
* ```contiguous```
* ```volatile```

So lets come up with a convention for the order of attributes.

Suggestion so far for ```STYLE_GUIDE.md```
```
# Order of attributes

Similiar to the guidlines for indentation and whitespace, specifiying the order of appearance of attributes can  help reviewing code and git-diffs.

1. ```allocatable```
2. ```dimension```
3. ```intent(<>)```
4. ...
```

jvdp1:
Good point.
I usually write things in this order:

1. ```dimension```
2. ```allocatable```
3. ```intent(intent-spec)```
4. ```optional```

Some combinations of attributes are not possible.
Could you maybe submit a PR with suggested modifications for further discussions?

certik:
I do not follow a strict order, although typically `intent()` is at the end, the `allocatable` and `dimension` I use in any order, and optional is typically around `intent()` in either order.

Is it important to always follow a certain order? It seems maybe we do not need to standardize this?

milancurcic:
I tend to write these attributes in the same order as @jvdp1. More lately, I prefer to omit `dimension`, so:

```fortran
real, dimension(:), allocatable :: a
real, dimension(:,:), allocatable :: b
```

becomes just:

```fortran
real, allocatable :: a(:), b(:,:)
```

Not only is the code less verbose by not including `dimension`, but also arrays of different size, shape, or rank can be declared on the same line.

Like @certik, I don't think the order is as important. Rather than enforcing the order, I'd prefer seeing these two clauses:

* Don't use `dimension` attribute to declare arrays
* Always specify `intent` for dummy arguments

If we did indeed specify order of attributes in the style guide, I'd personally like to see:

* `intent` attribute should be specified last in the declaration statement
* If `optional` attribute is used to declare a dummy argument, it should follow the `intent` attribute

or similar.

certik:
[...]

PR #168, opened 2020-03-31
specified order of attributes #167

JHenneberg:
Lets discuss if this is need or giving to much overhead. For now there is positive (@jvdp1 ) and negative (@certik ) feedback.

In my opinion a standard library should not only provide functionalty, it also should give example how to use a programming language in an elegant way. We decided already in the STYLE_GUIDE.md to include indentation and whitespaces conventions which have of course bigger impact but specifiying that at an end of a scope there should be a closing statment is quite a minor thing so why to draw a line when specifiying the order of attritbutes?

### DIFF
--- a/STYLE_GUIDE.md
+++ b/STYLE_GUIDE.md
@@ -59,7 +59,7 @@ focus on the semantics of the proposed changes rather than style and formatting.
   real, dimension(:,:), allocatable :: b
   ```
 
-  When defining a lot of arrays of the same dimension ```dimension``` can be used as an exception.
+When defining many arrays of the same dimension, ```dimension``` can be used as an exception if it makes the code less verbose.
 * If ```optional``` attribute is used to declare a dummy argument, it should follow the intent attribute
 
 ## End <scope> block closing statements
--- a/STYLE_GUIDE.md
+++ b/STYLE_GUIDE.md
@@ -44,7 +44,7 @@ focus on the semantics of the proposed changes rather than style and formatting.
 
 ## Attributes
 
-* Always specify ```intent``` for dummy arguments
+* Always specify ```intent``` for dummy arguments.
 * Don't use ```dimension``` attribute to declare arrays because it is less verbose.
   Use this:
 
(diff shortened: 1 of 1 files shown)


STYLE_GUIDE.md
```diff
@@ -42,14 +42,25 @@ focus on the semantics of the proposed changes rather than style and formatting.
...
+  ```
+
+  instead of:
+
+  ```
+  real, dimension(:), allocatable :: a
+  real, dimension(:,:), allocatable :: b
+  ```
+
+  When defining a lot of arrays of the same dimension ```dimension``` can be used as an exception.
```
milancurcic, line 62:
```suggestion
When defining many arrays of the same dimension, ```dimension``` can be used as an exception if it makes the code less verbose.
```

STYLE_GUIDE.md
```diff
@@ -42,14 +42,25 @@ focus on the semantics of the proposed changes rather than style and formatting.
 * Where conventional and appropriate shortening of a word is used then the underscore may be omitted,
   for example `linspace` is preferred over `lin_space`
 
-## Order of attributes
+## Attributes
 
-Similiar to the guidlines for indentation and whitespace, specifiying the order of appearance of attributes can  help reviewing code and git-diffs.
+* Always specify ```intent``` for dummy arguments
```
milancurcic, line 47:
```suggestion
* Always specify ```intent``` for dummy arguments.
```

STYLE_GUIDE.md
```diff
@@ -42,14 +42,25 @@ focus on the semantics of the proposed changes rather than style and formatting.
...
+
+  instead of:
+
+  ```
+  real, dimension(:), allocatable :: a
+  real, dimension(:,:), allocatable :: b
+  ```
+
+  When defining a lot of arrays of the same dimension ```dimension``` can be used as an exception.
+* If ```optional``` attribute is used to declare a dummy argument, it should follow the intent attribute
```
milancurcic, line 63:
```suggestion
* If the ```optional``` attribute is used to declare a dummy argument, it should follow the `intent` attribute.
```

milancurcic:
Merging into master, thank you!

Issue #169, opened 2020-04-07
diag, eye, and trace


ivan-pi:
Related to #10 

Convenience functions for use in linear algebra. The functions would go into `stdlib_experimental_linalg.f90`. Simple implementation by looping over the diagonal. Type-generic (integer, real, complex) implementation using the fypp preprocessor. 

A draft is available here: https://github.com/ivan-pi/stdlib/blob/ivan-pi/src/stdlib_experimental_linalg.fypp (based on the version from fortran-utils #103 )

## eye
Returns the identity matrix
```fortran
function eye(n) result(res)
integer, intent(in) :: n
real(dp), allocatable :: res(:  ,:)
end function
```
* Do we need an optional kind dummy variable for the return type (`eye(n [, kind])`)? Or just default to double precision? 
* Other names: `identity`, ... ?
* Generalize also to higher rank arrays (tensors)?
* Fixed-size or allocatable output?

## trace
Returns the sum of diagonal elements
```fortran
pure function trace(A) result(res)
real(dp), intent(in) :: A(:,:)
real(dp) :: res
end function
```
* How to deal with non-square matrices? Take the sum over the shorter dimension? Run-time shape assertion?
* Optional `axis` argument?
* Diagonal offset?
* Generalize to higher rank arrays (tensors)?

## diag
Create a diagonal matrix from a vector or extract a diagonal from a matrix

```fortran
! vector to matrix
function diag(v, k) result(res)
real(dp), intent(in) :: v(:)
integer, intent(in), optional :: k ! place elements of v on the k-th diagonal
real(dp), allocatable :: res(:,:)
end function
! matrix to vector
function diag(A, k) result(res)
real(dp), intent(in) :: A(:,:)
integer, intent(in), optional :: k ! extract elements from the k-th diagonal of A
real(dp), allocatable :: res(:)
end function
```
* Other names: `diagonal`, ...
* Specify which diagonal to use?
* How to deal with non-square matrices? Extract the diagonal along the shorter dimension? Run-time shape assertion?
* Generalize to higher rank arrays?
* Fixed-size or allocatable output?

## Other languages

The API's differ slightly between languages, some are more flexible than others.

* diag:
  * Julia: https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#LinearAlgebra.Diagonal
  * MATLAB: https://de.mathworks.com/help/matlab/ref/diag.html
[...]

PR #170, opened 2020-04-10
Addition of diag, eye, and trace

ivan-pi:
See proposal in #169 . This is a draft pull request.

This is my first time using the fypp preprocessor, so I decided to only focus on rank-2 arrays to keep things simple and submit a pull request before my focus shifts away from stdlib. Hopefully the API will not need to change if we decide to generalize these to higher rank-matrices, otherwise the generalizations could be placed into a `stdlib_tensor` module in the future.

I am still missing the tests, but thought it would be beneficial to get some feedback on the functions and documentation first.

@jvdp1 Did you use fypp to generate the tests for the statistical routines and then committed only the generated .f90 files?

### DIFF
--- a/src/stdlib_experimental_linalg.fypp
+++ b/src/stdlib_experimental_linalg.fypp
@@ -1,0 +1,80 @@
+#:include "common.fypp"
+#:set RCI_KINDS_TYPES = REAL_KINDS_TYPES + CMPLX_KINDS_TYPES + INT_KINDS_TYPES
+module stdlib_experimental_linalg
+  use stdlib_experimental_kinds, only: sp, dp, qp, &
+    int8, int16, int32, int64
+  implicit none
+  private
+
+  public :: diag
+  public :: eye
+  public :: trace
+
+  interface diag
+      !
+      ! Vector to matrix
+      !
+    #:for k1, t1 in RCI_KINDS_TYPES
+      module function diag_${t1[0]}$${k1}$(v) result(res)
+        ${t1}$, intent(in) :: v(:)
+        ${t1}$ :: res(size(v),size(v))
+      end function diag_${t1[0]}$${k1}$
+    #:endfor
+    #:for k1, t1 in RCI_KINDS_TYPES
+      module function diag_${t1[0]}$${k1}$_k(v,k) result(res)
+        ${t1}$, intent(in) :: v(:)
+        integer, intent(in) :: k
+        ${t1}$ :: res(size(v)+abs(k),size(v)+abs(k))
+      end function diag_${t1[0]}$${k1}$_k
+    #:endfor
+
+      !
+      ! Matrix to vector
+      !
+    #:for k1, t1 in RCI_KINDS_TYPES
+      module function diag_${t1[0]}$${k1}$_mat(A) result(res)
+        ${t1}$, intent(in) :: A(:,:)
+        ${t1}$ :: res(minval(shape(A)))
+      end function diag_${t1[0]}$${k1}$_mat
+    #:endfor
+    #:for k1, t1 in RCI_KINDS_TYPES
+      module function diag_${t1[0]}$${k1}$_mat_k(A,k) result(res)
+        ${t1}$, intent(in) :: A(:,:)
+        integer, intent(in) :: k
+        ${t1}$ :: res(minval(shape(A))-abs(k))
+      end function diag_${t1[0]}$${k1}$_mat_k
+    #:endfor
+  end interface
+
+  ! Matrix trace
+  interface trace
+    #:for k1, t1 in RCI_KINDS_TYPES
+      module procedure trace_${t1[0]}$${k1}$
+    #:endfor
+  end interface
+
+contains
+
+    function eye(n) result(res)
+      integer, intent(in) :: n
+      integer(int8) :: res(n, n)
+      integer :: i
+      res = 0
+      do i = 1, n
+         res(i, i) = 1
+      end do
+    end function eye
+
+
+    #:for k1, t1 in RCI_KINDS_TYPES
+      function trace_${t1[0]}$${k1}$(A) result(res)
+        ${t1}$, intent(in) :: A(:,:)
+        ${t1}$ :: res
+        integer :: i
+        res = 0
+        do i = 1, minval(shape(A))
+          res = res + A(i,i)
+        end do
+      end function trace_${t1[0]}$${k1}$
+    #:endfor
+end module
--- a/src/stdlib_experimental_linalg.md
+++ b/src/stdlib_experimental_linalg.md
@@ -1,0 +1,156 @@
+# Linear Algebra
+
+* [`diag` - Create a diagonal array or extract the diagonal elements of an array](#diag---create-a-diagonal-array-or-extract-the-diagonal-elements-of-an-array)
+* [`eye` - Construct the identity matrix](#eye---construct-the-identity-matrix)
+* [`trace` - Trace of a matrix](#trace---trace-of-a-matrix)
+
+## `diag` - Create a diagonal array or extract the diagonal elements of an array
+
+### Description
+
+Create a diagonal array or extract the diagonal elements of an array
+
+### Syntax
+
+`d = diag(a [, k])`
+
+### Arguments
+
+`a`: Shall be a rank-1 or or rank-2 array. If `a` is a rank-1 array (i.e. a vector) then `diag` returns a rank-2 array with the elements of `a` on the diagonal. If `a` is a rank-2 array (i.e. a matrix) then `diag` returns a rank-1 array of the diagonal elements.
+
+`k` (optional): Shall be a scalar of type `integer` and specifies the diagonal. The default `k = 0` represents the main diagonal, `k > 0` are diagonals above the main diagonal, `k < 0` are diagonals below the main diagonal.
+
+### Return value
+
+Returns a diagonal array or a vector with the extracted diagonal elements.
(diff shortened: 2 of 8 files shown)


src/stdlib_experimental_linalg.fypp
```diff
@@ -0,0 +1,132 @@
...
+      do i = 1, n
+         res(i, i) = 1
+      end do
+    end function eye
+
+    #:for k1, t1 in RCI_KINDS_TYPES
+      function diag_${t1[0]}$${k1}$(v) result(res)
+        ${t1}$, intent(in) :: v(:)
+        ${t1}$ :: res(size(v),size(v))
+        integer :: i, sz
```
jvdp1, line 54:
```suggestion
        integer :: i
```

src/stdlib_experimental_linalg.md
```diff
@@ -0,0 +1,156 @@
...
+
+Create a diagonal array or extract the diagonal elements of an array
+
+### Syntax
+
+`d = diag(a [, k])`
+
+### Arguments
+
+`a`: Shall be a rank-1 or or rank-2 array. If `a` is a rank-1 array (a vector)then `diag` returns a rank-2 array with the elements of `a` on the diagonal. If `a` is a rank-2 array (a matrix) then `diag` returns a rank-1 array of the diagonal elements.
```
jvdp1, line 19:
```suggestion
`a`: Shall be a rank-1 or or rank-2 array. If `a` is a rank-1 array (i.e. a vector) then `diag` returns a rank-2 array with the elements of `a` on the diagonal. If `a` is a rank-2 array (i.e. a matrix) then `diag` returns a rank-1 array of the diagonal elements.
```

src/stdlib_experimental_linalg.md
```diff
@@ -0,0 +1,156 @@
...
+
+`k` (optional): Shall be a scalar of type `integer` and specifies the diagonal. The default `k = 0` represents the main diagonal, `k > 0` are diagonals above the main diagonal, `k < 0` are diagonals below the main diagonal.
+
+### Return value
+
+Returns a diagonal array or the extracted diagonal.
```
jvdp1, line 25:
```suggestion
Returns a diagonal array or a vector with the extracted diagonal elements.
```

src/tests/linalg/test_linalg.f90
```diff
@@ -0,0 +1,25 @@
...
+  do i = 1, size(d,1)
+    write(*,*) (d(i,j),j=1,size(d,2))
+  end do
+  
+  a = diag(d,0)
+  print *, "a = ", a
+
+  print *, eye(5)
+
+  print *, all(eye(4) == diag([1,1,1,1]))
```
jvdp1, line 23: Could the `check` function be used here?

ivan-pi: I will add more tests with calls to `check`. This was just temporary for development purposes.

ivan-pi:
> * It would maybe good to move the functions in submodules. It avoids the compilation of alll functions when modifying only some of them.
> 
I agree. I expect that in the future we will likely have several linear algebra submodules (array creation routines, solvers, matrix properties...). 

> * I think it could be easily extended to ranks >2 with the same API quite easily using fypp. However, I am the target user for such functions.

I don't think a tensor version of `eye` is trivial, at least not under a common interface like `eye(n [,rank])`. We could have multiple versions for different ranks, e.g. `eye` (equal to `eye2`),  `eye3`, `eye4`, `eye5`, etc.?

I can try to make `trace` rank-generic for the main diagonal, and also have a version similar to numpy 
```python
numpy.trace(a, offset=0, axis1=0, axis2=1, dtype=None, out=None)[source]
```
which calculates the trace of each two-dimensional slice returning an array. But since the name `trace` is typically linked with linear algebra, for the tensor "trace" it might be a better idea to have a general `tensor_contraction` function (https://en.wikipedia.org/wiki/Tensor_contraction).

With `diag` comes again the question, how to specify the rank of the output? Also the diagonal specifier is not immediately clear. Is the aim to operate on the individual 2d slices, or somehow set/extract a diagonal/slice in >=3 dimensions?
image (image taken from [here](http://www.math.brown.edu/~banchoff/Beyond3d/chapter3/section02.html))

> * I wrote the tests by hand.

I applaud your effort. :clap:

jvdp1:
> Is the aim to operate on the individual 2d slices, or somehow set/extract a diagonal/slice in >=3 dimensions?

Good point. I didn't think so far.
Anyway, this is a nice proposal that could help many people as is.

src/tests/linalg/test_linalg.f90
milancurcic: Nice. For internal stdlib tests, I suggest providing custom message in each `check` call so we can identify which failed, and also with `warn=.true.` to allow multiple tests to fail in a single test program.

jvdp1: > and also with `warn=.true.` to allow multiple tests to fail in a single test program.

@milancurcic With `warn = .true.` multiple tests can fail, but `ctest` will report that all tests passed. It is ok when we develop it. But before merging into master, we should probably have `warn=.false.`.

milancurcic: You're right, I forgot about that. We shouldn't do it then. It's important that ctest correctly reports failure.

ivan-pi:
In the routine to extract the `k`-th diagonal from a matrix, I have the following interface:
```
     function diag_${t1[0]}$${k1}$_mat_k(A,k) result(res)
        ${t1}$, intent(in) :: A(:,:)
        integer, intent(in) :: k
        ${t1}$ :: res(minval(shape(A))-abs(k))
    end function
```
Interestingly, if `abs(k) > minval(shape(A))`, the size of `res` is negative, however this gets correctly translated a zero-size array. I've double checked this with the following program:
```fortran
implicit none
real :: res(-5)
print *, size(res)
end
```
which prints "0" for both the ifort and gfortan compilers  without any warnings. So it looks like Fortran had my back covered :sunglasses: .

ivan-pi:
Is there any interest to extend `diag` also to logical arrays?

jvdp1:
> Is there any interest to extend `diag` also to logical arrays?

Not from my side. For what would you use such a matrix?

ivan-pi:
It could be used as a mask to unpack values into a certain diagonal or more generally as a mask in other functions. The numpy diag also allows it, in MATLAB I think logical masks are binary arrays, so it probably also just works.

jvdp1:
> It could be used as a mask to unpack values into a certain diagonal or more generally as a mask in other functions. The numpy diag also allows it, in MATLAB I think logical masks are binary arrays, so it probably also just works.

If other languages support it, it is probably good to support it too. Up to you to decide. I don't think I would need such a logical function.

ivan-pi:
I was planning to have a go at the logical kind version, but would like to clarify first if it would suffice to support only the default logical kind or all those supported by the compiler (for some odd reason?):
```
ipribec@ipribec-T530:~/fortran$ cat test_logical_kinds.f90 
use iso_fortran_env
print *, logical_kinds
end
ipribec@ipribec-T530:~/fortran$ gfortran test_logical_kinds.f90 && ./a.out
           1           2           4           8          16
ipribec@ipribec-T530:~/fortran$ ifort test_logical_kinds.f90 && ./a.out
           1           2           4           8
ipribec@ipribec-T530:~/fortran$ 
```
Any thoughts?

If not, pending a review of the tests, I would merge this.

certik:
I think we can merge this now. We are in experimental, so any such changes / modifications can always be done later.

src/tests/linalg/test_linalg.f90
```diff
@@ -0,0 +1,436 @@
...
+    integer :: i
+    logical, allocatable :: mask(:,:)
+    write(*,*) "test_diag_int16"
+    a = reshape([(i,i=1,n**2)],[n,n])
+    mask = merge(.true.,.false.,eye(n) == 1)
+    call check(all(diag(a) == pack(a,mask)), &
+          msg="all(diag(a) == pack(a,mask))", warn=warn)
+    call check(all(diag(diag(a)) == merge(a,0_int16,mask)), &
+          msg="all(diag(diag(a)) == merge(a,0_int16,mask)) failed.", warn=warn)
+    a = unpack(int([1,2,3,4],int16),eye(n)==1,a)
```
jvdp1, line 207: I believe that this line could be removed.

src/tests/linalg/test_linalg.f90
```diff
@@ -0,0 +1,436 @@
...
+    write(*,*) "test_trace_rdp_nonsquare"
+
+    !  1   25   81
+    !  4   36  100  
+    !  9   49  121  
+    ! 16   64  144  
+    a = reshape([(i**2,i=1,n*(n-1))],[n,n-1])
+    ans = sum([1._dp,36._dp,121._dp])
+
+    call check(abs(trace(a) - ans) < epsilon(1.0_dp), &
```
jvdp1, line 308: When submitting tests for `mean` and `var`, it was mentioned that `epsilon()` is too strict. Therefore, you should multiply it by e.g., 1000 (see [here](https://github.com/fortran-lang/stdlib/blob/c95f7a9a6d031f9871e09564432ebcdd7035c179/src/tests/stats/test_mean.f90#L9) for the current implementation) to avoid any issues.

ivan-pi:
Thanks @jvdp1, @milancurcic, and @certik for the review and approvals. I've squashed the commits and will merge after the tests are completed.

PR #171, opened 2020-04-11
variance_correction: small issue with debug mode of gfortran 7

jvdp1:
Gfortran 7 with debug options (but not Gfortran 9) failed on, e.g., 
```fortran
i32 = d
```
So, I replaced the line by:
```fortran
allocate(i32, source = int(d)
```

### DIFF
--- a/src/stdlib_experimental_stats_var.fypp
+++ b/src/stdlib_experimental_stats_var.fypp
@@ -101,7 +101,7 @@ contains
             end do
           #:endfor
           case default
-            call error_stop("ERROR (mean): wrong dimension")
+            call error_stop("ERROR (var): wrong dimension")
         end select
         res = res / (n - merge(1, 0, optval(corrected, .true.)))
 
@@ -140,7 +140,7 @@ contains
             end do
           #:endfor
           case default
-            call error_stop("ERROR (mean): wrong dimension")
+            call error_stop("ERROR (var): wrong dimension")
         end select
         res = res / (n - merge(1, 0, optval(corrected, .true.)))
 
@@ -229,7 +229,7 @@ contains
             end do
           #:endfor
           case default
-            call error_stop("ERROR (mean): wrong dimension")
+            call error_stop("ERROR (var): wrong dimension")
         end select
         res = res / (n - merge(1, 0, (optval(corrected, .true.) .and. n > 0)))
 
@@ -264,7 +264,7 @@ contains
             end do
           #:endfor
           case default
-            call error_stop("ERROR (mean): wrong dimension")
+            call error_stop("ERROR (var): wrong dimension")
         end select
         res = res / (n - merge(1, 0, (optval(corrected, .true.) .and. n > 0)))
 
--- a/src/tests/stats/test_var.f90
+++ b/src/tests/stats/test_var.f90
@@ -57,7 +57,7 @@ program test_var
 
     !2dim
     print*,' test_sp_2dim'
-    s = d
+    allocate(s, source = real(d))
     call check( abs(var(s) - 13) < sptol)
     call check( all( abs( var(s, 1) - [20. / 3., 20. / 3., 5. / 3.]) < sptol))
     call check( all( abs( var(s, 2) - [19.0, 43. / 3., 31. / 3. , 7.0]) < sptol))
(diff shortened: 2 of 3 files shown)


ivan-pi:
In https://github.com/fortran-lang/stdlib/issues/169#issuecomment-611131737 @milancurcic reckoned we should avoid workarounds for what is supposed to be valid code.

At the same time, it would be a nuisance to bring in preprocessor macros to differentiate between compiler versions and avoid spurious warnings. Then again, your change is valid Fortran code, just a bit more explicit, so I suppose this is the easier solution in the short-term.

milancurcic:
@jvdp1 What exactly fails here for gfortran 7?

If gfortran 7 didn't support reallocation on assignment (I thought it did), I don't see a good reason to babysit the compiler here--the feature is reasonably mature (F2003). I'd rather recommend upgrading the compiler to a later release.

However, if the feature is supported but has a bug, we should file a bug report rather than inserting a workaround into stdlib.

We don't need to support all earlier releases of compilers. We just need to make a pragmatic decision what we want and don't want to support. 

Of course, this is all just my opinion. I'm curious to hear what others think, @certik @nncarlson @nshaffer

jvdp1:
> @jvdp1 What exactly fails here for gfortran 7?

Here are more details:
```
$ gfortran --version
GNU Fortran (GCC) 7.2.0
Copyright © 2017 Free Software Foundation, Inc.
```
Additional options used for compilation:
```
-fimplicit-none -Wall -fcheck=all -fbacktrace
```
Issue with `test_far` (and similarly for `test_varn`):
```
At line 60 of file /home/WUR/vande018/stdlib/src/tests/stats/test_var.f90
Fortran runtime error: Array bound mismatch for dimension 1 of array 's' (1/4)

Error termination. Backtrace:
#0  0x402096 in ???
#1  0x43ad54 in ???
#2  0x2aaaab81b544 in ???
#3  0x401538 in ???
#4  0xffffffffffffffff in ???
<end of output>
```
Correspinding line 60 in `test_var.f90`:
```fortran
s = d
```

certik:
It's this automatic reallocation of LHS, which used to be buggy in gfortran in various scenarios. We either require GFortran > 7, or we fix it (one way or another).

nncarlson:
I agree with @certik.  It's unfortunate but necessary.

I'd also add that the set of supported compilers will evolve in time, with some older compilers being dropped while new ones are added. In this regard it will be important to somehow tie individual workarounds to specific compiler versions and not just to the compiler (e.g., gfortran).  That will allow workarounds for unsupported compilers to be easily removed from the code base. Otherwise the code base will become more and more littered with workarounds as time goes on.

nncarlson:
> However, if the feature is supported but has a bug, we should file a bug report rather than inserting a workaround into stdlib.

Unless one is dealing with the NAG compiler where they almost always fix bugs within a couple of weeks, this is not an either-or proposition.  Definitely **always** file a bug report.  But then one will also need to implement a workaround if one wants to go ahead using the buggy feature. Every other compiler I've dealt with takes many months, sometimes years, to fix reported bugs.

jvdp1:
Thank you for your feedbacks.

The same issue seems to be also present in gfortran 8!
I will report the bug.

For this PR, I let you decide what is best for the current and future situations.

jvdp1:
Bug reported here:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=94585

certik:
Thanks!

They already responded as wontfix. 

I would be ok to drop GFortran 7 support. By the time stdlib is useful, hopefully most people will be able to use GFortran 8.

jvdp1:
Because the GFortran 7-branch is closed, I am also ok to drop GFortran 7 support. If more people think that too, this PR can be closed without being merged.

certik:
@milancurcic any comments?

Btw I think you have some unrelated change in this PR that can go in (you should post it as a separate PR).

milancurcic:
I agree with not supporting GFortran 7 with this workaround.

jvdp1:
> Btw I think you have some unrelated change in this PR that can go in (you should post it as a separate PR).

I will open another PR for these!

jvdp1:
> > Btw I think you have some unrelated change in this PR that can go in (you should post it as a separate PR).
> 
> I will open another PR for these!

Done with PR #173

jvdp1:
I close this PR because it seems there is an agreement to not support GFortran 7 with this workaround,

PR #172, opened 2020-04-12
Covariance of array elements

jvdp1:
Draft proposal for covariance of array elements opened for discussion

Related to #113 

To be discussed:
* Name: the currently proposed name is `cov`. But what about `covariance`,....?
* API: similar to the API of `variance`
* Implementation: only for arrays of rank 1 and rank 2. 
* Implementation: the API accepts only 1 array.
An implementation to compute the covariance among 2 arrays still needs to be done.

Others:
[Octave `cov`](https://octave.sourceforge.io/octave/function/cov.html)
[Matlab `cov`](https://www.mathworks.com/help/matlab/ref/cov.html)
[Julia `cov`](https://docs.julialang.org/en/v1/stdlib/Statistics/#Statistics.cov)
[R `cov`](http://www.r-tutor.com/elementary-statistics/numerical-measures/covariance)
[Numpy `cov`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.cov.html)

### Description

Returns the covariance of the elements of `array` along dimension `dim` if the corresponding e
lement in `mask` is `true`.

Per default, the covariance is defined as:

```
 cov(array) = 1/(n-1) sum_i (array(i) - mean(array) * (array(i) - mean(array)))
```
where n is the number of elements.

The scaling can be changed with the logical argument `corrected`. If `corrected` is `.false.`,
 then the sum is scaled with `n`, otherwise with `n-1`.

### Syntax

`result = cov(array, dim [, mask [, corrected]])`

### DIFF
--- a/src/stdlib_experimental_stats.md
+++ b/src/stdlib_experimental_stats.md
@@ -10,12 +10,12 @@
 
 ### Description
 
-Returns the covariance of the elements of `array` along dimension `dim` if the corresponding element in `mask` is `true` (if `mask` is provided).
+Returns the covariance of the elements of `array` along dimension `dim` if the corresponding element in `mask` is `true`.
 
 Per default, the covariance is defined as:
 
 ```
- cov(array) = 1/(n-1) sum_i (array(i) - mean(array) * (array(i) - mean(array)) )
+ cov(array) = 1/(n-1) sum_i (array(i) - mean(array) * (array(i) - mean(array)))
 ```
 
 where n is the number of elements.
@@ -29,7 +29,7 @@ The scaling can be changed with the logical argument `corrected`. If `corrected`
 
 ### Arguments
 
-`array`: Shall be a 1-rank or a 2-rank array of type `integer`, `real`, or `complex`.
+`array`: Shall be a rank-1 or a rank-2 array of type `integer`, `real`, or `complex`.
 
 `dim`: Shall be a scalar of type `integer` with a value in the range from 1 to n, where n is the rank of `array`.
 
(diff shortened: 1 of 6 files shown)


src/stdlib_experimental_stats.md
```diff
@@ -1,9 +1,67 @@
...
+
+### Syntax
+
+`result = cov(array, dim [, mask [, corrected]])`
+
+### Arguments
+
+`array`: Shall be a rank-1 or a rank-2 array of type `integer`, `real`, or `complex`.
+
+`dim`: Shall be a scalar of type `integer` with a value in the range from 1 to n, where n is the rank of `array`.
```
ivan-pi, line 34: The `` are missing on n -> `n`.

src/stdlib_experimental_stats.md
```diff
@@ -1,9 +1,67 @@
...
+
+`mask` (optional): Shall be of type `logical` and either by a scalar or an array of the same shape as `array`.
+
```
ivan-pi, line 37: `mask` (optional): Shall be of type `logical` and either *a* scalar or an array of the same shape as `array`.

ivan-pi:
I am fine with both `cov` or `covariance`. In case we decide to go for the long version, I would suggest renaming `var` to `variance`. They can always be renamed at import if desired, using `use stdlib_stats, only: cov => covariance`.

Overall a nice proposal, so I am fine with merging.

certik:
Looks good. Yes `cov` seems fine, consistent with other languages.

src/stdlib_experimental_stats.md
jvdp1: Thank you.  Changed!

src/stdlib_experimental_stats.md
jvdp1: Changed!

jvdp1:
It seems that all of you agreed with `cov` as name for this function.
If there are no other comments about the implementation and API, it could be merged.

Note: An implementation to compute the covariance between 2 arrays still needs to be done, but it will be for another PR.

jvdp1:
I will merge this PR. Thank you for your reviews and approvals.

PR #183, opened 2020-05-14
Add automatic API-doc generation & deployment

zbeekman:
EDIT: Preview the documentation here: https://fortran-lang.github.io/stdlib-docs/index.html
NOTE: Autodeploy failed, not sure if it's due to this being a PR from a fork, or if `${{ secrets.ACTIONS_DEPLOY_KEY }}` was never setup, or if it is incompatible with cross-repository deployment.

PR to automatically generate and deploy documentation using [FORD](https://github.com/Fortran-FOSS-Programmers/ford#readme) published via GH pages.

Notes:
- I think we should move the specs to a dedicated directory, and then they can be rendered by FORD.
- It would be nice to deploy these to something like `stdlib.fortran-lang.org`.
  - I can make this happen but we need a DNS entry so I can add a CNAME file in the deployed [gh-pages repo](https://github.com/fortran-lang/stdlib-docs) CC: @milancurcic @certik 
- May want to move the WORKFLOW.md and STYLEGUIDE.md documents into the FORD pages dir

For now this PR is a WIP.

Begins to address issues:

- #4 
- #182 
- #94

PR #184, opened 2020-05-14
Zb documentation

certik:
Opening this per @zbeekman's request at #183.


zbeekman:
I think #183 is in good shape now. I was able to test deployment by adding the SSH secret key to my fork.

Issue #38, opened 2019-12-22
Sparse matrix support


certik:
Prior art:

* SciPy: https://docs.scipy.org/doc/scipy/reference/sparse.html
* Matlab: https://www.mathworks.com/help/matlab/ref/sparse.html
* Julia: https://docs.julialang.org/en/v1/stdlib/SparseArrays/index.html
* @certik: https://github.com/certik/hfsolver/blob/b4c50c1979fb7e468b1852b144ba756f5a51788d/src/sparse.f90
* @jvdp1: https://github.com/jvdp1/libsparse

There are probably many more implementations. I really like the SciPy simple non-OO implementation, and I have ported it to modern Fortran in the link above. If people also want an OO implementation, then it can be build on top as an option.

One thing that I found out is that one must [sort the indices](https://github.com/certik/hfsolver/blob/b4c50c1979fb7e468b1852b144ba756f5a51788d/src/sparse.f90#L185) and the overall speed very much depends on how quickly one can sort it. I ended up using `quicksort`, but it might be even faster to use some specialized sorting algorithm (such as [Timsort](https://en.wikipedia.org/wiki/Timsort)) because in practice, the indices have subsections that are already sorted (typically coming from some local to global mapping as in finite elements), but overall it is not sorted.

jvdp1:
A non-OO implementation could be an easier start indeed.
An issue we may need to discuss about sparse matrices is that NNZ may be larger than integer(4).
Also we will need to discuss the formats to support. COO and CRS3 could be a good start.

> One thing that I found out is that one must [sort the indices](https://github.com/certik/hfsolver/blob/b4c50c1979fb7e468b1852b144ba756f5a51788d/src/sparse.f90#L185) and the overall speed very much depends on how quickly one can sort it. I ended up using `quicksort`, but it might be even faster to use some specialized sorting algorithm (such as [Timsort](https://en.wikipedia.org/wiki/Timsort)) because in practice, the indices have subsections that are already sorted (typically coming from some local to global mapping as in finite elements), but overall it is not sorted.

In my field, entries are added to a sparse matrix in a quite random way. Therefore, the indices are far to be sorted, even in a row. Therefore, I use `quicksort` as implemented in `LAPACK`.

certik:
[...]

PR #189, opened 2020-05-16
Initial implementation of COO / CSR sparse format

certik:
The key of this is the API, as shown with examples in test_sparse.f90.
In particular, I am proposing the following functions / API:

Higher level API:

Conversion Dense <-> COO:

* dense2coo
* coo2dense

Conversion COO -> CSR:

* coo2csr

CSR functionality:

* csr_matvec
* csr_getvalue

Dense functionality:

* getnnz

Lower level API:

* csr_has_canonical_format
* csr_sum_duplicates
* csr_sort_indices

In these algorithms, it seems it is possible to just have one (best)
implementation with one exception: sorting of indices (as implemented by
`csr_sort_indices`), where different algorithms give different
performance and it depends on the actual matrix. As such, it might make
sense to provide a default in `csr_sort_indices` that is called by
`coo2csr` that is as fast as we can make it for most cases (currently it
uses quicksort, but there are other faster algoritms for our use case
that we should switch over) and then provide other implementations such
as `csr_sort_indices_mergesort`, `csr_sort_indices_timsort`, etc. for
users so that they can choose the algorithm for sorting indices, or even
provide their own.

The file stdlib_experimental_sparse.f90 provides an example
implementation of these algorithms that was inspired by SciPy. We can
also discuss the details of that, but the key that I would like to
discuss is the API itself.

See #38.


certik:
@sfilippone I would really be interested in your feedback on the API, based on your work on psblas3. This is a serial API, which I think we should have, in addition to a parallel API that we'll tackle later.

src/stdlib_experimental_sparse.f90
```diff
@@ -0,0 +1,335 @@
...
+end do
+end subroutine
+
+function csr_matvec(Ap, Aj, Ax, x) result(y)
+! Compute y = A*x for CSR matrix A and dense vectors x, y
+integer, intent(in) :: Ap(:), Aj(:)
+real(dp), intent(in) :: Ax(:), x(:)
+real(dp) :: y(size(Ap)-1)
+integer :: i
+!$omp parallel default(none) shared(Ap, Aj, Ax, x, y) private(i)
```
zerothi, line 201: Generally I think `omp` should be orphaned. In that way it is the user who controls the parallelization, and not the stdlib.

src/stdlib_experimental_sparse.f90
```diff
@@ -0,0 +1,335 @@
...
+        if (abs(B(i, j)) < tiny(1._dp)) cycle
+        Ai(idx) = i
+        Aj(idx) = j
+        Ax(idx) = B(i, j)
+        idx = idx + 1
+    end do
+end do
+end subroutine
+
+integer function getnnz(B) result(nnz)
```
zerothi, line 30: Also, perhaps these methods should allow `integer(8)` to  allow very large matrices.

Perhaps this should be `get_nnz`?

src/stdlib_experimental_sparse.f90
```diff
@@ -0,0 +1,335 @@
...
+function csr_matvec(Ap, Aj, Ax, x) result(y)
+! Compute y = A*x for CSR matrix A and dense vectors x, y
+integer, intent(in) :: Ap(:), Aj(:)
+real(dp), intent(in) :: Ax(:), x(:)
+real(dp) :: y(size(Ap)-1)
+integer :: i
+!$omp parallel default(none) shared(Ap, Aj, Ax, x, y) private(i)
+!$omp do
+do i = 1, size(Ap)-1
+    y(i) = dot_product(Ax(Ap(i):Ap(i+1)-1), x(Aj(Ap(i):Ap(i+1)-1)))
```
zerothi, line 204: This will create a temporary which *could* potentially be very large. I would try to avoid this since the temporary still needs to traverse the copied elements.

src/stdlib_experimental_sparse.f90
```diff
@@ -0,0 +1,335 @@
...
+    if (A(idx) < val) then
+        l = idx
+    else
+        i = idx
+    end if
+end do
+end function
+
+
+real(dp) function csr_getvalue(Ap, Aj, Ax, i, j) result(r)
```
zerothi, line 238: `csr_get_value`?

I have never really liked `get` when intent is obvious, I may be wrong.
But `csr_value` is just as clear to me. ;)

certik:
Great comments! Thank you.

I'll reply later, I'll just point out a general design that we figured out that will satisfy almost everybody in the community: have a low level API that is procedural with no side effects (some people call it functional) as in this PR. And then have a high level API that is object oriented and hides complexity in a class or derived types and that uses the low level API to do the work.

If you want, you can help suggest some good high level API for this.

src/stdlib_experimental_sparse.f90
```diff
@@ -0,0 +1,335 @@
...
+        Ai(idx) = i
+        Aj(idx) = j
+        Ax(idx) = B(i, j)
+        idx = idx + 1
+    end do
+end do
+end subroutine
+
+integer function getnnz(B) result(nnz)
+real(dp), intent(in) :: B(:, :)
```
jvdp1, line 31: This is for dense matrices.
Since we are in a module for sparse matrices, should it return nnz only for sparse matrices?

src/stdlib_experimental_sparse.f90
```diff
@@ -0,0 +1,335 @@
...
+! COO
+
+subroutine coo2dense(Ai, Aj, Ax, B)
+integer, intent(in) :: Ai(:), Aj(:)
+real(dp), intent(in) :: Ax(:)
+real(dp), intent(out) :: B(:, :)
+integer :: n
+B = 0
+do n = 1, size(Ai)
+    B(Ai(n), Aj(n)) = B(Ai(n), Aj(n)) + Ax(n)
```
jvdp1, line 51: This implies that all elements in `Ai` and `Aj` are filled, which is not always the case because COO is often used as a temporary format.

src/stdlib_experimental_sparse.f90
```diff
@@ -0,0 +1,335 @@
...
+end do
+end function
+
+! COO
+
+subroutine coo2dense(Ai, Aj, Ax, B)
+integer, intent(in) :: Ai(:), Aj(:)
+real(dp), intent(in) :: Ax(:)
+real(dp), intent(out) :: B(:, :)
+integer :: n
```
jvdp1, line 48: I would suggest to always use `integer(int64)` when going through a sparse matrix.

src/stdlib_experimental_sparse.f90
```diff
@@ -0,0 +1,335 @@
...
+end do
+end subroutine
+
+subroutine coo2csr(Ai, Aj, Ax, Bp, Bj, Bx)
+! Converts from COO (Ai, Aj, Ax) into CSR (Bp, Bj, Bx)
+! Row and column indices are *not* assumed to be ordered.
+! Duplicate entries are carried over to the CSR representation.
+integer, intent(in) :: Ai(:), Aj(:)
+real(dp), intent(in) :: Ax(:)
+integer, intent(out) :: Bp(:), Bj(:)
```
jvdp1, line 61: I suggest to declare `Bp(:)` as `integer(int64)` to avoid any troubles with large sparse matrices.
However, we may want to be compatible to e.g., psblas3 (I didn't check what is the default declaration in psblas3).

src/stdlib_experimental_sparse.f90
```diff
@@ -0,0 +1,335 @@
...
+real(dp), allocatable, intent(out) :: Bx(:)
+logical, optional, intent(in) :: verbose
+integer :: Bj_(size(Ai))
+real(dp) :: Bx_(size(Ai))
+integer :: nnz
+logical :: verbose_
+verbose_ = .false.
+if (present(verbose)) verbose_ = verbose
+allocate(Bp(maxval(Ai)+1))
+if (verbose_) print *, "coo2csr"
```
jvdp1, line 113: `optval` can be used here ;)

src/stdlib_experimental_sparse.f90
```diff
@@ -0,0 +1,335 @@
...
+call coo2csr(Ai, Aj, Ax, Bp, Bj_, Bx_)
+if (verbose_) print *, "csr_sort_indices"
+call csr_sort_indices(Bp, Bj_, Bx_)
+if (verbose_) print *, "csr_sum_duplicates"
+call csr_sum_duplicates(Bp, Bj_, Bx_)
+if (verbose_) print *, "done"
+nnz = Bp(size(Bp))-1
+allocate(Bj(nnz), Bx(nnz))
```
jvdp1, line 121: This line can be removed with Fortran2003 and>, right?

src/stdlib_experimental_sparse.f90
```diff
@@ -0,0 +1,335 @@
...
+    do j = Ap(i)+1, Ap(i+1)-1
+        if (Aj(j-1) >= Aj(j)) return
+    end do
+end do
+r = .true.
+end function
+
+subroutine csr_sum_duplicates(Ap, Aj, Ax)
+! Sum together duplicate column entries in each row of CSR matrix A
+! The column indicies within each row must be in sorted order.
```
jvdp1, line 147:
```suggestion
! The column indices within each row must be in sorted order.
```

src/stdlib_experimental_sparse.f90
```diff
@@ -0,0 +1,335 @@
...
+    r1 = Ap(i)
+    r2 = Ap(i+1)-1
+    l = r2-r1+1
+    idx(:l) = iargsort_quicksort(Aj(r1:r2))
+    Aj(r1:r2) = Aj(r1+idx(:l)-1)
+    Ax(r1:r2) = Ax(r1+idx(:l)-1)
+end do
+end subroutine
+
+function csr_matvec(Ap, Aj, Ax, x) result(y)
```
jvdp1, line 195: I suggest to follow the API of Sparse BLAS.

src/stdlib_experimental_sparse.f90
```diff
@@ -0,0 +1,335 @@
...
+if (offset <= row_end) then
+    if (Aj(offset) == j) then
+        r = Ax(offset)
+        return
+    end if
+end if
+r = 0
+end function
+
+pure elemental subroutine swap_int(x,y)
```
jvdp1, line 257: I think this subroutine should be moved to another module of `stdlib`.

src/tests/sparse/test_sparse.f90
```diff
@@ -0,0 +1,117 @@
...
+call check(all(abs(Bx(:nnz) - [5, 7, 1, 3]) < 1e-12_dp))
+call check(csr_has_canonical_format(Bp, Bj(:nnz)))
+
+call coo2csr_canonical(Ai, Aj, Ax, Bp, Bj, Bx)
+call check(all(Bp == [1, 2, 4, 5]))
+call check(all(Bj == [1, 3, 4, 2]))
+call check(all(abs(Bx - [5, 7, 1, 3]) < 1e-12_dp))
+call check(csr_has_canonical_format(Bp, Bj))
+
+call check(abs(csr_getvalue(Bp, Bj, Bx, 1, 1) - 5) < tiny(1._dp))
```
jvdp1, line 102: By experience, `tiny(1._dp)` can be too severe.

sfilippone:
Re: int32 vs int64.
You'll need both. Let me elaborate.
The most common application of sparse matrices is in the inner steps of
some PDE solver, dealing with a discretization on a mesh. Such problems can
indeed reach sizes that need indices with int64; however in those cases
they are tackled through parallel solution strategies mostly based on MPI.
The parallelization strategy is to assign a subdomain to each MPI process;
each subdomain is most likely of a size that fits comfortably in an int32,
provided that you apply a local numbering scheme, as well as a global
numbering scheme. And you occasionally need to convert the indices of the
local portion into global numbering for some processing, and back; examples
include data exchanges among processes within a transpose operation.
The solution that I have now in the development branch of PSBLAS is the
following: there are 4 integer kinds:
int32==psb_mpk_ <= psb_ipk_ <= psb_lpk_ <= psb_epk_== int64
IPK is the integer kind used for *LOCAL indices, which are the ones for the
local portion of the matrices involved in numerical operators such as
Matrix-Vector product,
LPK is the integer kind used for GLOBAL indices, which are used in pre- and
post-processing.
IPK and LPK can be specified at configure time within the constraints abov,
with the default being IPK=int32 and LPK=int64.

Salvatore

jvdp1:
> Re: int32 vs int64. You'll need both. Let me elaborate. The most common application of sparse matrices is in the inner steps of some PDE solver, dealing with a discretization on a mesh. Such problems can indeed reach sizes that need indices with int64; however in those cases they are tackled through parallel solution strategies mostly based on MPI. The parallelization strategy is to assign a subdomain to each MPI process; each subdomain is most likely of a size that fits comfortably in an int32, provided that you apply a local numbering scheme, as well as a global numbering scheme. And you occasionally need to convert the indices of the local portion into global numbering for some processing, and back; examples include data exchanges among processes within a transpose operation. The solution that I have now in the development branch of PSBLAS is the following: there are 4 integer kinds: int32==psb_mpk_ <= psb_ipk_ <= psb_lpk_ <= psb_epk_== int64 IPK is the integer kind used for *LOCAL indices, which are the ones for the local portion of the matrices involved in numerical operators such as Matrix-Vector product, LPK is the integer kind used for GLOBAL indices, which are used in pre- and post-processing. IPK and LPK can be specified at configure time within the constraints abov, with the default being IPK=int32 and LPK=int64. Salvatore

Thank you Salvatore for your input. I should have a closer look to psblas.

For `stdlib`: would it not be of interest to integrate psblas3 (if allowed) in `stdlib` directly? Or through `fpm`? In comparison to the other implementations, psblas3 is well tested, is associated to peer-reviewed publications and is parallelized (also with Coarray Fortran?) if we want to support parallelization in stdlib. If something is missing, we could identify it and maybe collaborate with psblas to implement the missing parts there?
This is just a question to be sure that we don't start into something that is similar to reinventing the wheel.

sfilippone:
I am OK with integrating PSBLAS in stdlib (with proper handling of
copyright for me and my collaborators), at least the parts that make sense
today (I suspect I have a lot of stuff that should not be included just
yet).  In the project there are a set of facilities marked at SERIAL
(admittedly, they should be packaged better because they are split in
two-three subdirs) which are the ones you are probably most interested in.
Note that I also use a home-grown set of sed scripts to have a crude
emulation of macros/templates to avoid rewriting things multiple times; if
that could be morphed into something robust I'd be more than happy, while
waiting for the Fortran standard committee to decide on a template facility
(I have seen more than one proposal in this direction, but it's not yet
decided AFAIK).

A few additional comments:
1. Storage formats.
I have a set of additional storage formats in psblas3-ext; they also
support CUDA, accessed through the  C interoperability.

2. Coarrays.
I have a preliminary version with CoArrays (in fact, I am one of the
founders of the OpenCoArrays project which is used to provide CoArrays in
GNU). However, there is a big issue and that is constraints C825. The
effect of the constraint is that the set of entities that can have a
coarray component is fixed at compile time; in other words, you cannot have
an allocatable array of derived types that contains a(n allocatable)
coarray component.
I have reported this to the Fortran standard committee  (see
https://j3-fortran.org/doc/year/18/18-280r1.txt ) and indeed my proposal
has been approved; however there is no telling when it will appear in
compilers, hence I am not going to use this in production code any time
soon.

Salvatore

certik:
Finally got to my computer. Thanks everybody for the feedback, I'll first address the high level feedback: do we want this at all, or should stdlib just use psblas3?

My own view is that stdlib should not be competing with or reimplementing well established libraries, whether LAPACK, psblas3, or others.

Just linking against psblas3 but not providing any additional functionality or API think is also not worth it. As the Fortran Package Manager becomes more successful, it will be very easy to depend on any library such as psblas3 in users projects, and we should be encouraging people to do exactly that. It will be good for users, good for developers of libraries and the whole ecosystem.

Where I see the role of stdlib is to standardize API and provide a reference implementation to functionalities that are repeated from project to project with slightly different API each. There are many implementations of CSR and other functionality, see #38 for a partial list, and there are many codes that use this in some form. We should see if there is a way to figure out an API that each of these codes could use.

The other aspect where I see stdlib as contributing is when people just want to quickly use some functionality in their research / new projects, or later on interactively in the Jupyter notebook for example, having a nice and well documented API that is very easy to use for new users would be beneficial. Later on, for big specialized production codes, one can always use specialized libraries. As an example, most electronic structure codes have their own custom specialized eigensolver. But I think it is still worth having a nice API to dense eigensolver via Lapack, and a sparse eigensolver on top of this sparse functionality, so that people who just want to prototype some algorithm in Fortran can do so as easily as with SciPy, Matlab or Julia.

In order to move the conversation forward, we have to have both non-OO low level api as well as a high level OO API. The reason is that I know many people (myself included) who do not want to use the OO API, just simple subroutines that do the job. But I also know a lot of people who would prefer the OO API. By providing both, we can capture most of the Fortran community and make stdlib useful for them.

@sfilippone at the very least I can see that we can use your sorting algorithms. I am hoping there is more that we can figure out how to reuse.

Also, I just want to start with a serial implementation, we can overload the subroutines to work with both int32 and int64 etc. I am really hoping we can figure out an API that we would all agree would make sense to have in stdlib.

jvdp1:
> But I think it is still worth having a nice API to dense eigensolver via Lapack, and a sparse eigensolver on top of this sparse functionality, so that people who just want to prototype some algorithm in Fortran can do so as easily as with SciPy, Matlab or Julia.

@certik I totally agree with your comments.However, I wonder if this could be achieved by interfacing with the serial version of psblas, even partially. If it is not possible (e.g., because of no non-OO low level API), then I am fine with this, and I will help as good as I can.

milancurcic:
I don't think that stdlib should just gulp up existing libraries and expose them under different name. 

First, decide if sparse matrix algebra is in scope or not. I'm not the target user (although I use SMM through ESMF under the hood daily), but if we agreed that a SciPy-like functionality is in scope, then this is too. 

Second, if you do add it to stdlib, I think that it should provide added value over any existing library. In this case, I see that the added value could be ease of installation, and ease of use (simpler API). 

This resonated with me personally the most:

> The other aspect where I see stdlib as contributing is when people just want to quickly use some functionality in their research / new projects, or later on interactively in the Jupyter notebook for example, having a nice and well documented API that is very easy to use for new users would be beneficial. Later on, for big specialized production codes, one can always use specialized libraries.

milancurcic:
Just to be clear, I personally think that SciPy-like scope is included in the scope of Fortran stdlib.

src/stdlib_experimental_sparse.f90
certik: Good comment. Let's discuss openmp at #213.

I forgot that I used openmp here, I am completely fine to remove it, that is an orthogonal issue to sparse matrices. If we decide to use openmp in stdlib, we can keep it here, if we decide not to, we can remove it. And I am happy to remove it to merge this PR, and we can always add it back in. Let's discuss more in #213.

zerothi: Great. :)

awvwgk:
What is the current status of this patch? We have a couple of merge conflicts in the build files, which need a merge/rebase against the default branch. Also, there are a couple of unaddressed review comments. @certik, is there anything we can do to help you move this pull request forward?

milancurcic:
@certik would you like to continue working on this or should we close it for the time being?

awvwgk:
I marked this patch as *wontfix* since you are not intending to continue working on it. But the implementation here is of course free to use for other contributors to continue working on this project.

certik:
I wrote

> I don't have the time right now, but **I want to finish this**, or help others finish it.

So I disagree with the statement:

> since you are not intending to continue working on it.

Furthermore, I think marking things as "wontfix" is not encouraging to others (or me!) to pick it up and finish it.

We need to help each other finish these things and get them into stdlib to move stdlib forward.

You don't want such COO functionality in there? 

If you do, then let's plan how to do it.

awvwgk:
Sorry, that was not what I meant. This patch seemed stale and open for others to pickup on, at least that was the situation I usually applied this label for my own patches.

milancurcic:
FWIW, the "wontfix" label sounds demotivating to me as well. Like the whole project is yelling at me "we don't want this". I don't mind a PR staying open for a long time, if that's what it takes to find people to see it through.

awvwgk:
Be assured that I want to move project forward. Frankly, I don't think *this* patch can be moved forward in its current form, even after resolving the merge conflicts.

I have been looking into this patch a couple of times now, and stopped again because it has significantly diverged from the latest default branch, which makes it quite involved to work with the implementation because a lot of changes in the general structure and build files are just missing.

Instead I would suggest to salvage the implementation and start fresh from the latest default branch, addressing the feedback provided already. We have done so in the past with other patches like the stringlist and were quite successful with this approach.

certik:
@awvwgk awesome. I know that's not what you meant and that you want to move the project forward. But that is how I perceive it and I think many others too. To move forward, I think the label that you just used "help wanted" is perfect. Let's keep this PR open until a better one is up.

vickysharma0812:
@certik 

In the following subroutine

https://github.com/fortran-lang/stdlib/blob/89feb9167b8ed0cd95a1c8a027646ca1fec362a1/src/stdlib_experimental_sparse.f90#L44-L53

you can exploit `do concurrent`

```fortran
do concurrent(n = 1:size(Ai))
    B(Ai(n), Aj(n)) = B(Ai(n), Aj(n)) + Ax(n)
end do
```

loiseaujc:
Following #760, I guess this PR could be closed?

(Yep, I'm digging very far into the list of open PR to just because I think it'd be good to close those which are no longer relevant or stalled for too long just to increase the clarity of what is currently being worked on in `stdlib` :] )

PR #192, opened 2020-05-17
Addition of tests and clarification for cov

jvdp1:
I added some tests for `cov` and modified 2 functions (when a mask array is provided).


milancurcic:
Thank you!

PR #194, opened 2020-05-18
Addition of description for Ford

jvdp1:
Addition of small descriptions of the different modules for Ford.

Questions: Should we add descritpions for the different functions in the code? Or only in the specs? Or is it possible to combine specs and code through Ford?

@zbeekman @milancurcic @certik


milancurcic:
Thank you @jvdp1. This a small and straightforward PR, so I will merge it.

zbeekman:
LGTM. I typically put spaces between the double bang and text but as long as it builds and renders correctly I have no objections.

milancurcic:
> I typically put spaces between the double bang and text

Indeed, I agree, we should aim for the docstrings to be readable from the code as well.

PR #199, opened 2020-05-23
Addition of links for Ford

jvdp1:


doc/specs/stdlib_experimental_io.md
```diff
@@ -6,14 +6,14 @@ title: experimental_IO
 
 [TOC]
 
-## `loadtxt` - load a 2D array from a text file
+## `[[stdlib_experimental_io(module):loadtxt(interface)]]` - load a 2D array from a text file
```
zbeekman, line 9: I know I'm to blame for this, since I started putting links in the headings, but, JUST as an FYI and point for further discussion, [this breaks the TOC markdown extension](https://stdlib.fortran-lang.org/page/specs/stdlib_experimental_error.html) and in the TOC there are no longer links to the heading, only the link from within the heading to the linked FORD documentation.

IMO, this is a bug in the python markdown extra package (or perhaps FORD, but this is less likely) but I just wanted to point out this issue.

doc/specs/index.md
```diff
@@ -11,24 +11,17 @@ This is and index/directory of the specifications (specs) for each new module/fe
...
- - [quadrature](./stdlib_experimental_quadrature.html)
-   - Numerical integration
- - [stats](./stdlib_experimental_stats.html)
-   - Descriptive Statistics
+ - [error](./stdlib_experimental_error.html) - Catching and handling errors
+ - [IO](./stdlib_experimental_io.html) - Input/output helper & convenience
+ - [linalg](./stdlib_experimental_linalg.html) - Linear Algebra
+ - [optval](./stdlib_experimental_optval.html) - Fallback value for optional arguments
+ - [quadrature](./stdlib_experimental_quadrature.html) - Numerical integration
+ - [stats](./stdlib_experimental_stats.html) - Descriptive Statistics
```
zbeekman, line 19: These changes are an improvement IMO, cheers!

doc/specs/index.md
```diff
@@ -11,24 +11,17 @@ This is and index/directory of the specifications (specs) for each new module/fe
...
 
 ## Missing specs
 
-@todo
-Populate this section with a list of missing specs and link to their
-[sources on GH](https://github.com/fortran-lang/stdlib/tree/master/src).
+ - [ascii](https://github.com/fortran-lang/stdlib/blob/master/src/stdlib_experimental_ascii.f90)
+ - [kinds](https://github.com/fortran-lang/stdlib/blob/master/src/stdlib_experimental_kinds.f90)
```
zbeekman, line 24: Thanks for filling these in!

src/stdlib_experimental_error.f90
```diff
@@ -1,5 +1,5 @@
 module stdlib_experimental_error
-    !! Provide support for catching and handling errors ([spec](../page/specs/stdlib_experimental_error.html))
+    !! Provide support for catching and handling errors.
```
zbeekman, line 2: Yes, this was a bit redundant wasn't it. Thanks!

src/stdlib_experimental_io.fypp
```diff
@@ -157,20 +165,25 @@ contains
...
+    !!
+    !! __Read the [specification here](../page/specs/stdlib_experimental_io.html#description_1)__
+    !!
+    !!##### Behavior
+    !!
+    !!
+    !! To open a file to read:
+    !!
+    !! u = open("somefile.txt")        # The default `mode` is "rt"
+    !! u = open("somefile.txt", "r")
```
zbeekman, line 178: Just wondering, why is `#` used for a comment here? You should be able to use the normal Fortran comment character, `!`. Also, if you don't want to embed FORD links/crossrefs,  then I would recommend putting this in a pre-formatted code block. This will turn on syntax highlighting (using pygments) in the generated documentation.

e.g.;
 ```suggestion
    !!``` fortran
    !! u = open("somefile.txt")        # The default `mode` is "rt"
    !! u = open("somefile.txt", "r")
    !!```
```

**Comment by zbeekman** [comment]
Also, just FYI, you can download the FORD documentation generated to get a preview and double check formatting etc. The build for your latest push is here: https://github.com/fortran-lang/stdlib/actions/runs/113339579 and if you [click the artifact archive it will download to your local machine.](https://github.com/fortran-lang/stdlib/suites/711257100/artifacts/6935317)

File: doc/specs/stdlib_experimental_io.md (hunk 1 above)
Comment (jvdp1) [reply doc/specs/stdlib_experimental_io.md]: I removed them. Thanks for the explanation.
I think it would be good to have a link from the spec to the code. Weprobalbly could add something like:

See the code for loadtxt here

after the title.

**Comment by jvdp1** [comment]
> Also, just FYI, you can download the FORD documentation generated to get a preview and double check formatting etc. The build for your latest push is here: https://github.com/fortran-lang/stdlib/actions/runs/113339579 and if you [click the artifact archive it will download to your local machine.](https://github.com/fortran-lang/stdlib/suites/711257100/artifacts/6935317)

Thanks for the tip.

**Comment by jvdp1** [comment]
@zbeekman Thank you for your review. I modified the files as suggested.

File: doc/specs/stdlib_experimental_io.md (hunk 1 above)
Comment (zbeekman) [reply doc/specs/stdlib_experimental_io.md]: I like this idea!

File: src/stdlib_experimental_error.f90 [hunk 6]
```diff
@@ -22,7 +22,7 @@ end subroutine error_stop
 contains
 
 subroutine check(condition, msg, code, warn)
-    !! Checks the value of a logical condition. ([spec](../page/specs/stdlib_experimental_error.html#description))
```
zbeekman, line 25: @jvdp1 By the way, this *is* redundant, but unless you add the `!! summary: ...` meta data to the docstrings explicitly then the first paragraph is used for summary text on pages [like this one](https://stdlib.fortran-lang.org/lists/modules.html). (Look at the "description" column.) So I added an abbreviated link to the spec in the summary/first paragraph so that you could easily navigate to the spec from the [overview page](https://stdlib.fortran-lang.org/lists/modules.html).

I'm happy either way, but I thought it was convenient to have there.

jvdp1: Thanks @zbeekman for the review. I agree with your point-of-view, and also prefer that the link to the spec appears in the "description" column.  So I modified all the files such that the link to the spec appears in the description column for all modules and procedures (when available).

zbeekman: I haven't looked at the generated documentation, but your latest commit looks like it will do what you intend it to! Thanks again for working on this!

jvdp1:
Good. I'll merge it.

PR #200, opened 2020-05-24
Small re-organizations and additions

jvdp1:


certik:
I think the CoC should stay in the root location for better visibility and also GitHub I think links to it from the UI in some circumstances.

jvdp1:
> I think the CoC should stay in the root location for better visibility and also GitHub I think links to it from the UI in some circumstances.

I inserted a note about the CoC in the first page.

certik:
Ah, my bad -- I thought you moved the CoC file from the root directory, but you only moved the one from doc/ directory. However, your small note that you added at the main API page helps, I think it fits there.

So +1 to merge as is.

PR #205, opened 2020-05-30
test for standard conformance

MarDiehl:
There are several compiler options that help to detect non-standard conformant code.
I have included some of them (only for Intel and GNU compiler).
Unfortunately, the result is that the code does not compile:

    Intel compiler:
    /tmp/stdlib/build/src/stdlib_experimental_linalg_diag.f90(7): error #6645: The name of the module procedure conflicts with a name in the encompassing scoping unit.   [DIAG_RSP]
          function diag_rsp(v) result(res)
    GNU compiler
    /tmp/stdlib/src/tests/optval/test_optval.f90:252:15:
    
      252 |     z = optval(x, [2.0_qp, -2.0_qp])
            |               1
            Error: ‘x’ at (1) is an array and OPTIONAL; IF IT IS MISSING, it
            cannot be the actual argument of an ELEMENTAL procedure unless
            there is a non-optional argument with the same rank (12.4.1.5)
            [-Werror=pedantic]

There is also a high number of warnings related to type conversions.


milancurcic:
Martin, thank you and welcome! I think this will be a great step forward. I suggest that we work our way through the errors raised in this PR and fix them until all tests pass.

certik:
In Fortran, are you supposed to use `cmplx` or `complex`?

milancurcic:
`cmplx` is an [intrinsic conversion function](https://gcc.gnu.org/onlinedocs/gfortran/CMPLX.html) (like `int`, `real`, and `char`) and `complex` is the name of the type. I confused this myself many times before but the compiler always corrected me (I don't remember which one). In this case, it's a [GNU extension](https://gcc.gnu.org/onlinedocs/gfortran/COMPLEX.html).

certik:
Another pitfall the compiler should provide a nice error / warning message. Thanks!

jvdp1:
Thank you for starting this.
I am not a huge fan of `-Wconversion` because it reports many implicit conversions. Others will help us to avoif pitfalls like `complex` vs `cmplx`.

MarDiehl:
> Martin, thank you and welcome! I think this will be a great step forward. I suggest that we work our way through the errors raised in this PR and fix them until all tests pass.

For gfortran, the `optval` tests fail:
```
Error: ‘x’ at (1) is an array and OPTIONAL; IF IT IS MISSING, it cannot be the actual argument of an ELEMENTAL procedure unless there is a non-optional argument with the same rank (12.4.1.5)
```

I'm also not sure if you are even allowed to pass in an optional argument into a function. Is `x` in `optval` present or not?

MarDiehl:
> Thank you for starting this.
> I am not a huge fan of `-Wconversion` because it reports many implicit conversions. Others will help us to avoif pitfalls like `complex` vs `cmplx`.

I would say for a statically typed language, these warnings are reasonable. Once just has to carefully set the types (5 is not 5.)

milancurcic:
> I'm also not sure if you are even allowed to pass in an optional argument into a function. Is x in optval present or not?

You can pass it to a procedure if the corresponding dummy argument is also declared `optional`.

It looks like we'll need to remove the `elemental` attribute from `optval` and use fypp to create specific procedures for all ranks.

jvdp1:
> It looks like we'll need to remove the `elemental` attribute from `optval` and use fypp to create specific procedures for all ranks.

I am currently trying this option. However, we then must check is the actual arguments have the same dimension, and use `stop error` if it is not the case. This was ensure by the `elementatal` function. The `pure` statement might need to be removed if a message is provided.

MarDiehl:
> > It looks like we'll need to remove the `elemental` attribute from `optval` and use fypp to create specific procedures for all ranks.
> 
> I am currently trying this option. However, we then must check is the actual arguments have the same dimension, and use `stop error` if it is not the case. This was ensure by the `elementatal` function. The `pure` statement might need to be removed if a message is provided.

To be honest, I have problems to understand the error of gfortran. What does `there is` means? Does it refer to the calling functions or to the called (elemental) function? `optval` has a second, non-optional argument.

jvdp1:
> > > It looks like we'll need to remove the `elemental` attribute from `optval` and use fypp to create specific procedures for all ranks.
> > 
> > 
> > I am currently trying this option. However, we then must check is the actual arguments have the same dimension, and use `stop error` if it is not the case. This was ensure by the `elementatal` function. The `pure` statement might need to be removed if a message is provided.
> 
> To be honest, I have problems to understand the error of gfortran. What does `there is` means? Does it refer to the calling functions or to the called (elemental) function? `optval` has a second, non-optional argument.

My first impression was that the message was clear (to me at least). To be sure, I went to the [gcc source code](https://github.com/gcc-mirror/gcc/blob/master/gcc/fortran/resolve.c).
Now with [this comment in this code](https://github.com/gcc-mirror/gcc/blob/dd38c765a04d06c775134a135f68b18c3b7c9c78/gcc/fortran/resolve.c#L2245):
```c
/* If it is an array, it shall not be supplied as an actual argument
     to an elemental procedure unless an array of the same rank is supplied
     as an actual argument corresponding to a nonoptional dummy argument of
     that elemental procedure(12.4.1.5).  */
```
, I am doubting.

MarDiehl:
> > > > It looks like we'll need to remove the `elemental` attribute from `optval` and use fypp to create specific procedures for all ranks.
> > > 
> > > 
> > > I am currently trying this option. However, we then must check is the actual arguments have the same dimension, and use `stop error` if it is not the case. This was ensure by the `elementatal` function. The `pure` statement might need to be removed if a message is provided.
> > 
> > 
> > To be honest, I have problems to understand the error of gfortran. What does `there is` means? Does it refer to the calling functions or to the called (elemental) function? `optval` has a second, non-optional argument.
> 
> My first impression was that the message was clear (to me at least). To be sure, I went to the [gcc source code](https://github.com/gcc-mirror/gcc/blob/master/gcc/fortran/resolve.c).
> Now with [this comment in this code](https://github.com/gcc-mirror/gcc/blob/dd38c765a04d06c775134a135f68b18c3b7c9c78/gcc/fortran/resolve.c#L2245):
> 
> ```c
> /* If it is an array, it shall not be supplied as an actual argument
>      to an elemental procedure unless an array of the same rank is supplied
>      as an actual argument corresponding to a nonoptional dummy argument of
>      that elemental procedure(12.4.1.5).  */
> ```
> 
> , I am doubting.

Same here. In case that an optional argument is not given, there is still the other argument to determine the properties of the return value.
We should discuss report it to the gfortran developers. I will also check if it compiles with Intel.

jvdp1:
From the [Fortran Standard](http://isotc.iso.org/livelink/livelink?func=ll&objId=19442438&objAction=Open&viewType=1) (15.5.2.12):

```
3 An optional dummy argument that is not present is subject to the following restrictions.
.....
(6) If it is an array, it shall not be supplied as an actual argument to an elemental procedure unless an
array of the same rank is supplied as an actual argument corresponding to a nonoptional dummy
argument of that elemental procedure.
....
```
I understand the same from this sentence.

> We should discuss report it to the gfortran developers. I will also check if it compiles with Intel.

I agree. Btv, which version of gfortran did you use (I used GFortran9.3)?

MarDiehl:
> I agree. Btv, which version of gfortran did you use (I used GFortran9.3)?

10.1

jvdp1:
> Intel compiler:
> /tmp/stdlib/build/src/stdlib_experimental_linalg_diag.f90(7): error #6645: The name of the module procedure conflicts with a name in the encompassing scoping unit.   [DIAG_RSP]
>       function diag_rsp(v) result(res)

This issue should be solved by #208

MarDiehl:
Gfortran issue reported as bug on https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95446

MarDiehl:
in general, ready to merge.
But please discuss whether you are ok with all the warnings. I usually prefer to turn on all compiler warnings and remove them step by step. A few false positives will always remain.
The most of the warnings here are related to type conversions and can be easily fixed by introducing explicit type conversions.

src/CMakeLists.txt
```diff
@@ -31,7 +31,7 @@ set(SRC
     stdlib_experimental_ascii.f90
     stdlib_experimental_error.f90
     stdlib_experimental_kinds.f90
-    stdlib_experimental_system.F90
+    stdlib_experimental_system.F90 # Captital extension for a reason?
```
jvdp1, line 34:
```suggestion
    stdlib_experimental_system.F90
```

The extension `.F90` triggers the C preprocessor before compilation

MarDiehl:
> For the false positives, would it be possible/adequate to output a warning for the users?
Which false positives do you mean?

From my side, it was just a general comment that sometimes compilers give warnings despite the code is valid code. For example there are warnings about uninitialized variables even though the the choice of variables of the calling function makes it impossible to use a variable without initialization.
There are at least three ways to solve this:
1. Write some boilerplate code that silences the compiler
1. Turn off the warning selectively (not always possible)
1. Have two build configurations. A picky one for development (developers need to get used to a certain amount of warnings) and one for production that does not scares normal users.

jvdp1:
> > For the false positives, would it be possible/adequate to output a warning for the users?
> > Which false positives do you mean?
> 
> From my side, it was just a general comment that sometimes compilers give warnings despite the code is valid code. For example there are warnings about uninitialized variables even though the the choice of variables of the calling function makes it impossible to use a variable without initialization.


```
# prevent false positive (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95446)
if(CMAKE_Fortran_COMPILER_ID STREQUAL GNU)
  set_source_files_properties("test_optval.f90" PROPERTIES COMPILE_FLAGS "-Wno-error=pedantic")
endif()
```

I meant such false positives, that are due to, e.g. a bug in the compiler. If someone introduces new procedures in this test, he/she might want to know that these warnings are removed. However, if there is no warnings, he will need to check the CMakefile to know it.

> There are at least three ways to solve this:
> 
> 1. Write some boilerplate code that silences the compiler
> 2. Turn off the warning selectively (not always possible)
> 3. Have two build configurations. A picky one for development (developers need to get used to a certain amount of warnings) and one for production that does not scares normal users.

IMO we should go for the 3rd option. The 1st option should never be taken, and the 2nd option should be avoided as much as possible.

certik:
3 positive reviews, merging.

PR #206, opened 2020-05-30
proposition for solving an issue with optval

jvdp1:
Proposition to solve the issue raised in [the comment of PR #205](https://github.com/fortran-lang/stdlib/pull/205#issue-425496020) by @MarDiehl

The `elemental` is removed and `fypp` is used to generate the different ranks.

@milancurcic 
We should probably match that the shape of `x` is the same as of `default`. The `pure` might need to be removed too.

__Note__: this is only a proposition for a possible issue. However, the issue must be clarified first.
__Note__: the issue seems to be a bug in Gfortran (see [bug 95446](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95446). Therefore, this issue should be closed without being merged.

### DIFF
--- a/src/common.fypp
+++ b/src/common.fypp
@@ -71,6 +71,27 @@ ${prefix + joinstr.join([line.strip() for line in txt.split("\n")]) + suffix}$
 #:enddef
 
 
+#! Brace enclosed, comma separated Fortran expressions for a shape.
+#!
+#! Args:
+#!   varname (str): Name of the variable to be used as origin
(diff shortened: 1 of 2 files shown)


milancurcic:
Indeed, based on the development and Martin's bug report, it looks like this won't be needed after all, which I'm very happy about.

jvdp1:
Because it is due to a bug in [gfortran](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95446), I close this PR.

PR #207, opened 2020-05-30
explicitly state that unit tests are required

MarDiehl:
@milancurcic 

as discussed in #5


milancurcic:
There haven't been objections to this so I will merge it.

PR #210, opened 2020-06-04
fix build failure with nagfor7.0 because of using "complex" function

sakamoti:
"complex" function which is not a fortran standard causes an error like below in nagfor.

(for example)
"Error: test_linalg.f90, line 445: Implicit type for COMPLEX in TEST_TRACE_CQP"

So, "complex" is replaced with "cmplx" which is one of the fortran standard function.
At the same time, I specified the "kind" value so that it has the same precision as the "cmplx" function argument.

### DIFF
--- a/src/tests/io/test_savetxt.f90
+++ b/src/tests/io/test_savetxt.f90
@@ -86,13 +86,13 @@ contains
     character(*), intent(in) :: outpath
     complex(sp) :: d(3, 2), e(2, 3)
     complex(sp), allocatable :: d2(:, :)
-    d = cmplx(1, 1)* reshape([1, 2, 3, 4, 5, 6], [3, 2])
+    d = cmplx(1, 1,kind=sp)* reshape([1, 2, 3, 4, 5, 6], [3, 2])
     call savetxt(outpath, d)
     call loadtxt(outpath, d2)
     call check(all(shape(d2) == [3, 2]))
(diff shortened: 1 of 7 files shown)


milancurcic:
Thank you @sakamoti! This is in part covered by #205, and to your PR also adds the explicit `kind` parameter to `cmplx`. 

I suggest we wait to merge #205, then rebase this with master and merge. @certik does this sound reasonable?

sakamoti:
The latest version "48b28e8" resolved compilation error. But some "Questionable" message appeare from nagfor.

(example)
"Questionable: /home/saka/source/stdlib/fortran-lang_stdlib/src/tests/io/test_savetxt.f90, line 106: Intrinsic function CMPLX with double precision argument and no KIND= argument returns single precision result"

So, adds the explicit kind parameter to cmplx is still meaningfull I think. How do you think about this?

If it is preffer to add kind parameter, I'll try it according to compiler message and send new PR.

certik:
Yes, I think if the standard says that the result can be single precision (I think that's correct), we have to add the `kind` parameters. If you wouldn't mind doing it, we would really appreciate it. Thank you!

sakamoti:
Please wait a moment, I just editing. (It is my first time to rebase...)
when finish, I will add the comment.

sakamoti:
Rebase is over. Please review.

PR #211, opened 2020-06-04
Remove some unused variables and simplify use of ieee_value

jvdp1:
I removed some unused variables and I modified the call to `ieee_value` in `stdlib_experimental_stats_mean.fypp` to be in agreement with the other submodules.


src/stdlib_experimental_stats_mean.fypp
```diff
@@ -19,7 +19,7 @@ contains
         ${t1}$ :: res
 
         if (.not.optval(mask, .true.)) then
-          res = ieee_value(real(res, kind=${k1}$), ieee_quiet_nan)
+          res = ieee_value(1._${k1}$, ieee_quiet_nan)
```
jvdp1, line 22: Note: this line (as written before) generated a Run-Time Check Failure with `ifort` because `res` is used without being defined.

jvdp1:
Thanks @certik .
Because it only removes some unused variables and fixes an issue for ifort, I'll merge it.

PR #222, opened 2020-07-22
Install GFortran macOS in doc-deployment CI

jvdp1:
Issue: gfortran was not found for the doc-deployment script.
Installing it solved the problem.


certik:
Yes. The actually problem is that only `gfortran-9` is installed, but not `gfortran`. For `fpm` I fixed this in here this way:

https://github.com/fortran-lang/fpm/pull/132

For stdlib, if this PR works (it seems it does), then +1.

jvdp1:
Good. I will merge it.

Issue #216, opened 2020-07-05
Suggested change in directory structure


wclodius2:
I find the current directory/source file structure awkward. Rather than a src directory with lots of files with the term experimental in their names and module names, I suggest two directories: src with the validated source files, and experimental_src, with the source code under development. Only a few people should have commit privileges in the src directory, everyone in the experimental_src. None of the files/modules should have the term experimental in their names. That way when they have matured the developers can notify the src maintainers, and the maintainers can simply copy the files unmodified from experimental_src to src to commit. FWIW I suggest that the more mature src files be committed to src once they have been renamed and had their use statements modified to deal with the renaming.

milancurcic:
Hi WIlliam, I don't find the structure awkward, but I find the "experimental" in file names and modules unnecessary. We discussed this some time ago but now I can't find the thread. I think there was majority agreement to go with the "experimental" naming scheme, and it wasn't a big deal for me. However, I don't know why it's needed.

Actually, I personally find your proposal awkward as well. My preferred solution is to:

* Drop the "experimental" from file and module names
* Keep the source file where they are
* Start tracking the semantic version, e.g. 0.x.x is always "experimental", 1.x.x becomes stable.
* Begin making tagged and versioned releases

The agreed upon [Workflow](https://github.com/fortran-lang/stdlib/blob/master/WORKFLOW.md) remains the same.

certik:
Once we reach version 1.0.0, and if we drop the experimental namespace, then when we merge a PR, it immediately becomes part of stable, so we have to support it forever.

We need an experimental namespace, where we land all new features and then we graduate them into main.

How we manage the experimental namespace is for discussion, we can use a different scheme.

jvdp1:
I think the experimental namespace is important. I also like having the word "experimental" in the name of the module. It reminds the user that changes might happen in the API and implementations of the procedures.
However, having the word "experimental" in the name of the module might be an issue when we will reach the version 1.0.0 (for example, how will we add a new procedure (in the experimental namespace) to a module released in version 1.0.0?).

certik:
You do it by having both stable io and experimental_io modules.

jvdp1:
> You do it by having both stable io and experimental_io modules.

I agree with that. However, if there are experimental and released procedures of a same topic, we could end up with something like:
```fortran
use stdlib_stats, only: mean
use stdlib_experimental_stats, only: var
```
This situation could be awkward.
[...]

PR #223, opened 2020-07-22
Changes directory structure of stdlib as discussed in #216

jvdp1:
This PR changes the directory structure of stdlib as discussed in #216 opened by @wclodius2

In short:
* all "experimental" terms have been removed from the names of the modules
* a version ("experimental") tag has been added to all procedures and also in the specs. FORD uses the version tag as meta-data.

### DIFF
--- a/.github/workflows/doc-deployment.yml
+++ b/.github/workflows/doc-deployment.yml
@@ -24,6 +24,8 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0 # Full history to get tag and commit info
+      - name: Install GFortran macOS
+        run: brew install gcc || brew upgrade gcc || true
       - name: Install Dependencies
         run: |
           pip3 install --prefer-binary --no-clean --disable-pip-version-check --progress-bar off lxml fypp
--- a/.github/workflows/doc-deployment.yml
+++ b/.github/workflows/doc-deployment.yml
@@ -24,8 +24,6 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0 # Full history to get tag and commit info
-      - name: Install GFortran macOS
-        run: brew install gcc || brew upgrade gcc || true
       - name: Install Dependencies
         run: |
           pip3 install --prefer-binary --no-clean --disable-pip-version-check --progress-bar off lxml fypp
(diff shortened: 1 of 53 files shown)


jvdp1:
The FORD documentation can be dowloaded [here[(https://github.com/fortran-lang/stdlib/suites/951359601/artifacts/11866410).

jvdp1:
This is a major change in terms of structure for `stdlib`. 
Is everybody ok with that? It should be easier for the users to use `stdlib` now. However, is it still clear enough that the procedures are `experimentatl`?

certik:
I think we are ok with the general idea to do it like this.

I think the only thing to figure out is how exactly to document that all those routines are experimental. Should we also add a comment next to each subroutine in the source file? I think I would prefer that --- we want to convey as clearly as possible that the subroutine is experimental.

leonfoks:
@certik I believe having the version: experimental as @jvdp1 has done should suffice.  Having the experimental tag in the Ford  doc comments covers all bases.  Is there anywhere else that would more clearly indicate the experimental nature of a function?

Could we have a script that simply parses out "version: experimental" and lists them in the docs?  Maybe this is something we could add to FORD?

certik:
Given how important the change is, I am CCing others who committed the most code so far: @zbeekman, @nshaffer, @scivision, @fiolj, @MarDiehl, @aradi, @ivan-pi, to make sure everybody is on board.

fiolj:
I've been away for a while. I am ok with the change in structure.

jvdp1:
> Could we have a script that simply parses out "version: experimental" and lists them in the docs? Maybe this is something we could add to FORD?

The version of each procedure is mentioned in FORD website when you open the page of a procedure.
I think it would be also nice if the version would be added in the table that lists all the procedures.

nshaffer:
So in the generated docs, experimental interfaces get a subsection like

## Version
Experimental

What would this say for stable interfaces, "stable"? Seems strange to me. Might just be the word "version". If we are only classifying interfaces as "experimental" or "stable", I think maybe calling this section "status" makes more sense? If I'm a user and I see "Version: stable", I wonder "OK, so is there and unstable version as well?" (a la how some Linux distributions classify packages as stable/unstable/testing).

Other than that detail, I support this change.

certik:
I agree, using `Status` instead of `Version` is better I think.

jvdp1:
I agree with the comments on `status` vs `version`. I changed the title `Version` in the specs to `Status` in [this commit](https://github.com/fortran-lang/stdlib/pull/223/commits/4754c544527b5e9b35a7ade0638c86061a282552).

scivision:
Yes this is a good and necessary change

jvdp1:
It seems there is an agreement on the proposed change of the structure of `stdlib`.
Thank you all for your feedback. I'll merge.

Issue #69, opened 2020-01-02
String handling routines


ivan-pi:
Let's start a discussion on routines for string handling and manipulation. The thread over at [j3-fortran](https://github.com/j3-fortran/fortran_proposals/issues/96) already collected some ideas:

* **split** - given a separator, splits the string into some form of array
* **upper**/**lower** - convert a character string to all upper/lower case

The discussion also mentioned the proposed `iso_varying_string` module, which was supposed to include some string routines. I found three distinct implementations of this module: 

* [ISO_VARYING_STRING Module](http://www.astro.wisc.edu/~townsend/static.php?ref=iso-varying-string) by Rich Townsend
* [iso_varying_string](https://gitlab.com/everythingfunctional/iso_varying_string) implementation by Brad Richardson (@everythingfunctional)
* [ISO_VARYING_STRING](https://cyber.dabamos.de/programming/modernfortran/iso_vsta.f95) due to J.L.Schonfelder (author of the `iso_varying_string` proposal; the module dates back to 1998)

I also found the following Fortran libraries targetting string handling:

* [Strings For Fortran](https://gitlab.com/everythingfunctional/strff) by Brad Richardson @everythingfunctional
* [Fortran Character String Utilities](https://gbenthien.net/strings/str-index.html) by George Benthien
* [M_STRINGS](https://github.com/urbanjost/M_strings) from Urban Jost @urbanjost (part of [General-Purpose Fortran](http://www.urbanjost.altervista.org/LIBRARY/libGPF/GPF.html) tools)
* [String_Functions](http://fortranwiki.org/fortran/show/String_Functions) by David Frank
* [StringiFor](https://github.com/szaghi/StringiFor) by Stefano Szaghi @szaghi
* [fortranString](https://github.com/bceverly/fortranString) from @bceverly
* [fortran-string-utility-module](https://github.com/tomedunn/fortran-string-utility-module) by @tomedunn
* [fortran-string](https://github.com/dongli/fortran-string) from @dongli
* [strings](https://github.com/jchristopherson/strings) by @jchristopherson
* [fortran_libstring](https://github.com/koiking213/fortran_libstring) from @koiking213
* [fortran-strings](https://github.com/eengl/fortran-strings) by @eengl
* [flibs](http://flibs.sourceforge.net/) by @arjenmarkus contains several modules for handling strings

It is likely that several of the tools in the [list of popular Fortran projects](
[...]

PR #235, opened 2020-09-24
Update ascii - second round: clean up the code and implement robust versions of to_lower and to_upper

arjenmarkus:
The comments to the previous pull request are now all solved:
- Extra parentheses have been removed (for readability those around the logical disjunctions and conjunctions have been kept)

- Some clean-up of the comments (more explicitly stating the ranges)

- The functions to_lower and to_upper have been made robust (perhaps slower than before, but the performance is certainly not bad)

Suggestion for further improvement: allow to_lower and to_upper to convert strings of arbitrary length - this seems to me to be the most common usage.

### DIFF
--- a/src/stdlib_ascii.f90
+++ b/src/stdlib_ascii.f90
@@ -145,7 +145,7 @@ contains
         character(len=1), intent(in) :: c !! The character to test.
         integer :: ic
         ic = iachar(c)                    ! '~'
-        is_printable = c >= ' ' .and. ic <= int(z'7E')
+        is_printable = ic >= iachar(' ') .and. ic <= int(z'7E')
     end function
 
     !> Checks whether `c` is a lowercase ASCII letter (a .. z).
--- a/src/stdlib_ascii.f90
+++ b/src/stdlib_ascii.f90
@@ -60,6 +60,9 @@ module stdlib_ascii
     character(len=*), public, parameter :: lowercase = letters(27:) !! a .. z
     character(len=*), public, parameter :: whitespace = " "//TAB//VT//CR//LF//FF !! ASCII _whitespace
 
+    character(len=26), parameter, private :: lower_case = 'abcdefghijklmnopqrstuvwxyz'
+    character(len=26), parameter, private :: upper_case = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'
+
 contains
 
     !> Checks whether `c` is an ASCII letter (A .. Z, a .. z).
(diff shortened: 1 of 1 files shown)


src/stdlib_ascii.f90
```diff
@@ -181,31 +186,39 @@ pure logical function is_blank(c)
...
+        character(len=1)             :: t
+        integer :: k
+
+        k = index( upper_case, c )
+
+        if ( k > 0 ) then
+            t = lower_case(k:k)
+        else
+            t = c
+        endif
```
14NGiestas, line 205: I agree all the other changes... but I'm not quite sure about this one. The `to_lower` and `to_upper` functions got around 4x slower (0.05 s vs 0.20 s) and this change is a little bit overkill since the old one [already works with both ASCII and EBICID char tables](https://en.wikipedia.org/wiki/Letter_case#Methods_in_programming).
However a ASCII to EBICID function could be easily implemented, if desired, using a [static lookup table](https://people.cs.umu.se/isak/Snippets/a2e.c)

ivan-pi:
> Suggestion for further improvement: allow to_lower and to_upper to convert strings of arbitrary length - this seems to me to be the most common usage.

This has been requested as part of the [strings](#69) module. I have wondered already if it would be easier to just have a single function with the interface:
```Fortran
pure function upper(str) result(ustr)
character(len=len(str)) :: ustr
end function
```
and get rid of the single character case conversions entirely. As you say, the most common usage will be arbitrary length strings, so I am not sure if we need a specialization for single letters. The arbitrary length version can be used on single characters too (hopefully the compilers are capable enough to remove the loop for characters with `len=1`).

The reason I included the case conversions in the first plase was just to cover the same functionality as the [<ctype.h>](http://www.cplusplus.com/reference/cctype/) header file from C.

milancurcic:
@arjenmarkus @ivan-pi @14NGiestas Is this PR good to go or should [this comment](https://github.com/fortran-lang/stdlib/pull/235#discussion_r494660488) still be addressed?

src/stdlib_ascii.f90
arjenmarkus: I was not aware of that property of EBCDIC, but in that case, of course, the old implementation requires no replacement :).

arjenmarkus:
Argh, I had intended to rely to these remarks, but never got around to it. I agree: in C the situation is completely different - there are characters and there are arrays of characters. In Fortran we have character strings but not a separate character type. The suggested interface (and almost trivial implementation) looks good to me.

milancurcic:
Okay, I will merge this then. The suggested interface can go in as a new PR.

Issue #219, opened 2020-07-16
Standard library error reporting


wclodius2:
Currently my default approach to dealing with errors is to have an
optional status flag for the procedure in which the error is found,
one or more routines for writing an error message, and a branch for
either setting the status flag or reporting the error while stopping
processing, e.g., the psuedo code

    subroutine example(..., status )
        ...
        integer, intent(out), optional :: status
        ...
   	    if (present(status)) status = success ! Set default value
	    ...
	    if (bad stuff is found) go to label
	    ...
	! Error handling is placed at the end of the routine
	label if (present status) then
	        status = bad_stuff_flag
		    return
	    else
	        call report_error_and_stop(...)
	    end if
        ...
    end subroutine example

I want to write some code for the Standard Library that will have to
handle errors. This prompts the following questions:

1. Is this a good approach for handling errors in the Standard Library
   given the limitations of Fortran?

2. If this is not a good approach what is a better approach?

3. If this is a good approach should the Standard Library have a
   module supporting this approach?

4. If it should have a module what should be the API of that module?

In thinking about the API I have been looking at the error handling of
Futility and its inspiration FLIBS. My examination so far has been
superficial. My current approach is to send all information to the
`STD_ERR` of `ISO_FORTRAN_ENV`. Futility and FLIBS send it to a log
file. Futility has an object based system that allows it to send
messages to multiple log files. This prompts the following questions:

1. Is supporting a log file separate  from STD_ERR useful?

2. If it is useful should a log file be required or should the default
   be STD_ERR?

3. Would more than one log file be useful?

4. Should there be an option to report to both the log file and
   `STD_ERR`?

Futility and FLIBS also focus on the severity of the error, providing
five levels of error reporting: INFORMATION, WARNING, ERROR, FATAL
ERROR, and FAILURE. I have tended to have just WARNING and the
equivalent of FATAL ERROR. Are multiple levels of error that useful?

I have focussed on reporting the location where the error was
discovered, providing optional MODULE and PROCEDURE arguments for
reporting the MODULE and PROCEDURE in which the call was made> Would
others find that useful? If the error is discovered in an internal
procedure of a module procedure would an optional INTERNAL_PROCEDURE
argument be useful?

I have also focussed on enumerating various sources of errors, e.g.,
`ALLOC_FAULT`, `CONSISTENCY_FAULT`, `CONVERGENCE_FAILURE`,
`MISSING_FILE`, etc., to provide a consistent nomenclature for the
[...]

PR #239, opened 2020-09-30
Bitsets3

wclodius2:
Branch to implement bitset data types. Committed the following new source files:
src/stdlib_bitsets.f90
src/stdlib_bitset_64.f90
src/stdlib_bitset_large.f90

Added the following new test files:
src/tests/bitsets/CMakeLists.txt
src/tests/bitsets/Makefile.manual
src/tests/bitsets/test_stdlib_bitset_64.f90
src/tests/bitsets/test_stdlib_bitset_large.f90

Added the following new documentation file:
doc/specs/stdlib_bitssets.md

Modified the following compilation files:
src/CMakeLists.txt
src/Makefile.manual
src/tests/CMakeLists.txt
src/tests/Makefile.manual


doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,1975 @@
...
+        end if
+        set1 = set0
+        if ( set1 == set0 ) &
+            write(*,*) 'Initialization by assignment succeeded'
+        logical2 = set1
+        if ( all( logical2 ) ) then
+            write(*,*) 'Initialization of logical(int32) succeeded.'
+        end if
+    end program demo_assignment
+
```
jvdp1, line 164: the ``` are missing
```suggestion
```
```

File: doc/specs/stdlib_bitsets.md [hunk 2]
```diff
@@ -0,0 +1,1975 @@
...
+        use stdlib_bitsets
+        type(bitset_large) :: set0, set1
+        call set0 % init(166)
+        call set0 % set(100,150)
+        call extract( set1, set0, 100, 150)
+        if ( set1 % bits() == 51 ) &
+            write(*,*) 'SET1 has the proper size.'
+        if ( set1 % all() ) write(*,*) 'SET1 has the proper values.'
+    end program demo_extract
+
```
jvdp1, line 661: The ``` are missing

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,1975 @@
...
+        use stdlib_bitsets
+        type(bitset_large) :: set0
+        call set0 % init(166)
+        if ( set0 % none() ) write(*,*) 'SET0 is properly initialized.'
+        call set0 % flip(165)
+        if ( set0 % test(165) ) write(*,*) 'Bit 165 is flipped.'
+        call set0 % flip(0,164)
+        if ( set0 % all() ) write(*,*) 'All bits are flipped.'
+    end program demo_flip
+
```
jvdp1, line 718: the ``` are missing

jvdp1 (commented):
Thank you @wclodius2 for this PR. I will try to review it this weekend or next week.

src/tests/Makefile.manual
```diff
@@ -10,6 +11,7 @@ all:
 
 test:
 	$(MAKE) -f Makefile.manual --directory=ascii test
+	$(MAKE) -f Makefile.manual --directoru=bitsets test
```
jvdp1, line 14:
```suggestion
	$(MAKE) -f Makefile.manual --directory=bitsets test
```
This should allow the Github actions to be successful

14NGiestas (commented):
There's a misspelling of the word 'occurred' in this file: "occured" (line 1092)
EDIT: [this file](https://github.com/wclodius2/stdlib/blob/bitsets3/doc/specs/stdlib_bitsets.md)

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,1990 @@
...
+`status` (optional): shall be a scalar default integer variable. It is
+an `intent(out)` argument. If present on return it will have the value
+of `success` or `write_failure`. If absent and it would not have the
+value of `success` then processing will stop with an informative stop
+code. The two code values have the meaning:
+
+* `success` - no problem found
+
+* `write_failure` - a failure occured in a write statement.
```
jvdp1, line 1101:
```suggestion
* `write_failure` - a failure occurred in a write statement.
```

wclodius2:
Looks good.

14NGiestas:
We should really discuss the error handling further in #219 . The 'go to' usage in [this function](https://github.com/wclodius2/stdlib/blob/bitsets3/src/stdlib_bitset_large.f90#L960)... I'm not a fan xD. I think only one labeled part is needed to achieve the 'return with status' vs 'stop program with error message' desired behavior, with the aid of a error handling module, we could set the message and a flag to switch between the two possible actions, reusing the code. I will bring some working code snippet soon (I hope).
Also, in the assigment functions I wonder if it is possible to use the template system (as in the statistics module)?

wclodius2:
What is xD? The error handling has several aspects that could be commented on:

1. Rather than intermingling the error handling with the main logic I jump to error handling at the end of the procedure. This increases the amount of code slightly, and can obscure what the error handling is doing, but I think makes the main logic clearer.

2. I am handling some "errors" that possibly shouldn't be handled: memory allocation problems and user errors. With virtual memory, its not clear to me that allocation can ever fail, and with lazy allocation on Linux such failure will not occur on the allocate statement.. While not important in the procedure you call out, I also attempt to provide a fallback for some user errors where simple failure might be best.

3. Since we lack an error handling procedure I am relying explicitly on an if branch with `error stop`. An error handling procedure could incorporate the if branch and slightly simplify the logic by incorporating both a message and a status argument. There are however a lot of issues to be dealt with in developing an error handling module.

    * Should it be a separate module or part of the logger?

    * Should it include a set of error codes?

    * What should the arguments be to the main error handling procedure?
        - message
        - module and procedure names
        - `status`/`stat` flag
        - error code

    * How should processing be terminated, an `error stop`: with an F2018 runtime determined stop code, with an error code determined compile time stop code, generic compile time stop code?

4. Including the module and procedure names in the `error stop` makes the stop code longer and harder to read

5. Making the stop code message a character string parameter would shorten the error handling code, at the expense of cluttering up other portions of the code.

It is of course possible to use fypp preprocessing to generate templates for the assignments to and from logical arrays. The assignment procedures are few enough and simple enough that I thought it best to avoid a dependence on the preprocessor, but that is a matter of taste. The addition of more assignments, say to and from integers would probably change my opinion.

14NGiestas:
> What is xD?

ASCII version of 😆smiling face with closed eyes

1-5. Yes, the current way increases the amount of code, however the main logic can be still clear if we do something like that:

```fortran
    subroutine fail_behavior(..., status)
        ! doing a lot of stuff here
        ! ...
        ! case X: found some fatal error
            error_handler % status = ERROR_CODE_TO_X
            error_handler % message = "Oh no! X happened"
            go to 404
        ! case Y: found other general error
            error_handler % status = ERROR_CODE_TO_Y
            error_handler % message = "Oh no! Y happened"
            go to 404

    404 if (present(status)) then
            status = error_handler % status
            return
        else
            error stop error_handler % message
        end if
    end subroutine
``` 
This is a way we can avoid the code duplication, reusing the go to branching. Note that error_handler should be a object that couples the information about what happened (not reflecting in any way the API I got in mind for a final version).
The error handling should be a separate module with some default general usage errors and a abstract one, that would allow it to be extended when needed by other modules, this way they can provide their own specific codes constants and messages.

> It is of course possible to use fypp preprocessing to generate templates for the assignments to and from logical arrays. The assignment procedures are few enough and simple enough that I thought it best to avoid a dependence on the preprocessor, but that is a matter of taste. The addition of more assignments, say to and from integers would probably change my opinion.

IMHO, we should use the template system every time that not using it leads to code duplication, since it gets easier to modify and extend the code in the future. (DRY principle).

wclodius2:
FWIW for the error handling what I thought you were proposing was something like

    subroutine fail_behavior(..., status)
        ! doing a lot of stuff here
        ! ...
        ! case X: found some fatal error
            call error_handler(message="Oh no! X happened", error=error_code_to_x, status=status)
            return
        ! case Y: found other general error
            call error_handler(message="Oh no! Y happened", error=error_code_to_y, status=status)
            return
    end subroutine

where `error_handler` assigns `error` to `status` if `status` is `present` and stops processing with a message to `error_unit` otherwise.

I can change the code to use `fypp`, but it probably requires starting a new branch. I can see how to add or modify files in the current branch, but not how to rename or remove existing files, which I need to change the `*.f90` files to `*.fypp`.

14NGiestas:
> FWIW for the error handling what I thought you were proposing was something like (...)
> where `error_handler` assigns `error` to `status` if `status` is `present` and stops processing with a message to `error_unit` otherwise.

Indeed, actually this is even better because we don't need retype the branching logic with "go to" statements in every function that may fail and is doable [(working example here)](https://gist.github.com/14NGiestas/8ebf6753e6d4680ecd044faa2c2bed85). To avoid having any boilerplate along the function's main logic, the module that have a specific need should build a function that calls this base function `error_handler`, Ex:
```fortran
subroutine value_error(expected, got, status)
     ....
     write(message, '("value_error: Expected ",G0,", but got ",G0," instead.")') expected, got
     call error_handler(message, ...) 
end subroutine
```

> I can change the code to use `fypp`, but it probably requires starting a new branch. I can see how to add or modify files in the current branch, but not how to rename or remove existing files, which I need to change the `*.f90` files to `*.fypp`.

You can clone your repo [here](https://github.com/wclodius2/stdlib/tree/bitsets3), modify it locally and push the commits (It will probably sync with the open PR)

jvdp1:
> I can see how to add or modify files in the current branch, but not how to rename or remove existing files, which I need to change the `*.f90` files to `*.fypp`.

You can rename and remove a file locally with your OS, and then add/remove, commit, and push with git.

wclodius2:
When I try

    git clone https://github.com/wclodius2/stdlib/tree/bitsets3 my-bitsets

I get the error message

    Cloning into 'my-bitsets'...
    fatal: repository 'https://github.com/wclodius2/stdlib/tree/bitsets3/' not found

Is there something wrong with my syntax/repository name?

14NGiestas:
This is the wrong URL, please do the following:
```
git clone https://github.com/wclodius2/stdlib.git my-bitsets
cd my-bitsets
git checkout bitsets3
```

14NGiestas (approved):

src/stdlib_bitset_large.fypp
```diff
@@ -1396,17 +1342,11 @@ contains
...
                    string
         end if
-        if (ierr /= 0) go to 999
-
-        return
-
-999     if ( present(status) ) then
-            status = write_failure
+        if (ierr /= 0) then
+            call error_handler( 'Failure on a WRiTE statement for UNIT.', &
```
14NGiestas, line 1346:
```suggestion
            call error_handler( 'Failure on a WRITE statement for UNIT.', &
```

src/stdlib_bitset_large.fypp
```diff
@@ -807,31 +785,29 @@ contains
...
         return
 
-999     if ( present(status) ) then
-            status = write_failure
-            return
-        else
-            error stop module_name // ' % ' // procedure // ' had a ' // &
-                'failure in the write to UNIT.'
-        end if
+999     call error_handler( 'Failure on a WRiTE statement for UNIT.', &
```
14NGiestas, line 788:
```suggestion
999     call error_handler( 'Failure on a WRITE statement for UNIT.', &
```

14NGiestas (approved):

wclodius2:
I have modified, committed and pushed the code making the following changes:
1. Changed  the main file names from stdlib_bitset*.f90 to stdlib_bitset*.fypp
2. Used the preprocessor to autogenerate the assignment defs for logical arrays
3. Added a subroutine error_handler to stdlib_bitsets.fypp
4. Used the subroutine to handle errors in stdlib_bitset_64.fypp and stdlib_bitset_large.fypp
5. Used bulleted lists to document the use of error codes in stdlib_bitset*.fypp and stdlib_bitsets.md
6. Added the error code char_string_too_large_error
7. After pushing realized I forgot to check for trailing blanks. Removed trailing blanks and pushed again.

wclodius2:
Failed to run tests on the previous update.  I had accidentally removed a return statement causing failure of test_stdlib_bitset_64. Fixed and pushed.

14NGiestas (approved):

14NGiestas:
Thank you, @wclodius2!

src/stdlib_bitset_64.fypp
```diff
@@ -0,0 +1,1145 @@
...
+        bits = len(string, kind=int64)
+        if ( bits > 64 ) then
+            call error_handler( 'STRING was too long for a ' //      &
+                                'BITSET_64 SELF.',                   &
+                                char_string_too_large_error, status, &
+                                module_name, procedure )
+            return
+        end if
+        self % num_bits = bits
+        do bit = 1, bits
```
14NGiestas, line 314: I get this warning compiling it:
```
/home/ian/stdlib/src/stdlib_bitset_64.f90:437:19:

  437 |         do bit = 1, bits
      |                   1
Warning: Possible change of value in conversion from INTEGER(8) to INTEGER(4) at (1) [-Wconversion]
```
Maybe changing to this:
```suggestion
        do bit = 1_bits_kind, bits
```
Or modifying the kind of the variable `bit` to other than `bits_kind` since it's a simple iterator

src/stdlib_bitset_large.fypp
```diff
@@ -0,0 +1,1372 @@
...
+            return
+        end if
+
+        call init_zero_large( self, int(bits, kind=bits_kind), status )
+
+        if ( present(status) ) then
+            if ( status /= success ) return
+        end if
+
+        do bit = 1_bits_kind, bits
```
14NGiestas, line 430: Here there's a similar warning, this time the other way around:
```
/home/ian/stdlib/src/stdlib_bitset_large.f90:562:29:

  562 |         do bit = 1_bits_kind, bits
      |                             1
Warning: Possible change of value in conversion from INTEGER(8) to INTEGER(4) at (1) [-Wconversion]
```
I believe `bits` and `bit` should have the same kind, other than `bits_kind`, and since they are iterators there's no problem using `int64` as kind here.

src/stdlib_bitset_64.fypp
wclodius2: If you check on the use of bits, it is declared int64 to always ensure that an int32 for bit is adequate. The warning is not for the 1 in `do bit = 1, bits`, but rather for the conversion of `bits` to `bit`. The following tests ensure that int32 is adequate.

        bits = len(string, kind=int64)
        if ( bits > 64 ) then
 
in stdlib_bitset_64.fypp or

        bits = len(string, kind=int64)
        if ( bits > huge(0_bits_kind) ) then

in stdlib_bitset_large.fypp. So the warning is spurious. I can, if you want, shut off the warning by making bit an int64.

14NGiestas: That's correct, it's `bit` vs `bits` conversion that raises this warning. Is there any reason to `bit` be of kind `bits_kind`, in this function? If not, I think it should be a `int64` as well.

14NGiestas (approved):

14NGiestas (approved):

wclodius2:
I have made some aesthetic changes in the source code: deleting a couple of unnecessary returns, and regularizing the code fences in the FORD documentation of stdlib_bitsets.fypp. I now plan to take a break of a few days from edits of the source code, unless they are specifically requested.

jvdp1:
> I have made some aesthetic changes in the source code: deleting a couple of unnecessary returns, and regularizing the code fences in the FORD documentation of stdlib_bitsets.fypp. I now plan to take a break of a few days from edits of the source code, unless they are specifically requested.

Thank you @wclodius2 . I will try to review it next week,

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+#### Arguments
+
+`set1`: shall be a scalar `bitset_64` or `bitset_large` variable. It
+is an `intent(inout)` argument. On return the values of the bits in
+`set1` are the bitwise `and` of the original bits in `set1` with the
+corresponding negation of the bits in `set2`.
+
+`set2`: shall be a scalar expression of the same type as `set1`. It is
+an `intent(in)` argument. Note that it should also have the same
+number of bits as `set1` otherwise the result is undefined.
```
jvdp1, line 371:
```suggestion
number of bits as `set1`, otherwise the result is undefined.
```

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+`intent(in)` argument.
+
+#### Result character
+
+The result is a default logical scalar.
+
+#### Result value
+
+The result is `.true.` if any bits in `self` are set, otherwise it
+is `.false.`.
```
jvdp1, line 426: These 2 sections could be merged. Currently it is confusing.
```suggestion
#### Result value

The result is a default logical scalar. The result is `.true.` if any bits in `self` are set, otherwise it
is `.false.`.
```

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+`intent(in)` argument.
+
+#### Result character
+
+The result is an integer scalar of kind `bits_kind`.
+
+#### Result value
+
+The result is the number of bits that are set in `self`.
+
```
jvdp1, line 478:
```suggestion
#### Result value

The result is an integer scalar of kind `bits_kind`, equal to the number of bits that are set in `self`.

```

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+* If only `pos` is present, clears the bit with position `pos` in
+`self`.
+
+* If `start_pos` and `end_pos` are present with `end_pos >= start_pos`
+clears the bits with positions from `start_pos` to `end_pos` in `self`.
+
+* if `start_pos` and `end_pos` are present with `end_pos < start_pos`
+`self` is unmodified.
+
+Note: Positions outside the range 0 to `BITS(SET) -1` are ignored.
```
jvdp1, line 563: What happens with the range  -9 to 'bits(set)+10`?
```suggestion
Note: Positions outside the range 0 to `bits(set) -1` are ignored.
```

jvdp1: Reminder for myself: Are there some tests with a range outside 0-`bits(set)-`?

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+value `index_invalid_error`, otherwise processing stops with an
+informative message.
+
+#### Syntax
+
+`call [[stdlib_bitsets(module):extract(interface)]](new, old, start_pos, stop_pos, status )`
+
+#### Class
+
+Subroutine
```
jvdp1, line 628: Why is this a subroutine instead a function, e.g.:
`new = extract(old, start_pos,...)`?

Note: I am fine (and even prefer it) as a subroutine. It is mainly to document the PR.

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+
+#### Arguments
+
+`self`: shall be a scalar `bitset_64` or `bitset_large` variable. It
+is an `intent(out)` argument.
+
+`bits` (optional): shall be a scalar integer expression of kind
+`bits_kind`. It is an `intent(in)` argument that if present
+specifies the number of bits in `set`. A negative value or a value
+greater than 64 if `self` is of type `bitset_64` is an error.
```
jvdp1, line 817:
```suggestion
specifies the number of bits in `set`. A negative value, or a value
greater than 64 if `self` is of type `bitset_64`, is an error.
```

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+### `input` - reads a bitset from an unformatted file
+
+#### Status
+
+Experimental
+
+#### Description
+
+Reads a bitset from its binary representation in an unformatted
+file.
```
jvdp1, line 856: Is it not worthwhile to mention the type of access used for unformatted files?

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+
+`unit`: shall be a scalar default integer expression. It is an
+`intent(in)` argument. Its value must be that of a logical unit
+number for an open unformatted file with `READ` or `READWRITE`
+access positioned at the start of a BITSET value written by a
+`bitset_type` `output` subroutine by the same processor.
+
+`status` (optional): shall be a scalar default integer variable. If
+present its value shall be of one of the error codes defined in this
+module. IF absent and it would have had a value other than `success`
```
jvdp1, line 879:
```suggestion
module. If absent and it would have had a value other than `success`
```

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+            bits_1   = '000000000000000000000000000000001', &
+            bits_33  = '100000000000000000000000000000000'
+        integer :: unit
+        type(bitset_64) :: set0, set1, set2, set3, set4, set5
+        call set0 % from_string( bits_0 )
+        call set1 % from_string( bits_1 )
+        call set2 % from_string( bits_33 )
+        open( newunit=unit, file='test.bin', status='replace', &
+            form='unformatted', action='write' )
+        call set2 % output(unit)
```
jvdp1, line 909: Maybe I missed it, but would it be possible to use:
```fortran
 call set2 % output('test.bin')
```
and similarly for `input`?

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+  `intent(in)` argument.
+
+#### Result character
+
+The result is a default logical scalar.
+
+#### Result value
+
+The result is `.true.` if no bits in `self` are set, otherwise it is
+`.false.`.
```
jvdp1, line 959:
```suggestion
#### Result value

The result is a default logical scalar.
The result is `.true.` if no bits in `self` are set, otherwise it is
`.false.`.
```

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+
+Experimental
+
+#### Description
+
+Performs the logical complement on the bits of `self`.
+
+#### Syntax
+
+`result = self % [[bitset_type(class):not(bound)]] ()`
```
jvdp1, line 993:
```suggestion
`call self % [[bitset_type(class):not(bound)]] ()`
```

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+
+#### Status
+
+Experimental
+
+#### Description
+
+Replaces the original bits of `set1` with the bitwise `or` of those
+bits with the bits of `set2`. Note `set1` and `set2` must have the
+samee number of bits, otherwise the result is undefined.
```
jvdp1, line 1033:
```suggestion
same number of bits, otherwise the result is undefined.
```

jvdp1, line 1033: How is `undefined` defined? `NaN? (I didn't look to the code yet, but I think it would be good to define "undefined" in the specs.)

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+
+#### Arguments
+
+`set1`: shall be a scalar `bitset_64` or `bitset_large` variable. It
+ is an `intent(inout)` argument. On return the values of the bits in
+ `setf` are the bitwise `or` of the original bits in `set1` with the
+ corresponding bits in `set2`.
+
+`set2`: shall be a scalar expression of the same type as `set1`. It is
+an `intent(in)` argument. Note `bits(set2)` must equal `bits(set1)`
```
jvdp1, line 1051:
```suggestion
an `intent(in)` argument. Note that `bits(set2)` must equal `bits(set1)`
```

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+
+### `output` - Writes a binary representation of a bitset to a file
+
+#### Status
+
+Experimental
+
+#### Description
+
+Writes a binary representation of a bitset to an unformatted file.
```
jvdp1, line 1085: What kind of access is used? It might be important to mention it because the user must open the file himself.

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+
+`call [[stdlib_bitsets(module):init(interface)]] (self, bits [, status])`
+
+#### Class
+
+Subroutine.
+
+#### Arguments
+
+`self`: shall be a scalar `bitset_64` or `bitset_large` variable. It
```
jvdp1, line 811: to what are set the bits? All `0`? All `1`? Uninitialized?

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+`pos`: shall be a scalar integer expression of kind `bits_kind`. It is
+an `intent(in)` argument.
+
+#### Result character
+
+The result is a default logical scalar.
+
+#### Result value
+
+The result is `.true.` if the bit at `pos`, in `self` is set,
```
jvdp1, line 1358:
```suggestion
The result is `.true.` if the bit at `pos` in `self` is set,
```

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+otherwise it is `.false.`. If `pos` is outside the range
+`0... bits(self)-1` the result is `.false.`.
+
```
jvdp1, line 1361: " If `pos` is outside the range ... `.false.`.": does this make sense? As a user, I would prefer that the program stops, or at least, returns something corresponding to `NaN`

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+Experimental
+
+#### Description
+
+Represents the value of `self` as a binary literal in `string`.
+
+#### Syntax
+
+`call self % [[bitset_type(class):to_string(bound)]](string[, status])
```
jvdp1, line 1390:
```suggestion
`call self % [[bitset_type(class):to_string(bound)]](string[, status])`
```

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+                " into NEW_STRING."
+        end if
+    end program demo_to_string
+```
+
+### `value` - determine the value of a bit
+
+#### Status
+
+Experimeental
```
jvdp1, line 1439:
```suggestion
Experimental
```

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+#### Result character
+
+The result is a default logical scalar.
+
+#### Result value
+
+The result is one if the bit at `pos` in `self` is set, otherwise it
+is zero. If `pos` is outside the range `0... bits(set)-1` the result
+is zero.
+
```
jvdp1, line 1470:
```suggestion
#### Result value

The result is a default logical scalar.
The result is 1 if the bit at `pos` in `self` is set, otherwise it
is 0. If `pos` is outside the range `0... bits(set)-1` the result
is 0.

```

Again, does it make sense to return 0 if pos is outside the range?

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+Experimental
+
+#### Description
+
+Returns `.true.` if all bits in `set1` and `set2` have the same value,
+`.false.`  otherwise.
+
+#### Syntax
+
+`Result = set1 [[stdlib_bitsets(module):==(interface)]] set2
```
jvdp1, line 1655:
```suggestion
`result = set1 [[stdlib_bitsets(module):==(interface)]] set2
```

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+argument.
+
+#### Result character
+
+The result is a default logical scalar.
+
+#### Result value
+
+The result is `.true.` if the bits in both bitsets are set
+  to the same value, otherwise the result is `.FALSE.`.
```
jvdp1, line 1677:
```suggestion
#### Result value

The result is a default logical scalar.
The result is `.true.` if the bits in both bitsets are set
  to the same value, otherwise the result is `.false.`.
```

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+
+#### Result character
+
+The result is a default logical scalar.
+
+#### Result value
+
+The result is `.true.` if the bits in `set1` and `set2` are the same
+or the highest order different bit is set to 1 in `set1` and to 0 in
+`set2`, `.false.`.  otherwise.
```
jvdp1, line 1798:
```suggestion
`set2`, `.false.`  otherwise.
```

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+
+#### Status
+
+Experimental
+
+#### Description
+
+Returns `.true.` if the bits in `set1` and `set2` differ and the
+highest order different bit is set to 1 in `set1` and to 0 in `set2`,
+`.false.`.  otherwise. The sets must be the same size otherwise the
```
jvdp1, line 1832:
```suggestion
`.false.`  otherwise. The sets must be the same size otherwise the
```

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+results are undefined
+
+#### Syntax
+
+`Result = set1 [[stdlib_bitsets(module):>(interface)]] set2`
```
jvdp1, line 1837:
```suggestion
`result = set1 [[stdlib_bitsets(module):>(interface)]] set2`
```

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+
+#### Status
+
+Experimental
+
+#### Description
+
+Returns `.true.` if the bits in `set1` and `set2` are the same or the
+highest order different bit is set to 0 in `set1` and to 1 in `set2`,
+`.false.`.  otherwise. The sets must be the same size otherwise the
```
jvdp1, line 1893:
```suggestion
`.false.`  otherwise. The sets must be the same size otherwise the
```

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+results are undefined
+
+
+#### Syntax
+
+`Result = set1 [[stdlib_bitsets(module):<=(interface)]] set2`
```
jvdp1, line 1899:
```suggestion
`result = set1 [[stdlib_bitsets(module):<=(interface)]] set2`
```

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+
+#### Result character
+
+The result is a default logical scalar.
+
+#### Result value
+
+The result is `.true.` if the bits in `set1` and `set2` are the same
+or the highest order different bit is set to 0 in `set1` and to 1 in
+`set2`, `.false.`.  otherwise.
```
jvdp1, line 1922:
```suggestion
`set2`, `.false.`  otherwise.
```

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+
+#### Status
+
+Experimental
+
+#### Description
+
+Returns `.true.` if the bits in `set1` and `set2` differ and the
+highest order different bit is set to 0 in `set1` and to 1 in `set2`,
+`.false.`.  otherwise. The sets must be the same size otherwise the
```
jvdp1, line 1956:
```suggestion
`.false.`  otherwise. The sets must be the same size otherwise the
```

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+results are undefined
+
+
+#### Syntax
+
+`Result = set1 [[stdlib_bitsets(module):<(interface)]] set2`
```
jvdp1, line 1962:
```suggestion
`result = set1 [[stdlib_bitsets(module):<(interface)]] set2`
```

src/CMakeLists.txt
```diff
@@ -2,6 +2,9 @@
 
 # Create a list of the files to be preprocessed
 set(fppFiles
+    stdlib_bitset_64.fypp
+    stdlib_bitset_large.fypp
```
jvdp1, line 6: I suggest to call these 2 files as follows:
```suggestion
    stdlib_bitsets_64.fypp
    stdlib_bitsets_large.fypp
```

jvdp1 (commented):
Thank you @wclodius2 for this work.
I just reviewed the specs and it looks good to me. I  still need to review the source code and play with it to get more opinions.

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+equivalently be considered a sequence of logical values or a subset of
+the integers 0 ... `bits(bitset)-1`. The bits are indexed from 0 to
+`bits(bitset)-1`. A bitset is used when space savings are critical in applications that require a large number
+of closely related logical values.
+It may also improve performance by reducing memory traffic. To implement bitsets the module
+defines three bitset types, multiple constants, a character string
+literal that can be read to and from strings and formatted files, a
+simple character string literal that can be read to and from strings,
+assignments, procedures, methods, and operators. Note that the module
+assumes two's complement integers, but all current Fortran 95+ processors use such integers.
```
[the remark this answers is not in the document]
wclodius2: It is necessary if users are concerned about the portability of their code. There are still some Unisys machines in operation, and they used one's complement, but never had a Fortran 90+ compiler. There may also be some microcontrollers with one's complement or signed magnitude implementations.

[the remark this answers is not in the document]
jvdp1: Thank you. I 'll remove my suggestion.

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+`bitset_large`. `bitset_type` defines one method, `bits`, all of its
+other methods are deferred to its extensions. `bitset_64` is a bitset
+that can handle up 64 bits. `bitset_large` is a bitset that can handle
+up `huge(0_bits_kind)` bits. All attributes of the bitset types are
+private. The various types each define a sequence of binary values: 0
+or 1. In some cases it is useful to associate a logical value, `test`,
+for each element of the sequence, where `test` is `.true.` if the value
+is 1 and `.false.` otherwise. The number of such values in an entity
+of that type is to be termed, `bits`. The bits are ordered in terms of
+position, that, in turn, is indexed from 0 to `bits-1`. `bitset_type` is
```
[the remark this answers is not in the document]
wclodius2: It can be used to declare a class variable (`class(bitset_type)`) that can be used to represent either bitset_64 or bitset_large types.

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+        if ( set1 == set0 ) &
+            write(*,*) 'Initialization by assignment succeeded'
+        logical2 = set1
+        if ( all( logical2 ) ) then
+            write(*,*) 'Initialization of logical(int32) succeeded.'
+        end if
+    end program demo_assignment
+```
+
+### Table of the `bitset_type` methods
```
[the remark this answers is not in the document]
wclodius2: OK

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+
+
+### Table of the non-member comparison operations
+The comparison operators with two arguments of type `bitset_large` or
+`bitset_64` must have both arguments of the same known type which
+prevents them from being methods. The operands must also have the same
+number of bits otherwise the results are undefined. These operators
+are summarized in the following table:
+
+|Operator|Description|
```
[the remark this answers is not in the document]
wclodius2: My understanding is that ==, etc. are synonyms for .EQ., etc. and defining one automatically defines the other. If I remember correctly gfortran objected when I tried to define both pairs separately on another project. I could list them in this table for thoroughness.

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+`result = self % [[bitset_type(class):all(bound)]]()`
+
+#### Class
+
+Elemental function.
+
+#### Argument
+
+`self`: shall be a scalar expression of class `bitset_type`. It is an
+`intent(in)` argument.
```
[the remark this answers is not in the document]
wclodius2: The form I use parallels that used in the standard. Your form would be awkward for optional arguments.

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+
+#### Status
+
+Experimental
+
+#### Description
+
+Sets the bits in `set1` to the bitwise `and` of the original bits in
+`set1` and `set2`. Note that `set1` and `set2` must have the same
+number of bits, otherwise the result is undefined.
```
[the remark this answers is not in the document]
wclodius2: NaN is solely for floating point (real) numbers. The value of the individual bits in the result is undefined. In practice if `set1` has more bytes than `set2`, then you can have a segmentation fault. If `set2` has more bits than `set1` then bits in `set1` beyond `bits(set1)-1` may be modified, which can have an effect if `set1` is used as the second operand in a binary operation. If `set1` has more bits than `set2`, but the same number of bytes, then the bits beyond `bits(set2)-1` will be used in the operation.

doc/specs/stdlib_bitsets.md
wclodius2: OK

doc/specs/stdlib_bitsets.md
wclodius2: The use of a subroutine:
1. Makes it easier to enforce that `new` must be the same type as `old`
2. Allows me to return a status flag, which is useful now, and necessary if I ever want to allow `new` and `old` to be mixed types

doc/specs/stdlib_bitsets.md
wclodius2: I had been assuming sequential access, but it doesn't matter as long as the read is in the right position in the file. To ensure that you probably want to use the same access in the read as was used in the write.

doc/specs/stdlib_bitsets.md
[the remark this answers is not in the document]
jvdp1: Ok. If defining one automatically defines the other, then I think it would be good to add the other ones in the table too.

doc/specs/stdlib_bitsets.md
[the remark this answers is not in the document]
jvdp1: Ok. I 'll remove my comment. Thank you for the explanation.

doc/specs/stdlib_bitsets.md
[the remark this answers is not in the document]
jvdp1: thank you for the explanation. IMO it would be good to add these explanations to the specs. Let see what others think about it.

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2002 @@
...
+other methods are deferred to its extensions. `bitset_64` is a bitset
+that can handle up 64 bits. `bitset_large` is a bitset that can handle
+up `huge(0_bits_kind)` bits. All attributes of the bitset types are
+private. The various types each define a sequence of binary values: 0
+or 1. In some cases it is useful to associate a logical value, `test`,
+for each element of the sequence, where `test` is `.true.` if the value
+is 1 and `.false.` otherwise. The number of such values in an entity
+of that type is to be termed, `bits`. The bits are ordered in terms of
+position, that, in turn, is indexed from 0 to `bits-1`. `bitset_type` is
+not used in source code. The syntax for using the types are:
```
jvdp1, line 63:
```suggestion
not used in source code of `stdlib`, but is publicaly avaliable fo users to represent either `bitset_64` or `bitset_large`.
The syntax for using the types are:
```
Related to [this comment](https://github.com/fortran-lang/stdlib/pull/239/files#r504891872)

src/CMakeLists.txt
wclodius2: What are your reasons for changing the file names? Changing them affects five files, the two files whose names are changed, them CMakeLists.txt file, the Makefile.manual file, and the markdown document.

jvdp1: So they would have the same prefix ("stdlib_bitsets") as the file stdlib_bitsets.fypp. I always find such things easier to edit (e.g. I don't need to think if a "s" is needed or not). If others have no objections against the current names, it is also fine for me.

wclodius2:
@jvdp1 I have just committed edits to stdlib_bitsets.md based on most of your suggestions in your first posting. I will now rename the fypp files, change the names of their submodules to match the file names, and change the corresponding CMakeLists.txt and Makefile.manual files

src/stdlib_bitsets.fypp
```diff
@@ -0,0 +1,2126 @@
...
+    private
+
+    integer, parameter ::                    &
+        block_size  = bit_size(0_block_kind)
+
+    integer(block_kind), private, parameter :: all_zeros  = 0_block_kind
+    integer(block_kind), private, parameter :: all_ones   = not(all_zeros)
+
+    character(*), parameter, private :: module_name = "STDLIB_BITSETS"
+    integer, parameter, private :: &
```
jvdp1, line 30:
```suggestion
    integer(block_kind),  parameter :: all_zeros  = 0_block_kind
    integer(block_kind),  parameter :: all_ones   = not(all_zeros)

    character(*), parameter :: module_name = "STDLIB_BITSETS"
    integer, parameter :: &
```
`private` is not needed because the default in the module is `private`

src/stdlib_bitsets.fypp
```diff
@@ -0,0 +1,2126 @@
...
+!! Public procedures
+
+    public :: error_handler
+
+    type, abstract :: bitset_type
+!! version: experimental
+!!
+!! Parent type for bitset_64 and bitset_large
+        private
+        integer(bits_kind) :: num_bits
```
jvdp1, line 90: Currently `bits_kind` is equal to `int32`.
I am interested by `stdlib_bitsets` because I have a real case for such a module. However, limiting the number of bits by using `int32` would not allow me to use `stdlib_bitsets` in my applications. The max of int32 is easily reached. What would be the costs to use `int64` instead of `int32`?

src/stdlib_bitsets.fypp
```diff
@@ -0,0 +1,2126 @@
...
+!! Version: experimental
+!!
+!! Creates the bitset, `self`, of size `bits`, with all bits initialized to
+!! zero. `bits` must be non-negative.  If an error occurs and `status` is
+!! absent then processing stops with an informative stop code. `status`
+!! will have one of the values;
+!! * `success` -  if no problems were found,
+!! * `alloc_fault` - if memory allocation failed
+!! * `array_size_invalid_error` - if `bits` is either negative or larger
+!!   than 64 with `self` of class `bitset_64`, or
```
jvdp1, line 933: This sentence is not finished, right?

src/stdlib_bitsets.fypp
```diff
@@ -0,0 +1,2126 @@
...
+!!
+!! Returns the number of non-zero bits in `self`.
+            integer(bits_kind)           :: bit_count
+            class(bitset_64), intent(in) :: self
+        end function bit_count_64
+
+        elemental module subroutine clear_bit_64(self, pos)
+!! Version: experimental
+!!
+!! Sets to zero the bit at `pos` position in `self`. If 'pos` is less than
```
jvdp1, line 1246:
```suggestion
!! Sets to zero the bit at `pos` position in `self`. If `pos` is less than
```

src/stdlib_bitsets.fypp
```diff
@@ -0,0 +1,2126 @@
...
+    end interface assignment(=)
+
+
+    interface and
+
+        elemental module subroutine and_large(set1, set2)
+!! Version: experimental
+!!
+!! Sets the bits in `set1` to the bitwise `and` of the original bits in `set1`
+!! and `set2`. The sets mmust have the same number of bits
```
jvdp1, line 1532:
```suggestion
!! and `set2`. The sets must have the same number of bits
```

src/stdlib_bitsets.fypp
```diff
@@ -0,0 +1,2126 @@
...
+
+    end interface and_not
+
+    interface extract
+
+        module subroutine extract_large(new, old, start_pos, stop_pos, status)
+!! Version: experimental
+!!
+!! Creates a new bitset, `new`, from a range, `start_pos` to `stop_pos`, in
+!! bitset `old`. If 'start_pos` is greater than `stop_pos` the new bitset is
```
jvdp1, line 1626:
```suggestion
!! bitset `old`. If `start_pos` is greater than `stop_pos` the new bitset is
```

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2006 @@
...
+
+## The module's constants
+
+The module defines several public constants all integers, almost all
+intended to serve as error codes in reporting problems through an
+optional `stat` argument. One constant, `bits_kind` is
+the integer kind value for indexing bits and reporting counts of
+bits. The other constants that are error codes are summarized below:
+
+|Error Code|Summary|
```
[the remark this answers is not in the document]
wclodius2: Wait until the reviewers have seen what I did in the source code.

src/stdlib_bitsets.fypp
```diff
@@ -0,0 +1,2126 @@
...
+
+    end interface xor
+
+
+    interface operator(==)
+
+        elemental module function eqv_large(set1, set2) result(eqv)
+!! Version: experimental
+!!
+!! Returns `.true`. if all bits in `set1` and `set2` have the same value,
```
jvdp1, line 1771:
```suggestion
!! Returns `.true` if all bits in `set1` and `set2` have the same value,
```

src/stdlib_bitsets.fypp
```diff
@@ -0,0 +1,2126 @@
...
+    end interface operator(/=)
+
+
+    interface operator(>)
+
+        elemental module function gt_large(set1, set2) result(gt)
+!! Version: experimental
+!!
+!! Returns `.true.` if the bits in `set1` and `set2` differ and the
+!! highest order different bit is set to 1 in `set1` and to 0 in `set2`.
```
jvdp1, line 1864:
```suggestion
!! highest order different bit is set to 1 in `set1` and to 0 in `set2`,
```

jvdp1 (commented):
Here are some other small comments.

src/stdlib_bitsets.fypp
```diff
@@ -0,0 +1,2126 @@
+#:include "common.fypp"
+module stdlib_bitsets
+!! Implements zero based bitsets of size up to `huge(0_int32)`.
+!! The current code uses 64 bit integers to store the bits and uses all 64 bits.
+!! The code assumes two's complement integers, and treats negative integers as
+!! having the sign bit set.
+
+    use, intrinsic ::            &
+        iso_fortran_env, only:   &
```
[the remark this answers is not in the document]
wclodius2: Force of habit. I will change it.

doc/specs/stdlib_bitsets.md
wclodius2: I don't think this is correct. I have changed the text to

used only as a `class` to define entities that can be either a `bitset_64` or
a `bitset_large`. The syntax for using the types are:

Committed and pushed.

src/stdlib_bitsets.fypp
wclodius2: You have a need for greater than 2**31-1 bits in a single variable, as opposed to blocks of 64 bits? I will have to check the code to be certain I was consistent in using bits_kind, but it was intended that it could be changed if desired by the user. The variable will be greater than (2**31-1)/8 bytes so the caches will be too small to hold a single variable so unless you maintain locality of reference (move through the array sequentially) memory traffic might be high.

I have checked the source code. There were problems with stdlib_bitsets_large.fypp,, primarily in not requiring size to consistently return a `bits_kind` integer `kind`, but also a string `len` and the subroutine digit_count. I also made sure the do loop bounds were consistently `bits_kind`. I think I have fixed all the problems, and the typos you found in stdlib_bitsets.fypp, and have committed  and pushed the revised code.

wclodius2: My fix for the subroutine digit_count works for gfortran, but not for ifort where it reports errors at compile time. It doesn't like case values out of range of the select value's kind. gfortran only gives warnings. I have ideas for a fix, but don't have time today.

wclodius2: I fixed the ifort problem, replacing the select case construct with a do loop. I have committed and pushed the resulting code. Now the code is passing all three MacOS builds and failing all three Ubuntu builds with the message:

Error: /home/runner/work/stdlib/stdlib/build is not a directory
Error: Process completed with exit code 1.

I only have a Mac to test on, and don't know the configuration of the Ubuntu systems, so I am at a loss for how to fix things.

jvdp1: > I fixed the ifort problem, replacing the select case construct with a do loop. I have committed and pushed the resulting code. Now the code is passing all three MacOS builds and failing all three Ubuntu builds with the message:
> 
> Error: /home/runner/work/stdlib/stdlib/build is not a directory
> Error: Process completed with exit code 1.
> 
> I only have a Mac to test on, and don't know the configuration of the Ubuntu systems, so I am at a loss for how to fix things.

Thank you. I will check that later on my Linux computer.

wclodius2: I suspect the problem was that I hadn't updated Makefile.manual to reflect the use of stdlib_kinds.f90. Upon updating that, and the other codes so that they compile and run with `bits_kind=int64`, the committed and pushed codes do not report any errors in the past two hours. Usually it only takes a few minutes for them to report a problem.

jvdp1 (commented):
Some typos (I will commit most of them directly).

src/stdlib_bitsets.fypp
wclodius2: As near as I could tell the dangling ", or" was inappropriate and my revisions have replaced it with a "."

src/stdlib_bitsets_64.fypp
```diff
@@ -0,0 +1,1120 @@
...
+            select case( iachar( string(pos:pos) ) )
+            case(ia0:ia9)
+                digits = digits + 1
+                if ( digits == 10 .AND. bits > 2_bits_kind**30/5 ) go to 996
+!! May not be quite right
+                if ( digits > 10 ) go to 996
+                bits = bits*10 + iachar( string(pos:pos) ) - ia0
+                if ( bits < 0 ) go to 996
+            case(iachar('b'), iachar('B'))
+                go to 100
```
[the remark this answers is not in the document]
wclodius2: For some reason I missed this and the following post. I will replace the `go to 100` with a simple `exit` for the moment. Let me know if that is unsatisfactory for some reason.

src/stdlib_bitsets_64.fypp
```diff
@@ -0,0 +1,1120 @@
...
+            pos = pos + 1
+
+        end do
+
+100     if ( bits > 64 ) then
+            call error_handler( 'BITS in STRING was greater than 64.', &
+                                char_string_too_large_error, status,   &
+                                module_name, procedure )
+            return
+        end if
```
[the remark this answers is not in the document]
wclodius2: I recently introduced `max_digits` and `overflow_bits` in stdlib_bitsets.fypp
      integer, parameter ::                   &
          max_digits = 10 ! bits_kind == int32
    !        max_digits = 20 ! bits_kind == int64

      integer(bits_kind), parameter ::     &
          overflow_bits = 2_bits_kind**30/5 ! bits_kind == int32
    !        overflow_bits = 2_bits_kind**62/5 ! bits_kind == int64

so that

                if (  digits == 10 .AND. bits > 2_bits_kind**30/5 ) &

can be written

                if (  digits == max_digits .AND. bits > overflow_bits ) &

so that overflow checks for reading and writing bitsets can be easily modified for bits_kind=int64. For some reason, this introduction has revealed that gfortran 10.2's runtime code generation is fragile. In order to get the test cases to run to completion I had to introduce the statement

    public :: max_digits, overflow_bits

in stdlib_bitsets.fypp and change

        if ( pos < 0 .OR. pos > self % num_bits-1 ) return
        self % block = ibset( self % block, pos )

in stdlib_bitsets_64.fypp to

        integer(block_kind) :: dummy
        if ( pos < 0 .OR. pos > self % num_bits-1 ) return
        dummy = ibset( self % block, pos )
        self % block = dummy

in order for the test cases to run to completion. I don't think that either change is required by the standard. Certainly other parameters used by the submodules do not have to be `public`, and the comparable code in stdlib_bitsets_large.fypp

        if ( pos < 0 .OR. pos > self % num_bits-1 ) return

        set_block = pos / block_size + 1
        block_bit = pos - (set_block - 1) * block_size
        self % blocks(set_block) = ibset( self % blocks(set_block), block_bit )

still seems to be working.

src/stdlib_bitsets_large.fypp
```diff
@@ -0,0 +1,1348 @@
...
+        pos = pos + 1
+        bits = 0
+        digits = 0
+
+        do
+            select case( iachar( string(pos:pos) ) )
+            case(ia0:ia9)
+                digits = digits + 1
+                if ( digits == max_digits .AND. bits > overflow_bits ) go to 996
+!! May not be quite right
```
jvdp1, line 838:
```suggestion
!! May not be quite right
```
What does this comment mean? What could be not quite right?

src/tests/bitsets/test_stdlib_bitset_64.f90
```diff
@@ -0,0 +1,744 @@
+program test_stdlib_bitset_64
+    use, intrinsic :: iso_fortran_env, only : int8, int16, int32, int64
```
jvdp1, line 2:
```suggestion
    use stdlib_kinds, only : int8, int16, int32, int64
```

src/tests/bitsets/test_stdlib_bitset_64.f90
```diff
@@ -0,0 +1,744 @@
...
+    call test_bitset_operations()
+
+contains
+
+    subroutine test_string_operations()
+        character(*), parameter:: procedure = 'TEST_STRING_OPERATIONS'
+
+        write(*,*)
+        write(*,*) 'Test string operations: from_string, read_bitset, ' // &
+            'to_string, and write_bitset'
```
jvdp1, line 33:
```suggestion
        write(*,'(/a)') 'Test string operations: from_string, read_bitset, ' // &
            'to_string, and write_bitset'
```

src/tests/bitsets/test_stdlib_bitset_64.f90
```diff
@@ -0,0 +1,744 @@
...
+
+        call set0 % from_string( bitstring_0 )
+        if ( bits(set0) /= 33 ) then
```
jvdp1, line 36: Suggestion only (does not need to be applied if not wanted/too much work):
```suggestion
        call check(bits(set0) /= 33, msg =  error stop procedure // ' from_string failed to interpret ' // &
                'bitstring_0 size properly.' )
```

src/tests/bitsets/test_stdlib_bitset_64.f90
```diff
@@ -0,0 +1,744 @@
...
+            error stop procedure // ' failed to interpret bitstring_all ' // &
+                'value properly.'
+        else
+            write(*,*) 'from_string transferred bitstring_all properly ' // &
+                'into set1'
+        end if
+
+        call set3 % read_bitset( bitstring_0, status )
+        if ( status /= success ) then
+            write(*,*) 'read_bitset_string failed with bitstring_0 as expected.'
```
jvdp1, line 69: Is an `error stop` not missing here?

src/tests/bitsets/test_stdlib_bitset_64.f90
```diff
@@ -0,0 +1,744 @@
...
+        if ( set5 /= set0 .or. set4 /= set1 .or. set3 /= set2 ) then
+            error stop procedure // ' transfer to and from units using ' // &
+                ' bitset literals with advance == "no" failed.'
+        else
+            write(*,*) 'Transfer to and from units using ' // &
+                'write_bitset_unit and read_bitset_unit with ' // &
+                'advance=="no" succeeded.'
+        end if
+
+        open( newunit=unit, file='test.bin', status='replace', &
```
jvdp1, line 191: I suggest to add some tests with `access='stream', form = 'unformatted'`

src/tests/bitsets/test_stdlib_bitset_64.f90
```diff
@@ -0,0 +1,744 @@
...
+        end if
+
+        call set0 % not()
+        do i=0, set0 % bits() - 1
+            if ( set0 % test(i) ) go to 100
+        end do
+
+        write(*,*) 'As expected set0 had no bits set.'
+
+        go to 110
```
jvdp1, line 412: Is this `go to 110` really needed?

src/tests/bitsets/test_stdlib_bitset_64.f90
```diff
@@ -0,0 +1,744 @@
...
+
+100     error stop procedure // ' against expectations set0 has ' // &
```
jvdp1, line 414: `100` is only used once Therefore, I suggest to replace the `go to 100` by the `error stop` directly in te code.

src/tests/bitsets/test_stdlib_bitset_64.f90
```diff
@@ -0,0 +1,744 @@
...
+            if ( .not. set1 % test(i) ) go to 200
+        end do
+
+        write(*,*) 'As expected set1 had all bits set.'
+
+        go to 210
+
+200     error stop procedure // ' against expectations set1 has ' // &
+                   'at least 1 bit unset.'
+210     continue
```
jvdp1, line 429: Same comment as previous: replace 200 by the error stop.

src/tests/bitsets/test_stdlib_bitset_large.f90
```diff
@@ -0,0 +1,1476 @@
...
+            error stop procedure // ' failed to interpret bitstring_all ' // &
+                '// bitstring_all value properly.'
+        else if ( .not. set11 % any() ) then
+            error stop procedure // ' failed to interpret bitstring_all ' // &
+                '// bitstring_all value properly.'
+        else if ( .not. set11 % all() ) then
+            error stop procedure // ' failed to interpret bitstring_all ' // &
+                '// bitstring_all value properly.'
+        else
+            write(*,*) 'from_string transferred bitstring_all // ' // &
```
jvdp1, line 97:
```suggestion
            write(*,*) 'from_string transferred bitstring_all ' // &
```

src/tests/bitsets/test_stdlib_bitset_large.f90
```diff
@@ -0,0 +1,1476 @@
...
+            write(*,*) 'read_bitset_string failed with bitstring_0 ' // &
+                '// bitstring_0 as expected.'
+        end if
+
+        call set3 % read_bitset( 's33b' // bitstring_0, status )
+        if ( bits(set3) /= 33 ) then
+            error stop procedure // ' read_bitset_string failed to ' // &
+                'interpret "s33b" // bitstring_0 size properly.'
+        else if ( .not. set3 % none() ) then
+            error stop procedure // ' failed to interpret "s33b" // ' // &
```
jvdp1, line 117:
```suggestion
            error stop procedure // ' failed to interpret "s33b" ' // &
```

src/tests/bitsets/test_stdlib_bitset_large.f90
```diff
@@ -0,0 +1,1476 @@
...
+                'bitstring_0 value properly.'
+        else
+            write(*,*) 'read_bitset_string transferred "s33b" // ' // &
```
jvdp1, line 120:
```suggestion
            write(*,*) 'read_bitset_string transferred "s33b" ' // &
```

src/tests/bitsets/test_stdlib_bitset_large.f90
```diff
@@ -0,0 +1,1476 @@
...
+            error stop procedure // ' read_bitset_string failed to ' // &
+                'interpret "s33b" // bitstring_all size properly.'
+        else if ( set4 % none() ) then
+            error stop procedure // ' read_bitset_string failed to ' // &
+                'interpret "s33b" // bitstring_all value properly.'
+        else if ( .not. set4 % any() ) then
+            error stop procedure // ' read_bitset_string failed to // ' // &
+                'interpret "s33b" bitstring_all value properly.'
+        else if ( .not. set4 % all() ) then
+            error stop procedure // ' read_bitset_string failed to // ' // &
```
jvdp1, line 148:
```suggestion
            error stop procedure // ' read_bitset_string failed to ' // &
```

src/tests/bitsets/test_stdlib_bitset_large.f90
```diff
@@ -0,0 +1,1476 @@
...
+
+        call set4 % read_bitset( 's33b' // bitstring_all )
+        if ( bits(set4) /= 33 ) then
+            error stop procedure // ' read_bitset_string failed to ' // &
+                'interpret "s33b" // bitstring_all size properly.'
+        else if ( set4 % none() ) then
+            error stop procedure // ' read_bitset_string failed to ' // &
+                'interpret "s33b" // bitstring_all value properly.'
+        else if ( .not. set4 % any() ) then
+            error stop procedure // ' read_bitset_string failed to // ' // &
```
jvdp1, line 145:
```suggestion
            error stop procedure // ' read_bitset_string failed to ' // &
```

src/tests/bitsets/test_stdlib_bitset_large.f90
```diff
@@ -0,0 +1,1476 @@
...
+            error stop procedure // ' transfer to and from units using ' // &
+                ' bitset literals for bitss > 64 with advance == "no" failed.'
+        else
+            write(*,*) 'Transfer bits > 64 to and from units using ' // &
+                'write_bitset_unit and read_bitset_unit with ' // &
+                'advance=="no" succeeded.'
+        end if
+
+        open( newunit=unit, file='test.bin', status='replace', &
+            form='unformatted', action='write' )
```
jvdp1, line 341: It would be nice to have the same tests with `access=stream`

src/tests/bitsets/test_stdlib_bitset_large.f90
```diff
@@ -0,0 +1,1476 @@
...
+        end do
+
+        write(*,*) 'As expected set0 had no bits set.'
+
+        go to 110
+
+100     error stop procedure // ' against expectations set0 has ' // &
+                    'at least 1 bit set.'
+
+110     continue
```
jvdp1, line 735: These series of `go to` are not needed, right?

jvdp1 (commented):
Here are some additional comments on the tests. All options seem to be covered by the tests. I added some suggestions.
On overall I am pleased with this PR. Thank you @wclodius2!

wclodius2:
It means I wrote the code when I was tired and wanted someone else to check my reasoning. It is parsing a decimal representation of the size in bits, and I want to be certain that the size does not exceed what can be represented by an integer of `bits_kind`. That could happen for example if parsing a file with `bits_kind==int32` and the file was written with `bits_kind==int64`. We want to be sure that the representation for `bits_kind==int32` does not exceed `2**31-1`. My reasoning is that it will exceed that value if:
1. it has ten digits and is about to read another digit increasing its value by a factor of ten
2. It has nine digits and has a value greater than `2**31/10`
3. It had a value of `2**31/10` and reading the next digit causes an integer overflow of nine or less resulting in a negative value.

wclodius2:
I have now resolved my qualms about my handling of potential integer overflows on reads. I have changed `max_digits` for `bits_kind==int64` from 20 to 19 (`2**63-1` is about `9.2E18`) and convinced myself that the rest of my logic was correct. I have move the comments that the code may not be quite right. (I also fixed a typo in an error message.)

src/tests/bitsets/test_stdlib_bitset_64.f90
wclodius2: OK

src/tests/bitsets/test_stdlib_bitset_64.f90
wclodius2: Doing it everywhere seems like a lot of work, particularly for those cases with `if...then...else if...` cascades, and doing it for the few cases without the `if...then...else if...` cascades is liable to result in an inconsistent style of error reporting.

src/tests/bitsets/test_stdlib_bitset_64.f90
wclodius2: I guess an error stop if the expected failure does not occur is appropriate.

src/tests/bitsets/test_stdlib_bitset_64.f90
wclodius2: OK

src/tests/bitsets/test_stdlib_bitset_64.f90
wclodius2: I don't see how to avoid using a go to here.

wclodius2: Now I do.

src/tests/bitsets/test_stdlib_bitset_64.f90
wclodius2: Done

src/tests/bitsets/test_stdlib_bitset_large.f90
wclodius2: The original code is correct if odd.

src/tests/bitsets/test_stdlib_bitset_large.f90
wclodius2: Creating a string about concatenation looks odd, but is correct.

src/tests/bitsets/test_stdlib_bitset_large.f90
wclodius2: Done

src/tests/bitsets/test_stdlib_bitset_large.f90
wclodius2: Yes they can be recoded and I have done so.

wclodius2:
@jvdp1 thanks for the thorough review.

src/tests/bitsets/test_stdlib_bitset_64.f90
jvdp1: I understand your comment. Good for me as it is.

src/tests/bitsets/test_stdlib_bitset_large.f90
jvdp1: Thank you.

src/tests/bitsets/test_stdlib_bitset_large.f90
jvdp1: I understand the code now. Sorry for the misunderstanding.

src/tests/bitsets/test_stdlib_bitset_large.f90
```diff
@@ -0,0 +1,1504 @@
...
+        end do
+
+        write(*,*) 'As expected set10 had no bits set.'
+
+        go to 710
+
+700     error stop procedure // ' against expectations set10 has ' // &
+                    'at least 1 bit set.'
+
+710     continue
```
jvdp1, line 890: These gotos and the following ones could be removed too, right?

wclodius2: Done. Sorry I missed them.

doc/specs/stdlib_bitsets.md
```diff
@@ -0,0 +1,2000 @@
...
+# The `stdlib_bitsets` module
+
+[TOC]
+
+## Introduction
+
+The `stdlib_bitsets` module implements bitset types. A bitset is a
+compact representation of a sequence of `bits` binary values. It can
+equivalently be considered as a sequence of logical values or as a
+subset of the integers 0 ... `bits-1`. The bits are indexed from 0 to
```
milancurcic, line 14: > or as a subset of the integers 0 ... `bits-1`.

This confuses me. The equivalent integer values are 0 and 1, not 0 through `bits-1`, correct?

Do you mean to say something like this?

"It can equivalently be considered as a sequence of logical values or as a sequence of integers 0 and 1 with indices in the range 0... `bits-1`?"

wclodius2:
A bit set for each index `bit` in the range `0…bits-1` has a value of 0 or 1.  A value of 1 for index `bit` can be considered as including the `bit` value in the subset, i.e., the value `1110` can be considered as defining the subset (1, 2, 3). This was how Wirth defined sets in Pascal.

milancurcic:
Okay, thanks, I understand now. All good.

milancurcic:
Your example "i.e., the value `1110` can be considered as defining the subset (1, 2, 3)" was what made it click for me. Do you agree that we include it in the spec? Something like this:

"It can equivalently be considered as a sequence of logical values or as a subset of the integers 0 ... `bits-1`. For example, the value `1110` can be considered as defining the subset of integers [1, 2, 3]."

wclodius2:
Sounds good to me.

milancurcic:
I will go ahead and merge this considering there aren't any objections. Thank you @wclodius2 and all the reviewers.

Issue #214, opened 2020-06-23
Cube root of a real number


ivan-pi:
Related to #150 (non-special mathematical functions)

### Description

Returns the cube root of the real number \(x\), that is a number \(y) such that \(y^3 = x\). 

### Syntax

`y = cbrt(x)`

### Arguments

`x`: A real number \(x\).

### Return value

Returns the value \(\sqrt[3]{x}\), the result is of the type `real` and has the same kind as `x`.

### Example

```fortran
program demo_cbrt
use stdlib_experimental_math, only: cbrt
print *, cbrt(27.), cbrt(-27.)  ! outputs 3, -3
end program
```
---

As seen from the discussion on [Discourse](https://fortran-lang.discourse.group/t/error-handling-for-negative-real-numbers/120) this function is semantically more accurate than writing  `x**(1./3.)` which only works for positive real numbers and returns NaN otherwise.

A possible extension would be to allow complex arguments, the return value would then be an array with 3 elements for the 3 cube roots (if the number is real and non-zero, there is one real root and a conjugate pair of complex roots; a complex non-zero value will have three distinct cube roots)

ivan-pi:
I've already got an implementation ready based upon the NSWC Mathematical Library version and using Fypp for templating of different real kinds.

jvdp1:
Thanks.

> A possible extension would be to allow complex arguments, the return value would then be an array with 3 elements for the 3 cube roots (if the number is real and non-zero, there is one real root and a conjugate pair of complex roots; a complex non-zero value will have three distinct cube roots)

I don't use usually complex numbers, neither cube roots. However, to be in agreement with the intrinsic function `sqrt`, I think it should accept complex arguments. In this case, is a function still possible, or should it be a subroutine?

ivan-pi:
Personally, I also don't have a foreseeable usage for complex cube roots. 

It should however, be possible to have the the same interface for real and complex cube roots:
```fortran
interface cbrt
function cbrt_sp(x) result(cbrt)
  complex(sp), intent(in) :: x
  complex(sp) :: cbrt
end function
function cbrt_complex_sp(x) result(cbrt)
  complex(sp), intent(in) :: x
  complex(sp) :: cbrt(3)
end function
end interface
```
A better option might be to follow the IMSL library [CBRT(X)](https://www.afs.enea.it/software/imsl/help/sfun/sfun.htm?turl=docs%2Fcbrt.htm) function:
[...]

PR #240, opened 2020-10-10
Initial round of probability distributions and statistical functions

Jim-215-Fisher:
This is the first round of probability distributions and statistical functions. Anyone is welcomed to review it. The following source files were added:

src/stats_distribution_rvs.fypp
src/stats_distribution.fypp
src/stats_distribution_implementation.fypp

The specs documentation was added:

doc/specs/stdlib_stats_distribution.md

The test program was added:

src/tests/stats/test_distribution.f90

The following compilation files were modified:

src/CMakeLists.txt
src/Makefile.manual
src/tests/stats/CMakeLists.txt
src/tests/stats/Makefile.manual
doc/specs/index.md

### DIFF
--- a/doc/specs/index.md
+++ b/doc/specs/index.md
@@ -17,6 +17,7 @@ This is and index/directory of the specifications (specs) for each new module/fe
  - [optval](./stdlib_optval.html) - Fallback value for optional arguments
  - [quadrature](./stdlib_quadrature.html) - Numerical integration
  - [stats](./stdlib_stats.html) - Descriptive Statistics
+ - [stats](./stdlib_stats_distribution.html) - Random Distributions and Statistical Functions
 
 ## Missing specs
 
--- a/doc/specs/index.md
+++ b/doc/specs/index.md
@@ -17,7 +17,7 @@ This is and index/directory of the specifications (specs) for each new module/fe
  - [optval](./stdlib_optval.html) - Fallback value for optional arguments
  - [quadrature](./stdlib_quadrature.html) - Numerical integration
  - [stats](./stdlib_stats.html) - Descriptive Statistics
- - [stats](./stdlib_stats_distribution.html) - Random Distributions and Statistical Functions
+ - [distributions](./stdlib_stats_distribution.html) - Random Distributions and Statistical Functions
 
 ## Missing specs
 
--- a/doc/specs/stdlib_stats_distribution.md
+++ b/doc/specs/stdlib_stats_distribution.md
@@ -288,7 +288,7 @@ Experimental
 
 Cumulative distribution function of the normal continuous distribution
 
-\[F(X)=\frac{1}{2}\left&space;[&space;1&space;&plus;&space;erf(\frac{x-\mu}{\sqrt{2}&space;\sigma})&space;\right&space;]]\
+\[F(X)=\frac{1}{2}\left&space;[&space;1&space;&plus;&space;erf(\frac{x-\mu}{\sqrt{2}&space;\sigma})&space;\right&space;]\]
 ![equation](https://latex.codecogs.com/gif.latex?F(X)=\frac{1}{2}\left&space;[&space;1&space;&plus;&space;erf(\frac{x-\mu}{\sqrt{2}&space;\sigma})&space;\right&space;])
 
 ### Syntax
--- a/doc/specs/stdlib_stats_distribution.md
+++ b/doc/specs/stdlib_stats_distribution.md
@@ -80,7 +80,7 @@ program demo_uniform_rvs
 
     loc = 0.0
     scale = 1.0
-    print *, uniform(loc, scale)            ! single standard uniform random variate in [0, 1)
+    print *, uniform(loc, scale)            ! single standard uniform random variate in [0, 1]
     rv = uniform(loc, scale, 10)            ! an array of 10 uniform random variates in [0, 1)
     loc = -1.0
     scale = 2.0
--- a/doc/specs/stdlib_stats_distribution.md
+++ b/doc/specs/stdlib_stats_distribution.md
@@ -288,7 +288,6 @@ Experimental
 
 Cumulative distribution function of the normal continuous distribution
 
-\( F(X)=\frac{1}{2}\left&space;[&space;1&space;&plus;&space;erf(\frac{x-\mu}{\sqrt{2}&space;\sigma})&space;\right&space;] \)
 ![equation](https://latex.codecogs.com/gif.latex?F(X)=\frac{1}{2}\left&space;[&space;1&space;&plus;&space;erf(\frac{x-\mu}{\sqrt{2}&space;\sigma})&space;\right&space;])
 
 ### Syntax
--- a/doc/specs/stdlib_stats_distribution.md
+++ b/doc/specs/stdlib_stats_distribution.md
@@ -81,7 +81,7 @@ program demo_uniform_rvs
     loc = 0.0
     scale = 1.0
     print *, uniform(loc, scale)            ! single standard uniform random variate in [0, 1]
-    rv = uniform(loc, scale, 10)            ! an array of 10 uniform random variates in [0, 1)
+    rv = uniform(loc, scale, 10)            ! an array of 10 uniform random variates in [0, 1]
     loc = -1.0
     scale = 2.0
     print *, uniform(loc, scale)            ! single uniform random variate in [-1,1)
--- a/doc/specs/stdlib_stats_distribution.md
+++ b/doc/specs/stdlib_stats_distribution.md
@@ -14,7 +14,7 @@ Experimental
 
 ### Description
 
-Must set the seed value before calling random distribution for variates. 
+Sets the seed value before calling random distribution for variates. 
 
 ### Syntax
 
--- a/doc/specs/stdlib_stats_distribution.md
+++ b/doc/specs/stdlib_stats_distribution.md
@@ -1,0 +1,456 @@
+---
+title: stats_distribution
+---
+
+# Statistical Distribution
+
+[TOC]
+
+## `random_seed` - set or get a value of seed to the random distribution psudorandom number generator
+
+### Status
+
+Experimental
+
+### Description
+
+Must set the seed value before calling random distribution for variates. 
+
+### Syntax
+
+`call random_seed(put [, get])`
+
+### Arguments
+
+`put`: argument has intent `in` and may be a scalar of type `integer` with kind of int32.
+
+`get`: optional argument has intent `out` and is a scalar of type `integer` with kind of int32.
+
+### Return value
+
+
+### Example
+
+```fortran
+program demo_norm_seed
+    use stdlib_stats_distribution, only : random_seed
+    use iso_fortran_env, only : int32
+    implicit none
+    integer(int32) :: seed, seed_value
+
+    seed = 1234567
+    call random_seed(seed)                   ! reset the value of seed
+    call random_seed(seed, seed_value)       ! get current value of seed
+end program demo_norm_seed
+```
+
+## `uniform_distribution_rvs` - uniform distribution random variates
+
+### Status
+
+Experimental
+
+### Description
+
+Using the parameters `loc` and `scale`, one obtains the uniformly distributed random variates on [loc, loc + scale]. Return a scalar of type `real` pseudorandom number or a rank one array of such numbers with a size of `array_size`.
+
+### Syntax
+
+`result = [[stdlib_stats_distribution(module):uniform_distribution_rvs(interface)]](loc, scale, [array_size])`
+
+### Arguments
+
+`array_size`: optional argument has intent `in` and is a scalar of type `integer`.
+
+`loc`: has intent `in` and is a scalar of type `real`.
+
+`scale`: has intent `in` and is a scalar of type `real`.
+
+### Return value
+
+The result is a scalar of type `real`, or is a rank one array of type `real` with size of `array_size`.
+
+### Example
+
+```fortran
+program demo_uniform_rvs
+    use stdlib_stats_distribution, only: uniform => uniform_distribution_rvs
+    implicit none
+    real :: loc, scale, rv(10)
+
+    loc = 0.0
+    scale = 1.0
+    print *, uniform(loc, scale)            ! single standard uniform random variate in [0, 1)
+    rv = uniform(loc, scale, 10)            ! an array of 10 uniform random variates in [0, 1)
+    loc = -1.0
+    scale = 2.0
+    print *, uniform(loc, scale)            ! single uniform random variate in [-1,1)
+    loc = -0.5
+    scale = 1.0
+    rv = uniform(loc, scale, 10)            ! an array of 10 uniform variates in [-0.5, 0.5)
+end program demo_uniform_rvs
+```
+
+## `uniform_distribution_pdf` - uniform probability density function
+
+### Status
+
+Experimental
+
+### Description
+
+The probability density function of the continuous uniform distribution.
+
+![equation](https://latex.codecogs.com/gif.latex?f(x)=\begin{cases}\frac{1}{scale}&loc\leqslant&space;x<loc&plus;scale\\\\0&x<loc,or,x>loc&plus;scale\end{cases})
+
+### Syntax
+
+`result = [[stdlib_stats_distribution(module):uniform_distribution_pdf(interface)]](x, loc, scale)`
+
+### Arguments
+
+`x`: has intent `in` and is a scalar or an array of type `real`.
+
+`loc`: has intent `in` and is a scalar of type `real`.
+
+`scale`: has intent `in` and is a scalar of type `real`.
+
+### Return value
+
+The result is a scalar of type `real` or an array of type `real` with a shape conformable to input `x`.
+
+### Example
+
+```fortran
+program demo_uniform_pdf
+    use stdlib_stats_distribution,  uniform_pdf => uniform_distribution_pdf, &
+                                    uniform => uniform_distribution_rvs
+    implicit none
+    real :: loc, scale, x(3,4)
+
+    loc = 0.0
+    scale = 1.0
+    print *, uniform_pdf(0.5, loc, scale)             ! a probability density at 0.5 in [0,1)
+    print *, uniform_pdf(0.7, -1.0, 2.0)              ! a probability density at 0.7 in [-1,1)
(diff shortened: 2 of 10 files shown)


Jim-215-Fisher:
Realized the Fortran standard functions 'random_seed' and 'random_number' may not have the same algorithm on different platforms and OS. Will update the random number generator.

TODO: modify the random_seed function to avoid using standard random_number function call
TODO: create new random number generator for random distribution.

milancurcic:
Thanks a lot @Jim-215-Fisher! I will review it. On first look I only noticed two things:

1. `random_seed` if imported as is will make the intrinsic `random_seed` unavailable. I think we should rename it.
2. Minor style nit-pick, let's not use any non-newline code separators like
```
!---------------------------------------------------------------------------------------------
!                   Random seed
!---------------------------------------------------------------------------------------------
```
as they don't add value. I know, this is not mentioned in the style guide yet. Good rule of thumb is to just mimic the style of existing modules.

More later.

Jim-215-Fisher:
The `random_seed` is an overloading to intrinsic, and need two arguments which is different from intrinsic. I thought it might be easier for users to use the same name. 
I will remove those separators.

milancurcic:
> The random_seed is an overloading to intrinsic, and need two arguments which is different from intrinsic. I thought it might be easier for users to use the same name.

Ah, okay, I didn't realize. I think this is fine if it extends the intrinsic one while preserving the original behavior.

jvdp1:
Thank you @Jim-215-Fisher for this PR. I will try to review it this week.

src/stdlib_stats_distribution.fypp
```diff
@@ -0,0 +1,226 @@
...
+    !! version: experimental
+    !!
+    !! Set or get seed value of probability distributions for psudorandom bumber generator 
+    !! ([Specification](../page/specs/stdlib_stats_distribution.html#description))
+    !! 
+        procedure random_distribution_seed
+    end interface random_seed
+!---------------------------------------------------------------------------------------------
+!                   Uniform Distribution
+!---------------------------------------------------------------------------------------------
```
14NGiestas, line 40:
```suggestion

```

Jim-215-Fisher: as per @milancurcic , these separators add no value and is not in style, will remove in the next round.

Jim-215-Fisher:
Just load modified version. 1)add 64bit random integer generator code, 2) modified `random_seed` procedure to use the new generator instead of `random_number`, 3) modified uniform distribution random generator, 4) add inversion transformation procedure for binomial random variate generator. It was noticed that both BTRD and BTPE from Alan Miller algorithm failed the chi-squared test.

ivan-pi (commented):
Could it be a that your hexadecimal constants are overflowing the int64 kind?

With gfortran you might need the flag -fno-range-check in that case (see here: https://github.com/fortran-lang/stdlib/issues/214#issuecomment-656640371).

Jim-215-Fisher:
I noticed that issue. I have changed that part into regular negative decimal. Right now, file stats_distribution_rv and stats_distribution have passed compilation, but stats_distribution_implementation fail to compile on ubuntu. The same file has passed on windows and Mac.

Jim-215-Fisher:
This is likely a bug on gcc. See Bug 91773 - Buffer overflow for long module/submodule names

doc/specs/index.md
```diff
@@ -17,6 +17,7 @@ This is and index/directory of the specifications (specs) for each new module/fe
  - [optval](./stdlib_optval.html) - Fallback value for optional arguments
  - [quadrature](./stdlib_quadrature.html) - Numerical integration
  - [stats](./stdlib_stats.html) - Descriptive Statistics
+ - [stats](./stdlib_stats_distribution.html) - Random Distributions and Statistical Functions
```
jvdp1, line 20: To avoid confusion, maybe:
```suggestion
 - [distributions](./stdlib_stats_distribution.html) - Random Distributions and Statistical Functions
```

doc/specs/stdlib_stats_distribution.md
```diff
@@ -0,0 +1,456 @@
+---
+title: stats_distribution
+---
+
+# Statistical Distribution
```
jvdp1, line 5:
```suggestion
# Statistical Distributions
```

doc/specs/stdlib_stats_distribution.md
```diff
@@ -0,0 +1,456 @@
...
+
+[TOC]
+
+## `random_seed` - set or get a value of seed to the random distribution psudorandom number generator
```
jvdp1, line 9:
```suggestion
## `random_seed` - set or get a value of seed to the random distribution pseudorandom number generator
```

doc/specs/stdlib_stats_distribution.md
```diff
@@ -0,0 +1,456 @@
...
+
+### Status
+
+Experimental
+
+### Description
+
+Must set the seed value before calling random distribution for variates.
```
jvdp1, line 17:
```suggestion
Sets the seed value before calling random distribution for variates. 
```

doc/specs/stdlib_stats_distribution.md
```diff
@@ -0,0 +1,456 @@
...
+
+Must set the seed value before calling random distribution for variates. 
+
+### Syntax
+
+`call random_seed(put [, get])`
+
+### Arguments
+
+`put`: argument has intent `in` and may be a scalar of type `integer` with kind of int32.
```
jvdp1, line 25:
```suggestion
`put`: argument has intent `in` and may be a scalar of type `integer` with kind of `int32`.
```

doc/specs/stdlib_stats_distribution.md
```diff
@@ -0,0 +1,456 @@
...
+
+`get`: optional argument has intent `out` and is a scalar of type `integer` with kind of int32.
```
jvdp1, line 27:
```suggestion
`get` (optional): argument has intent `out` and is a scalar of type `integer` with kind of `int32`.
```

doc/specs/stdlib_stats_distribution.md
```diff
@@ -0,0 +1,456 @@
...
+
+### Return value
```
jvdp1, line 29: Should be completed to describe `get`

doc/specs/stdlib_stats_distribution.md
```diff
@@ -0,0 +1,456 @@
...
+
+```fortran
+program demo_uniform_rvs
+    use stdlib_stats_distribution, only: uniform => uniform_distribution_rvs
+    implicit none
+    real :: loc, scale, rv(10)
+
+    loc = 0.0
+    scale = 1.0
+    print *, uniform(loc, scale)            ! single standard uniform random variate in [0, 1)
```
jvdp1, line 83:
```suggestion
    print *, uniform(loc, scale)            ! single standard uniform random variate in [0, 1]
```

doc/specs/stdlib_stats_distribution.md
```diff
@@ -0,0 +1,456 @@
...
+    rv = uniform(loc, scale, 10)            ! an array of 10 uniform random variates in [0, 1)
```
jvdp1, line 84:
```suggestion
    rv = uniform(loc, scale, 10)            ! an array of 10 uniform random variates in [0, 1]
```

doc/specs/stdlib_stats_distribution.md
```diff
@@ -0,0 +1,456 @@
...
+    loc = -1.0
+    scale = 2.0
+    print *, uniform(loc, scale)            ! single uniform random variate in [-1,1)
```
jvdp1, line 87:
```suggestion
    print *, uniform(loc, scale)            ! single uniform random variate in [-1,1]
```

doc/specs/stdlib_stats_distribution.md
```diff
@@ -0,0 +1,456 @@
...
+    loc = -0.5
+    scale = 1.0
+    rv = uniform(loc, scale, 10)            ! an array of 10 uniform variates in [-0.5, 0.5)
```
jvdp1, line 90:
```suggestion
    rv = uniform(loc, scale, 10)            ! an array of 10 uniform variates in [-0.5, 0.5]
```

doc/specs/stdlib_stats_distribution.md
```diff
@@ -0,0 +1,456 @@
...
+                                    uniform => uniform_distribution_rvs
+    implicit none
+    real :: loc, scale, x(3,4)
+
+    loc = 0.0
+    scale = 1.0
+    print *, uniform_pdf(0.5, loc, scale)             ! a probability density at 0.5 in [0,1)
+    print *, uniform_pdf(0.7, -1.0, 2.0)              ! a probability density at 0.7 in [-1,1)
+    x = reshape(uniform(loc, scale, 12),[3,4])        ! uniform random variates array in [0,1)
+    print *, uniform_pdf(x, loc, scale)               ! probability density array in [0,1)
```
jvdp1, line 136:
```suggestion
    print *, uniform_pdf(0.5, loc, scale)             ! a probability density at 0.5 in [0,1]
    print *, uniform_pdf(0.7, -1.0, 2.0)              ! a probability density at 0.7 in [-1,1]
    x = reshape(uniform(loc, scale, 12),[3,4])        ! uniform random variates array in [0,1]
    print *, uniform_pdf(x, loc, scale)               ! probability density array in [0,1]
```

doc/specs/stdlib_stats_distribution.md
```diff
@@ -0,0 +1,456 @@
...
+
+### Status
+
+Experimental
+
+### Description
+
+The probability density function of the continuous uniform distribution.
+
+![equation](https://latex.codecogs.com/gif.latex?f(x)=\begin{cases}\frac{1}{scale}&loc\leqslant&space;x<loc&plus;scale\\\\0&x<loc,or,x>loc&plus;scale\end{cases})
```
jvdp1, line 104: FORD supports Latex: see [here](https://github.com/Fortran-FOSS-Programmers/ford/wiki/Writing-Documentation#latex-support) for more details.
Something like (I didn't check the formula):
```suggestion
\( \frac{1}{scale}&loc\leqslant&space;x<loc&plus;scale\\\\0&x<loc,or,x>loc&plus;scale )\
```

doc/specs/stdlib_stats_distribution.md
```diff
@@ -0,0 +1,456 @@
...
+end program demo_norm_seed
+```
+
+## `uniform_distribution_rvs` - uniform distribution random variates
+
+### Status
+
+Experimental
+
+### Description
```
jvdp1, line 53: Would it be possible to return an muti-dimension array (e.g., a ran-3 array), e.g., as in [Matlab](https://www.mathworks.com/help/stats/unifrnd.html)

doc/specs/stdlib_stats_distribution.md
```diff
@@ -0,0 +1,456 @@
...
+                                   uniform => uniform_distribution_rvs
+    implicit none
+    real :: loc, scale, x(3,4)
+
+loc = 0.0
+    scale = 1.0
+    print *, uniform_cdf(0.5, loc, scale)             ! a cumulative at 0.5 in [0,1)
+    print *, uniform_cdf(0.7, -1.0, 2.0)              ! a cumulative at 0.7 in [-1,1)
+    x = reshape(uniform(loc, scale, 12),[3,4])        ! uniform random variates array in [0,1)
+    print *, uniform_cdf(x, loc, scale)               ! cumulative array in [0,1)
```
jvdp1, line 184:
```suggestion
    print *, uniform_cdf(0.5, loc, scale)             ! a cumulative at 0.5 in [0,1]
    print *, uniform_cdf(0.7, -1.0, 2.0)              ! a cumulative at 0.7 in [-1,1]
    x = reshape(uniform(loc, scale, 12),[3,4])        ! uniform random variates array in [0,1]
    print *, uniform_cdf(x, loc, scale)               ! cumulative array in [0,1]
```

jvdp1 (commented):
Thank you for this PR. Here are some comments on the specs.

Since there are mainly 3 very similar (in ternms of API) parts (`uniform`, `norm`, and `binomial`), I am wondering if it would not be easier to submit/review 3 different small PRs instead of this large one. What is your opinion @certik @milancurcic @ivan-pi ?
Meanwhile I will first focus on the `uniform` part.

Jim-215-Fisher:
> Thank you for this PR. Here are some comments on the specs.
> 
> Since there are mainly 3 very similar (in ternms of API) parts (`uniform`, `norm`, and `binomial`), I am wondering if it would not be easier to submit/review 3 different small PRs instead of this large one. What is your opinion @certik @milancurcic @ivan-pi ?
> Meanwhile I will first focus on the `uniform` part.

These three are just the initial part of whole statistical distributions, there are more distributions coming. All of them have similar API.

doc/specs/stdlib_stats_distribution.md
Jim-215-Fisher: > Would it be possible to return an muti-dimension array (e.g., a ran-3 array), e.g., as in [Matlab](https://www.mathworks.com/help/stats/unifrnd.html)

Yes. the functions are elemental except the the random number generators, they should work on multi-dimension arrays.

doc/specs/stdlib_stats_distribution.md
Jim-215-Fisher: > FORD supports Latex: see [here](https://github.com/Fortran-FOSS-Programmers/ford/wiki/Writing-Documentation#latex-support) for more details.
> Something like (I didn't check the formula):

I tried FORD LaTex support, It seems not working for my formulae.

doc/specs/stdlib_stats_distribution.md
Jim-215-Fisher: done

doc/specs/stdlib_stats_distribution.md
jvdp1: Ok. Thanks for trying. I will try it later too. IMO it is preferable to not rely on a third party app.

src/stdlib_stats_distribution.fypp
```diff
@@ -0,0 +1,218 @@
...
+
+    public :: random_seed
+    public :: uniform_distribution_rvs, uniform_distribution_pdf, uniform_distribution_cdf
+    public :: normal_distribution_rvs, normal_distribution_pdf, normal_distribution_cdf
+    public :: binomial_distribution_rvs, binomial_distribution_pmf, binomial_distribution_cdf
+
+    interface random_seed
+    !! version: experimental
+    !!
+    !! Set or get seed value of probability distributions for psudorandom bumber generator
```
jvdp1, line 31:
```suggestion
    !! Set or get seed value of probability distributions for pseudorandom bumber generator 
```

doc/specs/stdlib_stats_distribution.md
jvdp1: It would be nice that the random number generator functions ("_rvs") also support arrays up to rank 7 or 15 (depending on the compiler). This could be easily generated with Fypp (maybe in a next PR?).

Jim-215-Fisher: random number generator function itself has no issue with array support, it is elemental. The problem is how to initiate the function call. Normally, one uses one set of parameters to get either one or an array of random numbers. This is how current implementation worked. If we want array support up to rank 15, then we need a mask array as an optional argument to the function call. If peoples like the idea, we can certainly implement it next round.

jvdp1: I was thinking to extend the rank-1 function to multiple ranks, e.g. with multiple loops. Therefore, there is no need for optional arguments (there would rank+1 different procedures). But, as you mentioned, we can think about that in a next round if desired.

peteroupc:
Just for your information you should use implementations avoiding floating-point arithmetic when possible (or at least make such implementations an option).  For example, there are implementations of the exponential distribution that avoid comparisons (von Neumann's method, for one) as well as exact samplers of the Binomial distribution (e.g., Farach-Colton and Tsai; the "Internal DLA" paper by Bringmann and others).  Even the Poisson distribution can be implemented without using the exponential function (e.g. "On Buffon Machines and Numbers").  See my section on "[Specific Non-Uniform Distributions](https://peteroupc.github.io/randomfunc.html#Specific_Non_Uniform_Distributions)" in "Randomization and Sampling Methods".

Jim-215-Fisher:
> Just for your information you should use algorithms avoiding floating-point arithmetic when possible (or at least make such implementations an option). For example, there are algorithms of the exponential distribution that use comparisons only (von Neumann's method, for one) as well as exact samplers of the Binomial distribution (e.g., Farach-Colton and Tsai, "Exact Sublinear Binomial Sampling"; the "Internal DLA" paper by Bringmann and others). Even the Poisson distribution can be sampled without using the exponential function (e.g. "On Buffon Machines and Numbers"). Perhaps the most familiar example of a distribution of this kind is the discrete (and continuous) normal distribution sampler by Karney. See also my section on "[Specific Non-Uniform Distributions](https://peteroupc.github.io/randomfunc.html#Specific_Non_Uniform_Distributions)" in "Randomization and Sampling Methods".
> 
> In fact, as the authors of "Exact Sublinear Binomial Sampling" found, BTPE can oversample the tail of a binomial distribution (or at least the GNU Scientific Library implementation of BTPE at the time of the paper can).

Thanks for the info. I will take a look of it.

Jim-215-Fisher:
I am going to update for the second round. Because of major change in module structure, should I initiate a new PR or just continue in this old one?

jvdp1:
> I am going to update for the second round. Because of major change in module structure, should I initiate a new PR or just continue in this old one?

Does the new changes contain the same info? If yes, I suggest to update this PR.

Jim-215-Fisher:
> > I am going to update for the second round. Because of major change in module structure, should I initiate a new PR or just continue in this old one?
> 
> Does the new changes contain the same info? If yes, I suggest to update this PR.

Instead of one big module plus one submodule structure, I split the statistical distributions into each individual module, which will be easier to maintain. Also, in the new update, it will contain more distributions and new function calls.

jvdp1:
> > > I am going to update for the second round. Because of major change in module structure, should I initiate a new PR or just continue in this old one?
> > 
> > 
> > Does the new changes contain the same info? If yes, I suggest to update this PR.
> 
> Instead of one big module plus one submodule structure, I split the statistical distributions into each individual module, which will be easier to maintain. Also, in the new update, it will contain more distributions and new function calls.

Thank you for the details. Following your description, I would propose that you submit one PR per module, if possible. It would be easier to review than a very long PR.

PR #243, opened 2020-10-30
Logger: support of new_line() in the output of the logger

jvdp1:
A proposition to support `new_line` properly when inserted in a message as, e.g.,:
```fortran
call global % log_message( 'Possible issues: ' // new_line('a') // &
 ' 1) First possible issue' // new_line('a') // &
 ' 2) Second possible issue')

### DIFF
--- a/src/stdlib_logger.f90
+++ b/src/stdlib_logger.f90
@@ -546,8 +546,7 @@ contains
 
             if ( self % max_width == 0 .or.                                         &
                 ( length <= self % max_width .and.                                  &
-                index( string(1:min(length, self % max_width)), new_line('a')) == 0 &
-                )) then
+                index( string(1:length), new_line('a')) == 0 ) ) then
                 write( unit, '(a)', err=999, iostat=iostat, iomsg=iomsg ) &
                     string(1:length)
                 remain = 0
--- a/src/stdlib_logger.f90
+++ b/src/stdlib_logger.f90
@@ -523,7 +523,7 @@ contains
         character(*), intent(in)       :: procedure_name
         character(*), intent(in)       :: col_indent
 
-        integer :: count, indent_len, index, iostat, length, remain
+        integer :: count, indent_len, index_, iostat, length, remain
         character(256) :: iomsg
 
         length = len_trim(string)
@@ -551,11 +551,14 @@ contains
                 return
             else
 
-                do index=self % max_width, 1, -1
-                    if ( string(index:index) == ' ' ) exit
-                end do
+                index_ = index( string(1:self % max_width), new_line('a'))
+                if ( index_ == 0 ) then
+                    do index_=self % max_width, 1, -1
+                        if ( string(index_:index_) == ' ' ) exit
+                    end do
+                end if
 
-                if ( index == 0 ) then
+                if ( index_ == 0 ) then
                     write( unit, '(a)', err=999, iostat=iostat, iomsg=iomsg ) &
                         string(1:self % max_width)
                     count = self % max_width
@@ -563,8 +566,8 @@ contains
                     return
                 else
                     write( unit, '(a)', err=999, iostat=iostat, iomsg=iomsg ) &
-                        string(1:index-1)
-                    count = index
+                        string(1:index_-1)
+                    count = index_
                     remain = length - count
                     return
                 end if
@@ -585,11 +588,15 @@ contains
                 return
             else
 
-                do index=count+self % max_width, count+1, -1
-                    if ( string(index:index) == ' ' ) exit
-                end do
+                index_ = count + index( string(count+1:count+self % max_width), &
+                    new_line('a'))
+                if(index_ == count) then
+                    do index_=count+self % max_width, count+1, -1
+                        if ( string(index_:index_) == ' ' ) exit
+                    end do
+                end if
 
-                if ( index == count ) then
+                if ( index_ == count ) then
                     write( unit, '(a)', err=999, iostat=iostat, iomsg=iomsg ) &
                         string(count+1:count+self % max_width)
                     count = count + self % max_width
@@ -597,8 +604,8 @@ contains
                     return
                 else
                     write( unit, '(a)', err=999, iostat=iostat, iomsg=iomsg ) &
-                        string(count+1:index-1)
-                    count = index
+                        string(count+1:index_-1)
+                    count = index_
                     remain = length - count
                     return
                 end if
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jvdp1** [comment]
I add @wclodius2 as a reviewer

**Comment by jvdp1** [comment]
> @jvdp1:
> 
> It looks like you did some revisions before adding me as a reviewer so I am missing some context. My idea, when I suggested using new_line, was to recursively insert new_line in a single character string at the points where we had breaks in write statements, and use that string for a single write to each file. This would have changed `format_output_string` from a procedure that performs writes to one that returns output string. You seem to be going for a more incremental approach.

All changes can be found [here](https://github.com/fortran-lang/stdlib/pull/243/files).

I am not sure to understand what was your approach. Sorry.

My idea was to lead the logger writes something like:
```
ERROR: Possibles reasons:
    1) first reason
    2) second reason
```
with 
```fortran
call global%log_error('Possible reasons'//new_line('a')//'1) first reason'//new_line('a')//'2) second reason')
```
Therefore, the user puts some formatting in the message.

Was it your aim too?

jvdp1:
Here is an additional example of what should be nice to achieve:
* code:
```fortran
    call global % log_message(&
    'Example of an issue: the new_line() is considered as a character:' //   &
    new_line('a')//                                                                                         &
    ' -1) first bullet point: this is a long sentence on purpose. It should'  //   &
   'be longer than 72 characters;'//new_line('a')                                     /   /&
    ' -2) second bullet point' ) 
```


* Here is the current output (without the proposed changes):
```
2020-11-01 17:18:03.241: Example of an issue: the new_line() is
    considered as a character:
 -1) first bullet point: this is a long
    sentence on purpose. It should be longer than 72 characters;
 -2)
    second bullet point
```

* Here is the output with the proposed changes:
```
2020-11-01 17:23:07.982: Example of an issue: the new_line() is
    considered as a character:
     -1) first bullet point: this is a long sentence on purpose. It
    should be longer than 72 characters;
     -2) second bullet point
```

wclodius2:
The output as shown looks very good. My one concern is what might happen if the output unit is asynchronous. In that case the output from different write statements to the same unit might become separated by writes from other parts of the code. This is not a new problem. It was inherent in my original approach. A way around that is to write the output initially to a single string, inserting `new_line` for the formatting, and then output that string with a single write to each output unit. It becomes a bit tricky with `new_line` already in the input string, but I think it is doable. Why don't you continue to make the changes you want to get the formatting for your inputs. Then, after you commit your changes to the main branch, I start a new branch to implement the single output string using your mods as a guide for the treatment of `new_line` in the input string.

jvdp1:
> My one concern is what might happen if the output unit is asynchronous. In that case the output from different write statements to the same unit might become separated by writes from other parts of the code. This is not a new problem. It was inherent in my original approach. A way around that is to write the output initially to a single string, inserting `new_line` for the formatting, and then output that string with a single write to each output unit. It becomes a bit tricky with `new_line` already in the input string, but I think it is doable.

This issue was indeed present before the proposed changes. Note that the `new_line` is already provided by the user in the string. Therefore, I guess it could be considered as a character (as any other ones).

> Why don't you continue to make the changes you want to get the formatting for your inputs. Then, after you commit your changes to the main branch, I start a new branch to implement the single output string using your mods as a guide for the treatment of `new_line` in the input string.

Nice proposition. I believe I finished my changes and it is ready for reviewers. I tested them in a real program with several scenarios, and I was satisfied with the output. Hopefully I covered all possible scenarios.

src/stdlib_logger.f90
```diff
@@ -544,27 +544,33 @@ subroutine format_output_string( self, unit, string, procedure_name, &
 
         subroutine format_first_line()
 
-            if ( length <= self % max_width .or. self % max_width == 0 ) then
+            if ( self % max_width == 0 .or. &
+                ( length <= self % max_width .and. &
+                index( string(1:min(length, self % max_width)), new_line('a')) == 0 &
+                )) then
```
wclodius2: As `length <= self % max_width`, `string(1:min(length, self % max_width))` is equivalent to `string(1:length)`.

src/tests/logger/test_stdlib_logger.f90
```diff
@@ -134,6 +134,14 @@ subroutine test_logging_configuration()
             module = 'N/A',                                                &
             procedure = 'TEST_STDLIB_LOGGER' )
 
+        call global % log_information( 'This message should be output ' // &
+            'to OUTPUT_UNIT, unlimited in width, not preceded by ' //      &
+            'a blank line, then by a time stamp, then by MODULE % ' //     &
+            'PROCEDURE, be prefixed by INFO. ' // new_line('a') //         &
+            'This is a new line of the same log message.',                  &
+            module = 'N/A',                                                &
+            procedure = 'TEST_STDLIB_LOGGER' )
```
wclodius2, line 143: I would align the one ampersand that sticks out.

wclodius2:
Outside of the two minor comments, the code looks good to me.

src/stdlib_logger.f90
jvdp1: Thank you. I changed it as suggested.

jvdp1:
> Outside of the two minor comments, the code looks good to me.

I will wait that other users review the proposed changes too.

14NGiestas (approved):
Neat change, since `length` must be less than `max_width` already.

milancurcic (approved):
Sorry for the delay, I think this is an improvement, thank you!

milancurcic:
Please merge if/when ready.

jvdp1:
> Please merge if/when ready.

Thanks for the final review. With 3 reviewers, I will now merge it.

jvdp1:
I used it mainly in one of our Fortran programs and was in overal pleased by this implementation.

> A way around that is to write the output initially to a single string, i

@wclodius2 are you still interested in implementing such a proposition?
While using the logger, I got the following situation:

```fortran
...some code...
call global%log_message('Initial value is: '//value2char(rdummy)
...some code...
call global%log_message('New value is: '//value2char(rdummy)
...some code...
```
It would have been nice to have something like:
```fortran
...some code...
call global%append_to_buffer_message('Initial value is: '//value2char(rdummy))
...some code...
call global%appent_to_buffer_message('New value is: '//value2char(rdummy))
...some code...
call global%log_message()   !write he buffer to file unit(s), and initialise the buffer to''
```
This is probably further than your initial thoughs

wclodius2:
Yes it is further than my initial thoughts, which was to make the code thread/asynchronous IO "safe". I could do what you want as part of my rewrite, but what you are proposing has some ambiguities. In the appended strings what would you want done with date/time, module/procedure? What should be the formatting of the appended string?

jvdp1:
> Yes it is further than my initial thoughts, which was to make the code thread/asynchronous IO "safe". I could do what you want as part of my rewrite, but what you are proposing has some ambiguities. In the appended strings what would you want done with date/time, module/procedure? What should be the formatting of the appended string?

I would expect that date/time/module/procedure woud be initialized once when the string is still empty. Additional appended strings would not contain this information. It might be no sense, and could be considered later if others are interested.

At this stage, I think it is more important to have a code thread/asynchronous IO safe.

PR #245, opened 2020-11-16
Logger safe async io

wclodius2:
Changes in four files:
* `doc/specs/index.md` - added `stdlib_logger.html` to the index;
* `doc/specs/stdlib_logger.md` - described the changes in `stdlib_logger.f90`;
* `src/tests/logger/test-stdlib_logger.f90` - added a blank after a comma in the output;
* `src/stdlib_logger.f90` - numerous changes
   - the code handles `new_line` calls in the message string properly (added by @jvdp1, but undocumented in `stdlib_logger.md`)
   - a character string buffer was added to `logger_type` and a`len_buffer` integer to track the number of active characters in the buffer;
   - the code replaced multiple calls to write statements with single calls with line breaks generated by inserting `new_line` in the output string, in order to make the output thread and asynchronous IO safe;
   - shortened a number of lines most of them 80+ characters long
   - changed `Rrea_only_error` to `read_only_error`

### DIFF
--- a/doc/specs/stdlib_logger.md
+++ b/doc/specs/stdlib_logger.md
@@ -34,6 +34,10 @@ The logger variables have the option to:
 * indent subsequent lines of the messages; and
 * format the text to fit within a maximum column width.
 
+While every effort has been made to make the code thread and
+asynchronous I/O safe, it is always best to have each process write to
+its own dedicated logger file.
+
 Note: Loggers of type `logger_type` normally report their messages to I/O
 units in the internal list termed `log_units`. However if `log_units`
 is empty then the messages go to the `output_unit` of the intrinsic
--- a/doc/specs/stdlib_logger.md
+++ b/doc/specs/stdlib_logger.md
@@ -75,14 +75,16 @@ significant events encountered during the execution of a program.
 
 ### Private attributes
 
-| Attribute        | Type          | Description                                     | Initial value
-|------------------|---------------|-------------------------------------------------|--------------
-| `add_blank_line` | Logical       | Flag to precede output with a blank line        | `.false.`
-| `indent_lines`   | Logical       | Flag to indent subsequent lines by four columns | `.true.`
-| `log_units`      | Integer array | List of I/O units used for output               | empty
-| `max_width`      | Integer       | Maximum column width of output                  | 0
-| `time_stamp`     | Logical       | Flag to precede output by a time stamp          | `.true.`
-| `units`          | Integer       | Count of the number of active output units      | 0
+| Attribute        | Type          | Description                                     | Initial value |
+|------------------|---------------|-------------------------------------------------|--------------|
+| `add_blank_line` | Logical       | Flag to precede output with a blank line        | `.false.`    |
+|`buffer`          | Character(:)  | Buffer to build output string                   | Unallocated  |
+| `indent_lines`   | Logical       | Flag to indent subsequent lines by four columns | `.true.`     |
+| `len_buffer`     | Integer       | Number of valid characters in buffer            | 0            |
+| `log_units`      | Integer array | List of I/O units used for output               | Unallocated  |
+| `max_width`      | Integer       | Maximum column width of output                  | 0            |
+| `time_stamp`     | Logical       | Flag to precede output by a time stamp          | `.true.`     |
+| `units`          | Integer       | Count of the number of active output units      | 0            |
 
 ## The `stdlib_logger` variable
 
@@ -285,7 +287,7 @@ Pure subroutine
 
 #### Arguments
 
-`self`: shall be a scalar variable of type `logger_type`. It is an
+`self`: shall be a scalar expression of type `logger_type`. It is an
 `intent(in)` argument. It shall be the logger whose configuration is reported.
 
 `add_blank_line` (optional): shall be a scalar default logical
@@ -415,11 +417,13 @@ Subroutine
 
 #### Arguments
 
-`self`: shall be a scalar expression of type `logger_type`. It is an
-`intent(in)` argument. It is the logger used to send the message.
+`self`: shall be a scalar variable of type `logger_type`. It is an
+`intent(inout)` argument. It is the logger used to send the message.
 
 `message`: shall be a scalar default character expression. It is an
-  `intent(in)` argument.
+`intent(in)` argument.
+
+* Note `message` may have embedded new_line calls. 
 
 `module` (optional): shall be a scalar default character
   expression. It is an `intent(in)` argument. It should be the name of
@@ -496,11 +500,13 @@ Subroutine
 
 #### Arguments
 
-`self`: shall be a scalar expression of type `logger_type`. It is an
-`intent(in)` argument. It is the logger used to send the message.
+`self`: shall be a scalar variable of type `logger_type`. It is an
+`intent(inout)` argument. It is the logger used to send the message.
 
 `message`: shall be a scalar default character expression. It is an
   `intent(in)` argument.
+
+* Note `message` may have embedded new_line calls. 
 
 `module` (optional): shall be a scalar default character
   expression. It is an `intent(in)` argument. It should be the name of
@@ -565,11 +571,13 @@ written.
 Subroutine
 
 #### Arguments
-`self`: shall be a scalar expression of type `logger_type`. It is an
-`intent(in)` argument. It is the logger used to send the message.
+`self`: shall be a scalar variable of type `logger_type`. It is an
+`intent(inout)` argument. It is the logger used to send the message.
 
 `message`: shall be a scalar default character expression. It is an
   `intent(in)` argument.
+
+* Note `message` may have embedded new_line calls. 
 
 `module` (optional): shall be a scalar default character
   expression. It is an `intent(in)` argument. It should be the name of
@@ -641,11 +649,13 @@ Subroutine
 
 #### Arguments
 
-`self`: shall be a scalar expression of type `logger_type`. It is an
-`intent(in)` argument. It is the logger used to send the message.
+`self`: shall be a scalar variable of type `logger_type`. It is an
+`intent(inout)` argument. It is the logger used to send the message.
 
 `message`: shall be a scalar default character expression. It is an
   `intent(in)` argument.
+
+* Note `message` may have embedded new_line calls. 
 
 `module` (optional): shall be a scalar default character
   expression. It is an `intent(in)` argument. It should be the name of
@@ -715,8 +725,8 @@ Subroutine
 
 #### Arguments
 
-`self`: shall be a scalar expression of type `logger_type`. It is an
-`intent(in)` argument. It is the logger used to send the message.
+`self`: shall be a scalar variable of type `logger_type`. It is an
+`intent(inout)` argument. It is the logger used to send the message.
 
 `line`: shall be a scalar default character expression. It is an
   `intent(in)` argument. It should be the line of text in which the
@@ -861,11 +871,13 @@ Subroutine
 
 #### Arguments
 
-`self`: shall be a scalar expression of type `logger_type`. It is an
-`intent(in)` argument. It is the logger used to send the message.
+`self`: shall be a scalar variable of type `logger_type`. It is an
+`intent(inout)` argument. It is the logger used to send the message.
 
 `message`: shall be a scalar default character expression. It is an
   `intent(in)` argument.
+
+* Note `message` may have embedded new_line calls. 
 
 `module`: (optional) shall be a scalar default character
   expression. It is an `intent(in)` argument. It should be the name of
@@ -924,7 +936,7 @@ Subroutine
 
 #### Arguments
 
-`self`: shall be a scalar expression of type `logger_type`. It is an
+`self`: shall be a scalar variable of type `logger_type`. It is an
 `intent(inout)` argument. It is the logger whose `log_units` is to be
 modified.
 
--- a/doc/specs/stdlib_logger.md
+++ b/doc/specs/stdlib_logger.md
@@ -34,9 +34,11 @@ The logger variables have the option to:
 * indent subsequent lines of the messages; and
 * format the text to fit within a maximum column width.
 
-While every effort has been made to make the code thread and
+While every effort has been made to make the code process and
 asynchronous I/O safe, it is always best to have each process write to
 its own dedicated logger file.
+For thread parallelism (e.g., with OpenMP), it is advised to put the
+logger call in a guarding region (e.g., in an OpenMP critical region).
 
 Note: Loggers of type `logger_type` normally report their messages to I/O
 units in the internal list termed `log_units`. However if `log_units`
--- a/doc/specs/stdlib_logger.md
+++ b/doc/specs/stdlib_logger.md
@@ -78,7 +78,7 @@ significant events encountered during the execution of a program.
 | Attribute        | Type          | Description                                     | Initial value |
 |------------------|---------------|-------------------------------------------------|--------------|
 | `add_blank_line` | Logical       | Flag to precede output with a blank line        | `.false.`    |
-|`buffer`          | Character(:)  | Buffer to build output string                   | Unallocated  |
+| `buffer`         | Character(:)  | Buffer to build output string                   | Unallocated  |
 | `indent_lines`   | Logical       | Flag to indent subsequent lines by four columns | `.true.`     |
 | `len_buffer`     | Integer       | Number of valid characters in buffer            | 0            |
 | `log_units`      | Integer array | List of I/O units used for output               | Unallocated  |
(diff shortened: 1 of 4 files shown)


milancurcic:
@14NGiestas if your fix for docs build works, can you submit it as a separate PR? Then we can merge it and rebase this PR to master so that the docs build.

14NGiestas:
> @14NGiestas if your fix for docs build works, can you submit it as a separate PR? Then we can merge it and rebase this PR to master so that the docs build.

It is already a separate PR #246. I'm just waiting the CI to finish.

src/stdlib_logger.f90
```diff
@@ -236,8 +239,8 @@ subroutine add_log_unit( self, unit, stat )
 !! version: experimental
 
 !! Adds `unit` to the log file units in `log_units`. `unit` must be an `open`
-!! file, of `form` `"formatted"`, with `"sequential"` `access`, and an `action` of
-!! `"write"` or `"readwrite"`, otherwise either `stat`, if preseent, has a
+!! file, of `form` `"formatted"`, with `"sequential"` `access`, and an `action`
+!! of `"write"` or `"readwrite"`, otherwise either `stat`, if preseent, has a
```
14NGiestas, line 243:
```suggestion
!! of `"write"` or `"readwrite"`, otherwise either `stat`, if present, has a
```

doc/specs/stdlib_logger.md
```diff
@@ -75,14 +75,16 @@ significant events encountered during the execution of a program.
...
-| `add_blank_line` | Logical       | Flag to precede output with a blank line        | `.false.`
-| `indent_lines`   | Logical       | Flag to indent subsequent lines by four columns | `.true.`
-| `log_units`      | Integer array | List of I/O units used for output               | empty
-| `max_width`      | Integer       | Maximum column width of output                  | 0
-| `time_stamp`     | Logical       | Flag to precede output by a time stamp          | `.true.`
-| `units`          | Integer       | Count of the number of active output units      | 0
+| Attribute        | Type          | Description                                     | Initial value |
+|------------------|---------------|-------------------------------------------------|--------------|
+| `add_blank_line` | Logical       | Flag to precede output with a blank line        | `.false.`    |
+|`buffer`          | Character(:)  | Buffer to build output string                   | Unallocated  |
```
jvdp1, line 81:
```suggestion
| `buffer`         | Character(:)  | Buffer to build output string                   | Unallocated  |
```

src/stdlib_logger.f90
```diff
@@ -1032,32 +1054,43 @@ subroutine log_message( self, message, module, procedure, prefix )
...
+                        iomsg=iomsg ) &
+                    self % buffer(1:self % len_buffer)
+            end if
         else
-            do unit=1, self % units
-                call write_log_message( self % log_units(unit) )
-            end do
+            if ( self % add_blank_line ) then
+                do unit=1, self % units
+                    write( output_unit, '(a)', err=999, iostat=iostat, &
```
jvdp1, line 1075:
```suggestion
                    write( self % log_units(unit), '(a)', err=999, iostat=iostat, &
```

src/stdlib_logger.f90
```diff
@@ -1032,32 +1054,43 @@ subroutine log_message( self, message, module, procedure, prefix )
...
+                        iomsg=iomsg ) new_line('a') // &
+                        self % buffer(1:self % len_buffer)
+                end do
+            else
+                do unit=1, self % units
+                    write( output_unit, '(a)', err=999, iostat=iostat, &
```
jvdp1, line 1081:
```suggestion
                    write( self % log_units(unit), '(a)', err=999, iostat=iostat, &
```

jvdp1 (commented):
Two small issues discovered while testing the changes...

src/stdlib_logger.f90
wclodius2: Do it!

src/stdlib_logger.f90
wclodius2: Do it!

src/stdlib_logger.f90
```diff
@@ -599,31 +610,57 @@ subroutine format_subsequent_line()
...
 
         subroutine indent_format_subsequent_line()
+            integer :: new_len_buffer
+            character(:), allocatable :: dummy
 
             if ( index( string(count+1:length), new_line('a')) == 0 .and. &
                 remain <= self % max_width - indent_len ) then
-                write( unit, '(a)', err=999, iostat=iostat, iomsg=iomsg ) &
-                    col_indent // string(count+1:length)
+                new_len_buffer = self % len_buffer + index_ - 1 &
```
jvdp1, line 652:
```suggestion
                new_len_buffer = self % len_buffer + length &
```

src/stdlib_logger.f90
```diff
@@ -599,31 +610,57 @@ subroutine format_subsequent_line()
...
+                    - count + new_len + indent_len
+                if ( new_len_buffer > len( self % buffer ) ) then
+                    allocate( character( 2*len( self % buffer ) ) :: dummy )
+                    dummy = self % buffer
+                    call move_alloc( dummy, self % buffer )
+                end if
+                self % buffer( self % len_buffer+1: &
+                    self % len_buffer + length - count + new_len &
+                    + indent_len ) = &
```
jvdp1, line 661:
```suggestion
                self % buffer( self % len_buffer+1:new_len_buffer ) = &
```

src/stdlib_logger.f90
```diff
@@ -599,31 +610,57 @@ subroutine format_subsequent_line()
...
+                    new_line('a') // col_indent // string(count+1:index_-1)
```
jvdp1, line 662:
```suggestion
                    new_line('a') // col_indent // string(count+1:length)
```

jvdp1 (commented):
Issue: the end of a string with `indent=.true.` was not printed correctly.

The 3 proposed suggestions should solve this issue properly.

src/stdlib_logger.f90
wclodius2: Do it

src/stdlib_logger.f90
```diff
@@ -758,46 +813,29 @@ subroutine log_error( self, message, module, procedure, stat, errmsg )
...
-                    call format_output_string( self, unit,          &
-                                               'With errmsg = "' // &
-                                               trim(errmsg) // '"', &
-                                               procedure_name,      &
-                                               '    ' )
-                end if
+        if ( present(errmsg) ) then
+            if ( len_trim(errmsg) > 0 ) then
+                suffix( len_trim(suffix)+1: ) = &
+                    new_line('a') // 'With errmsg = "' // trim(errmsg) // '"'
```
jvdp1, line 826: If the length of `errmsg` is > 256-15, then it cannot be contained in `suffix`. Would it not be preferable to make `suffix` allocatable such that it would be possible to adapt its length depending on the length of `errmsg`?

src/stdlib_logger.f90
```diff
@@ -910,46 +949,29 @@ subroutine log_io_error( self, message, module, procedure, iostat, &
...
+        call self % log_message( trim(message) // trim(suffix), &
+                                 module = module,               &
+                                 procedure = procedure,         &
+                                 prefix = 'I/O ERROR' )
 
-999         call handle_write_failure( unit, procedure_name, iostat, iomsg )
+        return
 
-        end subroutine write_log_io_error
+        unit = -999
```
jvdp1, line 973: Is this line valid (I mean it is just after `return` and just before `999`)?

wclodius2: It should be the statement with the label 999, and the statement currently with that label should follow it.

src/stdlib_logger.f90
wclodius2: I guess what we could do is define two character variables, say suffix and dummy, with suffix an allocatable len variable. We could size dummy large enough to hold the output up to this point, and allocate suffix with the string assignment `trim(dummy) // new_line('a') // 'With errmsg = "' // trim(errmsg) // '"'`.  If you want I could implement this while you take a break.

wclodius2: This will have to be done in two locations: `log_error` and `log_io_error`.

wclodius2:
I have committed and pushed fixes for the `unit = -999` and the potentially too short `suffix` in both `log_error` and `log_io_error`.

jvdp1 (approved):
The changes sound good to me. Thank you @wclodius2 for these improvements!
IMO, this PR could be merged.

milancurcic (approved):
Thank you @wclodius2, I tested the changes locally. 

Is it feasible to add a test that confirms that it's indeed async I/O safe? As I understand how this is implemented--write multi-lines into a character string, then write it to output in a single write statement--it should work as advertised. But a test program would help ensure this. What do you think?

Do you expect that this is async I/O safe also for SPMD programs using MPI or coarrays? Specifically, in a pattern where multiple processes are logging to the same file, or stdout. Granted, a safer programming pattern in this case would be for each process to write to its own dedicated logger file. Should we make note of this in the docs?

jvdp1:
> Do you expect that this is async I/O safe also for SPMD programs using MPI or coarrays? Specifically, in a pattern where multiple processes are logging to the same file, or stdout. Granted, a safer programming pattern in this case would be for each process to write to its own dedicated logger file. Should we make note of this in the docs?

Would an OpenMP do loop with a logger do the trick? I have a similar example in one of my codes, however I put the logger in a critical OMP section with the first implementation.

wclodius2:
While a test would be useful I don't know how to write it. Parallel programming was not in demand for my work. In particular, how do we ensure that asynchronous calls aren't accidentally synchronous? Do we want to special case the cmake compiler flags and Makefile.manual flags for `stdlib_logger.f90` and `test_stdlib_logger.f90`?

A note to the effect of "While every effort has been made to make the code thread and asynchronous I/O safe, it is always best to have each process write to its own dedicated logger file" might be useful.

On a related note, should I add an `asynchronous` argument to the subroutine `add_log_file`?

jvdp1:
> While a test would be useful I don't know how to write it. Parallel programming was not in demand for my work. In particular, how do we ensure that asynchronous calls aren't accidentally synchronous? Do we want to special case the cmake compiler flags and Makefile.manual flags for `stdlib_logger.f90` and `test_stdlib_logger.f90`?

Could such a test be satisfying, assuming that most compilers support basic OpenMP directives?

```fortran
...
type(logger_type) :: logger
character (len = 30) :: cdumy
...

call logger % add_log_file( 'log.txt')
call logger % configure(indent = .true., time_stamp = .true.)
....
!$omp parallel default(none) shared(logger)
!$omp do
do i = 1, 100
  !$ write(cdummy, '(i0)') omp_get_thread_num()
  !$omp atomic
  call logger % log_message( ' Message from thread '//trim(cdummy) )'
enddo
!$omp end do
!$omp end parallel
....
```
Such a test would be more about thread then asynchronous safety.
To "mimick" MPI, an array of loggers could be used, and each thread would used its own logger.
However, what/how should be tested to have a proper test?

> A note to the effect of "While every effort has been made to make the code thread and asynchronous I/O safe, it is always best to have each process write to its own dedicated logger file" might be useful.

IMO it is a good idea to add it in the specs, especially if we claim it should asynchornous/thread/process-safe.

> On a related note, should I add an `asynchronous` argument to the subroutine `add_log_file`?

It would be good indeed. Probably it is best to do it in another PR independent from this one.

milancurcic:
> A note to the effect of "While every effort has been made to make the code thread and asynchronous I/O safe, it is always best to have each process write to its own dedicated logger file" might be useful.

@wclodius2, I agree. Do you mind adding this note in the spec document where you find it most appropriate?

I think this will then be good to merge.

wclodius2:
I have added the sentence to `stdlib_logger.md` just before the note at the end of the introduction.

@jvdp1 I have no experience with Open MP. My gut feeling is that an additional test wouldn't hurt, but it should probably be written by you as a separate PR. Things to think about if you should decide to do it.

1. It appears that compiling with `gfortran` for open MP requires the compiler flag `-fopenmp`. Would that flag apply to the library as a whole or just the test code? 
2. Do you envision the code as a separate "application" or part of the current test case?
3. I find going through a hundred lines of output a bit daunting. Would ten lines be a sufficient test?

14NGiestas:
Does this PR handles access to the buffer from different threads? Because having a buffer string without guarding the critical region does not guarantee one thread will not overwrite the buffer while the other thread was writing. I need to review the code to better understand that, but I belive It's up to the user put the logger call in a guarded region (omp critical, for e.g.).

jvdp1:
@wclodius2 I will propose another test later, when this is merged.

> Does this PR handles access to the buffer from different threads? Because having a buffer string without guarding the critical region does not guarantee one thread will not overwrite the buffer while the other thread was writing. I need to review the code to better understand that, but I belive It's up to the user put the logger call in a guarded region (omp critical, for e.g.).

Good point. No, there is no guarding within `stdlib_logger` that different threads access to the buffer at the same time (that is why I set it in an atomic section in the example above). We could add orphaned OpenMP directives in `stdlib_logger`. However, as @14NGiestas , I would think it is up to the user to put the logger call in a guarded region. In that case, a sentence in the specs should be added for multi-thread programs.

Note that this issue should not apppear in MPI/coarray-type programs.

doc/specs/stdlib_logger.md
```diff
@@ -34,6 +34,10 @@ The logger variables have the option to:
 * indent subsequent lines of the messages; and
 * format the text to fit within a maximum column width.
 
+While every effort has been made to make the code thread and
+asynchronous I/O safe, it is always best to have each process write to
+its own dedicated logger file.
```
jvdp1, line 39:
```suggestion
While every effort has been made to make the code process and
asynchronous I/O safe, it is always best to have each process write to
its own dedicated logger file.
For thread parallelism (e.g., with OpenMP), it is advised to put the logger call in a guarding region (e.g., in an OpenMP critical region.
```

Related to [this](https://github.com/fortran-lang/stdlib/pull/245#issuecomment-732691050) and [this](https://github.com/fortran-lang/stdlib/pull/245#issuecomment-732718840) comment.

@wclodius2 @14NGiestas @milancurcic We usually speek about "process" when related to MPI process. Thread parallelism is related to OpenMP and Pthreads.

wclodius2: I have revised the discussion as suggested, adding a closing parenthesis at the end.

wclodius2:
@14NGiestas FWIW the string buffer is in a logger object, and not a module variable. With a little extra work it could be made a procedure variable.

wclodius2:
FWIW with my latest push five of six CI/Builds passed. The one that failed "CI / Build (macos-latest, 9) (pull_request)" failed in running `test_stdlib_bitset_large` with a file read error. It passes that test on my system with gfortran 10.2.

jvdp1:
> FWIW with my latest push five of six CI/Builds passed. The one that failed "CI / Build (macos-latest, 9) (pull_request)" failed in running `test_stdlib_bitset_large` with a file read error. It passes that test on my system with gfortran 10.2.

I tested it on my computer (Fedora) with gfortran 9.3.1, and there were no issues.
I rerun the tests, but they still failed.

wclodius2:
FWIW the Mac OS run that fails uses `gfortran 9.3.0`. It fails with an unexpected End-of-File, which I have trouble understanding as I would expect that error to occur only if there were a coding error that affects all runs. A glance at the code in `stdlib_bitsets_large.f90` and `test_stdlib_bitset_large.f90` doesn't show anything wrong with the pertinent sections of code. Further, CI / Build (ubuntu-latest, 9) also uses `gfortran 9.3.0` and passes the tests. Maybe it is a bad disk sector involving the source code. I hate boggles!

jvdp1:
> FWIW the Mac OS run that fails uses `gfortran 9.3.0`. It fails with an unexpected End-of-File, which I have trouble understanding as I would expect that error to occur only if there were a coding error that affects all runs. A glance at the code in `stdlib_bitsets_large.f90` and `test_stdlib_bitset_large.f90` doesn't show anything wrong with the pertinent sections of code. Further, CI / Build (ubuntu-latest, 9) also uses `gfortran 9.3.0` and passes the tests. Maybe it is a bad disk sector involving the source code. I hate boggles!

I tested it again on my computer with different compilers and debug options. It always passed the tests! So no idea...

doc/specs/stdlib_logger.md
```diff
@@ -34,6 +34,12 @@ The logger variables have the option to:
 * indent subsequent lines of the messages; and
 * format the text to fit within a maximum column width.
 
+While every effort has been made to make the code process and
+asynchronous I/O safe, it is always best to have each process write to
+its own dedicated logger file.
+For thread parallelism (e.g., with OpenMP), it is advised to put the
+logger call in a guarding region (e.g., in an OpenMP critical region).
+
```
jvdp1, line 42: @milancurcic 
Are you satisfied with this note that covers multi-thread and multi-process programs? If yes, I am ok to merge this PR.

milancurcic: Yes, I'm happy with it so I'll go ahead and merge. Thank you all.

PR #246, opened 2020-11-16
Small fix to issue #245

14NGiestas:
It seems the command `set-env` is deprecated (since it is exploitable), so the Ford CI is failing.
I believe this small fix solves the issue: [GH Workflow environment files](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files)


milancurcic:
Great, thank you! Do you mind also submitting one to fpm, see [here](https://github.com/fortran-lang/fpm/issues/240)?

14NGiestas:
> Great, thank you! Do you mind also submitting one to fpm, see [here](https://github.com/fortran-lang/fpm/issues/240)?

Gladly :D

milancurcic:
This is a small technical change that seems to fix the problem, so I will go ahead and merge it.

PR #250, opened 2020-11-23
Fixes CI failing on MacOS

14NGiestas:
This fixes the MacOS CI issue, updating the gcc version to the right one.


milancurcic:
It seems to work, thank a lot! I will merge it.

Issue #254, opened 2020-11-29
stdlib_logger: proposition to add a logger for debug phases and levels among the different logs


jvdp1:
I am playing with `stdlib_logger` and I think it would be nice to have a subroutine for messages that should be printed only during debug phases.

#### Syntax
`call self % log_debug( message [,module, procedure, stat, errmsg] )`

### Log levels
Log levels are used for filtering the messages based on their level/severity.
Each subroutine `log_error`, `log_information`, `log_io_error`, `log_text_error`, `log_warning`, and the additional `log_debug` is associated with a level.

For example:

* `log_debug` associated with `debug_level`
* `log_information` associated with `information_level`
* `log_warning` associated with `warning_level`
* `log_error`, `log_io_error`, `log_text_error` associated with `error_level`

with `debug_level` < `information_level` < `warning_level` < `error_level`.

Since `log_message` is called by all other `log` subroutines, it is probably best to not assign a level to `log_message` IMO.

The level of the logger should be set with the method `configure`, e.g. `call self %configure( level =  information_level)`, meaning that all calls with a level < `information_level` would be ignored.

Such log level options are avaible in [Julia](https://docs.julialang.org/en/v1/stdlib/Logging/#Log-event-structure) or [Python](https://docs.python.org/3/howto/logging.html) loggers.

@wclodius2 , @milancurcic , @14NGiestas , @ivan-pi , @certik is that of interests? any thoughs, comments?

wclodius2:
A few comments. This is really two separate proposals" adding `log_debug` and adding `level`, and I think it would be best addressed by two separate PRs. Levels are also used by [FLIBS](http://flibs.sourceforge.net/) (`m_exception`). On the other hand, [Futility](https://github.com/CASL/Futility) (`exception_handler`) has separate flags for each "level", so that you can output "debug" while not outputting "information" or "warning". Testing the effects of `level` will be tricky. How do you verify what is not output? `log_text_error` is different from the other error levels. It indicates a problem not with the application, but rather with a file being read by the application. I suspect its output should be independent of `level`. Overall I am in favor of adding `log_debug`, but worry that adding `level` will be a bit tricky.

jvdp1:
> A few comments. This is really two separate proposals" adding `log_debug` and adding `level`, and I think it would be best addressed by two separate PRs.

I agree with you that it should be two separate PR.

Re: `log_debug`, its implementation is straighfoward. The main question is about its API. Should it be similar to e.g., `log_information`,

[...]

PR #256, opened 2020-12-04
debug_logger: add log_debug to logger

jvdp1:
Following the discussion in #254, I propose to add a subroutine call `log_debug` with the following API:

#### Description

Writes the string `message` to `self % log_units` with optional additional text.

#### Syntax

`call self % log_debug( message [, module, procedure ] )`

#### Behavior

If time stamps are active, a time stamp is written, followed
by `module` and `procedure` if present, and then
`message` is written with the prefix `'DEBUG: '`.

QUESTION: Is this API OK? Or should it be more similar to the API of [`log_error`](https://stdlib.fortran-lang.org/page/specs/stdlib_logger.html#status_5) with `stat` and `errmsg`?

### DIFF
--- a/doc/specs/stdlib_logger.md
+++ b/doc/specs/stdlib_logger.md
@@ -29,7 +29,7 @@ The logger variables have the option to:
   that prompted the log message;
 * follow a message with the `iostat` and `iomsg` of the I/O error
   report that prompted the log message;
-* label a message with one of `'INFO: '`, `'WARN: '`,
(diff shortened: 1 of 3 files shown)


wclodius2:
I think the API is good. In particular I would not add a stat or errmsg.

milancurcic (approved):
Looks okay to me as well.

jvdp1:
Thank you for the review. I'll merge it and submit the next PR mentioned in #254

PR #257, opened 2020-12-05
Improve check for F18 error stop

awvwgk:
This PR improves the Fortran 2018 error stop check by running the example and checking for a correctly reported return code from the program.

The current check gives a false positive result for `ifort` 21.1 BETA 20200827, which compiles F18 error stop syntax but does not propagate the error code correctly. Since this is probably a bug I reported this upstream as well:
https://community.intel.com/t5/Intel-Fortran-Compiler/Fortran-2018-error-stop/m-p/1234424#M153065

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,7 +35,7 @@ endif()
 # --- compiler feature checks
 include(CheckFortranSourceCompiles)
 include(CheckFortranSourceRuns)
-check_fortran_source_compiles("error stop i; end" f18errorstop SRC_EXT f90)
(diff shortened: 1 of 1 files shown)


Romendakil:
BTW, I checked our code (https://gitlab.tp.nt.uni-siegen.de/whizard/public) with ifort 21.1 BETA 20200827, and the new compiler has several problems, apparently with deallocation and garbage collection. I'm preparing a reproducer for at least one issue for Intel, but roughly 80% of our test suite fails with the new ifort, so I doubt that this is only one problem.

jvdp1 (approved):
Thanks!

milancurcic (approved):

PR #260, opened 2020-12-08
Add CI testing with Intel Fortran compiler

### LINKED ISSUE #249: Support for Intel oneAPI compilers
I just tried to setup a CI for the Intel oneAPI compilers for stdlib and noticed that `src/tests/io/test_savetxt.f90` is triggering an ICE in `ifort` (see https://github.com/awvwgk/fortran-stdlib/runs/1439256395) and `src/stdlib_ascii.f90` is triggering an ICE in the `ifx` compiler (see https://github.com/awvwgk/fortran-stdlib/runs/1439276984).

I have seen there are compile issues with PGI/NVHPC, NAG and flang already for stdlib. Does this mean stdlib currently only supports GCC?

awvwgk:
Read more on Intel oneAPI: https://fortran-lang.discourse.group/t/intel-releases-oneapi-toolkit-free-fortran-2018/471

This finally implements support for Intel oneAPI compilers to test stdlib.

Requires #257 
Fixes #249


certik:
This is amazing, thank you!!

certik:
It took only 1.5 min to install, and it overall finishes quickly. Very usable!

certik (approved):

awvwgk:
If you only require `ifort` it is an okay but still largish amount of package downloads. Note, if you also require MKL you won't get around caching the whole 10GB installation in `/opt/intel`.

jvdp1 (approved):
Great! Thank you!

certik:
The CI says `Need to get 393 MB of archives.`, so it is only 393 MB?

awvwgk:
Only `ifort`, `icc` and Intel MPI, if you add any of the Intel libraries to the mix, the download size quickly explodes (try installing `intel-oneapi-basekit` instead).

certik:
I tried to install it on my laptop running Ubuntu 18.04 using the exact commands in this PR. Everything installs fine. I load the variables. But then trying to compile any program gives:
```
$ ifort a.f90 
ifort: error #10105: ld: core dumped
ifort: warning #10102: unknown signal(-1408259920)
ifort: error #10106: Fatal error in ld, terminated by unknown
```

awvwgk:
That is a known issue, they broke binutils with the update from beta10 to the release version and supplied patches only for Ubuntu 20.04 and later. So you either have to downgrade manually to the beta10 version, which is broken for stdlib or upgrade your Ubuntu to 20.04 (sounds like fun).

milancurcic (approved):
Wonderful, thanks a lot. #257 is in so I'll go ahead and merge.

certik:
I am going to go ahead and merge this. We can then further improve in master.

certik:
@junghans, here is how we got Intel Fortran working, and it seems to work really nicely (small download, quick run).

awvwgk:
Quick follow-up, here are the release notes stating the binutils issue: https://software.intel.com/content/www/us/en/develop/articles/intel-oneapi-fortran-compiler-release-notes.html

PR #261, opened 2020-12-09
Proposition to add levels to stdlib_logger

jvdp1:
As discussed in #254, I propose to add a `level` option.
This allows to ignore logging messages: logging messages with a level less severe than the threshold will be ignored.

The threshold values for the different levels must still be discussed (e.g., especially for `text_error_level`).

WIP:
- [x] Update the specs.


src/stdlib_logger.f90
```diff
@@ -404,6 +428,8 @@ pure subroutine configuration( self, add_blank_line, indent, &
 !! A logical flag to add a preceding blank line
         logical, intent(out), optional              :: indent
 !! A logical flag to indent subsequent lines
+        integer, intent(out), optional              :: level
+!! The mimimum level for printing a message
```
jvdp1, line 432:
```suggestion
!! The minimum level for printing a message
```

jvdp1:
I think it is ok for review. @wclodius2 could you have another look,  please? Comments from other reviewers would be nice too ;)

jvdp1:
> While I would have `error_level` and `io_error_level` have the same severity as they both indicate problems with the application, I would have `text_error_level` have a different severity as it typically indicates a problem with the contents of an input file. I would have `text_error_level` have the higher severity. The rest of the edits I think look good.

Thanks @wclodius2 for your comment. I increased the severity level of `text_error_level`.

doc/specs/stdlib_logger.md
```diff
@@ -8,19 +8,20 @@ title: logger
...
 This module defines a derived type, its methods, a variable, and
-constants to be used for the reporting of errors and other
-information. The derived type, `logger_type`, is to be used to define
-both global and local logger variables. The `logger_type` methods serve
-to configure the loggers and use the logger variables to report
-messages to a variable specific list of I/O units termed
-`log_units`. The variable, `global_logger`, of type `logger_type`, is
-intended to serve as the default global logger. The constants serve as
+constants to be used for the reporting of errors, for filtering the 
+log messages, and other information. The derived type, `logger_type`,
```
ivan-pi, line 12: This comment might be off due to me not being a native English speaker. I am not sure if "filtering" is the most suitable word for what you are trying to describe. I would be happy to leave the concept of levels out in the first paragraph. In that case the sentence could finish as `"...to be used for the reporting of errors, displaying messages, and other information."`

doc/specs/stdlib_logger.md
```diff
@@ -8,19 +8,20 @@ title: logger
...
+variables to report messages to a variable specific list of I/O units
+termed `log_units`. The variable, `global_logger`, of type `logger_type`,
+is intended to serve as the default global logger. The constants serve as
 error flags returned by the optional integer `stat` argument.
 
 The logger variables have the option to:
 
 * change which units receive the log messages;
 * report which units receive the log messages;
+* change the level for filtering the log messages;
```
ivan-pi, line 24: Related to my previous comment, an alternative could be `"select which types of messages are logged."`

But I might be mistaken, and the current wording is clear.

ivan-pi (commented):
I have reviewed the changes introduced by levels. While I am not familiar with the full logger implementation, the changes are clear and I could not spot any errors. The test also appears to cover all of the changes.

Currently the levels are organized into a hierarchy. The other possibility would be to think of levels as switches. Assuming the processor uses a binary representation of integers, the levels can be set as:

```fortran
none_level = 0
text_error_level = shiftl(1,0) 
io_error_level = shiftl(1,1) ! or the error levels can have the same switch
error_level = shiftl(1,2)
warning_level = shiftl(1,3)
information_level = shiftl(1,4)
debug_level = shiftl(1,5)
all_level = text_error_level + io_error_level + error_level + warning_level + &
                 information_level + debug_level
```
The decimal values of the levels in the sequence are (0, 1, 2, 4, 8, 16, ...).

The benefit of this representation is that you can selectively turn on only certain combinations by addition of the switches (or subtracting from `all_levels`). On the other hand this might only be intuitive to users with sufficient knowledge of the underlying binary representation. 

I am not saying we need to do it this way, but thought it might be worth discussing what are the pros and cons. The decision we make here will serve as a pattern for similar usage cases appearing in other modules

jvdp1:
> I have reviewed the changes introduced by levels. While I am not familiar with the full logger implementation, the changes are clear and I could not spot any errors. The test also appears to cover all of the changes.
> 
> Currently the levels are organized into a hierarchy. The other possibility would be to think of levels as switches. Assuming the processor uses a binary representation of integers, the levels can be set as:
> 
> ```fortran
> none_level = 0
> text_error_level = shiftl(1,0) 
> io_error_level = shiftl(1,1) ! or the error levels can have the same switch
> error_level = shiftl(1,2)
> warning_level = shiftl(1,3)
> information_level = shiftl(1,4)
> debug_level = shiftl(1,5)
> all_level = text_error_level + io_error_level + error_level + warning_level + &
>                  information_level + debug_level
> ```
> 
> The decimal values of the levels in the sequence are (0, 1, 2, 4, 8, 16, ...).
> 
> The benefit of this representation is that you can selectively turn on only certain combinations by addition of the switches (or subtracting from `all_levels`). On the other hand this might only be intuitive to users with sufficient knowledge of the underlying binary representation.
> 
> I am not saying we need to do it this way, but thought it might be worth discussing what are the pros and cons. The decision we make here will serve as a pattern for similar usage cases appearing in other modules

Thank you for your review. Your approach is interesting.
I mainly follow the approach used in [Julia](https://docs.julialang.org/en/v1/stdlib/Logging/#Logging.LogLevel) and [Python](https://docs.python.org/3/howto/logging.html#logging-levels) loggers. 

One of my ideas was to potentially add an option through which the user could customize the different levels. Your approach could then be quite easy because the user could simply specify the rank of the level, instead of an abstract value.

doc/specs/stdlib_logger.md
jvdp1: "filtering" was used in the docs of both Julia and Python logging. I would be happy to change it to a more suitable word.
Anyway, I will change the sentence as you proposed.

doc/specs/stdlib_logger.md
jvdp1: I also like your sentence. However, I am also not a native English speaker. Any advices from a native English speaker would be nice ;)

wclodius2: Ivan-pi's suggestion is good.

doc/specs/stdlib_logger.md
jvdp1: Changed as proposed. The word `filtering` has been removed.

doc/specs/stdlib_logger.md
jvdp1: Changed as proposed.

jvdp1:
> > I have reviewed the changes introduced by levels. While I am not familiar with the full logger implementation, the changes are clear and I could not spot any errors. The test also appears to cover all of the changes.
> > Currently the levels are organized into a hierarchy. The other possibility would be to think of levels as switches. Assuming the processor uses a binary representation of integers, the levels can be set as:
> > ```fortran
> > none_level = 0
> > text_error_level = shiftl(1,0) 
> > io_error_level = shiftl(1,1) ! or the error levels can have the same switch
> > error_level = shiftl(1,2)
> > warning_level = shiftl(1,3)
> > information_level = shiftl(1,4)
> > debug_level = shiftl(1,5)
> > all_level = text_error_level + io_error_level + error_level + warning_level + &
> >                  information_level + debug_level
> > ```
> > 
> > 
> > The decimal values of the levels in the sequence are (0, 1, 2, 4, 8, 16, ...).
> > The benefit of this representation is that you can selectively turn on only certain combinations by addition of the switches (or subtracting from `all_levels`). On the other hand this might only be intuitive to users with sufficient knowledge of the underlying binary representation.
> > I am not saying we need to do it this way, but thought it might be worth discussing what are the pros and cons. The decision we make here will serve as a pattern for similar usage cases appearing in other modules
> 
> Thank you for your review. Your approach is interesting.
> I mainly follow the approach used in [Julia](https://docs.julialang.org/en/v1/stdlib/Logging/#Logging.LogLevel) and [Python](https://docs.python.org/3/howto/logging.html#logging-levels) loggers.
> 
> One of my ideas was to potentially add an option through which the user could customize the different levels. Your approach could then be quite easy because the user could simply specify the rank of the level, instead of an abstract value.

@certik @milancurcic @wclodius2 What do you think about @ivan-pi 's proposition?

milancurcic:
Thank you for this addition and sorry for the late review. I think levels will be very useful for larger applications and frameworks that use the logger under the hood but allow the user to control the verbosity.

I think @ivan-pi's idea is good as it would allow finer control into which logs exactly to allow and which to suppress. I think the "levels" approach is useful and should go forward as is. Then we should consider to also have a "switches" approach for finer control as a separate PR (but best to first discuss it in the issues).

milancurcic (approved):

jvdp1:
> I think @ivan-pi's idea is good as it would allow finer control into which logs exactly to allow and which to suppress. I think the "levels" approach is useful and should go forward as is. Then we should consider to also have a "switches" approach for finer control as a separate PR (but best to first discuss it in the issues).

Thank you @milancurcic for your comments. I'll merge this PR as is later today, and open an issue to discuss the 'switches" approach for finer control.

jvdp1:
I'll merge this PR. Thank you all for your comments.
I'll open an issues about @ivan-pi 's switches approach for finer control later.

PR #267, opened 2020-12-10
Use distinct file names

### LINKED ISSUE #264: Failure on macOS in master
There is an error in master on macOS 9:

https://github.com/fortran-lang/stdlib/runs/1532433788?check_suite_focus=true

I first noticed it in #263, but it couldn't have been caused by that PR. The first time it appeared seems to be after merging #260, but probably also unrelated.

We need to fix it.

certik:
This seems to avoid a filename race conditions on macOS.

Fixes #264.

### DIFF
--- a/src/tests/bitsets/test_stdlib_bitset_64.f90
+++ b/src/tests/bitsets/test_stdlib_bitset_64.f90
@@ -147,13 +147,13 @@ contains
             'write_bitset'
 
         call set2 % from_string( bitstring_33 )
-        open( newunit=unit, file='test.txt', status='replace', &
+        open( newunit=unit, file='test64_1.txt', status='replace', &
             form='formatted', action='write' )
         call set2 % write_bitset(unit)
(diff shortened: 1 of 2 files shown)


certik:
The actual problem seems to be that the `ctest` tests are executed in parallel, and the tests in `test_stdlib_bitset_64` and in `test_stdlib_bitset_large` where both reading/writing to `test.txt`. That is definitely a bug, that was fixed.

However, even after renaming the test file in `test_stdlib_bitset_large`, the test `test_stdlib_bitset_64` failed, even though I didn't find another test that would be writing to `test.txt`. I don't have an explanation for this. But I renamed the `test.txt` in `test_stdlib_bitset_64` also and that seemed to fix the problem.

certik:
I am going to merge this, as this seems to fix it. We will watch the master and see.

jvdp1:
Thank you. Hopefully it is fixed indeed!

Issue #268, opened 2020-12-17
Proposal for lists of strings


arjenmarkus:
Last tuesday, during the december call, it was mentioned that support for a list of strings might be useful. This is a straightforward proposal for such a feature in the standard library.

Fortran has supported variable-length strings since the 2003 standard, but it does not have a native type to handle collections of strings of different lengths. Such collections are quite useful though and the language allows us to define a derived type that can handle such collections.

This proposal considers the features that would be useful for this derived type, a list of strings. It does not prescribe the methods for implementing the type. Given the ease by which arrays are handled in Fortran, it is possible to use allocatable arrays, but perhaps a linked list may prove to be more efficient, at least in certain scenarios of use. Therefore the proposal concentrates on the usage.

A further limitation of the proposal: there is no provision for nested strings or for storing data with the string.

Methods for the list of strings:

* Method `insert` - insert a new string after the given index.
  Special index values: `head` and `end`, where `head` means insert the new string before the first one and `end` means insert it after the last one. Arithmetic with `end` is possible: `end-1` means insert before the last element and so on.

  Note: besides a string you can also insert another list or an ordinary array of strings (in the latter case all inserted strings will be of the same length as the array)

  Convenience method:
  `append` - same as insert with index = end

  Indices beyond the bounds of the list:
  Should they cause an error? Or should they be interpreted as insert at the start or at the end as if `head` or `end` were given? Or should the list simply grow to that length? This potentially causes holes.

* Method `delete` - delete a single string or a range of strings from the list. 
  `head` and `end` have similar meanings as with `insert`.

* Method `replace` - replace a string with a new value at the indicated location.

  Note: should we support replacement with multiple strings? My initial take at this is: no, if you want to do that, it can be done with a combination of `insert` and `delete`. The `replace` method is a convenient way to deal with individual strings.

* Method: `get` - get one element from the list and return it as a string

* Method: `range` - get all strings in a given range and return them as a new list.

* Method: `index` - find the index of the first occurrence of a string in the list. Returns 0 (zero) if there is no such string. May also look for the last.

* Method: `index_sub` - similar to `index` but rather than the entire string, a matching substring is sought for.

* Method: `destroy` - destroy the list

* Assignment: you can use lists in the context:

```fortran
  list1 = list2
```

  where list1 gets a copy of the entire list of strings.

[...]

PR #269, opened 2020-12-18
String list

arjenmarkus:
Here is a very preliminary implementation of a modules for handling lists of strings (see issue #268). The implementation is based on a derived type that holds an array of separate strings, but the suggestion made by @esterjo of using a single long string and an indexing array is well worth considering. Therefore consider this as a proposal for the behaviour and the interface.

### DIFF
--- a/doc/specs/stdlib_stringlist.md
+++ b/doc/specs/stdlib_stringlist.md
@@ -1,0 +1,467 @@
+---
+title: stringlist
+---
+# Lists of strings
+
+[TOC]
+
+## Introduction
+
+Fortran has supported variable-length strings since the 2003 standard,
+but it does not have a native type to handle collections of strings of
+different lengths. Such collections are quite useful though and the
+language allows us to define a derived type that can handle such
+collections.
+
+The `stdlib_stringlist` module defines a derived type that is capable of
+storing a list of strings and of manipulating them.
+
+Methods include:
+
+* inserting strings at a given position
+* replacing strings at a given position
+* deleting a single string or a range of strings
+* retrieving a string or a range of strings at a given position
+* finding the position of a particular string or a string which contains some substring
+* sorting the list
+
+## Positions in a list of strings
+
+The module implements what are effectively infinitely long lists: a position is
+represented as a positive integer, but there is no "out-of-bound" index. That is,
+the following piece of code will simply work:
+
+```fortran
+type(stringlist_type) :: list
+
+! Add two strings ...
+call list%insert( list_head, "The first string" )
+call list%insert( 20, "The last string" )
+
+write(*,*) 'The last:    ', list%get(list_end)
+write(*,*) 'Beyond that: ', list%get(30)
+```
+The special position `list_head` represents the position *before* the first element. Likewise,
+the special position `list_end` represents the position of the *last* element. You can
+use these positions to insert a string before the first string that is already in the
+list or to insert after the last string that has been inserted.
+
+If you specify a position beyond the last, the `list%get()` method simply returns an empty
+string.
+
+You can also specify *negative* positions, but they are interpreted as going back from the
+last inserted string. If you need the last but one string, you can do so innthis way:
+
+```fortran
+write(*,*) 'The last but onw: ', list%get(list_end-1)
+```
+
+So, it is possible to do simple arithmetic.
+
+*Note:* this does not work for the head of the list.
+
+## The derived type: stringlist_type
+
+### Status
+
+Experimental
+
+### Description
+
+The type holds a small number of components and gives access to a number of procedures,
+some of which are implemented as subroutines, others as functions or as operations.
+
+
+### Public `stringlist_type` methods
+
+The following methods are defined:
+
+Method               | Class      | Description
+---------------------|------------|------------
+[`delete`](./stdlib_stringlist.html#delete-delete_one_or_more_strings)                                                 | Subroutine | Delete one or more strings from the list
+[`destroy`](./stdlib_stringlist.html#destroy_destroy_all_strings_in_the_list)                                          | Subroutine | Destroy the contents of the list
+[`get`](./stdlib_stringlist.html#get-get_a_single_string_from_a_list)                                                  | Function   | Get a string from a particular position
+[`index`](./stdlib_stringlist.html#index-find_the_index_of_a_particular_string_in_the_list)                            | Function   | Find the index of a string in a list
+[`index_sub`](./stdlib_stringlist.html#index_sub-find_the_index_of_a_particular_string_containing_the_given_substring) | Function   | Find the index of a string containing a partilcar substring
+[`insert`](./stdlib_stringlist.html#insert-insert_one_or_more_strings_after_a_given_position)                          | Subroutine | Insert a string or a list after a given position
+[`length`](./stdlib_stringlist.html#length-return_the_length_of_the_list)                                              | Function   | Return the index of the last set position
+[`range`](./stdlib_stringlist.html#range-retrieve_a_range_of_string_from_the_list)                                     | Function   | Retrieve a range of strings from the list
+[`replace`](./stdlib_stringlist.html#replace-replace_one_or_more_strings_between_two_given_positions)                  | Subroutine | Replace one or more stringa between two positions
+[`sort`](./stdlib_stringlist.html#sort-return_a_sorted_list)                                                           | Function   | Sort the list and return the result as a new list
+[`=`](./stdlib_stringlist.html#assign-copy_the_contents_of_a_list)                                                     | Assignment | Copy a list
+[`//`](./stdlib_stringlist.html#//-concatenate_a_list_with_one_or_more_strings)                                        | Operation  | Concatenate a list with a string or concatenate two lists
+
+
+## Details of the methods
+
+### `delete` - delete one or more strings
+
+#### Status
+
+Experimental
+
+#### Description
+
+Delete one or more strings from the list via a given position or positions.
+
+#### Syntax
+
+`call list % [[stringlist_type(type):delete(bound)]]( first [, last] )`
+
+#### Class
+
+Subroutine
+
+#### Arguments
+
+`list`: the stringlist variable from which to delete one or more strings
+
+`first`: the index of the first string to be deleted
+
+`last` (optional): the index of the last string to be deleted. If left out, only one string is deleted.
+If the value is lower than that of `first`, the range is considered to be empty and nothing is deleted.
+
+
+### `destroy` - destroy all strings in the list
+
+#### Status
+
+Experimental
+
+#### Description
+
+Destroy the entire contents of the list. As the variable holding the list is simply a derived type, the variable
+itself is not destroyed.
+
+#### Syntax
+
+`call list % [[stringlist_type(type):destroy(bound)]]`
+
+#### Class
+
+Subroutine
+
+#### Arguments
+
+`list`: the stringlist variable from which to delete all strings
+
+
+### `get` - get a single string from the list
+
+#### Status
+
+Experimental
+
+#### Description
+
+Get the string at the given position.
+
+#### Syntax
+
+`string = list % [[stringlist_type(type):get(bound) ( idx )]]`
+
+#### Class
+
+Function
+
+#### Arguments
+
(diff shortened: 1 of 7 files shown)


src/stdlib_stringlist.f90
```diff
@@ -0,0 +1,148 @@
...
+    integer, intent(in)                       :: idx
+    character(len=*), intent(in)              :: string
+
+    integer                                   :: i
+    integer                                   :: idxnew
+    type(t_string)                            :: new_element
+    type(t_string), dimension(:), allocatable :: empty_strings
+
+    !
+    ! Initialise the list if necessary
```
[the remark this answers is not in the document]
milancurcic: Bad, bad review doggy!

jvdp1 (commented):
Nice progress. Here are a few minor comments.
It seems that the position of the elements starts at 0 until n-1? Did I understand it correctly?

milancurcic:
I agree with @jvdp1 regarding `t_string` -> `string_type` and `t_stringlist` -> `stringlist_type`.

The module should be called `stdlib_stringlist` (like the source file).

Would `stringarray` not be a more appropriate name for this derived type?

arjenmarkus:
I have no problem with the suggestions :). Just one remark: "initialise" is
quite acceptable according to a dictionary I checked (even though it is UK
spelling).

And I see why there are build failures: no documentation yet. Well, I will
work on that.

Op ma 21 dec. 2020 om 03:52 schreef Milan Curcic <notifications@github.com>:

jvdp1:
> And I see why there are build failures: no documentation yet. Well, I will work on that.

It is partly due to that. There is also the issue that gfortran is called now gfortran-10 in MacOS. This PR #270 solves this issue. So you might want to rebase your branch to pass this action.

milancurcic:
> Just one remark: "initialise" is quite acceptable according to a dictionary I checked (even though it is UK spelling).

It is, it's why I scolded the ReviewDog. :) I think we can just ignore it.

arjenmarkus:
Yes, my remark also reflects my personal grievances against those smart
aleck spelling checkings.

Op ma 21 dec. 2020 om 17:25 schreef Milan Curcic <notifications@github.com>:

src/stdlib_stringlist.f90
```diff
@@ -0,0 +1,487 @@
...
+    contains
+        procedure :: destroy   => destroy_list
+        procedure :: insert    => insert_string
+        procedure :: get       => get_string
+        procedure :: length    => length_list
+        procedure :: sort      => sort_list
+        procedure :: index     => index_of_string
+        procedure :: index_sub => index_of_substring
+        procedure :: delete    => delete_strings
+        procedure :: range     => range_list
```
jvdp1, line 44: First quick comment: shouldn't a `final` procedure (that potential calls the method `destroy`) be present?

doc/specs/stdlib_stringlist.md
```diff
@@ -0,0 +1,467 @@
...
+The special position `list_head` represents the position *before* the first element. Likewise,
+the special position `list_end` represents the position of the *last* element. You can
+use these positions to insert a string before the first string that is already in the
+list or to insert after the last string that has been inserted.
+
+If you specify a position beyond the last, the `list%get()` method simply returns an empty
+string.
+
+You can also specify *negative* positions, but they are interpreted as going back from the
+last inserted string. If you need the last but one string, you can do so innthis way:
```
awvwgk, line 53:
```suggestion
last inserted string. If you need the last but one string, you can do so in this way:
```

doc/specs/stdlib_stringlist.md
```diff
@@ -0,0 +1,467 @@
...
+
+! Add two strings ...
+call list%insert( list_head, "The first string" )
+call list%insert( 20, "The last string" )
+
+write(*,*) 'The last:    ', list%get(list_end)
+write(*,*) 'Beyond that: ', list%get(30)
+```
+The special position `list_head` represents the position *before* the first element. Likewise,
+the special position `list_end` represents the position of the *last* element. You can
```
awvwgk, line 45: This feels inconsistent to me. `list_head` represents an element outside of the list, while `list_end` represents an element in the list. I would prefer either having both special indices refer to elements inside or outside the list, but not mixed.

doc/specs/stdlib_stringlist.md
```diff
@@ -0,0 +1,467 @@
...
+list or to insert after the last string that has been inserted.
+
+If you specify a position beyond the last, the `list%get()` method simply returns an empty
+string.
+
+You can also specify *negative* positions, but they are interpreted as going back from the
+last inserted string. If you need the last but one string, you can do so innthis way:
+
+```fortran
+write(*,*) 'The last but onw: ', list%get(list_end-1)
```
awvwgk, line 56:
```suggestion
write(*,*) 'The last but one: ', list%get(list_end-1)
```

doc/specs/stdlib_stringlist.md
```diff
@@ -0,0 +1,467 @@
...
+
+`list`: the stringlist variable to retrieve a string from
+
+`string`: the string to be found in the list
+
+`back` (optional): logical argument indicating the first occurrence should be returned (`false`) or the last (`true`)
+
+#### Result value
+
+The result is either the index of the string in the list or -1 if the string was not found
```
awvwgk, line 207: Why is -1 chosen here? The intrinsics `index`/`scan`/`verify` return 0 on “failure” rather than -1.

doc/specs/stdlib_stringlist.md
```diff
@@ -0,0 +1,467 @@
...
+
+`list`: the stringlist variable to retrieve a string from
+
+`substring`: the substring in question
+
+`back` (optional): logical argument indicating the first occurrence should be returned (`false`) or the last (`true`)
+
+#### Result value
+
+The result is either the index of the string in the list or -1 if the string was not found
```
awvwgk, line 248: Same question on the return code as with `index`.

doc/specs/stdlib_stringlist.md
```diff
@@ -0,0 +1,467 @@
...
+
+#### Syntax
+
+`concatenatedlist = list // string`
+
+`concatenatedlist = string // list`
+
+#### Class
+
+Assignment
```
awvwgk, line 445:
```suggestion
Operator
```

awvwgk (commented):
I looked over the API documentations and left a few comments on the API design. Mainly the special positions `list_head` and `list_end` are inconsistent in my opinion.

arjenmarkus:
Thanks for the comments - I will give a more detailed reply later.

Op zo 3 jan. 2021 om 17:58 schreef Sebastian Ehlert <
notifications@github.com>:

doc/specs/stdlib_stringlist.md
awvwgk: One surprising consequence of the current choice of `list_head` (`0`) and `list_end` (`-1`) is that appending with `insert` is impossible:
```f90
call list%insert(list_end + 1, "last item")
```
Because it will evaluate to `list_head` and prepend instead.

awvwgk: One surprising consequence of the current choice of `list_head` (`0`) and `list_end` (`-1`) is that appending with `insert` is impossible:
```f90
call list%insert(list_end + 1, "last item")
```
Because it will evaluate to `list_head` and prepend instead.

arjenmarkus:
I have been thinking about the use of integers to indicate the head and the end of the list. Clearly this is a bad design choice, as there is no way to make it unambiguous. My current idea is to use instead a special derived type, something along the lines:

`type list_special_index_type`
`     logical :: head = .true.`
`     integer :: offset = 0 `
`end type`
with appropriate operations + and - for defining an offset. Then there can be no confusion about the meaning.

(Of course, a bit more work, but that is limited :)).

epagone:
> I have been thinking about the use of integers to indicate the head and the end of the list. Clearly this is a bad design choice, as there is no way to make it unambiguous. My current idea is to use instead a special derived type, something along the lines:
> 
> `type list_special_index_type`
> ` logical :: head = .true.`
> ` integer :: offset = 0`
> `end type`
> with appropriate operations + and - for defining an offset. Then there can be no confusion about the meaning.
> 
> (Of course, a bit more work, but that is limited :)).

I really like this solution: it seems to address at once most of the shortcomings that I have identified. I had a very limited amount of time to play with it but the current implementation seems to work already very well.

The aspects that I did not like much.
  - As already [commented](https://github.com/fortran-lang/stdlib/issues/268#issuecomment-752952937) in the related issue, I believe that the sorting method should not be part of this PR but it should be included in the `stdlib` sorting module and `stdlib_stringlist` should depend on it.
  - I would prefer to encapsulate into the derived type the control variables (e.g. `list_head`, `list_end` in the current implementation, `head` and `offset` in the future one) to not "pollute" the user namespace. It is a bit more verbose but I think it's a better choice.
  - I do not understand fully the need to hard code the `initial_size = 20` parameter to a somewhat arbitrary value that also "pollutes" the user namespace (see previous point).
  - I personally believe that the only way to let the user add an element to an already "initialised" list should be with the `append` method. Any other attempt to add new elements "out of bounds" should return an error. Quite strict, I know, but I think it's safe.

Thanks @arjenmarkus  for the doing this!

src/stdlib_stringlist.f90
```diff
@@ -0,0 +1,249 @@
...
+    integer, parameter, public :: list_head = 0
+    integer, parameter, public :: list_end  = -1
+    public :: t_stringlist
+
+
+    integer, parameter :: initial_size  = 20
+
+    type t_string
+        character(len=:), allocatable :: value
+    end type t_string
```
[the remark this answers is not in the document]
arjenmarkus: Done

src/stdlib_stringlist.f90
```diff
@@ -0,0 +1,249 @@
...
+    type(t_string), dimension(:), allocatable :: empty_strings
+
+    !
+    ! Initialise the list if necessary
+    !
+    if ( .not. allocated(list%string) ) then
+        allocate( list%string(initial_size) )
+        do i = 1,size(list%string)
+            list%string(i)%value = ''
+        enddo
```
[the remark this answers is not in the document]
arjenmarkus: Yes

src/stdlib_stringlist.f90
```diff
@@ -0,0 +1,249 @@
...
+
+! get_string --
+!     Get the string at a particular index
+!
+! Arguments:
+!     list                   The list of strings to retrieve the string from
+!     idx                    Index after which to insert the string
+!
+function get_string( list, idx )
+    class(t_stringlist), intent(inout) :: list
```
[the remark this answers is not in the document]
arjenmarkus: You're quite right

src/stdlib_stringlist.f90
arjenmarkus: It could actually be a dummy - allocatables get cleaned up automatically.

doc/specs/stdlib_stringlist.md
arjenmarkus: Yes, there is that ... I hope to solve this with special types rather than special values.

doc/specs/stdlib_stringlist.md
arjenmarkus: Yes, 0 is more consistent. Just corrected that

doc/specs/stdlib_stringlist.md
arjenmarkus: Indeed

doc/specs/stdlib_stringlist.md
arjenmarkus: The standard makes a difference between assignment and operator ... I can live with either, though.

ivan-pi:
@arjenmarkus I suspect you initiated this project in an old branch. The `stdlib_ascii` module is the old version you already modified (and is now in the master branch). The solution would be to rebase against the master branch.

arjenmarkus:
Is it? I recently did a rebase. Okay, I will have to check.

Op zo 10 jan. 2021 om 18:29 schreef Ivan Pribec <notifications@github.com>:

milancurcic:
Closing in favor of #311.

PR #270, opened 2020-12-18
Add gfortran 10 to CI

jvdp1:
To solve issues with Github actions (doc deployment) in #261 and #269.
Apparently `gfortran` (version 10) was renamed `gfortran-10` in MacOS. Therefore, I modified the different scripts to consider it correctly.
I also added `gfortran-10` to the current evaluated versions (7, 8, and 9).


certik (approved):
I think this change is fine. Thank you!

jvdp1:
There are several PRs with failed actions due to this issue. Therefore, I will merge it.

PR #273, opened 2020-12-21
Probability Distribution and Statistical Functions -- Normal Distribution Module 

Jim-215-Fisher:
This is the second round of probability distribution and statistical functions, continuing from #240. Since the whole modular structure has been changed, each distribution will have its own PR.
This PR contains normal distributions. Files uploaded are:
src/stdlib_stats_distribtuion_normal.fypp
src/CmakeLists.txt
src/Makefile.manual
src/tests/test_distribution_normal.f90
src/tests/CmakeLists.txt
src/tests/Makefile.manual
doc/specs/stdlib_stats_distribution_normal.md
doc/specs/index.md


ivan-pi:
Hello @Jim-215-Fisher ,

it's an impressive undertaking what you have here!

I just gave the files a quick glance and noticed certain calls to `real(val)` might not be what was desired. To preserve the precision of the real part of a complex number you need to use the full syntax: `real(cval,kind= ...)`. An alternative is to use `cval%re`. A discussion of this pitfall can be found [here](https://fortran-lang.discourse.group/t/the-counter-intuitive-rise-of-python-in-scientific-computing/469/38?u=ivanpribec).

Jim-215-Fisher:
> Hello @Jim-215-Fisher ,
> 
> it's an impressive undertaking what you have here!
> 
> I just gave the files a quick glance and noticed certain calls to `real(val)` might not be what was desired. To preserve the precision of the real part of a complex number you need to use the full syntax: `real(cval,kind= ...)`. An alternative is to use `cval%re`. A discussion of this pitfall can be found [here](https://fortran-lang.discourse.group/t/the-counter-intuitive-rise-of-python-in-scientific-computing/469/38?u=ivanpribec).

Thanks for the point. I will take a look of it. Originally, I was using cval%re and cval%im, but it did not pass the compilation on gfortran 7,8,9.

Jim-215-Fisher:
> Hello @Jim-215-Fisher ,
> 
> it's an impressive undertaking what you have here!
> 
> I just gave the files a quick glance and noticed certain calls to `real(val)` might not be what was desired. To preserve the precision of the real part of a complex number you need to use the full syntax: `real(cval,kind= ...)`. An alternative is to use `cval%re`. A discussion of this pitfall can be found [here](https://fortran-lang.discourse.group/t/the-counter-intuitive-rise-of-python-in-scientific-computing/469/38?u=ivanpribec).

I checked the Fortran 2018 Handbook, it states like this for real(A, [KIND]):
"if A is type complex and KIND is not present, the kind type parameter is the kind type parameter of A."
This is the result I desired in the code. Maybe I missed something.

ivan-pi:
> I checked the Fortran 2018 Handbook, it states like this for real(A, [KIND]):
> "if A is type complex and KIND is not present, the kind type parameter is the kind type parameter of A."
> This is the result I desired in the code. Maybe I missed something.

In that case consider my comment void. I was basing my comment of the comments I read on Fortran Discourse.

jvdp1:
Wrong button....I re-open this PR.

jvdp1:
Thank you @Jim-215-Fisher for the changes. I don't have additional major changes currently.
@ivan-pi @awvwgk @milancurcic @gareth-nx  could some of you review this PR to try to move it foward please?

src/tests/stats/test_distribution_normal.fypp
```diff
@@ -0,0 +1,235 @@
+
+#:include "common.fypp"
+#:set RC_KINDS_TYPES = REAL_KINDS_TYPES + CMPLX_KINDS_TYPES
+program test_distribution_normal
+    use stdlib_kinds
```
awvwgk, line 5: Can you make this import explicit?

src/tests/stats/test_distribution_normal.fypp
```diff
@@ -0,0 +1,235 @@
...
+    use stdlib_error, only : check
```
awvwgk, line 6: Might be worth to switch to using test-drive's `check` function for tests

awvwgk (approved):
Would be nice to switch the tests to the new framework before merging, but this can also happen in a separate PR.

Looks good to me.

doc/specs/stdlib_stats_distribution_normal.md
```diff
@@ -0,0 +1,262 @@
...
+
+### Status
+
+Experimental
+
+### Description
+
+The probability density function of the continuous normal distribution.
+
+$$f(x) = \frac{1}{\sigma \sqrt{2}} e^{-\frac{1}{2}(\frac{x-\mu}{\sigma})^{2}}$$
```
gareth-nx, line 107: Does this definition also cover the case of complex `x`? I am not immediately familiar with the normal density for a complex random variable, so it would be good to provide some reference.

doc/specs/stdlib_stats_distribution_normal.md
```diff
@@ -0,0 +1,262 @@
...
+
+### Status
+
+Experimental
+
+### Description
+
+Cumulative distribution function of the normal continuous distribution
+
+$$F(x)=\frac{1}{2}\left [ 1+erf(\frac{x-\mu}{\sigma \sqrt{2}}) \right ]$$
```
gareth-nx, line 190: Does this directly cover the complex case? In any case it would be good to provide a reference that covers that.

src/tests/stats/test_distribution_normal.fypp
Jim-215-Fisher: OK

src/tests/stats/test_distribution_normal.fypp
Jim-215-Fisher: Don't know what is test-drive. Where is it located?

doc/specs/stdlib_stats_distribution_normal.md
Jim-215-Fisher: I will elaborate the formula for the complex.

src/tests/stats/test_distribution_normal.fypp
jvdp1: See  e.g. [this example](https://github.com/fortran-lang/stdlib/blob/master/src/tests/stats/test_mean.fypp).
However, since the results of the different functions depend on the state of different global variables, it might not work until [this PR](https://github.com/fortran-lang/test-drive/issues/10) (or another similar) is merged.

for the moment, I propose to keep them as they are and to change them in another PR.

Jim-215-Fisher:
@jvdp1 @awvwgk @gareth-nx , the files have been updated. Especially documentation for complex variable case was changed.

src/stdlib_stats_distribution_normal.fypp
```diff
@@ -0,0 +1,332 @@
...
+
+
+
+    #:for k1, t1 in REAL_KINDS_TYPES
+    impure elemental function pdf_norm_${t1[0]}$${k1}$(x, loc, scale) result(res)
+    !
+    ! Normal distribution probability density function
+    !
+        ${t1}$, intent(in) :: x, loc, scale
+        real :: res
```
gareth-nx, line 277: Here I think `res` should be of type ${t1}$ ? I noticed differences between this function and the results from Rs `dnorm`, when usign double-precision in Fortran, that are likely due to this.

src/stdlib_stats_distribution_normal.fypp
```diff
@@ -0,0 +1,332 @@
...
+    end function pdf_norm_${t1[0]}$${k1}$
+
+    #:endfor
+
+
+
+    #:for k1, t1 in CMPLX_KINDS_TYPES
+    impure elemental function pdf_norm_${t1[0]}$${k1}$(x, loc, scale) result(res)
+        ${t1}$, intent(in) :: x, loc, scale
+        real :: res
```
gareth-nx, line 293: Again, should `res` be of type `${t1}$` ?

src/stdlib_stats_distribution_normal.fypp
```diff
@@ -0,0 +1,332 @@
...
+
+
+
+    #:for k1, t1 in REAL_KINDS_TYPES
+    impure elemental function cdf_norm_${t1[0]}$${k1}$(x, loc, scale) result(res)
+    !
+    ! Normal distribution cumulative distribution function
+    !
+        ${t1}$, intent(in) :: x, loc, scale
+        real :: res
```
gareth-nx, line 309: specify type of `res`.

src/stdlib_stats_distribution_normal.fypp
```diff
@@ -0,0 +1,332 @@
...
+    end function cdf_norm_${t1[0]}$${k1}$
+
+    #:endfor
+
+
+
+    #:for k1, t1 in CMPLX_KINDS_TYPES
+    impure elemental function cdf_norm_${t1[0]}$${k1}$(x, loc, scale) result(res)
+        ${t1}$, intent(in) :: x, loc, scale
+        real :: res
```
gareth-nx, line 324: as above, type of `res`

gareth-nx: Actually in this case, I believe res should be real (but with the same kind as the real part of `${t1}$`

src/stdlib_stats_distribution_normal.fypp
gareth-nx: No, I think `res` should be real but with the same kind as `${t1}$` (??).

src/stdlib_stats_distribution_normal.fypp
gareth-nx: Note: The same kind of bug also occurs in the equivalent `uniform` module

gareth-nx: Just confirming that when I make the obvious fix in my local branch of stdlib, the results agree with R to within `5e-016`, when checking 401 evenly spaced points from -10 to 10 standard deviations in double precision. Whereas the differences reach around `1e-08` otherwise.

gareth-nx: It would be good if the unit tests were able to catch such precision issues (easy mistake to make or introduce in future when editing).

src/stdlib_stats_distribution_normal.fypp
```diff
@@ -0,0 +1,332 @@
...
+    !
+    ! Normal distribution cumulative distribution function
+    !
+        ${t1}$, intent(in) :: x, loc, scale
+        real :: res
+        ${t1}$, parameter :: sqrt_2 = sqrt(2.0_${k1}$)
+
+        if(scale == 0._${k1}$) call error_stop("Error(cdf_norm): Normal"       &
+            //"distribution scale parameter must be non-zero")
+        res = (1.0_${k1}$ + erf((x - loc) / (scale * sqrt_2))) / 2.0_${k1}$
```
gareth-nx, line 314: While correct, this representation is prone to round-off error when the `erf` term is close to -1.0 - which in practice leads to greater relative errors when `x` is quite a few standard deviations below the mean.

To improve the results in that range of x, consider instead using the intrinsic `erfc`. I think the equation is `0.5*erfc(-(x-loc)/(scale*sqrt_2))`.

src/stdlib_stats_distribution_normal.fypp
Jim-215-Fisher: May be due to my limited probability knowledge, I purposely restrict the results of all probability function value to be single precision. To me, the **probability** of 0.1234567 (single precision) would be the same as **probability** of 0.12345678901234567 (double precision) practically, thought mathematically there are different. May be I am wrong.

Jim-215-Fisher: Should we provide all precision for probability?

gareth-nx: Yes, we should provide all precision for all probability. For these kids of functions, the need for precision is similar to other branches of applied math. 

While in some cases single-precision is fine, we also do things that need higher precision (e.g. suppose these functions appear in an expression that is then numerically differentiated).

I think the solution is to make `res` be of kind `real(${k1}$)` in all of the functions. Casual tests of `pdf_uniform` and `cdf_uniform` with this approach gave results very close to R's `dnorm` and `pnorm` in double precision.

src/tests/stats/test_distribution_normal.fypp
```diff
@@ -0,0 +1,235 @@
...
+        integer, parameter :: k = 5
+        integer :: i, n, seed, get
+        real :: res(3,5)
+        #:if t1[0] == "r"
+            #! for real type
+        real, parameter :: ans(15) =                                           &
+                         [0.215050772, 0.215050772, 0.215050772, 0.200537622,  &
+                          5.66161536E-02, 0.238986954, 0.265935957,0.262147546,&
+                          0.249866411, 3.98721099E-02, 0.265902370,0.161311597,&
+                          0.249177739, 0.237427220, 0.155696079]
```
gareth-nx, line 156: I think here, if the numbers are updated with quad-precision results, then the tests should be able to catch errors caused by loss of precision. They have a precision-dependent tolerance already - it might need tweaking if extended to higher precisions, but looks like it should just work.

src/tests/stats/test_distribution_normal.fypp
```diff
@@ -0,0 +1,235 @@
...
+        #:else
+            #! for complex type
+        real, parameter :: ans(15) =                                           &
+                         [0.129377320, 0.129377320,0.129377320,4.05915640E-02, &
+                          0.209143385,2.98881028E-02, 0.128679410, 0.177484736,&
+                          3.82205322E-02, 7.09915683E-02, 4.56126593E-02,      &
+                          6.57454133E-02,0.165161043,3.86104807E-02,0.196922958]
```
gareth-nx, line 163: As above, update RHS with quad-precision results, and use `real(${k1}$), parameter :: ans`

src/tests/stats/test_distribution_normal.fypp
```diff
@@ -0,0 +1,235 @@
...
+        integer :: i, n
+        integer :: seed, get
+        real :: res(3,5)
+        #:if t1[0] == "r"
+            #!for real type
+        real, parameter :: ans(15) =                                           &
+                         [7.50826299E-02, 7.50826299E-02, 7.50826299E-02,      &
+                          0.143119827, 0.241425425, 0.284345865, 0.233239830,  &
+                         0.341059506,0.353156865,6.81066737E-02,4.38792333E-02,&
+                          0.763679624, 0.363722175, 0.868187129, 0.626506805]
```
gareth-nx, line 203: As above, this can be modified to catch loss of precision in the results

src/tests/stats/test_distribution_normal.fypp
```diff
@@ -0,0 +1,235 @@
...
+        #:else
+            #! for complex type
+        real, parameter :: ans(15) =                                           &
+                         [1.07458131E-02, 1.07458131E-02, 1.07458131E-02,      &
+                          6.86483234E-02, 7.95486644E-02, 2.40523387E-02,      &
+                          3.35096754E-02,0.315778911,0.446311295, 0.102010213, &
+                          7.66918957E-02, 0.564691007, 0.708769500,            &
+                          6.40553832E-02, 5.39999157E-02]
```
gareth-nx, line 211: As above

src/stdlib_stats_distribution_normal.fypp
gareth-nx: Based on tests I ran just now, it should be: `real(${k1}$) :: res`

src/stdlib_stats_distribution_normal.fypp
gareth-nx: Based on tests I ran just now, it should be: `real(${k1}$) :: res`

src/stdlib_stats_distribution_normal.fypp
gareth-nx: Based on tests I ran just now, it should be: `real(${k1}$) :: res`

src/stdlib_stats_distribution_normal.fypp
gareth-nx: Just now I tried changing this to be 
```
        !res = (1.0_${k1}$ + erf((x - loc) / (scale * sqrt_2))) / 2.0_${k1}$
        res = erfc(-(x - loc) / (scale * sqrt_2)) / 2.0_${k1}$
```
All the tests passed, and a comparison with R showed much better agreement when `x` was around -8 standard deviations below `loc`, where we expect the previous formulation to have trouble.

src/stdlib_stats_distribution_normal.fypp
```diff
@@ -0,0 +1,332 @@
...
+    !
+    ! Normal distribution probability density function
+    !
+        ${t1}$, intent(in) :: x, loc, scale
+        real :: res
+        ${t1}$, parameter :: sqrt_2_pi = sqrt(2.0_${k1}$ * acos(-1.0_${k1}$))
+
+        if(scale == 0._${k1}$) call error_stop("Error(pdf_norm): Normal"       &
+            //"distribution scale parameter must be non-zero")
+        res = exp(- 0.5_${k1}$ * (x - loc) * (x - loc) / (scale * scale)) /    &
```
gareth-nx, line 282: Here, to reduce the risk of numerical problems when `x, loc, scale` are all very large magnitude, perhaps it would be better to write this as `exp(- 0.5_${k1}$ * ((x - loc)/scale) * ((x - loc) /scale))) `

src/stdlib_stats_distribution_normal.fypp
Jim-215-Fisher: So here we got a problem. The normal distribution is based on Marsaglia's fast ziggurat method which use double precision for the ziggurat table. It would be OK for the single and double precision probability but quardruple precision probability. For quardruple precision probability, any number beyond 16 significant figures will be meaningless.

gareth-nx: I see, thanks. In the first instance, I think it's OK if the random number generator does not give quad-precision accuracy - we can worry about that later if it's important. But for the pdf/cdf functions, we should be able to get full precision without any problems.

gareth-nx: But please add a statement in the specs about the precision of the random numbers being fundamentally limited to double precision by the algorithm (even if results are provided for quad precision).

src/stdlib_stats_distribution_normal.fypp
Jim-215-Fisher: Thanks.

src/stdlib_stats_distribution_normal.fypp
```diff
@@ -0,0 +1,332 @@
...
+module stdlib_stats_distribution_normal
+    use stdlib_kinds, only : sp, dp, xdp, qp, int32
+    use stdlib_error, only : error_stop
+    use stdlib_random, only : dist_rand
+    use stdlib_stats_distribution_uniform, only : uni=>rvs_uniform
+
+    implicit none
+    private
+
+    real(dp), parameter  :: HALF = 0.5_dp, ONE = 1.0_dp, TWO = 2.0_dp
```
gareth-nx, line 12: Here I suggest you add a comment `! These variables are only used in the random number generation, which is fundamentally double precision`.

doc/specs/stdlib_stats_distribution_normal.md
```diff
@@ -0,0 +1,272 @@
...
+
+`loc`: has `intent(in)` and is a scalar of type `real` or `complex`.
+
+`scale`: has `intent(in)` and is a scalar of type `real` or `complex`.
+
+All three arguments must have the same type.
+
+### Return value
+
+The result is a scalar or an array, with a shape conformable to arguments, of type `real`.
```
gareth-nx, line 222: Can be updated now that more precisions are supported.

doc/specs/stdlib_stats_distribution_normal.md
```diff
@@ -0,0 +1,272 @@
...
```
gareth-nx, line 135: Can be updated now that more precisions are supported

doc/specs/stdlib_stats_distribution_normal.md
```diff
@@ -0,0 +1,262 @@
...
+
+`loc`: optional argument has `intent(in)` and is a scalar of type `real` or `complex`.
+
+`scale`: optional argument has `intent(in)` and is a scalar of type `real` or `complex`.
+
+`loc` and `scale` arguments must have the same type.
+
+### Return value
+
+The result is a scalar or rank one array, with a size of `array_size`, of type `real` or `complex`.
```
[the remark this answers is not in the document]
gareth-nx: I suggest also mentioning that the internal computations produce double precision random numbers, so higher precision outputs are simply type-cast from these and will not have full-precision accuracy.

doc/specs/stdlib_stats_distribution_normal.md
```diff
@@ -0,0 +1,262 @@
...
+
+### Status
+
+Experimental
+
+### Description
+
+A normal continuous random variate distribution, also known as Gaussian, or Gauss or Laplace-Gauss distribution. The location `loc` specifies the mean or expectation. The `scale` specifies the standard deviation. 
+
+Without argument the function returns a standard normal distributed random variate N(0,1).
```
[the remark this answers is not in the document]
gareth-nx: I believe more precisions are supported now, and this can be updated (?).

Jim-215-Fisher:
@jvdp1 @awvwgk @gareth-nx All precision for probability were added according to @gareth-nx . Once this PR is done, I will set up a new PR for uniform distribution to correct similar problems.

gareth-nx (commented):
Thankyou! This is a very important contribution to stdlib.

src/stdlib_stats_distribution_normal.fypp
gareth-nx: @Jim-215-Fisher what do you think about this change?

Jim-215-Fisher: Sorry, did not see it.

jvdp1:
@Jim-215-Fisher @gareth-nx  is this PR ready for merging?

gareth-nx:
@jvdp1 @Jim-215-Fisher I think it just needs some minor documentation fixes mentioned above (or, a comment about why they are not needed -- I haven't looked in detail since reviewing). I was intending to approve once those points were addressed.

Jim-215-Fisher:
> @Jim-215-Fisher @gareth-nx is this PR ready for merging?

I think it is ready.

Jim-215-Fisher:
> @jvdp1 @Jim-215-Fisher I think it just needs some minor documentation fixes mentioned above (or, a comment about why they are not needed -- I haven't looked in detail since reviewing). I was intending to approve once those points were addressed.

I don't think we mention precision in documentation though. By default, all precision are supported.

gareth-nx:
> > @jvdp1 @Jim-215-Fisher I think it just needs some minor documentation fixes mentioned above (or, a comment about why they are not needed -- I haven't looked in detail since reviewing). I was intending to approve once those points were addressed.
> 
> I don't think we mention precision in documentation though. By default, all precision are supported.

Ahh, I see. I am getting confused between type and kind in Fortran. 

So at the moment the documentation indicates the output type, but not the kind, and it is assumed to be implied that all kinds are supported.

I think it would be nicer to be more explicit about this. The same point was made in [this review comment](https://github.com/fortran-lang/stdlib/pull/273/files/e786bd501fb6cdfac80fc48512081550eb93ae85#r725516236) by @jvdp1 on October 10. @Jim-215-Fisher would you mind to use phrasing like this when describing the output variables?

Jim-215-Fisher:
> > > @jvdp1 @Jim-215-Fisher I think it just needs some minor documentation fixes mentioned above (or, a comment about why they are not needed -- I haven't looked in detail since reviewing). I was intending to approve once those points were addressed.
> > 
> > 
> > I don't think we mention precision in documentation though. By default, all precision are supported.
> 
> Ahh, I see. I am getting confused between type and kind in Fortran.
> 
> So at the moment the documentation indicates the output type, but not the kind, and it is assumed to be implied that all kinds are supported.
> 
> I think it would be nicer to be more explicit about this. The same point was made in [this review comment](https://github.com/fortran-lang/stdlib/pull/273/files/e786bd501fb6cdfac80fc48512081550eb93ae85#r725516236) by @jvdp1 on October 10. @Jim-215-Fisher would you mind to use phrasing like this when describing the output variables?

done.

gareth-nx (approved):
Many thanks!

gareth-nx:
@jvdp1 This is ready to merge from my view.

jvdp1 (approved):
LGTM. Thank you @Jim-215-Fisher for this and to all reviewers. i'll merge it.

PR #284, opened 2020-12-29
Fix subproject build

awvwgk:
Required changes to use stdlib as subproject in CMake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,7 @@
 cmake_minimum_required(VERSION 3.14.0)
 project(stdlib Fortran)
 enable_testing()
+get_directory_property(IS_SUBPROJECT PARENT_DIRECTORY)
 
 include(${PROJECT_SOURCE_DIR}/cmake/stdlib.cmake)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,6 @@
 cmake_minimum_required(VERSION 3.14.0)
 project(stdlib Fortran)
 enable_testing()
-get_directory_property(IS_SUBPROJECT PARENT_DIRECTORY)
 
 include(${PROJECT_SOURCE_DIR}/cmake/stdlib.cmake)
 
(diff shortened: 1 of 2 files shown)


jvdp1 (approved):
Thank you @awvwgk for these improvements. Unfortunately my knowledge in CMake is really basic. So I am not sure to be the appropriate reviewer for these changes ;)
@certik @milancurcic Could you have a look, please?

For me (as far as I can check), it is good to be merge as is.

awvwgk:
I created a repo to ease the review process of this PR: https://github.com/awvwgk/stdlib-cmake-example
You can check the `subprojects/CMakeLists.txt` and switch between my fork and this repo to investigate the difference for CMake integration.

milancurcic:
I'm a true CMake n00b so I can't comment on the code. It seems good to me.

But I played with the stdlib-cmake-example and it works like a charm. Actually, CMake fetching stdlib and using it as a dependency felt like magic because I never did anything like that with CMake. So I think this will significantly lower the bar for using stdlib for some users.

I'm still a bit confused what this PR addresses exactly. If I'm not mistaken, the stdlib-cmake-example works as is with the current master. I think that is using the fetch method. Does this PR then allow using stdlib with the other method described in the stdlib-cmake-example README? Likely just me not understanding the CMake lingo.

Anyhow, whichever it is, I trust that this is good to go.

awvwgk:
> I'm still a bit confused what this PR addresses exactly. If I'm not mistaken, the stdlib-cmake-example works as is with the current master. I think that is using the fetch method. Does this PR then allow using stdlib with the other method described in the stdlib-cmake-example README? Likely just me not understanding the CMake lingo.

@milancurcic It does not, I currently let it fetch from my fork instead of the fortran-lang/stdlib repo (see https://github.com/awvwgk/stdlib-cmake-example/blob/623da353140578646daf125f6b6dd46981d9947a/subprojects/CMakeLists.txt#L11-L21). I force-pushed the repo a few times, the initial example fetching from `fortran-lang/stdlib` is still in the actions log at https://github.com/awvwgk/stdlib-cmake-example/runs/1636468980?check_suite_focus=true and has actually failed.

The repo README still explains it with the `fortran-lang/stdlib` repo, which will start to work once this PR is merged and I will than match the implementation with the docs again :wink:.

milancurcic:
@awvwgk Perfect, that's clear now. :)

awvwgk:
I'll go ahead and merge this PR, we can still revert or patch it later.

Issue #287, opened 2021-01-05
Using stdlib in meson


awvwgk:
Using stdlib is currently only possible with CMake by inclusion as subproject or by systemwide installation (either `/usr` or `/usr/local` prefix). The first approach excludes certain CMake setups and other build systems like meson, while the latter is not acceptable in case several compilers are used for development.

An usual approach to install packages in a local prefix is exporting CMake package files or pkg-config files, the latter makes the project discoverable by most non-CMake build systems.

Are there plans in the short-term to allow exporting of package files to make stdlib accessible for other projects?

milancurcic:
I don't know of any plans for this. I'm not too familiar with pkg-config files so I don't use them regularly. It would be helpful for people who use them to post here and express that this is a desired feature. But even if it's useful for only you, I'd say just go ahead because something like this increases ease of use without adding complexity for the end-user.

ivan-pi:
I used `pkg-config` recently to link with the Intel MKL libraries and found it easy enough to use. Essentially, using pkg-config requires writing a small file with some meta-data about the project, the required link flags and include folders of the library. 

One downside in case of Fortran is that the link flags might be compiler specific (in C this is less of a problem as `icc` borrows the `gcc` front-end). Intel MKL offers a "link line advisor" tool which helps generate the right flags that can be pasted into a pkg-config file (ideally you only do this once upon installation). The GNU Scientific library (and also it's Fortran wrapper FSGL) offer pkg-config files, meaning some potential users will be familiar with this already.

awvwgk:
This is working now just fine with meson. Therefore I will close this issue as resolved.

PR #288, opened 2021-01-06
Follow GNU install conventions

awvwgk:
Part 1/3 for #287, status: *optional*. If you do not like this approach we can do #287 without it.

*Changes:*

- remove hard coded subpaths for installation with CMake
- allow customization of install locations by [`GNUInstallDirs`](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) module
- enable project installation on CI testing (CMake targets are only generated on install)

*Reasoning:*

- allows customization of the include directory if it shouldn't go into the top-level include directory
- makes the life of packagers easier, because we don't have to patch the `CMakeLists.txt`, but can use command line arguments instead to apply our directory naming conventions


milancurcic (approved):

jvdp1 (approved):
LGTM. Thank you.

awvwgk:
Since this is a simple change I will go ahead and merge this PR with two approvals.

PR #294, opened 2021-01-08
Fix dependencies in manual Makefile

awvwgk:
This PR fixes the manual Makefile by including the correct dependencies. The manual Makefile build could fail in full parallel builds due to missing dependencies.

To test this use a slow computer and let make create as many build jobs as allowed by the dependency network with:

```sh
make -f Makefile.manual FYPPFLAGS=-DMAXRANK=4 -j
```

This error might be hard to catch in a sequential CI build due to the correct ordering of the source files in `SRC` variable.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -82,7 +82,7 @@ jobs:
     - name: Test manual makefiles
       if: contains(matrix.os, 'ubuntu') && contains(matrix.gcc_v, '10')
       run: |
-        make -f Makefile.manual FYPPFLAGS="-DMAXRANK=4"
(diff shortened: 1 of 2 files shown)


awvwgk:
Still fails occasionally for `stdlib_quadrature_trapzd.o` even if the dependency on `stdlib_quadrature.o` is explicitly specified in the Makefile.

jvdp1 (approved):
LGTM! Thank you for this improvement!

milancurcic (approved):
I tested it and it works. I will go ahead and merge. Thanks @awvwgk!

PR #300, opened 2021-01-16
Fix compilation error for NAG

awvwgk:
NAG detects a duplicated name for error_handler which is defined as module procedure as well as the name of the interface of said procedure:

    Error: src/stdlib_bitsets.f90, line 2131: ERROR_HANDLER previously declared as a generic procedure name
           detected at ERROR_HANDLER@(

See #108 for discussion on NAG.

### DIFF
--- a/src/stdlib_bitsets.fypp
+++ b/src/stdlib_bitsets.fypp
@@ -2006,7 +2006,7 @@ module stdlib_bitsets
 
     end interface operator(<=)
 
-    interface error_handler
+    interface
         module subroutine error_handler( message, error, status, &
             module, procedure )
             character(*), intent(in)           :: message
@@ -2015,7 +2015,7 @@ module stdlib_bitsets
             character(*), intent(in), optional :: module
             character(*), intent(in), optional :: procedure
         end subroutine error_handler
-    end interface error_handler
+    end interface
 
 contains
 


jvdp1:
Thanks @awvwgk  for this workaround. However, it a bug in the compiler, I am not sure if it is wise to introduce this workaround.

awvwgk:
I'm not sure if this is actually a compiler error, since it flags a name clash between an overloaded interface and a procedure name. Somebody with more detailed knowledge of the standard could probably help out here.

jvdp1:
> I'm not sure if this is actually a compiler error, since it flags a name clash between an overloaded interface and a procedure name. Somebody with more detailed knowledge of the standard could probably help out here.

I am not sure neither. I will just flag a few people: @certik @sblionel @milancurcic @everythingfunctional : any ideas/comments?

sblionel:
Could I see a small but complete example that shows the problem? I tried to figure it out from the Files Changed but am worried I might be missing something.

15.4.3.4.1p3 says "A generic name may be the same as any one of the procedure names in the generic interface, or the same as any accessible generic name." I haven't thought about this related to submodules.

awvwgk:
A minimal example reproducing the issue would be
```f90
module stdlib_bitsets
    implicit none
    private

    public :: error_handler

    interface error_handler
        module subroutine error_handler( message, error, status, &
            module, procedure )
            character(*), intent(in)           :: message
            integer, intent(in)                :: error
            integer, intent(out), optional     :: status
            character(*), intent(in), optional :: module
            character(*), intent(in), optional :: procedure
        end subroutine error_handler
    end interface error_handler

contains

    module subroutine error_handler( message, error, status, module, procedure )
        character(*), intent(in)           :: message
        integer, intent(in)                :: error
        integer, intent(out), optional     :: status
        character(*), intent(in), optional :: module
        character(*), intent(in), optional :: procedure
        ! <snip>
    end subroutine error_handler

end module stdlib_bitsets
```

Running `nagfor -c` yields:

```
NAG Fortran Compiler Release 7.0(Yurakucho) Build 7038
Error: example.f90, line 20: ERROR_HANDLER previously declared as a generic procedure name
       detected at ERROR_HANDLER@(
[NAG Fortran Compiler pass 1 error termination, 1 error]
```

sblionel:
Thanks. I don't see a problem with this code. ifort doesn't complain, but I can confirm a recent NAG does. I'll ask Malcolm about it.

sblionel:
Malcolm says, "Thanks for your bug report. I concur that this is a bug. It should be fixed in build 7041."

awvwgk:
I will close this PR since the issue is now fixed upstream. Thanks everybody.

jvdp1:
> Malcolm says, "Thanks for your bug report. I concur that this is a bug. It should be fixed in build 7041."

Thank you for this quick answer. Nice that it will be fixed.

PR #304, opened 2021-01-19
Add supported compilers MinGW 8, 9, 10

brocolis:
Add supported compilers MinGW 8, 9, 10 (x86_64 and i686) on Windows 10.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,8 +80,13 @@ GCC Fortran | 7, 8, 9, 10 | MacOS Catalina 10.15 | x86_64
 GCC Fortran | 8 | Windows Server 2019 | x86_64
 GCC Fortran (MSYS) | 10 | Windows Server 2019 | x86_64
 GCC Fortran (MinGW) | 10 | Windows Server 2019 | x86_64, i686
-GCC Fortran (MinGW) | 8, 9, 10 | Windows 10 | x86_64, i686
 Intel oneAPI classic | 2021.1 | Ubuntu 20.04 | x86_64
+
+The following combinations are known to work, but they are not tested in the CI:
+
+Name | Version | Platform | Architecture
+--- | --- | --- | ---
+GCC Fortran (MinGW) | 8.4.0, 9.3.0, 10.2.0 | Windows 10 | x86_64, i686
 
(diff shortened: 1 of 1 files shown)


milancurcic (approved):
Thanks for testing, change looks good to me.

README.md
```diff
@@ -80,6 +80,7 @@ GCC Fortran | 7, 8, 9, 10 | MacOS Catalina 10.15 | x86_64
 GCC Fortran | 8 | Windows Server 2019 | x86_64
 GCC Fortran (MSYS) | 10 | Windows Server 2019 | x86_64
 GCC Fortran (MinGW) | 10 | Windows Server 2019 | x86_64, i686
+GCC Fortran (MinGW) | 8, 9, 10 | Windows 10 | x86_64, i686
```
jvdp1, line 83: I propose to move this line an additional table, because these are not tested within the WindowsCI (if I am correct).
Therefore, if a new procedure is added to stdlib, these combinations OS-compiler will not tested for this new procedure.

awvwgk:
@brocolis Did you install the GCC versions from http://winlibs.com/? We might be able to download the corresponding GCC versions from https://github.com/brechtsanders/winlibs_mingw/releases/ and actually test them in the CI to keep supporting a wide range of compilers on Windows as well.

awvwgk:
Thanks for the clarification. Using this approach would be taxing on the CI runners, if you find a good way to (semi-)automatically install different GCC versions on Windows, let me know, I would be interested to extend the stdlib CI testing for those at some point.

brocolis:
Hello. I'm not familiar with CI setup. Anyway, I'll upload the binaries/build scripts to the internet in the next few days.

brocolis:
> Hello. I'm not familiar with CI setup. Anyway, I'll upload the binaries/build scripts to the internet in the next few days.

OK, mingw/gfortran-8.4.0, 9.3.0 and 10.2.0 for Windows (i686, x64) are now available at https://fortran.rocks/all.html

jvdp1 (approved):
Sorry for the delay.
I approve this change, as it was suggested.

PR #310, opened 2021-02-04
Extend stdlib_ascii module for handling character variables

awvwgk:
This PR implements some character handling routines discussed in #69, but purposely does not put them into the `stdlib_string` namespace. ~~There is one little caveat with this implementation due to a namespace collision with the `stdlib_ascii` module, which might require reconsidering the naming of the functions implemented here.~~

- extend `to_lower` function to work on character strings
- extend `to_upper` function to work on character strings
- implement `to_title` function
- implement `reverse` function

The change in `stdlib_ascii` might be breaking ABI compatibility with the default branch.

### DIFF
--- a/doc/specs/stdlib_ascii.md
+++ b/doc/specs/stdlib_ascii.md
@@ -34,9 +34,7 @@ Converts input character variable to all lowercase.
 
 #### Syntax
 
-```f90
-res = [stdlib_asciii(module):to_lower(function)]](string)
-```
+`res = [[stdlib_asciii(module):to_lower(function)]] (string)`
 
 #### Class
 
@@ -72,9 +70,7 @@ Converts input character variable to all uppercase.
 
 #### Syntax
 
-```fortran
-res = [[stdlib_ascii(module):to_upper(function)]](string)
-```
+`res = [[stdlib_ascii(module):to_upper(function)]] (string)`
 
 #### Class
 
@@ -112,9 +108,7 @@ All following characters will become lowercase.
 
 #### Syntax
 
-```
-res = [[stdlib_ascii(module):to_title(interface)]](string)
-```
+`res = [[stdlib_ascii(module):to_title(interface)]] (string)`
 
 #### Class
 
@@ -152,9 +146,7 @@ Reverses the order of all characters in the input character type.
 
 #### Syntax
 
-```f90
-res = [[stdlib_ascii(module):reverse(function)]](string)
-```
+`res = [[stdlib_ascii(module):reverse(function)]] (string)`
 
 #### Class
 
@@ -176,3 +168,4 @@ program demo_reverse
     implicit none
     print'(a)', reverse("Hello, World!") ! returns "!dlroW ,olleH"
 end program demo_reverse
+```
--- a/doc/specs/stdlib_ascii.md
+++ b/doc/specs/stdlib_ascii.md
@@ -38,7 +38,7 @@ Converts input character variable to all lowercase.
 
 #### Class
 
-Pure function.
+Elemental function.
 
 #### Argument
 
@@ -74,7 +74,7 @@ Converts input character variable to all uppercase.
 
 #### Class
 
-Pure function.
+Elemental function.
 
 #### Argument
 
@@ -112,7 +112,7 @@ The rest of the character sequence is transformed to lowercase.
 
 #### Class
 
-Pure function.
+Elemental function.
 
 #### Argument
 
@@ -150,7 +150,7 @@ Reverses the order of all characters in the input character type.
 
 #### Class
 
-Pure function.
+Elemental function.
 
 #### Argument
 
--- a/doc/specs/stdlib_ascii.md
+++ b/doc/specs/stdlib_ascii.md
@@ -34,7 +34,7 @@ Converts input character variable to all lowercase.
 
 #### Syntax
 
-`res = [[stdlib_asciii(module):to_lower(function)]] (string)`
+`res = [[stdlib_ascii(module):to_lower(function)]] (string)`
 
 #### Class
 
@@ -108,7 +108,7 @@ The rest of the character sequence is transformed to lowercase.
 
 #### Syntax
 
-`res = [[stdlib_ascii(module):to_title(interface)]] (string)`
+`res = [[stdlib_ascii(module):to_title(function)]] (string)`
 
 #### Class
 
--- a/doc/specs/stdlib_ascii.md
+++ b/doc/specs/stdlib_ascii.md
@@ -103,8 +103,8 @@ Experimental
 #### Description
 
 Returns a capitalized version of an input character variable.
-Only the first alphanumeric character is capitalized.
-All following characters will become lowercase.
+The first alphabetical character is transformed to uppercase unless it follows a numeral.
+The rest of the character sequence is transformed to lowercase.
 
 #### Syntax
 
--- a/doc/specs/stdlib_ascii.md
+++ b/doc/specs/stdlib_ascii.md
@@ -1,0 +1,149 @@
+---
+title: ASCII
+---
+
+# The `stdlib_ascii` module
+
+[TOC]
+
+## Introduction
+
+The `stdlib_ascii` module provides procedures for handling and manipulating
+intrinsic character variables and constants.
+
+
+## Constants provided by `stdlib_ascii`
+
+@note Specification of constants is currently incomplete.
+
+
+## Specification of the `stdlib_ascii` procedures
+
+@note Specification of procedures is currently incomplete.
+
+
+### `to_lower`
+
+#### Status
+
+Experimental
+
+#### Description
+
+Converts input character variable to all lowercase.
+
+#### Syntax
+
+```f90
+res = to_lower("HELLO!")
+! res == "hello!"
+```
+
+#### Class
+
+Pure function.
+
+#### Argument
+
+`string`: shall be an intrinsic character type. It is an `intent(in)` argument.
+
+#### Result value
+
+The result is an intrinsic character type of the same length as `string`.
+
+
+### `to_upper`
+
+#### Status
+
+Experimental
+
+#### Description
+
+Converts input character variable to all uppercase.
+
+#### Syntax
+
+```
+res = to_upper("hello!")
+! res == "HELLO!"
+```
+
+#### Class
+
+Pure function.
+
+#### Argument
+
+`string`: shall be an intrinsic character type. It is an `intent(in)` argument.
+
+#### Result value
+
+The result is an intrinsic character type of the same length as `string`.
+
+
+### `to_title`
+
+#### Status
+
+Experimental
+
+#### Description
+
+Returns capitalized version of input character variable.
+The first alphanumeric character is capitalized.
+
+#### Syntax
+
+```
+res = to_title("hello!")
+! res == "Hello!"
+res = to_title("'enquoted'")
+! res == "'Enquoted'"
+res = to_title("1st")
+! res == "1st"
+```
+
+#### Class
+
+Pure function.
+
+#### Argument
+
+`string`: shall be an intrinsic character type. It is an `intent(in)` argument.
+
+#### Result value
+
+The result is an intrinsic character type of the same length as `string`.
+
+
+### `reverse`
+
+#### Status
+
+Experimental
+
+#### Description
+
+Reverses the order of all characters in the input character type.
+
+#### Syntax
+
+```f90
+res = reverse("Hello, World!")
+! res == "!dlroW ,olleH"
+res = reverse(res)
+! res == "Hello, World!"
+```
+
+#### Class
+
+Pure function.
+
+#### Argument
+
+`string`: shall be an intrinsic character type. It is an `intent(in)` argument.
+
+#### Result value
+
+The result is an intrinsic character type of the same length as `string`.
--- a/doc/specs/stdlib_ascii.md
+++ b/doc/specs/stdlib_ascii.md
@@ -38,7 +38,7 @@ Converts input character variable to all lowercase.
 
 #### Class
 
-Elemental function.
+Pure function.
 
 #### Argument
 
@@ -74,7 +74,7 @@ Converts input character variable to all uppercase.
 
 #### Class
 
-Elemental function.
+Pure function.
 
 #### Argument
 
@@ -112,7 +112,7 @@ The rest of the character sequence is transformed to lowercase.
 
 #### Class
 
-Elemental function.
+Pure function.
 
 #### Argument
 
@@ -150,7 +150,7 @@ Reverses the order of all characters in the input character type.
 
 #### Class
 
-Elemental function.
+Pure function.
 
 #### Argument
 
--- a/doc/specs/stdlib_ascii.md
+++ b/doc/specs/stdlib_ascii.md
@@ -35,8 +35,7 @@ Converts input character variable to all lowercase.
 #### Syntax
 
 ```f90
-res = to_lower("HELLO!")
-! res == "hello!"
+res = [stdlib_asciii(module):to_lower(function)]](string)
 ```
 
 #### Class
@@ -51,6 +50,15 @@ Pure function.
 
 The result is an intrinsic character type of the same length as `string`.
 
+#### Example
+
+```fortran
+program demo_to_lower
+    use stdlib_ascii, only : to_lower
+    implicit none
+    print'(a)', to_lower("HELLo!") ! returns "hello!"
+ end program demo_to_lower
+``` 
 
 ### `to_upper`
 
@@ -64,9 +72,8 @@ Converts input character variable to all uppercase.
 
 #### Syntax
 
-```
-res = to_upper("hello!")
-! res == "HELLO!"
+```fortran
+res = [[stdlib_ascii(module):to_upper(function)]](string)
 ```
 
 #### Class
@@ -81,6 +88,15 @@ Pure function.
 
 The result is an intrinsic character type of the same length as `string`.
 
+#### Example
+
+```fortran
+program demo_to_upper
+    use stdlib_ascii, only : to_upper
+    implicit none
+    print'(a)', to_upper("hello!") ! returns "HELLO!"
+ end program demo_to_upper
+``` 
 
 ### `to_title`
 
@@ -90,18 +106,14 @@ Experimental
 
 #### Description
 
-Returns capitalized version of input character variable.
-The first alphanumeric character is capitalized.
+Returns a capitalized version of an input character variable.
+Only the first alphanumeric character is capitalized.
+All following characters will become lowercase.
 
 #### Syntax
 
 ```
-res = to_title("hello!")
-! res == "Hello!"
-res = to_title("'enquoted'")
-! res == "'Enquoted'"
-res = to_title("1st")
-! res == "1st"
+res = [[stdlib_ascii(module):to_title(interface)]](string)
 ```
 
 #### Class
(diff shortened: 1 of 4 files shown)


milancurcic:
Thank you, I think it's mostly a good addition, and much needed functions. A few questions and comments:

1. Why did you choose `stdlib_character` over `stdlib_string`? To me, character means a single character, and these functions work on strings. So I have a preference for `stdlib_string`, although I don't think `stdlib_character` is terrible, only misleading.
2. For `to_lower` and `to_upper`, I have a preference for just `lower` and `upper` (like Python).

milancurcic:
One more question: Why do you need interface definitions in the module?

awvwgk:
From the discussion in #69 I concluded that there will be potentially both, a string type and a module for handling intrinsic character variables. For the former the `stdlib_string` module should be reserved, therefore the latter might as well be in a separate module (I'm open for a better name).

milancurcic:
I just saw your post in #69. I like `lowercase`/`uppercase`/`titlecase` also. I understand the desire to have consistent naming with the ascii module, but I think deciding which names people like the best should take priority over consistency (and may be necessary due to name clash that you pointed out).

Considering planning for a `string_type`, I'd then prefer either of these two options, higher preference first:

1. Have both `string_type` and procedures that work on intrinsics in the same module called `stdlib_string`;
2. Have `stdlib_string_type` module for `string_type` and `stdlib_string` for the procedures that work on intrinsics.

awvwgk:
> One more question: Why do you need interface definitions in the module?

Mainly because I was trying to overload the `to_upper` / `to_lower` interface and get both single character manipulation and character string manipulation in one interface, which of course doesn't work. A future string type would potentially implement procedures under the same generic interface as well.

awvwgk:
Maybe the implementation could be moved into the `stdlib_ascii` module as well and just extend the existing implementations of `to_lower` and `to_upper` instead.

milancurcic:
> Maybe the implementation could be moved into the stdlib_ascii module as well and just replace the existing implementations of to_lower and to_upper instead.

I like this solution also.

awvwgk:
Moved the implementation to `stdlib_ascii` which seems much cleaner than creating a new module. To keep the repo history clean of the intermediary state the commits have been squashed.

src/tests/ascii/test_ascii.f90
```diff
@@ -540,4 +545,72 @@ pure logical function validation_func_interface(c)
         write(*,'(5X,12(I4))') (count(table(:,i)),i=1,12)
     end subroutine test_ascii_table
 
+    subroutine test_to_lower_string
+        character(len=:), allocatable :: dlc
```
ivan-pi, line 549: Is `dlc`  an acronym for something?

src/tests/ascii/test_ascii.f90
```diff
@@ -540,4 +545,72 @@ pure logical function validation_func_interface(c)
...
+        dlc = to_title(input)
+        call check(len(dlc) == 32)
+        call check(len_trim(dlc) == 5)
+        call check(trim(dlc) == "Title")
+
+        dlc = to_title(" s P a C e D !")
+        call check(dlc == " S p a c e d !")
+
+        dlc = to_title("1st, 2nd, 3rd")
+        call check(dlc == "1st, 2nd, 3rd")
```
ivan-pi, line 596: Is the purpose of this test to assure `to_title` does not modify alphabetical characters following numerals?

src/tests/ascii/test_ascii.f90
awvwgk: deferred length character

ivan-pi:
Thanks for fixing this. This was an initial blunder, the result of following the C routines to literally...

The `char_to_upper` and `char_to_lower` remain private to the module right? Do you think there is any benefit to just inline them? The public routines `to_upper`/`to_lover` cover the `character(len=1)` case anyways. For an array of fixed-length characters I hope an optimizing compiler would delete the calls to `len` and incrementing the loop iterator. In that case there is really no downside. 

I adopted the naming `to_lower`/`to_upper` initially in conformance with the C and D standard libraries (besides for the underscore, which was added). I am not really fixated to the current naming. If there is community interest we can change these to `lowercase`/`uppercase` or whatever else people prefer.

I am still slightly confused by the logic in the `to_title` routine, but will come back to it today. The rest of the pull request is fine. :+1:

awvwgk:
I have to admit I was deliberately opening this pull request to start a discussion on the features mentioned in #69.

I found a few additional caveats of the current `stdlib_ascii` implementation (especially with respect to arrays of fixed size character variables), which I would like to discuss separately (I can open a new issue or propose a fix with a different (later) PR and we discuss there).

> The `char_to_upper` and `char_to_lower` remain private to the module right? Do you think there is any benefit to just inline them?

I'm fully trusting the compiler to be capable of inlining a pure function in the same module scope. I can inline them of course, which will require some code duplication, since I'm currently able to reuse `char_to_upper` and `char_to_lower` in the `to_title` function.

ivan-pi:
With three functions, I agree it is cleaner this way. No need to inline.

With `reverse` I was for a second worried that it might be surprising to newcomers since it breaks the left alignment. Upon checking what Python allows, there is no reverse function, but the array slice syntax can be used:
```
>>> s = "abcd  "
>>> s[::-1] == "  dcba"
True
```
This is consistent with your proposed behavior, and honestly not a problem, once you are accustome to it.

ivan-pi:
> I found a few additional caveats of the current `stdlib_ascii` implementation (especially with respect to arrays of fixed size character variables), which I would like to discuss separately (I can open a new issue or propose a fix with a different (later) PR and we discuss there).

PR works fine for me.

src/tests/ascii/test_ascii.f90
```diff
@@ -540,4 +545,72 @@ pure logical function validation_func_interface(c)
...
+        character(len=32), parameter :: input = "reversed"
+
+        dlc = reverse("reversed")
+        call check(dlc == "desrever")
+
+        dlc = reverse(input)
+        call check(len(dlc) == 32)
+        call check(len_trim(dlc) == 32)
+        call check(trim(dlc) == "                        desrever")
+        call check(trim(adjustl(dlc)) == "desrever")
```
jvdp1, line 613: If I am correct, in Fortran 90, two strings are defined to be equal if they have the
exact same contents, except for trailing blanks, i.e.,
```fortran
"abc" == "abc    " !.true.
```
Therefore, I would expect that `reverse('abc') == reverse('abc    ')`.

The current behaviour is ok for me, if it would be mentioned in the specs.

jvdp1 (commented):
Thank you for this PR. Here are some comments.

doc/specs/stdlib_ascii.md
```diff
@@ -0,0 +1,149 @@
...
+### `to_title`
+
+#### Status
+
+Experimental
+
+#### Description
+
+Returns capitalized version of input character variable.
+The first alphanumeric character is capitalized.
```
ivan-pi, line 94: Capitalization does not make sense for numeric characters. From the tests, alphabetic characters following numerals are not capitalized (e.g. `1st`).

On the other hand the first alphabetic character following a *non-alphanumeric* character will be capitalized.

I haven't really studied other languages to know, what are sensible title case rules for an input like `++c`. With the current implementation the result will be `++C`.

ivan-pi (approved):
The intent of the `to_title` routine is clear to me now.

doc/specs/stdlib_ascii.md
```diff
@@ -0,0 +1,149 @@
...
+
+#### Syntax
+
+```
+res = to_title("hello!")
+! res == "Hello!"
+res = to_title("'enquoted'")
+! res == "'Enquoted'"
+res = to_title("1st")
+! res == "1st"
```
jvdp1, line 104:
```suggestion
res = [[stdlib_ascii(module):to_title(interface)]](string)
```

src/stdlib_ascii.f90
```diff
@@ -179,34 +176,94 @@ pure logical function is_blank(c)
...
+            t = uppercase(k:k)
         else
             t = c
         endif
-    end function
+    end function char_to_upper
+
+    !> Convert character variable to lower case
+    pure function to_lower(string) result(lower_string)
+        character(len=*), intent(in) :: string
```
jvdp1, line 211:
```suggestion
        !! version: experimental
        !!
        !! Converts input character variable to all lowercase.

        character(len=*), intent(in) :: string
```

src/stdlib_ascii.f90
```diff
@@ -179,34 +176,94 @@ pure logical function is_blank(c)
...
+
+        do i = 1, len(string)
+            lower_string(i:i) = char_to_lower(string(i:i))
+        end do
+
+    end function to_lower
+
+    !> Convert character variable to upper case
+    pure function to_upper(string) result(upper_string)
+        character(len=*), intent(in) :: string
```
jvdp1, line 223:
```suggestion
        !! version:experimental
        !!
        !! Converts input character variable to all uppercase.
        character(len=*), intent(in) :: string
```

src/stdlib_ascii.f90
```diff
@@ -179,34 +176,94 @@ pure logical function is_blank(c)
...
+        integer :: i
+
+        do i = 1, len(string)
+            upper_string(i:i) = char_to_upper(string(i:i))
+        end do
+
+    end function to_upper
+
+    !> Convert character variable to title case
+    pure function to_title(string) result(title_string)
```
jvdp1, line 234:
```suggestion
    pure function to_title(string) result(title_string)
    !! version: experimental
    !!
    !! Returns capitalized version of input character variable.
```

src/stdlib_ascii.f90
```diff
@@ -179,34 +176,94 @@ pure logical function is_blank(c)
...
+        end do
+
+        do i = n + 1, len(string)
+            title_string(i:i) = char_to_lower(string(i:i))
+        end do
+
+    end function to_title
+
+    !> Reverse the character order in the input character variable
+    pure function reverse(string) result(reverse_string)
```
jvdp1, line 257:
```suggestion
    pure function reverse(string) result(reverse_string)
        !! version: experimental
        !!
        !! Reverses the order of all characters in the input character type.
```

jvdp1 (commented):
The links to the specs in the source code should be added. But it could be done at the end of the review

doc/specs/stdlib_ascii.md
```diff
@@ -0,0 +1,149 @@
...
+Pure function.
+
+#### Argument
+
+`string`: shall be an intrinsic character type. It is an `intent(in)` argument.
+
+#### Result value
+
+The result is an intrinsic character type of the same length as `string`.
+
```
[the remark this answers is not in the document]
awvwgk:
````suggestion

#### Example
```fortran
program demo_to_upper
    use stdlib_ascii, only : to_upper
    implicit none
    print'(a)', to_upper("hello!") ! returns "HELLO!"
 end program demo_to_upper
``` 
````

File: src/stdlib_ascii.f90 (hunk 6 above)
Comment (awvwgk) [reply src/stdlib_ascii.f90:217]: I'll use predoc comments here to be consistent with the rest of the module.

File: doc/specs/stdlib_ascii.md (hunk 4 above)
Comment (awvwgk) [reply doc/specs/stdlib_ascii.md]: The `capitalize` implementation I found so far in Fortran usually went for the first alphanumeric character. Without this experience I would have gone only for the first character of the string. It kind makes sense to check and skip leading delimiters or quotation marks, `++c` is borderline case, but returning `++C` feels consistent with my expectation on a `capitalize` function.

File: src/tests/ascii/test_ascii.f90 (hunk 3 above)
Comment (awvwgk) [reply src/tests/ascii/test_ascii.f90:613]: Trailing whitespace is somewhat special, while leading whitespace is not. With reverse I would expect

```f90
call check(dlc == reverse(reverse(dlc)))
```

to always hold. Following the proposal to handle trailing whitespace special in reverse we would arrive at:

```f90
character(len=*), parameter :: input = "  abc "
character(len=:), allocatable :: dlc
dlc = reverse(input)
! dlc == "cba   "
dlc = reverse(dlc)
! dlc == "abc   " .and. .not. (dlc == input)
```

ivan-pi: In a previous comment I also alluded to the second interpretation of reverse which left aligns.

I agree with @awvwgk the specification such that
```
call check(dlc == reverse(reverse(dlc)))
```
is less surprising. It is also in agreement with at least one other language (Python).

doc/specs/stdlib_ascii.md
ivan-pi: I don't really have strong feelings about this. I think the most common use case is to capitalize individual words rather then sentences or even strings containing non-alpha numerical characters.

I still find the sentence `The first alphanumeric character is capitalized.` misleading. I understand the logic how it includes the ordinal numbers like `1st`, `2nd`, ... But dictionaries define capitalization only for letters, not for numerals.

awvwgk: How about:

*The first alphanumerical character is transformed to uppercase if it is a letter, the rest of the character sequences is transformed to lowercase.*

ivan-pi: I would prefer using a negative form: *The first alphabetical character is transformed to uppercase unless it follows/succeeds a numeral. The rest of the character sequence is transformed to lowercase.*

milancurcic (approved):
I promoted the four public functions to elemental (it's free lunch) and added tests for elementalness.

I think it's good to go, thank you!

awvwgk:
> I promoted the four public functions to elemental (it's free lunch) and added tests for elementalness.

Thanks, I was planning to look into making most of the `stdlib_ascii` module elemental while collecting the missing specs in a later pull request. But we can already have it here since it is an easy change.

ivan-pi:
> > I promoted the four public functions to elemental (it's free lunch) and added tests for elementalness.
> 
> Thanks, I was planning to look into making most of the `stdlib_ascii` module elemental while collecting the missing specs in a later pull request. But we can already have it here since it is an easy change.

The issue of elemental vs pure for the character validation functions was already raised in the original issue: https://github.com/fortran-lang/stdlib/issues/11#issuecomment-568794626
A downside was that if the procedures are elemental they can not be used as procedure pointers. On the other hand the pure procedures can be overloaded to work on rank-1, rank-2, or higher arrays.

If the procedures are made elemental, then the subroutine `test_ascii_table` at https://github.com/fortran-lang/stdlib/blob/a1c911ce003f47095cd5be796453045e55903a50/src/tests/ascii/test_ascii.f90#L487 needs to be refactored/removed.

milancurcic:
I see, I wasn't aware of this. It's not free lunch after all. I see 3 options going forward (don't have to be resolved in this PR).

1. We choose one or the other, we should decide which use case (use these functions on arrays or use them as procedure pointers) is more common or desired by the community. I vote for elemental.

2. We provide many specific procedures behind a generic name, that would work with arrays and as procedure pointers, as @ivan-pi suggested.

3. We provide both elemental and non-elemental versions under separate names. Then the user code could import them as:

```
use stdlib_ascii, only: to_lower => to_lower_elem, to_upper => to_upper_elem
```

or similar.

awvwgk:
I reverted e67026237c7cce2ea47cc2bcb6b68e94754782ee, but left it in the history. We should certainly discuss this separately and implement whatever solution we find most suitable with a different patch.

ivan-pi:
From my own experience, I have rarely resorted to using arrays of single characters. Therefore I am not strongly invested in making these procedures elemental. (After all the equivalent C routines are not elemental either...).

Then again, the usage case as procedure pointers is quite artificial. If one needs to switch between different operations, one could easily replace procedure pointers with a select case statement and an integer flag to perform the right operation.

milancurcic:
> From my own experience, I have rarely resorted to using arrays of single characters. 

Elemental here would apply to multi-char strings as well (the test that I included did). It could be useful to pass output of the upcoming (F202X) `split` to an elemental `to_lower`, `to_upper`, or `reverse`.

ivan-pi:
I agree that for multi-char strings the elemental makes sense. :+1: 

If I might suggest a different generalization, the character validation functions (`is_alpha`, `is_blank`, ...) could be modified to accept multi-char strings, but only test the first character. The intrinsic function `iachar` follows this type of behavior:  

> IACHAR(C) returns the code for the ASCII character in the *first character position* of C.

awvwgk:
What do you think? Are we ready to merge this PR?

milancurcic:
Yes, @jvdp1 please merge when ready.

doc/specs/stdlib_ascii.md
```diff
@@ -0,0 +1,171 @@
...
+Experimental
+
+#### Description
+
+Converts input character variable to all lowercase.
+
+#### Syntax
+
+`res = [[stdlib_asciii(module):to_lower(function)]] (string)`
```
jvdp1, line 37:
```suggestion
`res = [[stdlib_ascii(module):to_lower(function)]] (string)`
```

doc/specs/stdlib_ascii.md
```diff
@@ -0,0 +1,171 @@
...
+
+#### Description
+
+Returns a capitalized version of an input character variable.
+The first alphabetical character is transformed to uppercase unless it follows a numeral.
+The rest of the character sequence is transformed to lowercase.
+
+#### Syntax
+
+`res = [[stdlib_ascii(module):to_title(interface)]] (string)`
```
jvdp1, line 111:
```suggestion
`res = [[stdlib_ascii(module):to_title(function)]] (string)`
```

src/stdlib_ascii.f90
```diff
@@ -178,35 +179,103 @@ pure logical function is_blank(c)
...
-            t = upper_case(k:k)
+            t = uppercase(k:k)
         else
             t = c
         endif
-    end function
+    end function char_to_upper
+
+    !> Convert character variable to lower case
+    !>
```
jvdp1, line 214:
```suggestion
    !> ([Specification](../page/specs/stdlib_ascii.html#to_lower))
    !>
```

src/stdlib_ascii.f90
```diff
@@ -178,35 +179,103 @@ pure logical function is_blank(c)
...
+        integer :: i
+
+        do i = 1, len(string)
+            upper_string(i:i) = char_to_upper(string(i:i))
+        end do
+
+    end function to_upper
+
+    !> Convert character variable to title case
+    !>
```
jvdp1, line 242:
```suggestion
    !> ([Specification](../page/specs/stdlib_ascii.html#to_title))
    !>
```

src/stdlib_ascii.f90
```diff
@@ -178,35 +179,103 @@ pure logical function is_blank(c)
...
+        integer :: i
+
+        do i = 1, len(string)
+            lower_string(i:i) = char_to_lower(string(i:i))
+        end do
+
+    end function to_lower
+
+    !> Convert character variable to upper case
+    !>
```
jvdp1, line 228:
```suggestion
    !> ([Specification](../page/specs/stdlib_ascii.html#to_upper))
    !>
```

src/stdlib_ascii.f90
```diff
@@ -178,35 +179,103 @@ pure logical function is_blank(c)
...
+        end do
+
+        do i = n + 1, len(string)
+            title_string(i:i) = char_to_lower(string(i:i))
+        end do
+
+    end function to_title
+
+    !> Reverse the character order in the input character variable
+    !>
```
jvdp1, line 267:
```suggestion
    !> ([Specification](../page/specs/stdlib_ascii.html#reverse))
    !>
```

jvdp1 (approved):
Sorry for the delay. I corrected some minor typos in specs and added hyperlinks in the code.
I'll commit them and merge this PR after a last check.

jvdp1:
All checks have passed! I'll merge it. thank you @awvwgk for this PR.

Issue #277, opened 2020-12-22
Proposal: gaussian quadrature


hsnyder:
Hi everyone.

Would there be interest in adding methods to compute Gaussian quadrature points and weights (plus possibly Gauss-Lobatto points and weights)? I'd be happy to offer the implementation if there is interest.

arjenmarkus:
Sofar the library contains the trapezoid and Simpson rules only. More
sophisticated rules are certainly welcome, I'd say.

Op di 22 dec. 2020 om 06:13 schreef Harris Snyder <notifications@github.com

ivan-pi:
I'd be certainly interested in these. In a previous reply https://github.com/fortran-lang/stdlib/issues/112#issuecomment-576351981 I suggested these might be more suitable for a separate package. On second thought, since many Fortran users are physicists and engineers, at least for the classic Gaussian rules it might be beneficial to provide them here. If nothing else, the feedback from other stdlib developers and page visitors can help design a tidy API.

I have some very old Gaussian quadrature routines dating back to 1966 available here: https://github.com/ivan-pi/stroud_quad. If I am not mistaken they are accurate up to 10-11 digits for most cases.

jvdp1:
I agree wth @arjenmarkus and @ivan-pi: it would be good to have them in `stdlib` ;)

hsnyder:
Okay, sounds like the idea has some support. I'll wait another couple of days (holidays coming up anyway) and if there aren't any objections raised we can discuss an API

certik:
@hsnyder do not wait. We want this in `stdlib`, as we agreed the scope of `stdlib` can be roughly as `scipy`, and thus it definitely includes quadrature rules. Go ahead and start proposing the API. See also my comment with an implementation accurate to full double precision: https://github.com/fortran-lang/stdlib/issues/112#issuecomment-575963239

certik:
I think this is a duplicate of #112. But I'll keep it open, we can close both #277 and #112 once this is implemented in `stdlib`.

ivan-pi:
The NAG and IMSL mathematical libraries also offer Gaussian quadrature weights:
* [NAG](https://www.nag.com/numeric/nl/nagdoc_latest/flhtml/d01/d01conts.html)
* [IMSL)(https://help.imsl.com/fortran/current/html/fnlmath/index.html#page/FNLMath/mch4.06.01.html)

hsnyder:
Lots of options re: implementation it seems (depending on licenses)

[...]

PR #313, opened 2021-02-08
legendre polynomials and gaussian quadrature

hsnyder:
- Added preliminary implementation for Legendre polynomials
- Added preliminary implementation for Legendre polynomial first derivatives
- Added preliminary implementation for Gauss-Legendre quadrature points/weights
- Added preliminary implementation for Gauss-Legendre-Lobatto quadrature points/weights

link to proposal: https://github.com/fortran-lang/stdlib/issues/277


src/stdlib_quadrature_gauss.f90
```diff
@@ -0,0 +1,115 @@
...
+                    w(N/2+1)   = 2._dp/(N*(N+1._dp)*leg**2) 
+                end if
+                end block
+        end select
+        end associate
+        
+        if (present(interval)) then
+            associate ( a => interval(1) , b => interval(2) )
+            x = 0.5*(b-a)*x+0.5*(b+a)
+            w = 0.5*(b-a)*w
```
ivan-pi, line 111: I think here it should be `0.5_dp` to avoid mixed precision arithmetic. It might be worth checking that `x(1) == interval(1)` and `x(size(x)) == interval(2)`. I imagine this might not be the case due to rounding errors.

src/stdlib_quadrature_gauss.f90
```diff
@@ -0,0 +1,115 @@
...
+                    end do
+                    x(N-i+1) = -x(i+1)
+
+                    dleg = dlegendre(N+1,x(i+1))
+                    w(i+1)   = 2._dp/((1-x(i+1)**2)*dleg**2) 
+                    w(N-i+1) = w(i+1)
+                end do
+
+                if (mod(N,2) == 0) then
+                    x(N/2+1) = 0.0
```
ivan-pi, line 46: `x(N/2+1) = 0` (implicit type casting)

src/stdlib_quadrature_gauss.f90
```diff
@@ -0,0 +1,115 @@
...
+                integer :: i,j
+                real(dp) :: leg, dleg, delta
+
+                x(1)   = -1._dp
+                x(N+1) =  1._dp
+                w(1)   =  2._dp/(N*(N+1._dp))
+                w(N+1) =  2._dp/(N*(N+1._dp))
+
+                do i = 1, int(floor((N+1)/2._dp)-1)
+                    x(i+1) = -cos( (i+0.25_dp)*PI/N  - 3/(8*N*PI*(i+0.25_dp)))
```
ivan-pi, line 83: A comment containing the source for the initial guess could be helpful.

src/stdlib_quadrature_gauss.f90
```diff
@@ -0,0 +1,115 @@
...
+
+    pure module subroutine gauss_legendre_fp64 (x, w, interval)
+        real(dp), intent(out) :: x(:), w(:)
+        real(dp), intent(in), optional :: interval(2)
+
+        associate (N => size(x)-1 )
+        select case (N)
+            case (0)
+                x = 0._dp
+                w = 2._dp
```
ivan-pi, line 20: I think this should be
```
x(1) = 0
w(1) = 2
```

src/stdlib_quadrature_gauss.f90
```diff
@@ -0,0 +1,115 @@
...
+            case (1)
+                x = [-sqrt(1._dp/3._dp), sqrt(1._dp/3._dp)]
+                w = [1._dp, 1._dp]
```
ivan-pi, line 23:
```
x(1) = -sqrt(1.0_dp/3.0_dp)
x(2) = -x(1)
w = 1
```
Perhaps this is not worth bothering about. Hopefully, the compiler does not generate temporaries.

certik:
I have a comment regarding the implementation --- what is the accuracy? I had bad experience with higher order (say `p=64`) that wasn't accurate to machine accuracy with such methods. So we resorted to calculate the weights and points using high precision and generate code that hard codes the numbers valid to all digits. But this comment is an implementation, which is minor.

Let's mainly discuss the API first. I would remove `interval` as doing `[-1, 1]` seems like a default?

ivan-pi:
> Let's mainly discuss the API first. I would remove `interval` as doing `[-1, 1]` seems like a default?

The `interval` is an optional argument: 
```
pure module subroutine gauss_legendre_fp64 (x, w, interval)
        real(dp), intent(out) :: x(:), w(:)
        real(dp), intent(in), optional :: interval(2)
```

The default is `[-1,1]`. At least in (chemical) engineering problems, shifted Legendre polynomials are sometimes used. I find the `interval` argument handy for such cases. Of course one could argue, that the user of the quadrature should be responsible for such things.

I am afraid the `intent(out)` attribute for `x`, and `w` is wrong. If these are allocatable arrays in the calling program, then `gauss_legendre` will not work as intended. Adding an integer variable, and using fixed size arrays would be the most flexible solution.

nncarlson:
> I am afraid the `intent(out)` attribute for `x`, and `w` is wrong. If these are allocatable arrays in the calling program, then `gauss_legendre` will not work as intended.

How so?  I think you are mistaken here. Whether the actual arguments are allocatable is irrelevant. Are you thinking that the `intent(out)` causes them to be deallocated?

ivan-pi:
@nncarlson, you are right.

milancurcic:
Thank you @hsnyder. I didn't do a proper review yet, but two things stood out on a cursory skim through:

1. stdlib_functions.f90 for the source file and `stdlib_functions` for the module are not informative IMO. Is there some descriptive name we can use? What kind of functions are these?
2. Please use [lowercase for all variables](https://github.com/fortran-lang/stdlib/blob/master/STYLE_GUIDE.md#variable-and-procedure-naming). I think uppercase is okay for constants (`parameter`) although we don't have an established position on this yet.

ivan-pi:
>   2. Please use [lowercase for all variables](https://github.com/fortran-lang/stdlib/blob/master/STYLE_GUIDE.md#variable-and-procedure-naming). I think uppercase is okay for constants (`parameter`) although we don't have an established position on this yet.

Unfortunately, the lower-case `l` is sometimes frowned upon due to its potential for mix up with `1`. Perhaps `p` (for the polynomial) and `dp` (for its derivative) can be used as substitutes.

awvwgk:
> Perhaps `p` (for the polynomial) and `dp` (for its derivative) can be used as substitutes.

`dp` will unfortunately clash with the kind parameter for double precision reals. More than one or two letters might be advisable instead.

ivan-pi:
> `dp` will unfortunately clash with the kind parameter for double precision reals. More than one or two letters might be advisable instead.

Right... My next suggestion then are `pn` and `dpn`, which are unsurprisingly also used in the code by [Stroud & Secrest](https://github.com/ivan-pi/stroud_quad/blob/b36291d94ffa8a81e5e8aed6d7afb37bc4c0b0e7/legacy/recur.f#L1).

hsnyder:
Thanks everyone for the input so far!

> Thank you @hsnyder. I didn't do a proper review yet, but two things stood out on a cursory skim through:
> 
>     1. stdlib_functions.f90 for the source file and `stdlib_functions` for the module are not informative IMO. Is there some descriptive name we can use? What kind of functions are these?

The only thing I can really think of is "special functions". Acceptable?

arjenmarkus:
"Special functions" is the general term that is used for this part of
mathematics. So I'd say it is certainly acceptable.

Op ma 8 feb. 2021 om 21:04 schreef Harris Snyder <notifications@github.com>:

arjenmarkus (commented):
Just a remark: the do-loop at line 33 runs from 0 to newton_iters-1. As the do-loop variable is not explicitly used, I would say: let it run from 1 to newton_iters instead.

hsnyder:
> Just a remark: the do-loop at line 33 runs from 0 to newton_iters-1. As the do-loop variable is not explicitly used, I would say: let it run from 1 to newton_iters instead.

haha you can tell I'm a C programmer... I will fix this

hsnyder:
sorry for the delay I have been very busy. other than resolving the makefile conflicts, what are the next steps towards getting this merged? I will try to get to the conflicts in the makefiles sometime this week but if I can't and someone else wants to, please feel free :)

src/stdlib_quadrature_gauss.f90
```diff
@@ -0,0 +1,123 @@
+submodule (stdlib_quadrature) stdlib_quadrature_gauss
+    use stdlib_kinds, only: sp, dp, qp
```
awvwgk, line 2: Remove this line to solve the compilation error with GCC 10.2
```suggestion
```

src/stdlib_specialfunctions_legendre.f90
```diff
@@ -0,0 +1,72 @@
...
+                    integer :: i 
+
+                    leg_down1  = x
+                    dleg_down1 = 1
+
+                    leg_down2  = 1
+                    dleg_down2 = 0
+
+                    do i = 2, n
+                        leg = (2*i-1)*x*leg_down1/i - (i-1)*leg_down2/i
```
awvwgk, line 31: I get four warnings on implicit conversion here:
```
../src/stdlib_specialfunctions_legendre.f90:31:31:

   31 |                         leg = (2*i-1)*x*leg_down1/i - (i-1)*leg_down2/i
      |                               1
Warning: Conversion from INTEGER(4) to REAL(8) at (1) [-Wconversion-extra]
../src/stdlib_specialfunctions_legendre.f90:31:51:

   31 |                         leg = (2*i-1)*x*leg_down1/i - (i-1)*leg_down2/i
      |                                                   1
Warning: Conversion from INTEGER(4) to REAL(8) at (1) [-Wconversion-extra]
../src/stdlib_specialfunctions_legendre.f90:31:55:

   31 |                         leg = (2*i-1)*x*leg_down1/i - (i-1)*leg_down2/i
      |                                                       1
Warning: Conversion from INTEGER(4) to REAL(8) at (1) [-Wconversion-extra]
../src/stdlib_specialfunctions_legendre.f90:31:71:

   31 |                         leg = (2*i-1)*x*leg_down1/i - (i-1)*leg_down2/i
      |                                                                       1
Warning: Conversion from INTEGER(4) to REAL(8) at (1) [-Wconversion-extra]
```
Can we make those integer real conversion explicit?

src/stdlib_specialfunctions_legendre.f90
```diff
@@ -0,0 +1,72 @@
...
+                        dleg = dleg_down2 + (2*i-1)*leg_down1
```
awvwgk, line 32:
```
../src/stdlib_specialfunctions_legendre.f90:32:45:

   32 |                         dleg = dleg_down2 + (2*i-1)*leg_down1
      |                                             1
Warning: Conversion from INTEGER(4) to REAL(8) at (1) [-Wconversion-extra]
```
Can we make the integer real conversion explicit?

src/stdlib_specialfunctions_legendre.f90
```diff
@@ -0,0 +1,72 @@
...
+            case default
+                block
+                    real(dp) :: leg_down1, leg_down2
+                    integer :: i 
+
+                    leg_down1  = x
+                    leg_down2  = 1
+
+                    do i = 2, n
+                        leg = (2*i-1)*x*leg_down1/i - (i-1)*leg_down2/i
```
awvwgk, line 62: Another bunch of conversion warnings, can we make those explicit?
```
../src/stdlib_specialfunctions_legendre.f90:62:31:

   62 |                         leg = (2*i-1)*x*leg_down1/i - (i-1)*leg_down2/i
      |                               1
Warning: Conversion from INTEGER(4) to REAL(8) at (1) [-Wconversion-extra]
../src/stdlib_specialfunctions_legendre.f90:62:51:

   62 |                         leg = (2*i-1)*x*leg_down1/i - (i-1)*leg_down2/i
      |                                                   1
Warning: Conversion from INTEGER(4) to REAL(8) at (1) [-Wconversion-extra]
../src/stdlib_specialfunctions_legendre.f90:62:55:

   62 |                         leg = (2*i-1)*x*leg_down1/i - (i-1)*leg_down2/i
      |                                                       1
Warning: Conversion from INTEGER(4) to REAL(8) at (1) [-Wconversion-extra]
../src/stdlib_specialfunctions_legendre.f90:62:71:

   62 |                         leg = (2*i-1)*x*leg_down1/i - (i-1)*leg_down2/i
      |                                                                       1
Warning: Conversion from INTEGER(4) to REAL(8) at (1) [-Wconversion-extra]
```

src/stdlib_quadrature_gauss.f90
```diff
@@ -0,0 +1,122 @@
...
+                real(dp) :: leg, dleg, delta
+
+                x(1)   = -1._dp
+                x(n+1) =  1._dp
+                w(1)   =  2._dp/(n*(n+1._dp))
+                w(n+1) =  2._dp/(n*(n+1._dp))
+
+                do i = 1, (n+1)/2 - 1
+                    ! initial guess from an approximate form given by SV Parter (1999)
+                    x(i+1) = -cos( (i+0.25_dp)*pi/n  - 3/(8*n*pi*(i+0.25_dp)))
```
awvwgk, line 88: More implicit conversion warnings here:
```
../src/stdlib_quadrature_gauss.f90:88:36:

   88 |                     x(i+1) = -cos( (i+0.25_dp)*pi/n  - 3/(8*n*pi*(i+0.25_dp)))
      |                                    1
Warning: Conversion from INTEGER(4) to REAL(8) at (1) [-Wconversion-extra]
../src/stdlib_quadrature_gauss.f90:88:66:

   88 |                     x(i+1) = -cos( (i+0.25_dp)*pi/n  - 3/(8*n*pi*(i+0.25_dp)))
      |                                                                  1
Warning: Conversion from INTEGER(4) to REAL(8) at (1) [-Wconversion-extra]
../src/stdlib_quadrature_gauss.f90:88:60:

   88 |                     x(i+1) = -cos( (i+0.25_dp)*pi/n  - 3/(8*n*pi*(i+0.25_dp)))
      |                                                            1
Warning: Conversion from INTEGER(4) to REAL(8) at (1) [-Wconversion-extra]
../src/stdlib_quadrature_gauss.f90:88:51:

   88 |                     x(i+1) = -cos( (i+0.25_dp)*pi/n  - 3/(8*n*pi*(i+0.25_dp)))
      |                                                   1
Warning: Conversion from INTEGER(4) to REAL(8) at (1) [-Wconversion-extra]
```

src/stdlib_quadrature_gauss.f90
```diff
@@ -0,0 +1,122 @@
...
+                x(2) =  1
+                w = 1
+            case default
+                block
+                integer :: i,j
+                real(dp) :: leg, dleg, delta
+
+                x(1)   = -1._dp
+                x(n+1) =  1._dp
+                w(1)   =  2._dp/(n*(n+1._dp))
```
awvwgk, line 83: This doesn't look too problematic, but we should minimize the compiler warnings generated by stdlib as much as possible:
```
../src/stdlib_quadrature_gauss.f90:83:36:

   83 |                 w(1)   =  2._dp/(n*(n+1._dp))
      |                                    1
Warning: Conversion from INTEGER(4) to REAL(8) at (1) [-Wconversion-extra]
../src/stdlib_quadrature_gauss.f90:83:34:

   83 |                 w(1)   =  2._dp/(n*(n+1._dp))
      |                                  1
Warning: Conversion from INTEGER(4) to REAL(8) at (1) [-Wconversion-extra]
```

src/stdlib_quadrature_gauss.f90
```diff
@@ -0,0 +1,122 @@
...
+                w(n+1) =  2._dp/(n*(n+1._dp))
```
awvwgk, line 84: Again, not too problematic, but better explicit than implicit:
```
../src/stdlib_quadrature_gauss.f90:84:36:

   84 |                 w(n+1) =  2._dp/(n*(n+1._dp))
      |                                    1
Warning: Conversion from INTEGER(4) to REAL(8) at (1) [-Wconversion-extra]
../src/stdlib_quadrature_gauss.f90:84:34:

   84 |                 w(n+1) =  2._dp/(n*(n+1._dp))
      |                                  1
Warning: Conversion from INTEGER(4) to REAL(8) at (1) [-Wconversion-extra]
```

src/stdlib_quadrature_gauss.f90
```diff
@@ -0,0 +1,122 @@
...
+                        leg  = legendre(n+1,x(i+1)) - legendre(n-1,x(i+1))
+                        dleg = dlegendre(n+1,x(i+1)) - dlegendre(n-1,x(i+1))
+                        delta = -leg/dleg
+                        x(i+1) = x(i+1) + delta
+                        if ( abs(delta) <= tolerance * abs(x(i+1)) )  exit
+                    end do
+                    x(n-i+1) = -x(i+1)
+
+                    leg = legendre(n, x(i+1))
+                    w(i+1)   = 2._dp/(n*(n+1._dp)*leg**2)
```
awvwgk, line 99: More implicit conversion here
```
../src/stdlib_quadrature_gauss.f90:99:41:

   99 |                     w(i+1)   = 2._dp/(n*(n+1._dp)*leg**2)
      |                                         1
Warning: Conversion from INTEGER(4) to REAL(8) at (1) [-Wconversion-extra]
../src/stdlib_quadrature_gauss.f90:99:39:

   99 |                     w(i+1)   = 2._dp/(n*(n+1._dp)*leg**2)
      |                                       1
Warning: Conversion from INTEGER(4) to REAL(8) at (1) [-Wconversion-extra]
```

src/stdlib_quadrature_gauss.f90
```diff
@@ -0,0 +1,122 @@
...
+                    leg = legendre(n, x(i+1))
+                    w(i+1)   = 2._dp/(n*(n+1._dp)*leg**2) 
+                    w(n-i+1) = w(i+1)
+                end do
+
+                if (mod(n,2) == 0) then
+                    x(n/2+1) = 0
+
+                    leg = legendre(n, 0.0_dp)
+                    w(n/2+1)   = 2._dp/(n*(n+1._dp)*leg**2)
```
awvwgk, line 107:
```
../src/stdlib_quadrature_gauss.f90:107:43:

  107 |                     w(n/2+1)   = 2._dp/(n*(n+1._dp)*leg**2)
      |                                           1
Warning: Conversion from INTEGER(4) to REAL(8) at (1) [-Wconversion-extra]
../src/stdlib_quadrature_gauss.f90:107:41:

  107 |                     w(n/2+1)   = 2._dp/(n*(n+1._dp)*leg**2)
      |                                         1
Warning: Conversion from INTEGER(4) to REAL(8) at (1) [-Wconversion-extra]
```

src/stdlib_quadrature_gauss.f90
```diff
@@ -0,0 +1,122 @@
...
+                x(2) = -x(1)
+                w = 1
+            case default
+                block
+                integer :: i,j
+                real(dp) :: leg, dleg, delta
+
+                do i = 0, (n+1)/2 - 1
+                    ! use Gauss-Chebyshev points as an initial guess
+                    x(i+1) = -cos((2*i+1)/(2._dp*n+2._dp) * pi)
```
awvwgk, line 31:
```
../src/stdlib_quadrature_gauss.f90:31:50:

   31 |                     x(i+1) = -cos((2*i+1)/(2._dp*n+2._dp) * pi)
      |                                                  1
Warning: Conversion from INTEGER(4) to REAL(8) at (1) [-Wconversion-extra]
../src/stdlib_quadrature_gauss.f90:31:35:

   31 |                     x(i+1) = -cos((2*i+1)/(2._dp*n+2._dp) * pi)
      |                                   1
Warning: Conversion from INTEGER(4) to REAL(8) at (1) [-Wconversion-extra]
```

awvwgk (commented):
GCC flags a lot of implicit conversions here, they are probably all okay and not harmful. Still, can we make those explicit to reduce the noise to signal ratio of compiler warnings generated by stdlib?

awvwgk:
Most of the GCC warnings are generated because we are always enforcing it from our build files with
https://github.com/fortran-lang/stdlib/blob/0886501d057656cedf1806ebcd1ba2c2afc21e6a/CMakeLists.txt#L24

We get 23 possibly “false positive” warnings for this patch from `-Wconversion-extra` with GCC. There are only two ways forward IMO, either we did enable this warning because we want to avoid implicit conversion, than this patch has to account for this and make all conversions explicit, or we regard those warnings as not useful because they are false positives and delete this option from build files.

milancurcic:
I don't see a problem with these warnings. We could have the flag only in development and remove it for release build. I personally prefer implicit type casts where safe as it allows for cleaner code. But I'm not opposed to explicit conversion if others prefer it. Something to think about to add to the style guide.

ivan-pi:
In this usage case I'm in favor of implicit casts because they make sure the expression match closely the original mathematical form of the [recurrence relations](https://en.wikipedia.org/wiki/Legendre_polynomials#Recurrence_relations).

awvwgk:
> We could have the flag only in development and remove it for release build. I personally prefer implicit type casts where safe as it allows for cleaner code

This is actually not true, the `-Wconversion-extra` is added unconditionally for all builds. Therefore, I assumed it should always be honored, thanks for clarifying that this is not the case. I'll open a new issue to discuss our compilation profiles than.

awvwgk (approved):
With my confusion about the strictness of warnings resolved, this patch looks good to be merged.

certik:
I think this looks good, but we need to add tests, showing how to use it and also ensuring it actually works as expected for some rules.

hsnyder:
awvwgk, thanks very much for fixing those build issues. @certik I can add some tests and documentation, though probably not for another week or so. if you have any specific tests you'd like me to add please let me know.

certik:
I don't have any specific tests, just any tests that would run this say for order 3 and 5 and 64 and check the correct result/accuracy.

awvwgk:
What is the current status of this patch? @hsnyder can we do anything to help you move this pull request forward?

hsnyder:
@awvwgk sorry for the delay. I think test cases and a spec are the outstanding items. In principle anyone could write them but it would obviously be easiest for me to do so. I’ve been very busy for the past few weeks but things are starting to ease up. I think I can finish this off this week. I will post again when I’ve done so.

hsnyder:
Sorry for the delay! I've added some tests and specs for the new quadrature functions and the special functions module. Please let me know if any revisions are required.

*note*: I wasn't sure how to validate the the spec format is correct. I tried to follow the examples already present but I may have made errors and am not sure how to check it.

doc/specs/stdlib_quadrature.md
```diff
@@ -186,3 +186,87 @@ program demo_simps_weights
...
+If not specified, the default integral is -1 to 1.
+
+### Example
+
+```fortran
+integer, parameter :: N = 6
+real(dp), dimension(N) :: x,w
+call gauss_legendre(x,w)
+integral = sum(x**2 * w)
+```
```
awvwgk, line 229: This should be preferably a standalone demo program

doc/specs/stdlib_quadrature.md
```diff
@@ -186,3 +186,87 @@ program demo_simps_weights
...
+Using the nodes `x` and weights `w`, you can compute the integral of some function `f` as follows:
+`integral = sum(f(x) * w)`.
+
+Only double precision is supported - if lower precision is required, you must do the appropriate conversion yourself.
+Accuracy has been validated up to N=64 by comparing computed results to tablulated values known to be accurate to machine precision
+(maximum difference from those values is 2 epsilon).
+
+### Syntax
+
+`subroutine [[stdlib_quadrature(module):gauss_legendre_lobatto(interface)]](x, w [, interval])`
```
awvwgk, line 250:
```suggestion
`subroutine [[stdlib_quadrature(module):gauss_legendre_lobatto(interface)]] (x, w[, interval])`
```

doc/specs/stdlib_specialfunctions.md
```diff
@@ -0,0 +1,64 @@
...
+### Description
+
+Computes the value of the n-th Legendre polynomial at a specified point.
+Currently only 64 bit floating point is supported.
+
+This is an `elemental` function.
+
+### Syntax
+
+`result = [[stdlib_specialfunctions(module):legendre(interface)]](n, x)`
```
awvwgk, line 24:
```suggestion
`result = [[stdlib_specialfunctions(module):legendre(interface)]] (n, x)`
```

doc/specs/stdlib_specialfunctions.md
```diff
@@ -0,0 +1,64 @@
...
+### Description
+
+Computes the value of the first derivative of the n-th Legendre polynomial at a specified point.
+Currently only 64 bit floating point is supported.
+
+This is an `elemental` function.
+
+### Syntax
+
+`result = [[stdlib_specialfunctions(module):dlegendre(interface)]](n, x)`
```
awvwgk, line 53:
```suggestion
`result = [[stdlib_specialfunctions(module):dlegendre(interface)]] (n, x)`
```

awvwgk (approved):
Thanks a lot, looks good to me now. Only a few comments regarding the markup.

doc/specs/stdlib_quadrature.md
```diff
@@ -222,10 +222,15 @@ If not specified, the default integral is -1 to 1.
 ### Example
 
 ```fortran
-integer, parameter :: N = 6
-real(dp), dimension(N) :: x,w
-call gauss_legendre(x,w)
-integral = sum(x**2 * w)
+program integrate
+	use iso_fortran_env, dp => real64
+	implict none
```
awvwgk, line 227:
```suggestion
	implicit none
```

doc/specs/stdlib_quadrature.md
```diff
@@ -263,8 +268,13 @@ If not specified, the default integral is -1 to 1.
 ### Example
 
 ```fortran
-integer, parameter :: N = 6
-real(dp), dimension(N) :: x,w
-call gauss_legendre_lobatto(x,w)
-integral = sum(x**2 * w)
+program integrate
+	use iso_fortran_env, dp => real64
+	implict none
```
awvwgk, line 273:
```suggestion
	implicit none
```

milancurcic (approved):
Thank you @hsnyder and reviewers!

I have one last minute question and suggestion: What do you think about renaming everything that's called `stdlib_specialfunctions*` to `stdlib_special*`? I find `specialfunctions` quite long and a bit hard to read because two words are together, and `special` is clear enough what it means IMO.

I think this PR can go forward either way, but I'd prefer if we called this just `stdlib_special`.

hsnyder:
Hi @milancurcic , I prefer specialfunctions because of clarity, but I don't have strong feelings about it. On the one hand, I agree that "special" is shorter, easier to type and read, but on the other hand it's not very descriptive (I don't think "special" on it's own is very clear - the stdlib includes more than just math, so I think "special what?" is a perfectly valid reaction).

I'll wait ~ 24 hours for any additional opinions re: the name change and then I'll merge the request. Thanks very much for reviewing!

arjenmarkus:
An alternative might be "stdlib_specfunc" - though perhaps that is too
reminiscent of other packages/libraries in other languages?

Op di 29 jun. 2021 om 00:31 schreef Harris Snyder ***@***.***

hsnyder:
So it turns out I cannot merge this myself as I don't have write access. @milancurcic are you able to do so? Feel free to change the name if you want (though I still prefer "specialfunctions", or Arjen's suggestion instead of just "special"). 

Thanks!

milancurcic:
Thanks, @hsnyder, I'll merge as is.

Issue #312, opened 2021-02-06
Derived type IO


awvwgk:
With types like the `bitset_type` it might be a really useful feature to use them in normal IO (`read` and `write` data transfer). The `bitset_type` for example implements a `write_bitset` interface, but could as well define the very same function as derived type IO:

```f90
! exported as generic interface
interface write(formatted)
  module procedure :: write_formatted
end interface write(formatted)
interface read(formatted)
  module procedure :: read_formatted
end interface read(formatted)
interface write(unformatted)
  module procedure :: write_unformatted
end interface write(unformatted)
interface read(unformatted)
  module procedure :: read_unformatted
end interface read(unformatted)
! or type-bound as
generic :: write(formatted) => write_formatted
procedure, private :: write_formatted
generic :: read(formatted) => read_formatted
procedure, private :: read_formatted
generic :: write(unformatted) => write_unformatted
procedure, private :: write_unformatted
generic :: read(unformatted) => read_unformatted
procedure, private :: read_unformatted
```

```f90
subroutine write_formatted(self, unit, iotype, v_list, iostat, iomsg)
  !> Instance of the object to write.
  class(*), intent(in) :: self
  !> Formatted unit for output.
  integer, intent(in) :: unit
  !> Type of formatted data transfer, has the value "LISTDIRECTED" for fmt=*, "NAMELIST" for namelist output or starts with "DT" for derived type output.
  character(len=*), intent(in) :: iotype
  !>Rank one array of default integer type containing the edit descriptors for derived type output.
  integer, intent(in) :: v_list(:)
  !> Status identifier to indicate success of output operation.
  integer, intent(out) :: iostat
  !> Buffer to return error message in case of failing output operation.
  character(len=*), intent(inout) :: iomsg
end subroutine write_formatted
```

```f90
subroutine write_formatted(self, unit, iotype, v_list, iostat, iomsg)
  !> Instance of the object to write.
  class(*), intent(in) :: self
  !> Formatted unit for output.
  integer, intent(in) :: unit
  !> Type of formatted data transfer, has the value "LISTDIRECTED" for fmt=*, "NAMELIST" for namelist output or starts with "DT" for derived type output.
  character(len=*), intent(in) :: iotype
  !>Rank one array of default integer type containing the edit descriptors for derived type output.
  integer, intent(in) :: v_list(:)
  !> Status identifier to indicate success of output operation.
  integer, intent(out) :: iostat
  !> Buffer to return error message in case of failing output operation.
[...]

PR #320, opened 2021-02-13
Implement non-fancy functional string type

awvwgk:
This patch proposes a non-fancy string type, which basically provides the same functionality as a deferred length character but can be used in an `elemental` rather than a `pure` way. The idea was to have a scaffold for the string type in stdlib which can be extended later but already provides everything we are used to have from the deferred length character without the rough edges, like not being able to read into an unallocated deferred length character variable or being unable to have arrays.

#### Related issues

- String support in general (https://github.com/fortran-lang/stdlib/issues/69)
- User defined derived type input and output (https://github.com/fortran-lang/stdlib/issues/312)
- Better deferred length character support (https://github.com/j3-fortran/fortran_proposals/issues/9)
- Initial discussion on this implementation (https://github.com/awvwgk/stdlib_string/pull/1)

#### Goals

- add a non-extendible derived type representing a character sequence (`string_type`)
- support all intrinsic operators for character variables (`llt`, `lgt`, `lle`, `lge`, `len`, `len_trim`, `trim`, `adjustl`, `adjustr`, `index`, `scan`, `verify`, `repeat`, `ichar`, `iachar`)
- index and slice operations are implemented with the `char` function, which also allows to return a `string_type` as fixed length character variable
- support user defined derived type IO (formatted and unformatted)
- implement assignment from character values
- implement constructing string from character values

#### Non-goals (for this patch)

- add new functionality beyond the intrinsic functionality of deferred length character variables
- define the internal representation of the represented character sequence

#### FPM project

To ease review I created an fpm project at https://github.com/awvwgk/stdlib_string, which I will keep in sync with this patch. It also provides the API documentation and the specs at https://awvwgk.github.io/stdlib_string.


milancurcic:
Thank you, I think this will be a very nice addition.

I learned something new. If you define and import, say, `operator(.gt.)`, you get `operator(>)` for free. Do you know that this is standard behavior? I was surprised to see it.

Why did you implement the operators as module procedures instead of type bound methods? Currently, you need to explicitly import the operators from the module if you want to use them:

```
    use stdlib_string_type, only : string_type, assignment(=), len, &
        operator(.gt.), operator(.lt.), operator(.ge.), operator(.le.), &
        operator(.ne.), operator(.eq.), operator(//)
```

But if they were type-bound, they'd come with the type. Another advantage is that they would work with types that extend `string_type` (assuming the dummy arguments are declared as `class(string_type)`). There may be some disadvantage to the type-bound operator approach that I'm not aware of.

I also suggest adding an option to assign a `string_type` to an allocatable character variable:

```
    !> Assign a character sequence to a string and vice versa.
    interface assignment(=)
        module procedure :: assign_string_char
        module procedure :: assign_char_string
    end interface assignment(=)
...
    !> Assign a string to a character sequence.
    elemental subroutine assign_char_string(lhs, rhs)
        character(len=*), intent(inout) :: lhs
        type(string_type), intent(in) :: rhs
        lhs = rhs%raw
    end subroutine assign_char_string
```
If you agree, let me know if you'd like me to add the implementation and tests for it to this PR.

One thing that may be a non-goal for this PR, but I think useful, is a function (type-bound or otherwise) that returns a raw Fortran character string without having to assign it. This will allow `string_type` instance to be easily passed to procedure that expect an ordinary `character` variable. This would allow a "progressive" (incremental) incorporation of `string_type` into larger code that works with ordinary strings.

awvwgk:
> Why did you implement the operators as module procedures instead of type bound methods? Currently, you need to explicitly import the operators from the module if you want to use them:

None of the operators is defined for a polymorphic object so far, therefore they cannot be type-bound. Also the `string_type` is a non-extendible derived type and can therefore not have type-bound procedures, it would be cool to have a non-extendible derived type carry generic interfaces but we don't have those in Fortran. Also, it's a design choice aiming for an functional rather than for an object oriented design, for two reasons:

- non of the existing (intrinsic) string routines are type-bound, they all operate in a functional way rather than on the character sequence itself, defining type bound procedures for some (the operators) is inconsistent in my opinion
- type bound procedures are resolved at runtime, a procedure pointer is usually carried in the object instance, which makes the string object larger than it has to be

I did want to explore a string class at some point (maybe we can call it `string_class`) which allows to work on the character sequence it is representing in place. It comes which much more additional headache I wanted to avoid on the first pass. `ftlString` is such an example, even there not all operator interfaces can be type-bound so it won't be a complete package in any case.

---

> I also suggest adding an option to assign a string_type to an allocatable character variable:

There is the usual tie between fixed length characters and deferred length characters, you can only define one interface but not both.

You can either have a `pure` assignment to a deferred length character, but this disallows the assignment to a fixed length character, because it doesn't have the `allocatable` attribute

```f90
pure subroutine assign_char_string(lhs, rhs)
  character(len=:), allocatable, intent(inout) :: lhs
  type(string_type), intent(in) :: rhs
end subroutine assign_char_string
```

Or you can have an `elemntal` assignment to a fixed length character, which works nicely for allocated deferred length characters, but segfaults for unallocated deferred length characters (`iso_varying_string` has such an interface and it annoys me every time I have to move an `varying_string` type to a deferred length character):

```f90
elemental subroutine assign_char_string(lhs, rhs)
  character(len=*), intent(inout) :: lhs
  type(string_type), intent(in) :: rhs
end subroutine assign_char_string
```

The best option in my opinion was to drop this case entirely and use the automatic left-hand-side reallocating by explicitly returning the character sequences from the `string_type` with the `char` method. It's still not as elegant as it wish it were, but with the current standard this is how far we can get. Also the `char` method could have been so much nicer if GCC 8 had support for my original implementation.

awvwgk:
> One thing that may be a non-goal for this PR, but I think useful, is a function (type-bound or otherwise) that returns a raw Fortran character string without having to assign it.

Have a look at the [`char`](https://awvwgk.github.io/stdlib_string/interface/char.html) procedure, it implements exactly this behaviour together with the index access and slice functionality.

milancurcic:
Excellent, I didn't put two and two together and realize that this also implements `char` which works for both use cases (assign and pass as argument).

milancurcic:
> Also the string_type is a non-extendible derived type and can therefore not have type-bound procedures, it would be cool to have a non-extendible derived type carry generic interfaces but we don't have those in Fortran.

Is `sequence` the only thing making it non-extendable? I thought `sequence` was deprecated a while ago. Modern Fortran Explained says so in A.2. But I can't find a similar statement in the standard so I'm confused about its status.

> it's a design choice aiming for an functional rather than for an object oriented design, for two reasons:
>
> * non of the existing (intrinsic) string routines are type-bound, they all operate in a functional way rather than on the character sequence itself, defining type bound procedures for some (the operators) is inconsistent in my opinion
> * type bound procedures are resolved at runtime, a procedure pointer is usually carried in the object instance, which makes the string object larger than it has to be

Well, operators being type-bound is an implementation detail. The UI would still be functional.

The 2nd argument (run-time resolution) is valid, IMO. So it's important to note that this choice would be a compromise between performance (compile-time resolution of operators) and UI (having to import operators explicitly). 

As long as this module exports only its type and the procedure interfaces that work on it, I'd be happy to recommend users just doing `use stdlib_string_type` and be done with it.

awvwgk:
> Is `sequence` the only thing making it non-extendable? I thought `sequence` was deprecated a while ago. Modern Fortran Explained says so in A.2. But I can't find a similar statement in the standard so I'm confused about its status.

See https://j3-fortran.org/doc/year/18/18-007r1.pdf#subsubsection.573, no mention about any deprecation here or elsewhere in this document. To my knowledge, `sequence` and `bind(c)` are the only ways to make a type non-extendible. That said, the current specification would allow a C struct and implementation and only provide `iso_c_binding` interfaces in `stdlib_string_type`.

ivan-pi:
> As long as this module exports only its type and the procedure interfaces that work on it, I'd be happy to recommend users just doing `use stdlib_string_type` and be done with it.

I think this is a good general rule we can adopt in a future Fortran style guide. Particularly in the case when the procedures overload intrinsic functions and provide equivalent functionality.

epagone:
> > As long as this module exports only its type and the procedure interfaces that work on it, I'd be happy to recommend users just doing `use stdlib_string_type` and be done with it.
> 
> This is a good general rule we can adopt in a future Fortran style guide. Particularly in the case when the procedures overload intrinsic functions and provide equivalent functionality.

I agree on the fact that it should be a general rule, but I don't know if we have a majority on this. Maybe @milancurcic wants to consider this case-by-case, considering his pre-clause above (as long as...).

He might not be alone. For example, I did [raise a related issue](https://github.com/fortran-lang/stdlib/pull/269#issuecomment-755221591) while reviewing the string list PR (together with other observations) where the namespace was polluted with a `list_end` variable and I suggested to encapsulated it within the derived type. Well, although we [got to string list "new"](https://github.com/fortran-lang/stdlib/pull/311), well, let's say that my comment [did not get much traction](https://github.com/fortran-lang/stdlib/blob/90b06ff89ac3dce00e6c1d2f770beca38b837e34/src/stdlib_stringlist.f90#L26) :smiley:

milancurcic:
> See https://j3-fortran.org/doc/year/18/18-007r1.pdf#subsubsection.573, no mention about any deprecation here or elsewhere in this document. To my knowledge, sequence and bind(c) are the only ways to make a type non-extendible. 

I just asked about this on [Discourse](https://fortran-lang.discourse.group/t/what-is-the-status-of-the-sequence-attribute-and-statement/720?u=milancurcic).

But I'm still unclear what are you trying to do with `sequence`. Do you use it only to forbid extending the type? From my reading of `sequence`, it's meant only for enforcing the storage order of derived type components (please correct me if I'm wrong). But `string_type` has only one scalar component. So does `sequence` do anything useful here?

awvwgk:
> Do you use it only to forbid extending the type?

It's the only surefire method I know to make a type non-extendible, that's the reason I'm applying it here. But maybe there is a better way.

The problem is that extending another class from the `string_type` wouldn't work as expected given the current implementation, the new class would basically become useless unless one implements all functionality from `stdlib_string_type` again. For this reason I decided to it should be non-extendible.

Designing an abstract base class that can support the full functionality of the intrinsic deferred length character is not easy, also it is not a lightweight object anymore. I will upload a `stdlib_string_class.f90` to [awvwgk/stdlib_string](https://github.com/awvwgk/stdlib_string) once I'm done writing the ABC for an extendible string object.

milancurcic:
@awvwgk okay, all good, I understand now. I agree with this design choice. :+1:

src/stdlib_string_type.f90
```diff
@@ -0,0 +1,1102 @@
...
+    public :: write(formatted), write(unformatted)
+    public :: read(formatted), read(unformatted)
+
+
+    integer, parameter :: long = selected_int_kind(18)
+
+
+    !> String type holding an arbitrary sequence of characters.
+    type :: string_type
+        sequence
```
milancurcic, line 33:
```suggestion
    type :: string_type
        ! Use the sequence statement below as a hack to prevent extending this type.
        ! It is not used for storage association.
￼        sequence
```

src/tests/string/test_string_operator.f90
```diff
@@ -0,0 +1,123 @@
...
+
+contains
+
+    subroutine test_gt
+        type(string_type) :: string
+        logical :: res
+
+        string = "bcd"
+        res = string > "abc"
+        call check(res .eqv. .true.)
```
milancurcic, line 17: I'm curious, why `call check(res .eqv. .true.)` instead of `call check(res)`? And similar for `call check(res .eqv. .false.)` instead of `call check(.not. res)`. Is there any practical difference or is it just style?

milancurcic:
An idea: What do you think about calling this `str_type`? It's more concise. If this type is successful, it will be used often. `str` is also ubiquitous to mean "string" in programming. But `string_type` spells out exactly what it is, which I also think is good. I like them both, but `str_type` slightly better.

src/tests/string/test_string_operator.f90
awvwgk: Yes that could be simplified, I created those tests from the examples in the specs where I added a comment after each operation to show the current state / content of the variable and batch translated them with vim into `check` functions.

milancurcic: I also suggest to further isolate each test. As they are now, these tests also depend on the assignment operator (`string = "bcd"`). So if there was a regression that broke the assignment, leading to, say, `string` remaining empty after `string = "bcd", the comparison test would fail, but it wouldn't be the fault of the comparison. So for these tests I suggest a more concise:

```
call check(string_type("bcd") > "abc")
```

and similar for others.

awvwgk: Which in turn would dependent on the custom constructor instead of the assignment. The only safe way would be to test both string types created by assignment and custom constructor.

awvwgk: I decided to leave the current tests in as usage examples and have a little bit of Fortran procedure pointer magic generate tests for broader functionality coverage.

awvwgk:
> An idea: What do you think about calling this `str_type`? It's more concise. If this type is successful, it will be used often. `str` is also ubiquitous to mean "string" in programming. But `string_type` spells out exactly what it is, which I also think is good. I like them both, but `str_type` slightly better.

IMO, being explicit rather than implicit should be preferred, similar like we prefer `_type` over `_t` I would prefer `string_type` over `str_type` here. Either way it is already a huge gain over the deferred length character in length and expressiveness:

```f90
character(len=:), allocatable :: str
type(string_type) :: str
type(str_type) :: str
```

doc/specs/stdlib_string_type.md
```diff
@@ -0,0 +1,1607 @@
...
+is implementation dependent and not visible for the user of the module.
+
+#### Status
+
+Experimental
+
+
+## Procedures and methods provided
+
+Procedures returning `string_type` instances can usually used in elemental
```
jvdp1, line 33:
```suggestion
Procedures returning `string_type` instances can usually be used in elemental
```

doc/specs/stdlib_string_type.md
```diff
@@ -0,0 +1,1607 @@
...
+
+Elemental function.
+
+#### Argument
+
+None.
+
+#### Result value
+
+The result is an instance of `string_type`.
```
jvdp1, line 65:
```suggestion
The result is an instance of `string_type` with a lenght equal to 0.
```

doc/specs/stdlib_string_type.md
```diff
@@ -0,0 +1,1607 @@
...
+
+#### Example
+
+```fortran
+use stdlib_string_type
+implicit none
+type(string_type) :: string
+string = string_type()
+! len(string) == 0
+end
```
jvdp1, line 75:
```suggestion
program demo_string_type
    use stdlib_string_type
    implicit none
    type(string_type) :: string
   string = string_type()
   ! len(string) == 0
end program demo_string_type
```

doc/specs/stdlib_string_type.md
```diff
@@ -0,0 +1,1607 @@
...
+
+```fortran
+use stdlib_string_type
+implicit none
+type(string_type) :: string
+string = string_type("Sequence")
+! len(string) == 8
+string = string_type(" S p a c e d ")
+! len(string) == 9
+end
```
jvdp1, line 121:
```suggestion
program demo_string_type_char
use stdlib_string_type
implicit none
type(string_type) :: string
string = string_type("Sequence")
! len(string) == 8
string = string_type(" S p a c e d ")
! len(string) == 9
end program demo_string_type_char
```

doc/specs/stdlib_string_type.md
```diff
@@ -0,0 +1,1607 @@
...
+string = "Fortran"
+code = ichar(string)
+end
+```
+
+
+<!-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -->
+### Iachar function
+
+Code in ASCII collating sequence.
```
jvdp1, line 601: I suggest to remove this line.
```suggestion
```

doc/specs/stdlib_string_type.md
```diff
@@ -0,0 +1,1607 @@
...
+string = "Fortran"
+code = iachar(string)
+end
+```
+
+
+<!-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -->
+### Index function
+
+Position of a *substring* within a *string*.
```
jvdp1, line 645: This is a repetiton of the description. I suggest to remove this line.

doc/specs/stdlib_string_type.md
```diff
@@ -0,0 +1,1607 @@
...
+
+
+<!-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -->
+### Verify function
+
+Scan a string for the absence of a set of characters. Verifies that all
+the characters in string belong to the set of characters in set.
+
+#### Description
+
```
jvdp1, line 759:
```suggestion
#### Description

Scan a string for the absence of a set of characters. Verifies that all
the characters in string belong to the set of characters in set.

```

doc/specs/stdlib_string_type.md
```diff
@@ -0,0 +1,1607 @@
...
+```
+
+
+<!-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -->
+### Lgt function (lexical greater)
+
+Lexically compare the order of two character sequences being greater.
+
+#### Description
+
```
jvdp1, line 820:
```suggestion
#### Description

Lexically compare the order of two character sequences being greater.


```

doc/specs/stdlib_string_type.md
```diff
@@ -0,0 +1,1607 @@
...
+<!-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -->
+### Llt function (lexical less)
+
+Lexically compare the order of two character sequences being less.
+
+#### Description
+
+The left-hand side, the right-hand side or both character sequences can
+be represented by a string type.
+This defines three procedures overloading the intrinsic `llt` procedure.
```
jvdp1, line 876:
```suggestion
This defines three procedures overloading the intrinsic `llt` procedure.
```
Is this sentence needed for a user?

doc/specs/stdlib_string_type.md
```diff
@@ -0,0 +1,1607 @@
...
+<!-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -->
+### Comparison operator greater
+
+Compare the order of two character sequences being greater.
+
+#### Description
+
+The left-hand side, the right-hand side or both character sequences can
+be represented by a string type.
+This defines three procedures overloading the intrinsic `operator(.gt.)`.
```
jvdp1, line 1035: Related to the discussion during the monthly call, `.gt.` could become `>`

src/stdlib_string_type.f90
```diff
@@ -0,0 +1,1104 @@
...
+    public :: len, len_trim, trim, index, scan, verify, repeat, adjustr, adjustl
+    public :: lgt, lge, llt, lle, char, ichar, iachar
+    public :: assignment(=)
+    public :: operator(.gt.), operator(.ge.), operator(.lt.), operator(.le.)
+    public :: operator(.eq.), operator(.ne.), operator(//)
+    public :: write(formatted), write(unformatted)
+    public :: read(formatted), read(unformatted)
+
+
+    integer, parameter :: long = selected_int_kind(18)
```
jvdp1, line 28: Why did you not use one of `iso_fortran_env`?

doc/specs/stdlib_string_type.md
```diff
@@ -0,0 +1,1607 @@
...
+
+## Procedures and methods provided
+
+Procedures returning `string_type` instances can usually used in elemental
+context, while procedures returning scalar character values can only be
+used in a pure way.
+
+
+<!-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -->
+### Constructor for empty string
```
jvdp1, line 39: The two constructors could be combine into one single section, as:

`res = [[stdlib_string_type(module):string_type(interface)]] ([string])`

with `string` being `optional`

jvdp1 (commented):
Nice work. Thank you @awvwgk .
For the specs, I suggest to explicitely mention `program demo_....` and `end program demo_...`.
While it is correct like that, it might be confusing for beginners who would read these specs.

doc/specs/stdlib_string_type.md
awvwgk: That's how it is implemented, but for the API I found it clearer to specify the construction from a deferred length character separately from the empty string construction.

src/stdlib_string_type.f90
awvwgk: I've been following @sblionel's recommendation to use `selected_*_kind` rather than the `iso_fortran_env` kind parameters so far (see https://stevelionel.com/drfortran/2017/03/27/doctor-fortran-in-it-takes-all-kinds/). This just stuck with me ever since, but I'm can change to use `iso_fortran_env` here as well.

jvdp1: I understand the issue (I believe we had a similar discussion when proposing stdlib_kinds). Since it is only used inside this module, it is fine for me as it is.

doc/specs/stdlib_string_type.md
jvdp1: It was first a bit confusing to me. Let see what others think about it.

doc/specs/stdlib_string_type.md
```diff
@@ -0,0 +1,1607 @@
...
+end
+```
+
+
+<!-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -->
+### Len function
+
+#### Description
+
+Returns the length of the character sequence represented by the string.
```
ivan-pi, line 165: Is their a need to carry always the "character sequence" or couldn't this be simply "Returns the length of the string instance"? I am wondering whether we need this distinction.

doc/specs/stdlib_string_type.md
ivan-pi: How come? The other functions also have a short one line description.

doc/specs/stdlib_string_type.md
```diff
@@ -0,0 +1,1607 @@
...
+! pos == 0
+end
+```
+
+
+<!-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -->
+### Scan function
+
+Scan a *string* for the presence of a *set* of characters. Scans a *string* for
+any of the characters in a *set* of characters.
```
ivan-pi, line 700: The two sentences are very similar. Is it written similarly in the documentation of the intrinsic function?

doc/specs/stdlib_string_type.md
```diff
@@ -0,0 +1,1607 @@
...
+! pos == 0
+end
+```
+
+
+<!-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -->
+### Verify function
+
+Scan a string for the absence of a set of characters. Verifies that all
+the characters in string belong to the set of characters in set.
```
ivan-pi, line 756: The scan function used italic for string and set variable names. I think it should be here too for consistency or?

doc/specs/stdlib_string_type.md
```diff
@@ -0,0 +1,1607 @@
...
+well as operators for working with character variables and constants.
+
+
+## Derived types provided
+
+
+<!-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -->
+### The `string_type` derived type
+
+The `string_type` is defined as a non-extenable derived type representing a
```
ivan-pi, line 22:
```suggestion
The `string_type` is defined as a non-extendible derived type representing a
```

doc/specs/stdlib_string_type.md
ivan-pi: Watch out for "length" in the proposed correction.
```suggestion
The result is an instance of `string_type` with zero length.
```

doc/specs/stdlib_string_type.md
ivan-pi: Perhaps a more philosophical question, but if it is clearer this way in specs, why not also implement two constructors with a generic interface? (Answer is obviously to avoid code duplication).

Doesn't the non-fancy functional string behave as an empty string by default if not explicitly initialized? Are there any situations where I might want to explicitly use `string_type()` instead of just `''`?

doc/specs/stdlib_string_type.md
ivan-pi: Personally, I like the short example without the explicit `program` block. Is this worthy of a documentation meta-issue?

I might even feel tempted to drop the `implicit none` for brevity, although I imagine this is a controversial stance. :see_no_evil:

doc/specs/stdlib_string_type.md
```diff
@@ -0,0 +1,1607 @@
...
+! len(string) == 0
+end
+```
+
+
+<!-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -->
+### Constructor from character scalar
+
+The module defines a default constructor to create a string type
+from a character scalar.
```
ivan-pi, line 83: Just a question: since the internal representation is private, the default structure constructor would not work correctly?

doc/specs/stdlib_string_type.md
```diff
@@ -0,0 +1,1607 @@
...
+#### Example
+
+```fortran
+use stdlib_string_type
+implicit none
+type(string_type) :: string
+string = string_type("Sequence")
+! len(string) == 8
+string = string_type(" S p a c e d ")
+! len(string) == 9
```
ivan-pi, line 120:
```suggestion
! len(string) == 13
```

doc/specs/stdlib_string_type.md
```diff
@@ -0,0 +1,1607 @@
...
+! pos == 7
+
+pos = verify(string, string)
+! pos == 0
+end
+```
+
+
+<!-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -->
+### Lgt function (lexical greater)
```
ivan-pi, line 815:
```suggestion
### Lgt function (lexical greater than)
```

doc/specs/stdlib_string_type.md
```diff
@@ -0,0 +1,1607 @@
...
+! res .eqv. .false.
+
+res = lgt(string, "cde")
+! res .eqv. .false.
+end
+```
+
+
+<!-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -->
+### Llt function (lexical less)
```
ivan-pi, line 868:
```suggestion
### Llt function (lexical less than)
```

doc/specs/stdlib_string_type.md
```diff
@@ -0,0 +1,1607 @@
...
+! res .eqv. .false.
+
+res = llt(string, "cde")
+! res .eqv. .true.
+end
+```
+
+
+<!-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -->
+### Lge function (lexical greater or equal)
```
ivan-pi, line 921:
```suggestion
### Lge function (lexical greater than or equal)
```

doc/specs/stdlib_string_type.md
```diff
@@ -0,0 +1,1607 @@
...
+! res .eqv. .true.
+
+res = lge(string, "cde")
+! res .eqv. .false.
+end
+```
+
+
+<!-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -->
+### Lle function (lexical less or equal)
```
ivan-pi, line 974:
```suggestion
### Lle function (lexical less than or equal)
```

doc/specs/stdlib_string_type.md
ivan-pi: Maybe the syntax section could display both the new and old forms for completeness?

ivan-pi (commented):
I haven't managed to go through the implementation yet (not that I expect any wrongdoing), but I left a few comments in the specs.

doc/specs/stdlib_string_type.md
awvwgk: I left out the program block to help focus on the short example rather than on boilerplate. I can add program blocks to make it more clear that we have standalone programs here.

Regarding `implicit none`, good practice wants to be enforced and we don't want to start another lengthy thread on this old topic *again*.

In any case, let's discuss this elsewhere, GitHub PRs get clutter quickly once a discussion starts.

doc/specs/stdlib_string_type.md
awvwgk: Just some boilerplate I started with and carried on with afterwards. I tried to avoid calling it string because at the point I was not sure if it is going to be *the* string we agree on and looked for something more generic.

doc/specs/stdlib_string_type.md
awvwgk: Most descriptions are oriented at the descriptions of the intrinsic functions, since they are in fact implementing supplements to the intrinsics.

doc/specs/stdlib_string_type.md
awvwgk: Default constructors for allocatables have been buggy for several compilers in my experience. Best to avoid those whenever possible.

doc/specs/stdlib_string_type.md
awvwgk: I think there is an issue with two separate constructors, the only way to implement the specs is using the allocatable to optional mapping.

While I like this mapping it might not be apparent for somebody not too familiar with Fortran. Therefore, I thought explaining the resulting cases as separate constructors is clearer. It depends on whether specs/docs here are targeted at implementors or users whatever to prefer.

If you have declared it as variable, yes it does. In case you need to pass a string to a function that only accepts `string_type` instances the empty constructor becomes handy.

doc/specs/stdlib_string_type.md
awvwgk: Yes, it's stating the obvious, but I wouldn't consider it too harmful.

doc/specs/stdlib_string_type.md
awvwgk: Those short descriptions were meant as man page like synopsis while the description itself is more concrete on the actual functionality and argument qualifiers and as an attempt to fill the space between the function name header and the description header.

doc/specs/stdlib_string_type.md
jvdp1: Sorry for forgetting the  `implicit none`. It should be in the suggestion. I wil add it ;)

jvdp1: > Personally, I like the short example without the explicit `program` block. Is this worthy of a documentation meta-issue?

I believe it has been discussed somewhere when opening one of the first PR (I cannot retrieve it). The idea was to have some tools that would extract them, compile them and test them.

IMO adding program blocks doesn't add complexity (it is only one additional line ). However, my main issue with this short form is that it might not be obvious for Fortran beginners.

doc/specs/stdlib_string_type.md
jvdp1: > How come? The other functions also have a short one line description.

Well, I would remove all of them. There is a section "description".Therefore, I didn't understand why it was described before the section "description".

doc/specs/stdlib_string_type.md
jvdp1: Mmmm, if I look for a description of a procedure, I would go to the section "Description",  not before it. I just found that strange.

The idea for these specs was also to have a structure similar to the one describing the procedure in the [Fortran Standard](https://j3-fortran.org/doc/year/18/18-007r1.pdf). If we go for your proposition, all other specs should be adapted too, for consistency.

doc/specs/stdlib_string_type.md
milancurcic: I think "string" is fine, simpler language.

src/stdlib_string_type.f90
```diff
@@ -0,0 +1,1104 @@
...
+module stdlib_string_type
+    implicit none
+    private
+
+    public :: string_type
+    public :: len, len_trim, trim, index, scan, verify, repeat, adjustr, adjustl
+    public :: lgt, lge, llt, lle, char, ichar, iachar
+    public :: assignment(=)
+    public :: operator(.gt.), operator(.ge.), operator(.lt.), operator(.le.)
+    public :: operator(.eq.), operator(.ne.), operator(//)
```
milancurcic, line 23: @awvwgk What do you think about my suggestion to call these `operator(>)`, `operator(>=)`, and so on?

milancurcic (approved):
Thank you, I'm happy with this PR!

doc/specs/stdlib_string_type.md
ivan-pi: > I think there is an issue with two separate constructors, the only way to implement the specs is using the allocatable to optional mapping.

Is this due to the choice of deferred-length character for the character storage sequence?

From my point of view, it is okay two keep the two constructor forms separate in the specs.

doc/specs/stdlib_string_type.md
ivan-pi: I feel inclined to drop the word "default" in the first sentence. Given that the constructors over-ride the *default* structure constructor, I think it makes the sentence more confusing. I suppose the meaning of default above was in the sense of "most commonly used" constructor.

doc/specs/stdlib_string_type.md
awvwgk: > Is this due to the choice of deferred-length character for the character storage sequence?

Not really, just the usual Fortran madness about deferred length character vs fixed length character handling, sigh.

doc/specs/stdlib_string_type.md
ivan-pi:
```suggestion
Scan a *string* for the presence of a *set* of characters. 
```
Personally, I would drop the second sentence. I think the fact the dummy argument is called `set` and not `substring` is clear enough to indicate that we are scanning for *any* of the characters.

ivan-pi (approved):
+1 to merge.

awvwgk:
Ready to go from my side.

ivan-pi:
@jvdp1 do you have any unresolved comments? If not I suggest we go ahead and merge this later today.

jvdp1 (approved):
Sorry for the delay. It is ok from my side. It can be merged. Thank you @awvwgk for this work!

awvwgk:
Thanks a lot your comments and the review. The pushed the follow up for the non-fancy string type: an abstract base class for string objects (see https://github.com/fortran-lang/stdlib/pull/333), which basically provides the same functionality as the string type but in an object-oriented rather than a functional way.

PR #336, opened 2021-03-12
Add functions to convert integer/logical values to character values

awvwgk:
- use fypp in `stdlib_ascii` to generate functions for all kind values

### DIFF
--- a/src/stdlib_ascii.fypp
+++ b/src/stdlib_ascii.fypp
@@ -20,17 +20,17 @@ module stdlib_ascii
 
     ! Character conversion functions
     public :: to_lower, to_upper, to_title, reverse
-    public :: char_value
+    public :: to_char
 
     !> Version: experimental
     !>
     !> Create a character string representing the value of the provided variable.
-    interface char_value
+    interface to_char
     #:for kind in INT_KINDS
         module procedure :: integer_${kind}$_to_char
         module procedure :: logical_${kind}$_to_char
     #:endfor
-    end interface char_value
+    end interface to_char
 
     ! All control characters in the ASCII table (see www.asciitable.com).
     character(len=1), public, parameter :: NUL = achar(int(z'00')) !! Null
--- a/src/stdlib_ascii.fypp
+++ b/src/stdlib_ascii.fypp
@@ -27,8 +27,8 @@ module stdlib_ascii
     !> Create a character string representing the value of the provided variable.
     interface to_char
     #:for kind in INT_KINDS
-        module procedure :: integer_${kind}$_to_char
-        module procedure :: logical_${kind}$_to_char
+        module procedure :: to_char_integer_${kind}$
+        module procedure :: to_char_logical_${kind}$
     #:endfor
     end interface to_char
 
@@ -328,7 +328,7 @@ contains
 
     #:for kind in INT_KINDS
     !> Represent an integer of kind ${kind}$ as character sequence
-    pure function integer_${kind}$_to_char(val) result(string)
+    pure function to_char_integer_${kind}$(val) result(string)
         integer, parameter :: ik = ${kind}$
         integer(ik), intent(in) :: val
         character(len=:), allocatable :: string
@@ -359,18 +359,18 @@ contains
         end if
 
         string = buffer(pos:)
-    end function integer_${kind}$_to_char
+    end function to_char_integer_${kind}$
     #:endfor
 
     #:for kind in INT_KINDS
     !> Represent an logical of kind ${kind}$ as character sequence
-    pure function logical_${kind}$_to_char(val) result(string)
+    pure function to_char_logical_${kind}$(val) result(string)
         integer, parameter :: ik = ${kind}$
         logical(ik), intent(in) :: val
         character(len=1) :: string
 
         string = merge("T", "F", val)
-    end function logical_${kind}$_to_char
+    end function to_char_logical_${kind}$
     #:endfor
 
 end module stdlib_ascii
--- a/src/stdlib_ascii.fypp
+++ b/src/stdlib_ascii.fypp
@@ -20,17 +20,17 @@ module stdlib_ascii
 
     ! Character conversion functions
     public :: to_lower, to_upper, to_title, reverse
-    public :: to_char
+    public :: to_string
 
     !> Version: experimental
     !>
     !> Create a character string representing the value of the provided variable.
-    interface to_char
+    interface to_string
     #:for kind in INT_KINDS
-        module procedure :: to_char_integer_${kind}$
-        module procedure :: to_char_logical_${kind}$
+        module procedure :: to_string_integer_${kind}$
+        module procedure :: to_string_logical_${kind}$
     #:endfor
-    end interface to_char
+    end interface to_string
 
     ! All control characters in the ASCII table (see www.asciitable.com).
     character(len=1), public, parameter :: NUL = achar(int(z'00')) !! Null
@@ -328,7 +328,7 @@ contains
 
     #:for kind in INT_KINDS
     !> Represent an integer of kind ${kind}$ as character sequence
-    pure function to_char_integer_${kind}$(val) result(string)
+    pure function to_string_integer_${kind}$(val) result(string)
         integer, parameter :: ik = ${kind}$
         integer(ik), intent(in) :: val
         character(len=:), allocatable :: string
@@ -359,18 +359,18 @@ contains
         end if
 
         string = buffer(pos:)
-    end function to_char_integer_${kind}$
+    end function to_string_integer_${kind}$
     #:endfor
 
     #:for kind in INT_KINDS
     !> Represent an logical of kind ${kind}$ as character sequence
-    pure function to_char_logical_${kind}$(val) result(string)
+    pure function to_string_logical_${kind}$(val) result(string)
         integer, parameter :: ik = ${kind}$
         logical(ik), intent(in) :: val
         character(len=1) :: string
 
         string = merge("T", "F", val)
-    end function to_char_logical_${kind}$
+    end function to_string_logical_${kind}$
     #:endfor
 
 end module stdlib_ascii
--- a/src/stdlib_ascii.fypp
+++ b/src/stdlib_ascii.fypp
@@ -1,8 +1,11 @@
+#:include "common.fypp"
+
 !> The `stdlib_ascii` module provides procedures for handling and manipulating
 !> intrinsic character variables and constants.
 !>
 !> The specification of this module is available [here](../page/specs/stdlib_ascii.html).
 module stdlib_ascii
+    use stdlib_kinds, only : int8, int16, int32, int64
 
     implicit none
     private
@@ -17,6 +20,17 @@ module stdlib_ascii
 
     ! Character conversion functions
     public :: to_lower, to_upper, to_title, reverse
+    public :: char_value
+
+    !> Version: experimental
+    !>
+    !> Create a character string representing the value of the provided variable.
+    interface char_value
+    #:for kind in INT_KINDS
+        module procedure :: integer_${kind}$_to_char
+        module procedure :: logical_${kind}$_to_char
+    #:endfor
+    end interface char_value
 
     ! All control characters in the ASCII table (see www.asciitable.com).
     character(len=1), public, parameter :: NUL = achar(int(z'00')) !! Null
@@ -311,5 +325,52 @@ contains
         end do
 
     end function reverse
+
+    #:for kind in INT_KINDS
+    !> Represent an integer of kind ${kind}$ as character sequence
+    pure function integer_${kind}$_to_char(val) result(string)
+        integer, parameter :: ik = ${kind}$
+        integer(ik), intent(in) :: val
+        character(len=:), allocatable :: string
+        integer, parameter :: buffer_len = range(val)+2
+        character(len=buffer_len) :: buffer
+        integer :: pos
+        integer(ik) :: n
+        character(len=1), parameter :: numbers(0:9) = &
+            ["0", "1", "2", "3", "4", "5", "6", "7", "8", "9"]
+
+        if (val == 0_ik) then
+            string = numbers(0)
+            return
+        end if
+
+        n = abs(val)
+        buffer = ""
+
+        pos = buffer_len+1
+        do while (n > 0_ik)
+            pos = pos - 1
+            buffer(pos:pos) = numbers(mod(n, 10_ik))
+            n = n/10_ik
+        end do
+        if (val < 0_ik) then
+            pos = pos - 1
+            buffer(pos:pos) = '-'
+        end if
+
+        string = buffer(pos:)
+    end function integer_${kind}$_to_char
+    #:endfor
+
+    #:for kind in INT_KINDS
+    !> Represent an logical of kind ${kind}$ as character sequence
+    pure function logical_${kind}$_to_char(val) result(string)
+        integer, parameter :: ik = ${kind}$
+        logical(ik), intent(in) :: val
+        character(len=1) :: string
+
+        string = merge("T", "F", val)
+    end function logical_${kind}$_to_char
+    #:endfor
 
 end module stdlib_ascii
--- a/src/stdlib_ascii.fypp
+++ b/src/stdlib_ascii.fypp
@@ -347,7 +347,7 @@ contains
         n = abs(val)
         buffer = ""
 
-        pos = buffer_len+1
+        pos = buffer_len + 1
         do while (n > 0_ik)
             pos = pos - 1
             buffer(pos:pos) = numbers(mod(n, 10_ik))
--- a/doc/specs/stdlib_ascii.md
+++ b/doc/specs/stdlib_ascii.md
@@ -171,7 +171,7 @@ end program demo_reverse
 ```
 
 
-### `char_value`
+### `to_char`
 
 #### Status
 
@@ -183,7 +183,7 @@ Create a character string representing the value of the provided variable.
 
 #### Syntax
 
-`res = [[stdlib_ascii(module):char_value(interface)]] (string)`
+`res = [[stdlib_ascii(module):to_char(interface)]] (string)`
 
 #### Class
 
@@ -201,10 +201,10 @@ The result is an intrinsic character type.
 
 ```fortran
 program demo_char_value
-    use stdlib_ascii, only : char_value
+    use stdlib_ascii, only : to_char
     implicit none
-    print'(a)', char_value(-3) ! returns "-3"
-    print'(a)', char_value(.true.) ! returns "T"
-    print'(a)', char_value(42) ! returns "42"
+    print'(a)', to_char(-3) ! returns "-3"
+    print'(a)', to_char(.true.) ! returns "T"
+    print'(a)', to_char(42) ! returns "42"
 end program demo_char_value
 ```
--- a/doc/specs/stdlib_ascii.md
+++ b/doc/specs/stdlib_ascii.md
@@ -169,3 +169,42 @@ program demo_reverse
     print'(a)', reverse("Hello, World!") ! returns "!dlroW ,olleH"
 end program demo_reverse
 ```
+
+
+### `char_value`
+
+#### Status
+
+Experimental
+
+#### Description
+
+Create a character string representing the value of the provided variable.
+
+#### Syntax
+
+`res = [[stdlib_ascii(module):char_value(interface)]] (string)`
+
+#### Class
+
+Pure function.
+
+#### Argument
+
+`val`: shall be an intrinsic integer or logical type. It is an `intent(in)` argument.
+
+#### Result value
+
+The result is an intrinsic character type.
+
+#### Example
+
+```fortran
+program demo_char_value
+    use stdlib_ascii, only : char_value
+    implicit none
+    print'(a)', char_value(-3) ! returns "-3"
+    print'(a)', char_value(.true.) ! returns "T"
+    print'(a)', char_value(42) ! returns "42"
+end program demo_char_value
+```
--- a/doc/specs/stdlib_ascii.md
+++ b/doc/specs/stdlib_ascii.md
@@ -171,7 +171,7 @@ end program demo_reverse
 ```
 
 
-### `to_char`
+### `to_string`
 
 #### Status
 
@@ -183,7 +183,7 @@ Create a character string representing the value of the provided variable.
 
 #### Syntax
 
-`res = [[stdlib_ascii(module):to_char(interface)]] (string)`
+`res = [[stdlib_ascii(module):to_string(interface)]] (string)`
 
 #### Class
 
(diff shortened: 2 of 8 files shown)


src/stdlib_ascii.fypp
```diff
@@ -282,4 +296,51 @@ pure function reverse(string) result(reverse_string)
...
+
+        if (val == 0_ik) then
+            string = numbers(0)
+            return
+        end if
+
+        n = abs(val)
+        buffer = ""
+
+        pos = buffer_len+1
```
ivan-pi, line 320:
```suggestion
        pos = buffer_len + 1
```

src/stdlib_ascii.fypp
```diff
@@ -282,4 +296,51 @@ pure function reverse(string) result(reverse_string)
...
+
+        string = buffer(pos:)
+    end function integer_${kind}$_to_char
+    #:endfor
+
+    #:for kind in INT_KINDS
+    !> Represent an logical of kind ${kind}$ as character sequence
+    pure function logical_${kind}$_to_char(val) result(string)
+        integer, parameter :: ik = ${kind}$
+        logical(ik), intent(in) :: val
```
ivan-pi, line 339: Is this just for improved clarity of the generated code?

ivan-pi (commented):
In general I agree with the implementation and tests. I'm not sure about the name `char_value`. I think it is too long to be practical in building strings using concatenation, e.g. `filename = basename//char_value(timestep)//".txt"`. 

I guess the name `str` would be the typical name, but it risks clashing with variables. `char` cannot be used due to overlap with the intrinsic function.

src/stdlib_ascii.fypp
awvwgk: Yes, I didn't like the look with the `${kind}$` preprocessor all over the place

awvwgk:
I was thinking about something short, `str` or `string` would qualify but are likely to clash. `char` would be nice, but we can't have that one. There I went with `char_value` which can easily be renamed as wished by the user, but I'm open for better suggestions which are not usual variable names for strings.

epagone:
How about `to_char` or `2char`?

ivan-pi:
I just wanted to suggest `to_char` too. It is also very relatable to the C++ [`std::to_string`](http://www.cplusplus.com/reference/string/to_string/) function.

epagone:
I was thinking that there is also `to_str` (one character shorter than `to_char`) but the Fortran data type is `character`, so I still vote for `to_char`.

awvwgk:
Should we extend the constructor for the `string_type` as well? With the overloaded assignment we can already do this:
```f90
use stdlib_string_type, only : string_type, assignment(=), write(formatted)
use stdlib_ascii, only : to_char
implicit none
type(string_type) :: string
string = to_char(42)
print*, string
print*, string_type(to_char(42))
end
```
Do we want to extend the `string_type` constructor/assignment to allow this?
```f90
use stdlib_string_type, only : string_type, assignment(=), write(formatted)
implicit none
type(string_type) :: string
string = 42  # possibly harmful
print*, string
print*, string_type(42)  # maybe useful
end
```

epagone:
> Do we want to extend the `string_type` constructor/assignment to allow this?
> 
> ```fortran-modern
> use stdlib_string_type, only : string_type, assignment(=), write(formatted)
> implicit none
> type(string_type) :: string
> string = 42  ! possibly harmful
> print*, string
> print*, string_type(42)  ! maybe useful
> end
> ```

I'd [love](https://github.com/szaghi/StringiFor/issues/27) to allow the above. I agree that allowing the assignment `string = 42` it's quite loose and might be dangerous, but I would vote in favour of both.

milancurcic:
> I was thinking about something short, str or string would qualify but are likely to clash.

IMO `str` is an obvious choice. What would this clash with?

ivan-pi:
> > I was thinking about something short, str or string would qualify but are likely to clash.
> 
> IMO `str` is the obvious choice. What would this clash with?

```
type(string_type) :: str
```

milancurcic:
> I was thinking that there is also to_str (one character shorter than to_char) but the Fortran data type is character, so I still vote for to_char.

You're right, but the Fortran type name is surprising in general. "Character" is commonly used to refer to a single character, like in a C `char`, and "string" to refer to a character string of any length. But the fact that Fortran `character` type can be a multi-char string is an oddity and can be confusing to newcomers.

So the question is whether we should follow an existing, suboptimal convention, or choose the most appropriate and commonly used term. I'd prefer to use `str` over `to_char` here, because we're not only converting single characters.

milancurcic:
> `type(string_type) :: str`

@ivan-pi I'm confused. This is a snippet of a potential user code, right? So, an stdlib function name clashes with a user snippet?

Here's a scenario:

User: "I can't use the `str` function because I want to name my variable `str`."
stdlib: "Don't name your variable `str`! :)"

I totally get that this is a common and convenient pattern for single string variables, but if you use this argument, you're throwing out many perfectly good function names.

Is there some other scenario of a name clash?

epagone:
> But the fact that Fortran `character` type can be a multi-char string is an oddity and can be confusing to newcomers.

I agree that Fortran `character` type is quite unique. On the other hand, newcomers have to learn it anyway, if they want to code in Fortran. `to_str` or `str` is inconsistent with the name of the intrinsic data type that it refers to and creates confusion, IMO.

> So the question is whether we should follow an existing, suboptimal convention, or choose the most appropriate and commonly used term.

I would agree with you if we were planning on changing also the intrinsic name from `character` to `string` for multi-char strings, so that we would have name consistency with the functions of this PR. However, I would pass on such endeavour: it would be too disruptive even for me :smile:

ivan-pi:
@milan, That is precisely the scenario I had in mind. I recall there being a discussion somewhere of how `str` is a common variable name.

I found two examples where `str` is a variable name in Discourse:
- [Joining strings – problem with gfortran?](https://fortran-lang.discourse.group/t/joining-strings-problem-with-gfortran/492)
- [Reading a matrix of unknown size](https://fortran-lang.discourse.group/t/reading-a-matrix-of-unknown-size/706/13)

Here are a few Fortran libraries that use `str` as a variable name:
- [datetime-fortran](https://github.com/wavebitscientific/datetime-fortran/blob/53a91520eeb51841dfa1827a64d86760ede1ec96/src/datetime_module.f90#L180)
- [toml-f](https://github.com/toml-f/toml-f/search?q=str)
- [AMOS](https://github.com/certik/fortran-utils/blob/b43bd24cd421509a5bc6d3b9c3eeae8ce856ed88/src/legacy/amos/zbknu.f)
- [Fortran-Astrodynamics-Toolkit](https://github.com/jacobwilliams/Fortran-Astrodynamics-Toolkit/search?q=str)
- [M_CLI](https://github.com/urbanjost/M_CLI/search?q=str)
- [M_strings](https://github.com/urbanjost/M_strings/search?l=Fortran&q=str)
- [fpm](https://github.com/fortran-lang/fpm/blob/63a565fab4300e43fbd2d06b23d3c1f2a4e1976e/fpm/src/fpm/toml.f90#L90)
- [feq-parse](https://github.com/FluidNumerics/feq-parse/search?q=str)
- [fortran2018-examples](https://github.com/scivision/fortran2018-examples/search?q=str)
- [gtk-fortran](https://github.com/vmagnin/gtk-fortran/search?q=str)
- [ogpf](https://github.com/kookma/ogpf/search?q=str)
- [fortran-csv-module](https://github.com/jacobwilliams/fortran-csv-module/search?q=str)
- [json-fortran](https://github.com/jacobwilliams/json-fortran/search?q=str)
- [cp2k](https://github.com/cp2k/cp2k/search?l=Fortran&q=str)
- [dftbplus](https://github.com/dftbplus/dftbplus/search?l=Fortran&q=str)
- [Nek5000](https://github.com/Nek5000/Nek5000/search?l=Fortran&q=str)

And here are a few libraries which already use `str` for numerical to character conversion:
- [Coretran](https://github.com/leonfoks/coretran/blob/8c6b36946ff152956d6af55c4801c8b8e91b2513/docs/src/m_strings.f90#L29)
- [fortran-utils](https://github.com/certik/fortran-utils/blob/b43bd24cd421509a5bc6d3b9c3eeae8ce856ed88/src/utils.f90#L12)
- [PENF](https://github.com/szaghi/PENF/blob/b146a7d6444e0cd8a42a20452a7c4a4f4f16c667/src/lib/penf_stringify.F90#L38)
- [FEconv](https://github.com/victorsndvg/FEconv/blob/d0de78f8cad65987d2f31efc4f9bc868c96cf396/basicmod/source/vtk/IR_Precision.f90#L123) (actually uses it as variable name in some modules)

Let's look at what other languages have:
- Clojure: `(let [s (str i)])`
- C++: `auto s = std::to_string(i);
- C#: `string s = i.ToString()`
- D: `string s = i.to!string;`
- Go: `s := fmt.Sprintf("%d", i)`
- Javascript: `var s = i.toString();`
- Java: `String s=((Integer)i).toString();`
- Julia: [`string()`](https://docs.julialang.org/en/v1/base/strings/#Base.string)
- Kotlin: `val x = 42.toString()`
- Lua: `s = tostring(i)`
- PHP: `$s = (string)$i;`
- Python: `str()`
- Ruby: `s = i.to_s`
- Rust: `let s = i.to_string();`

ivan-pi:
> You're right, but the Fortran type name is surprising in general. "Character" is commonly used to refer to a single character, like in a C `char`, and "string" to refer to a character string of any length. But the fact that Fortran `character` type can be a multi-char string is an oddity and can be confusing to newcomers.

I'm not sure if using C is a good counter-argument. Fortran is the older language of the two, and the motivation for using Fortran is partially to avoid the pitfalls of C pointers and tedious array management. Isn't it then better to maintain the same root morpheme which is already part of language?

milancurcic:
> I'm not sure if using C is a good counter-argument.

Sorry for the confusion with that, I only meant to use C `char` as an illustration, not a counter-argument. The argument is: that in English language "character" means just one character; in computing, "character" refers to one character and "string" refers to a sequence of characters. When I wrote "C `char`" I didn't mean to say "because C does it like this, Fortran should too". I meant it as "C `char` is a an example of a programming term mapping to a term in the English language".

ivan-pi:
I understand your viewpoint. Quoting [Computer Hope](https://www.computerhope.com/jargon/c/char.htm):

> The abbreviation **char** is used as a reserved keyword in some programming languages, such as C, C++, C#, and Java. It is short for character, which is a data type that holds one character (letter, number, etc.) of data.

I also amended my list above with the name of the string object in different languages. Fortran *is* the odd one :joy:, which is exactly what would make the name `str` (or `to_str`) confusing. This is even more so once you consider we just added a `type(string_type)`.

milancurcic:
@ivan-pi Thank you for the comprehensive review of names in other languages. I think that's very helpful.

Regarding the Fortran libraries that use `str` as a variable, I don't think it's as easy to make the argument. If a program or procedure uses `str` as a variable name, there's no name clash unless there is also conversion to string happening in the same scope, and which would eventually motivate importing `str` from stdlib. So listing all the codes that have `str` as a variable name is not helpful on its own for the argument. 

Instead, let's narrow down to a scenario that would cause a name-clash:

1. Does a program or procedure declare a variable `str`? If yes, then
2. Is there a conversion from a non-string to a string in the same scope? If yes, then
3. There is a name clash if the user does `use stdlib_ascii` or `use stdlib_ascii, only: str`, and the user also wants to use `str` function to replace their home-cooked conversion. Then, the user has a choice:
  a) Rename their variable `str` in that scope, or
  b) Import `str` as `stdlib_ascii, only: to_str => str`

3a could be a major nuisance for the user, and 3b is a minor nuisance. Do you agree with my description of the scenario, and is my assessment of the level of nuisance fair?

But I think the consistency argument for `to_char` is good and it would probably be less confusing to experienced Fortran programmers than `str` or `to_str`. I'm not strongly convinced for the "popularity" (use string because that's what you mean) over the "consistency" (use `char` because Fortran calls it that) argument, but I lean toward the popular choice.

From your list, 9/15 use some kind of `to_` prefix, so I think `to_str` would also be a good name (and is consistent with existing conversion functions in `stdlib_ascii`). Currently, my preference is:

1. `str`
2. `to_str`
3. `to_char`

ivan-pi:
I did take the list of Fortran libraries a notch too far. I merely wanted to demonstrate that `str` is used in many real-world projects. As you suggest, existing codes that would need to import the conversion routine can always rename it. 

Regarding points 3a and 3b when developing new codes, as long as the code is modular, renaming is not a big issue with modern source code editors. In fact when preparing the list I noticed many codes use variables such as `file_str`, `key_str`, etc., which include a descriptive prefix (good practice), so 3a might be the "better" option. But in general I agree with your assessment.

One further downside of `to_char` is that it can be easily mistaken with the intrinsic function `char(i [,kind])` that returns the character represented by the integer `i`. At least for integer arguments this could go unnoticed until something breaks (e.g. strange symbols in output file names).

This has me quite torn now.

milancurcic:
I added Elixir, Erlang, Nim, and Zig to your table.

ivan-pi:
Just to throw a few other options on the table:

* A custom operator like `(.tc. 42)` for *to character*
* `repr` like the Python represent
* `jot` as in "jot down"
* `rec` as in "record"
* Functions named `write()` or `print()`

urbanjost:
Speaking of someone that already uses a STR() function name,one version I have that might be relevant (just uses internal reads) is in [M_msg]( https://github.com/urbanjost/M_msg) but uses class(\*) to allow the
useful ability to compose a CHARACTER variable  with something like STR(10.4,' is between',10,'and',11) and to create something similiar to WRITE  so you can call WRT([10,6],'the value is',V/2.0)  in a way somewhat similiar to list directed output. 

 There are other versions that use the old ITOA, FTOA, DTOA routines that are used for arrays as they are considerably faster than internal I/O; but I do not think I have a public version.  It raises the interesting issue of whether you want the routines to be elemental or not.

I find overloading helps make things much more compact, so assuming you have a function like STR, overloading // with that allows you to do things like "MESSAGE='the value is '//10//' and the limit is '//300.4//'.'".  Some would probably like to overload + to do that, but I prefer to overload + - / \* to convert strings to numbers, and to overload INT(), REAL(), and DBLE() to take character values and assume they are numeric.  So that makes for a symmetric set for converting to and from strings and numeric values.

I originally overloaded CHAR() to do something other than it does by default but decided that was a bad idea (I actually do not remember what I tried, but I remember I decided it was too confusing).

 So I did not see anything about overloading '//' or using class(*) variables so thought I would mention that works well for me.

epagone:
Any comments about `to_chars` or `chars` to clarify it's a multi-character function? How do you feel about it?

ivan-pi:
> So I did not see anything about overloading '//' or using class(*) variables so thought I would mention that works well for me.

Both Java and Javascript support the idiom of using the concatenation operator to build a string using

```Java
String s = "" + i;
```

If `operator(//)` were overload then one could do the same thing in Fortran:

```
character(len=:), allocatable :: s
s = ''//42   ! or 42//''
```

ivan-pi:
> Any comments about `to_chars` or `chars` to clarify it's a multi-character function? How do you feel about it?

FWIW, I find `to_chars` okay. The `to_` prefix helps clarify the intent IMO. Maybe the even shorter Ruby versions `to_s` or `to_c` would be of interest, even if they are a bit cryptic (`tos`/`toc` just seem confusing). 

I think ultimately we will need to go with the popular choice.

awvwgk:
Thanks for all the feedback, naming functions is hard, especially with the oddity of the Fortran `character`. I prefer to keep consistent with the Fortran intrinsic names here.

I also added the integer/logical to character conversion to the constructor of the `string_type`, not to the assignment yet, because I'm still undecided whether it will cause more harm than good. I'm not taking the overloaded assignment from the table, but would prefer to discuss it separately, because we might think about including an assignment from integer/logical to character as well.

milancurcic:
I realize now, both arguments--for `to_char` and for `to_str` or `str`--are "consistency" arguments. `to_char` is consistent with Fortran names. `to_str` or `str` are consistent with broader use. I see merits to both.

Even though "experimental" and can change, it's important to make a good choice here because these are such high impact functions.

Let's get more eyes on this. @certik @jvdp1 @everythingfunctional @arjenmarkus @rouson @LKedward @smeskos what do you think?

jvdp1:
Nice and important discussion! IMO the prefix "to_" is needed, as it explains the purpose of the function (like for `to_upper`, ...). I would be in favor of "to_char" because the result is a `character` variable. Similiarly, `to_string` could be used for the string DT. I find `to_str` a bit confusing: based on the name you could think that it would return a string DT, instead of a `character`.

smeskos:
I am in favor of the prefix "to_" as well.  Now as far as the rest, I have given it much thought and I am leaning towards "to_string", keeping in mind that it can be used for both characters as well as the string DT. 

Reasoning: I am inspired by C++ `std::to_string`. In a similar way as C++ has  `std::static_cast<char>(49)`  or `char(49)` that both will return the ascii "1", Fortran has `transform(49,'a')` and `char(49)` or  `achar(49)`, (`char()` being the compiler's default which I believe is almost always ascii and `achar()` for ascii) therefore Fortran misses something that will converts (un)signed integers and reals to a string with the same content and it's only logical to find something different than what Fortran programmers are already used to (the aforementioned options). For me, this leaves us with `to_char` or `to_string` and since the concept of string DT with the stdlib is also a recent proposal, why not use `to_string` for that purpose too. Bonus: it's natural to everyone who uses C++ as well, and I emphasize that since many programers who write Fortran, know and write in C++ as well, and every single one who writes in C++ for scientific purposes knows, or/and writes, or/and learned Fortran first...

everythingfunctional:
I vote for `to_char`. I've used the name `to_string` in my [strff](https://gitlab.com/everythingfunctional/strff) library to return `varying_string`. I assume that stdlib will want use `to_string` to return its own string type, and thus you can't use that name to return a `character`, or you'll having clashing interfaces for the generic resolution.

milancurcic:
> I assume that stdlib will want use to_string to return its own string type, and thus you can't use that name to return a character, or you'll having clashing interfaces for the generic resolution.

With d835ad8 (this PR), it will be `string_type(42)`.

awvwgk:
For derived types the natural choice is to extend the constructor IMO, this way no additional function name is required.

src/stdlib_ascii.fypp
```diff
@@ -17,6 +20,17 @@ module stdlib_ascii
...
     public :: to_lower, to_upper, to_title, reverse
+    public :: to_char
+
+    !> Version: experimental
+    !>
+    !> Create a character string representing the value of the provided variable.
+    interface to_char
+    #:for kind in INT_KINDS
+        module procedure :: integer_${kind}$_to_char
+        module procedure :: logical_${kind}$_to_char
```
milancurcic, line 31: Here, and everywhere else in the module, I suggest flipping this around so that the base name of the function comes first, type second, kind third:

```
        module procedure :: to_char_integer_${kind}$
￼        module procedure :: to_char_logical_${kind}$
```

This is consistent with other templates in stdlib, and the left-to-right reading order in most languages.

LKedward:
> I realize now, both arguments--for `to_char` and for `to_str` or `str`--are "consistency" arguments. `to_char` is consistent with Fortran names. `to_str` or `str` are consistent with broader use. I see merits to both.

My preference is for `str` or `to_str` since I personally associate char with a single character and have always conceptualised the Fortran `character()` declaration as a string anyway. `str` also matches closely with what you would expect if coming from another common language. I don't consider potential variable name clashes as a strong argument against since for small codebases the variable can be trivially renamed and for large codebases the function can be imported as an alias with `use , only:`. Seeking consistency with the Fortran language (for `to_char`) seems like a bad precedent to set since it potentially puts a lot of constraint on future naming schemes.

awvwgk:
From the discussion I conclude that `to_string`/`to_str` seems to find more acceptance than `to_char`/`to_chars`. So we arrive at the two new functions:

```f90
use stdlib_ascii, only : to_string  ! returns character(len=*)
use stdlib_string_type, only : string_type, &  ! overloaded constructor, returns type(string_type)
    & write(formatted), assignment(=)
implicit none
character(len=:), allocatable :: dlc
type(string_type) :: str
! Intrinsic character case
dlc = to_string(7)
print*, dlc, to_string(42)
! Derived type case
str = string_type(7)  ! str = to_string(7) works as well due to overloaded assignment(=)
print*, str, string_type(42)
end
```

milancurcic (approved):

LKedward (approved):
LGTM 👍

jvdp1 (approved):
LGTM! Thank you.

src/stdlib_ascii.fypp
```diff
@@ -312,4 +326,51 @@ pure function reverse(string) result(reverse_string)
...
+            buffer(pos:pos) = numbers(mod(n, 10_ik))
+            n = n/10_ik
+        end do
+        if (val < 0_ik) then
+            pos = pos - 1
+            buffer(pos:pos) = '-'
+        end if
+
+        string = buffer(pos:)
+    end function to_string_integer_${kind}$
```
everythingfunctional, line 362: Is there a reason not to make this a bit simpler and just use internal file IO? (like [this](https://gitlab.com/everythingfunctional/strff/-/blob/main/src/strff.f90#L477))

awvwgk: There was a brief discussion in https://github.com/fortran-lang/stdlib/issues/69#issuecomment-570563373 about internal IO vs alternative implementations and I got the impression that we want to avoid internal IO in such conversion routines.

everythingfunctional: Skimming that thread I didn't see anything saying that. Maybe I just missed it.

I think at one point I heard rumblings about internal file IO being really slow, but I don't think I've ever seen anybody actually demonstrate it.

LKedward: > I think at one point I heard rumblings about internal file IO being really slow, but I don't think I've ever seen anybody actually demonstrate it.

My proposed [improvement to the Julia benchmarks](https://fortran-lang.discourse.group/t/improving-fortran-results-in-the-julia-micro-benchmarks/198) found a significant improvement in `itoa` type functionality by avoiding internal write with gfortran. After skimming the gfortran runtime library I eventually came to the conclusion that internal write just has __a lot__ of overhead because it must handle so many different eventualities.

awvwgk: I did a quick benchmark at https://github.com/awvwgk/test-internal-io

The manual writing needs 3.9s while the internal IO needs 8.1s for the same workload (5 million integer to string conversions).

ivan-pi: Another reason, quoting from Jones & Crabtree (1988), Fortran Tools for VAX/VMS and MS-DOS, John Wiley & Sons: 

> The reason for not using internal files is mainly to avoid using Fortran I/O if possible on microcomputers with limited memory, with the hope that host compiler won't include Fortran I/O in the executable image; it is also kind of nice to be able to do it on your own. In our tests of IBM-PC program  that do not use Fortran I/O in the primitives, replacing versions of `itoa` and `atoi` using internal files with the ones shown here reduced the size of executable  programs by 21K bytes.

ivan-pi: Given that stdlib is both a specification and a *reference implementation* I think it is also fitting to demonstrate Fortran is capable of doing it's own string conversions efficiently (hopefully this turns out to be valid for floating point numbers too :roll_eyes:).

everythingfunctional: I stand corrected. It looks good to me then.

everythingfunctional (approved):

awvwgk:
Thanks everybody for the feedback. With four approvals I'll go ahead an merge.

certik:
I think that looks good. `to_string` is a good choice.

PR #339, opened 2021-03-13
string case conversion

Ketan2010:
I found many Fortran users are asking for built in methods to convert string into lower and upper case on stack overflow. 
Hence I have added two functions (upper, lower) to change case of string. I hope that my small contribution will help.

Looking forward to contribute more through Gsoc. 
Thanks!

### DIFF
--- a/src/stdlib_string_type.f90
+++ b/src/stdlib_string_type.f90
@@ -16,7 +16,7 @@ module stdlib_string_type
     private
 
     public :: string_type
-    public :: len, len_trim, trim, index, scan, verify, repeat, adjustr, adjustl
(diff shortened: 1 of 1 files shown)


awvwgk (changes requested):
Thanks for sharing this patch, please check the discussion at https://github.com/fortran-lang/stdlib/pull/310 for the agreed on naming of the case conversion routines.

There is an existing implementation available in the `stdlib_ascii` module which should be used to implement the case conversion for the `stdlib_string_type` to avoid code duplication.

Ketan2010:
Oh I see, Thanks!

PR #346, opened 2021-03-15
Added to_lower, to_upper, reverse and to_title function to stdlib_string_type.f90 file

### LINKED ISSUE #335: Implement to_lower, to_upper, to_title and reverse for string_type
The `stdlib_string_type` currently only implements the bare minimum functionality. To start extending the functionality the functions `to_lower`, `to_upper`, `to_title` and `reverse` implemented in `stdlib_ascii` should be implemented for `stdlib_string_type` as well.

Aman-Godara:
**Status**: open for review.

**Implemented**: the functions asked in #335, added unit tests for each of these functions in `test_string_functions.f90` file and documented the functions in `stdlib_string_type.md` file. (Fixes #335)
**Yet to be done**: https://github.com/fortran-lang/stdlib/pull/346#issuecomment-802004187
**Note**: Please have a look at the recent [suggestion comment ](https://github.com/fortran-lang/stdlib/issues/335#issuecomment-798815164)concerning `to_title` function

I am still trying to get the issues concerning my compiler fixed to run the code on my machine. The code written is **not** checked to run successfully on machine yet. Thus, I kindly request the reviewers to please also run the code once.

### DIFF
--- a/src/stdlib_string_type.f90
+++ b/src/stdlib_string_type.f90
@@ -16,7 +16,7 @@ module stdlib_string_type
 
     implicit none
     private
-    
+
     public :: string_type
     public :: len, len_trim, trim, index, scan, verify, repeat, adjustr, adjustl, to_lower, to_upper, to_title, reverse
     public :: lgt, lge, llt, lle, char, ichar, iachar
--- a/src/stdlib_string_type.f90
+++ b/src/stdlib_string_type.f90
@@ -12,13 +12,15 @@
 !>
 !> The specification of this module is available [here](../page/specs/stdlib_string_type.html).
 module stdlib_string_type
-    use stdlib_ascii, only: to_lower_ => to_lower, to_upper_ => to_upper, to_title_ => to_title, reverse_ => reverse
+    use stdlib_ascii, only: to_lower_ => to_lower, to_upper_ => to_upper, &
+                            to_title_ => to_title, reverse_ => reverse
 
     implicit none
     private
 
     public :: string_type
-    public :: len, len_trim, trim, index, scan, verify, repeat, adjustr, adjustl, to_lower, to_upper, to_title, reverse
+    public :: len, len_trim, trim, index, scan, verify, repeat, adjustr, &
+              adjustl, to_lower, to_upper, to_title, reverse
     public :: lgt, lge, llt, lle, char, ichar, iachar
     public :: assignment(=)
     public :: operator(>), operator(>=), operator(<), operator(<=)
--- a/src/stdlib_string_type.f90
+++ b/src/stdlib_string_type.f90
@@ -95,28 +95,32 @@ module stdlib_string_type
 
     !> Convert the character sequence hold by the input string to lower case
     !>
-    !> Elemental or pure ???? Does not return anything
+    !> This method is Elemental and returns a new string_type instance which holds lowercase
+    !> version of the character sequence hold by the input string
     interface to_lower
         module procedure :: to_lower_string
     end interface to_lower
 
     !> Convert the character sequence hold by the input string to upper case
     !>
-    !> Elemental or pure ???? Does not return anything
+    !> This method is Elemental and returns a new string_type instance which holds uppercase
+    !> version of the character sequence hold by the input string
     interface to_upper
         module procedure :: to_upper_string
     end interface to_upper
 
     !> Convert the character sequence hold by the input string to title case
     !>
-    !> Elemental or pure ???? Does not return anything
+    !> This method is Elemental and returns a new string_type instance which holds titlecase
+    !> version of the character sequence hold by the input string
     interface to_title
         module procedure :: to_title_string
     end interface to_title
 
     !> Reverse the character sequence hold by the input string
     !> 
-    !> Elemental or pure ???? Does not return anything
+    !> This method is Elemental and returns a new string_type instance which holds reversed
+    !> sequence of the character sequence hold by the input string
     interface reverse
         module procedure :: reverse_string
     end interface reverse
--- a/src/stdlib_string_type.f90
+++ b/src/stdlib_string_type.f90
@@ -12,7 +12,7 @@
 !>
 !> The specification of this module is available [here](../page/specs/stdlib_string_type.html).
 module stdlib_string_type
-    use stdlib_ascii, only: to_lower, to_upper, to_title, reverse
+    use stdlib_ascii, only: to_lower_ => to_lower, to_upper_ => to_upper, to_title_ => to_title, reverse_ => reverse
 
     implicit none
     private
@@ -482,7 +482,7 @@ contains
         type(string_type), intent(in) :: string
         type(string_type) :: lowered_string
 
-        lowered_string%raw = to_lower(string%raw)
+        lowered_string%raw = to_lower_(maybe(string))
 
     end function to_lower_string
 
@@ -492,7 +492,7 @@ contains
         type(string_type), intent(in) :: string
         type(string_type) :: uppered_string
 
-        uppered_string%raw = to_upper(string%raw)
+        uppered_string%raw = to_upper_(maybe(string))
 
     end function to_upper_string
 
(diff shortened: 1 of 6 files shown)


src/stdlib_string_type.f90
```diff
@@ -12,11 +12,13 @@
 !>
 !> The specification of this module is available [here](../page/specs/stdlib_string_type.html).
 module stdlib_string_type
+    use stdlib_ascii, only: to_lower, to_upper, to_title, reverse
```
awvwgk, line 15: Avoid namespace pollution:

```suggestion
    use stdlib_ascii, only: to_lower_ => to_lower, to_upper_ => to_upper, to_title_ => to_title, reverse_ => reverse
```

src/stdlib_string_type.f90
```diff
@@ -447,6 +477,46 @@ elemental function repeat_string(string, ncopies) result(repeated_string)
     end function repeat_string
 
 
+    !> Convert the character sequence hold by the input string to lower case
+    elemental function to_lower_string(string) result(lowered_string)
+        type(string_type), intent(in) :: string
+        type(string_type) :: lowered_string
+
+        lowered_string%raw = to_lower(string%raw)
```
awvwgk, line 485:
```suggestion
        lowered_string%raw = to_lower_(maybe(string))
```

src/stdlib_string_type.f90
```diff
@@ -447,6 +477,46 @@ elemental function repeat_string(string, ncopies) result(repeated_string)
...
+
+    end function to_lower_string
+
+
+    !> Convert the character sequence hold by the input string to upper case
+    elemental function to_upper_string(string) result(uppered_string)
+        type(string_type), intent(in) :: string
+        type(string_type) :: uppered_string
+
+        uppered_string%raw = to_upper(string%raw)
```
awvwgk, line 495:
```suggestion
        uppered_string%raw = to_upper_(maybe(string))
```

src/stdlib_string_type.f90
```diff
@@ -447,6 +477,46 @@ elemental function repeat_string(string, ncopies) result(repeated_string)
...
+
+    end function to_upper_string
+
+
+    !> Convert the character sequence hold by the input string to title case
+    elemental function to_title_string(string) result(titled_string)
+        type(string_type), intent(in) :: string
+        type(string_type) :: titled_string
+
+        titled_string%raw = to_title(string%raw)
```
awvwgk, line 505:
```suggestion
        titled_string%raw = to_title_(maybe(string))
```

src/stdlib_string_type.f90
```diff
@@ -447,6 +477,46 @@ elemental function repeat_string(string, ncopies) result(repeated_string)
...
+
+    end function to_title_string
+
+
+    !> Reverse the character sequence hold by the input string
+    elemental function reverse_string(string) result(reversed_string)
+        type(string_type), intent(in) :: string
+        type(string_type) :: reversed_string
+
+        reversed_string%raw = reverse(string%raw)
```
awvwgk, line 515:
```suggestion
        reversed_string%raw = reverse_(maybe(string))
```

awvwgk (commented):
Thanks for sharing. I suggest to rename the imports from `stdlib_ascii` to avoid polluting the namespace of `stdlib_string_type` and reexporting them accidentally.

I recommend to use the `maybe()` wrapper for safe memory access inside `stdlib_string_type` rather than directly accessing the raw value on the right-hand-side.

awvwgk:
> Should the newly created functions in stdlib_string_type.f90 module (to_lower_string, to_upper_string, etc.) be elemental or pure?

Elemental should be the way to go here. For `stdlib_ascii` we discussed the usage of `pure` vs `elemental` briefly here: https://github.com/fortran-lang/stdlib/pull/310#issuecomment-776978642. Those concerns are mainly related to the intrinsic character type and not the `string_type`.

Also the manual Makefile failure is related to the newly introduced dependency of `stdlib_string_type` on `stdlib_ascii` and can be fixed by adding
```make
stdlib_string_type.o: stdlib_ascii.o
```
to `src/Makefile.manual`.

Aman-Godara:
I will do the rest of the changes in the pull request as soon as my minor exams (15th to 18th) are over. Nevertheless, I remain active on github to check for the updates. I will start working on this issue full time from 18th until. If you anyone have any suggestion, please post them here. Thank you.
I will delete this comment once it becomes irrelevant.

Aman-Godara:
Can I create a new .f90 file in the `src/tests/string` to create test cases for the functions implemented above or should I edit an already existing file in the folder.

awvwgk:
> Can I create a new .f90 file in the `src/tests/string` to create test cases for the functions implemented above or should I edit an already existing file in the folder.

You can add a new file, but make sure to register the unit test in the CMake build file in the same directory than as well.

Aman-Godara:
What happens when a user gives a `string_type` object on the right hand side instead of a character sequence while handling the `assignment operator(=)` with `string_type` variables.

milancurcic (approved):
Looks good to me, thanks @Aman-Godara!

milancurcic:
> What happens when a user gives a string_type object on the right hand side instead of a character sequence while handling the assignment operator(=) with string_type variables. I think current implementation doesn't handle such possibility.

Assigning `string_type` to another `string_type` comes out-of-the box with the derived type. It should just work. Is that what you mean or something else?

awvwgk (commented):
@Aman-Godara Can you add the new procedures to the specification of the `string_type` in `doc/specs/stdlib_string_type.md` as well?

Aman-Godara:
@awvwgk Yes, I am working on documenting the functions created. I just needed a little help with https://github.com/fortran-lang/stdlib/pull/346#issuecomment-802004187 to proceed further because I am at the starting stage of learning CMake.

@milancurcic yes, you got it right. I wanted to talk more about it, but before that let me try to figure things out on my own by learning fortran.

src/stdlib_string_type.f90
```diff
@@ -12,11 +12,15 @@
...
+                            to_title_ => to_title, reverse_ => reverse
+
     implicit none
     private
 
     public :: string_type
-    public :: len, len_trim, trim, index, scan, verify, repeat, adjustr, adjustl
+    public :: len, len_trim, trim, index, scan, verify, repeat, adjustr, &
+              adjustl, to_lower, to_upper, to_title, reverse
     public :: lgt, lge, llt, lle, char, ichar, iachar
```
awvwgk, line 24: Just a bit of reordering here to keep the overloaded intrinsics together.
```suggestion
    public :: len, len_trim, trim, index, scan, verify, repeat, adjustr, adjustl
    public :: lgt, lge, llt, lle, char, ichar, iachar
    public :: to_lower, to_upper, to_title, reverse
```

awvwgk (approved):
Thanks, this looks good now.

milancurcic:
With two approvals, I will merge. Thank you, @Aman-Godara!

Aman-Godara:
> I recommend to use the `maybe()` wrapper for safe memory access inside `stdlib_string_type` rather than directly accessing the raw value on the right-hand-side.

@awvwgk does `maybe()` wrapper handle the case when `raw` (component of `string_type` derived type) is not allocated?

awvwgk:
Indeed, it does make sure that an unallocated `raw` component of the `string_type` behaves like an empty string:
https://github.com/fortran-lang/stdlib/blob/bc37bcc8f71605f78e51c3c79c2d914e7825a339/src/stdlib_string_type.f90#L1166-L1177

Aman-Godara:
May I ask why we didn't design the constructor to have an empty character sequence ("") when optional argument `string` is not given?
https://github.com/fortran-lang/stdlib/blob/bc37bcc8f71605f78e51c3c79c2d914e7825a339/src/stdlib_string_type.f90#L346-L352

awvwgk:
There is still the possibility to have an uninitialized `string_type` which will have an unallocated `raw` component by default as in
```f90
use stdlib_string_type
implicit none
type(string_type) :: str
print *, str
end

Aman-Godara:
Got it! 👍 Thanks!

PR #349, opened 2021-03-18
Simplify test makefile

awvwgk:
- use a single target for all, test and clean
- makes it easier to add a new subdirectory without having to modify multiple entries


milancurcic (approved):
Nice, I didn't know you could do that.

jvdp1 (approved):
Nice! Less error-prone. Thank you.
From my side, it can be merged!

Issue #327, opened 2021-02-26
Tolerant comparison of real numbers


arjenmarkus:
I have the source for a set of comparison functions and other functions such as floor and ceil that take into account that real numbers are not as well-behaved as, say, integers, by taking a margin into account. The source comes from Skip Knoble, once quite active on comp.lang.fortran, and builds on work by Hindmarsh and others. 

Is this something useful for stdlib? The source needs to be updated to the current standard, but I can do that :).

epagone:
Personally, I would be very interested in that. The naive approach that I currently use in my codes is the following:
```fortran
elemental logical function eqreal(real1, real2)
    real(wp), intent(in) :: real1, real2
    eqreal = abs(real1 - real2) < spacing( max( abs(real1), abs(real2) ) )
end function eqreal
```

I would be curious to learn what a proper approach would be.

arjenmarkus:
I am not sure if there is a proper approach ;). The best I can think of is a well-chosen tolerance and that may actually be dependent on the application. The code I was referring predates Fortran 90, by the way, but it is definitely worth considering, if only because it will provide some common ground based on much prior work.

ivan-pi:
I think this is a partial duplicate of https://github.com/fortran-lang/stdlib/issues/145.

arjenmarkus:
Indeed, the collection of functions I have includes the ones described
there.

Op vr 26 feb. 2021 om 17:35 schreef Ivan Pribec <notifications@github.com>:

jvdp1:
> Is this something useful for stdlib? 

I think it is indeed something for stdlib. It could even be used inside stdlib for the different tests with real values (and for which we already got multiple issues/discussions).

arjenmarkus:
I attach the original source code [eps_f90.txt](https://github.com/fortran-lang/stdlib/files/6062412/eps_f90.txt) - I have not take the opportunity yet to convert it to a modern package, so it is really old-school :). However, the output from the program may be of interest to give better insight:

 EPS=  0.22204460D-15  3CB0000000000000, EPS3=  0.66613381D-15  3CC8000000000000
 X=1.D0/          49 , Y=X*          49 , Z=1.D0
 Y=  0.99999999999999989       Z=   1.0000000000000000     
 X=3F94E5E0A72F0539 Y=3FEFFFFFFFFFFFFF Z=3FF0000000000000
 Fuzzy Comparisons: Y<>Z
 but TEQ(Y,Z) is .TRUE.
[...]

PR #353, opened 2021-03-20
Initial checkin for a module for tolerant comparison of reals

arjenmarkus:
The module is based on the idea to allow for a small margin when comparing reals. Such a margin can solve the problem that the finite precision of floating-point calculations make the results inaccurate. Two essentially equal real numbers may therefore appear to be different. This is in response to issue #327


src/CMakeLists.bak
```diff
@@ -0,0 +1,73 @@
+### Pre-process: .fpp -> .f90 via Fypp
```
awvwgk, line 1: `src/CMakeLists.bak` is not supposed to be committed here, is it?

arjenmarkus:
Certainly not - github desktop is all too keen on submitting all files
residing in the checkout directory.

Op za 20 mrt. 2021 om 12:42 schreef Sebastian Ehlert <
***@***.***>:

milancurcic:
Thanks @arjenmarkus, I think these will be useful. 

Is "tolerant" a commonly and uniquely used term? I wonder if most people would understand what it is just based on that word.

What do you think about `stdlib_comparisons`? But then it misses the point of the comparisons being tolerant. And `stdlib_tolerant_comparisons` is unwieldy.

I guess my main gripe with the word "tolerant" is that it's an adjective and not a noun, i.e. it doesn't say what is in the module, but what something is like in the module.

arjenmarkus:
The original source code mentions "fuzzy Fortran operators", that may be
more apt, but then I fear people would expect fuzzy sets and the like. I am
not too happy with the term tolerant either, so I am open to suggestions.

Op ma 22 mrt. 2021 om 20:46 schreef Milan Curcic ***@***.***>:

milancurcic:
Currently `stdlib_comparison` or `stdlib_comparisons` is my favorite. We can request feedback on the call tomorrow.

arjenmarkus:
The plural has my preference. As I will not be able to join today, I will
simply await the results.

Op di 23 mrt. 2021 om 23:35 schreef Milan Curcic ***@***.***>:

jvdp1 (commented):
Thank ypou @arjenmarkus for this PR. Here are some comments.
I agree with @milancurcic: `stdlib_comparison` or `stdlib_comparisons` would be better thatn `stdlib_tolerant`. However, they don't cover `tfloor`,,... What do you think about `stdlib_operators`?

General comments:
* Some hyperlinks mst still be added in the code, but this could be done at the end of the PR
* `fypp` could be used to generated the code for the different kinds of `real`.

awvwgk:
With https://github.com/fortran-lang/stdlib/pull/488 this functionality was implemented, therefore I'm going to close this PR.

Feel free to reopen this PR if you think #488 doesn't address all changes included in this patch.

PR #363, opened 2021-03-26
Added sort to stdlib_string_type.f90 file

ChetanKarwa:
I have implemented a sort function for a string. This function sorts the string's characters according to their ASCII values.

Eg: 
string - "This is to be sorted" 
sorted string - "    Tbdeehiioorssstt"

### DIFF
--- a/doc/specs/stdlib_ascii.md
+++ b/doc/specs/stdlib_ascii.md
@@ -169,3 +169,39 @@ program demo_reverse
     print'(a)', reverse("Hello, World!") ! returns "!dlroW ,olleH"
 end program demo_reverse
 ```
+
+### `sort`
+
+#### Status
+
+Experimental
+
+#### Description
+
+Sorts all the characters in the input character type according to their ascii values.
+
+#### Syntax
+
+`res = [[stdlib_ascii(module):sort(function)]] (string)`
+
+#### Class
+
+Pure function.
+
+#### Argument
+
+`string`: shall be an intrinsic character type. It is an `intent(in)` argument.
+
+#### Result value
+
+The result is an intrinsic character type of the same length as `string`.
+
+#### Example
+
+```fortran
+program demo_sort
+    use stdlib_ascii, only : sort
+    implicit none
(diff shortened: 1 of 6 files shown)


ivan-pi:
Hello @ChetanKarwa,

great to see you managed to build stdlib and get your new function working.

However, I am not totally convinced by the purpose of this function. The more common usage of a sort function would be to sort a list of strings (like the one in development in #311), say to output a sorted list of labels.

Do you have some usage cases that motivate a character sort function?

ChetanKarwa:
I implemented this because I've seen other languages supporting the sorting of a character array.
But they support a general sort function they don't have a separate sort function for the character array and that function is slow [time complexity - O(N*logN)]. So I thought of implementing something different for Fortran stdlib and so the one implemented in this PR is O(N).
Now giving a second thought to it currently I really can't figure out any use case of this function.

jvdp1:
@wclodius2 Is your implementation related to this PR?

wclodius2:
No, it isn't. My sort is of an array of elements of the `string_type` and not of the characters of a `string_type` scalar, or of a list of elements of the `string_type`. FWIW I also view sorting of the characters as of very limited use, and sorting a list rather than an array would require aa drastic rewrite of my code.

jvdp1:
I renamed this PR to avoid confusion with #408

ivan-pi:
What could be done is to provide the O(n) sorting algorithm of @ChetanKarwa as a specialization in #408 when the input is an array of `character(len=1)`, but I don't expect this to be used very often. 

The only usage I can think of currently is to find the [median](https://en.wikipedia.org/wiki/Median) letter (and maybe the [mode](https://en.wikipedia.org/wiki/Mode_(statistics))) in a character sequence.

awvwgk:
Also, consider that this implementation is likely to fail for non-ASCII input:

```f90
print *, sort("Ångström")
! Fortran runtime error: Index '195' of dimension 1 of array 'count' above upper bound of 128
```

Issue #298, opened 2021-01-16
Intel Fortran and -standard-semantics


awvwgk:
Compiling stdlib with Intel Fortran compilers sets the `-standard-semantics` flag. Unfortunately, this requires all Fortran projects in the dependency tree compiled by Intel Fortran to inherit this compilation flag since it changes the name mangling of the symbols. Setting the `-standard-semantic` flag by `add_compile_options` in the CMake build files makes it difficult to overwrite by dependent projects.

Is there a reason why stdlib requires `-standard-semantics`? Would it be possible to make this optional for projects that want to use the stdlib project but don't want to use `-standard-semantics` flag?

Romendakil:
This is for backward compatibility with Ifort 17 (and earlier). Ifort 18 and newer doesn’t need this any more. It is mainly for the automatic allocation I believe.

awvwgk:
In this case `-assume realloc_lhs` would be a more suitable option for Intel 17 and earlier without changing the name mangling for dependent project.

Also, I tested compiling stdlib with a range of Intel versions and found it can only be compiled with Intel 19 or newer right now (_notes to open an issue on this as well_) .

jvdp1:
I don't remember the reason. Was it @MarDiehl who introduced this option?

[Here ](https://community.intel.com/t5/Intel-Fortran-Compiler/standard-semantics-and-linking-with-external-libs/m-p/1174941/highlight/true#M147103) is a possible solution. Following this post, adding the option "-assume nostd_mod_proc_name" could solve the issue.

However, since it doesn't seem possible to compile stdlib with Intel ifort 18 and lower, is this option still a problem? On the other hand, it could be removed if not appropriate anymore for the newest versions of ifort.

awvwgk:
blame says it was introduced in https://github.com/fortran-lang/stdlib/pull/205 with 39bb56cb2a6738433343dc695353e3bd90a79b9b.

MarDiehl:
The general idea was to check for standard conformance. I was not aware of the fact that this also implies a different scheme for name mangling.  The name `-assume nostd_mod_proc_name` indicates that Intel thinks that this mangling scheme is standard conforming. I would assume that this is related to interoperability with C.

awvwgk:
What would be the best strategy to move forward here? Should the `-standard-semantic` be kept, removed from the options or potentially made optional?

Romendakil:
I would stick to the compiler default, so remove the flag.

PR #367, opened 2021-03-28
Add Intel compiler workflow for OSX

awvwgk:
Expand our compiler testing with Intel oneAPI classic 2021.1 on OSX. It also removes the `-standard-semantics` flag from the options for Intel Fortran.

GitHub actions workflow is taken from https://github.com/oneapi-src/oneapi-ci (MIT licensed).

Closes https://github.com/fortran-lang/stdlib/issues/369
Closes https://github.com/fortran-lang/stdlib/issues/298


jvdp1 (approved):
LGTM. It can be merged IMO

PR #372, opened 2021-03-30
To_Title is corrected

ChetanKarwa:
The to_title function was not working properly it is working like to_sentence:

input:- this text is some title
earlier output:- This text is some title
expected output:- This Text Is Some Title

I made the necessary changes.
Check it out.

### DIFF
--- a/src/stdlib_ascii.f90
+++ b/src/stdlib_ascii.f90
@@ -289,14 +289,14 @@ contains
                     title_string(i:i) = char_to_upper(string(i:i))
                     capitalize_flag = .FALSE.
                 else
-                    title_string(i:i) = string(i:i)
+                    title_string(i:i) = char_to_lower(string(i:i))
                 end if
             else
                 if(string(i:i)==" ") then
-                    title_string(i:i) = string(i:i)
+                    title_string(i:i) = char_to_lower(string(i:i))
                     capitalize_flag = .TRUE.
                 else
-                    title_string(i:i) = string(i:i)
+                    title_string(i:i) = char_to_lower(string(i:i))
                 end if
             end if
         end do
--- a/src/stdlib_ascii.f90
+++ b/src/stdlib_ascii.f90
@@ -277,21 +277,28 @@ contains
     pure function to_title(string) result(title_string)
         character(len=*), intent(in) :: string
         character(len=len(string)) :: title_string
+        logical:: capitalize_flag
         integer :: i, n
 
         n = len(string)
+        capitalize_flag = .TRUE.
         do i = 1, len(string)
-            if (is_alphanum(string(i:i))) then
-                title_string(i:i) = char_to_upper(string(i:i))
-                n = i
-                exit
-            else
-                title_string(i:i) = string(i:i)
-            end if
-        end do
 
-        do i = n + 1, len(string)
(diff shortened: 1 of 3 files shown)


Aman-Godara:
Titled version: The first alphabetical character of the input character sequence is transformed to uppercase unless it follows a numeral and the rest of the characters in the sequence are transformed to lowercase.

Please check this comment by arjenmarkus to understand this better : https://github.com/fortran-lang/stdlib/pull/346#commitcomment-48441307

ChetanKarwa:
Ok I get that. 
That is what we have implemented and written in description but the common implementation is capitalising each word. (There are exceptional words) 
Same is implemented in python language and on various online sentence to title convertors.

I have not checked in other languages but I suggest we should also implement to_title the similar way and the current implementation can be called 'to_sentence' or any other name that suits the operation. 

Just a suggestion to land on common ground with other languages.

jvdp1:
> I have not checked in other languages but I suggest we should also implement to_title the similar way and the current implementation can be called 'to_sentence' or any other name that suits the operation

I didn't think about it while reading the specs. However, I may agree with this comment. @awvwgk what do you think about it?

awvwgk:
I just checked python strings and they have both `.capitalize()` and `title()`, the current `to_title` implementation is a version of `capitalize` under the wrong name. `to_sentence` would be a better name and we can also have an actual `to_title` function.

@ChetanKarwa Would you adjust your patch to rename the current `to_title` function as `to_sentence` and than we can discuss the actual `to_title` implementation here?

ChetanKarwa:
Sure will do that.

ChetanKarwa:
For now, the PR holds the to_title() function and so I would go forward and implement the to_sentence() separately so that in case we all come down to the conclusion of implementing to_title() this can be used.

src/stdlib_ascii.f90
```diff
@@ -277,23 +277,30 @@ end function to_upper
...
+            if(capitalize_flag) then
+                if (is_alphanum(string(i:i))) then
+                    title_string(i:i) = char_to_upper(string(i:i))
+                    capitalize_flag = .FALSE.
+                else
+                    title_string(i:i) = char_to_lower(string(i:i))
+                end if
             else
-                title_string(i:i) = string(i:i)
+                if(string(i:i)==" ") then
```
awvwgk, line 295: This could be more general than just spaces, every non-word boundary might trigger capitalization, consider:
`to_title(".this.sentence.is.separated.with.dots.")`, would we expect `.This.sentence.is.separated.with.dots.` or  `.This.Sentence.Is.Separated.With.Dots.` for a titlecase return value?

jvdp1: Hmmm! I am not sure what to expect in such a case. My favor would go to capitalize each word. What do other languages?

Aman-Godara: I checked for python and realised that every character which is not an alphabet triggers capitalization including integers. Python mentions that even apostrophe(`'`) triggers this capitalization which is undesirable. May be we can take an optional set as the argument from the user which will consist of characters the user wants to the trigger capitalization for..

Aman-Godara:
I am assigning this task to myself as well.

PR #384, opened 2021-04-10
Implement starts_with and ends_with functions

awvwgk:
This patch implements a simple function to check whether a string starts or ends with a substring, which is a special variant of the intrinsic `index`. A similar `ends_with` function is currently used in the `fpm_strings` module.

```f90
program demo
  use stdlib_strings, only : starts_with, ends_with
  implicit none
  print '(a)', starts_with("pattern", "pat")  ! T
  print '(a)', starts_with("pattern", "ern")  ! F
  print '(a)', ends_with("pattern", "ern")  ! T
  print '(a)', ends_with("pattern", "pat")  ! F
end program demo
```

### DIFF
--- a/doc/specs/stdlib_strings.md
+++ b/doc/specs/stdlib_strings.md
@@ -108,3 +108,87 @@ program demo
   print'(a)', chomp("hello", substring="lo")   ! "hel"
 end program demo
 ```
+
+
+<!-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -->
+### `starts_with`
+
+#### Description
+
+Check if a *string* starts with a given *substring*.
+
+#### Syntax
+
+`string = [[stdlib_strings(module):starts_with(interface)]] (string, substring)`
+
+#### Status
+
+Experimental
+
+#### Class
+
+Pure function.
+
+#### Argument
+
+- `string`: Character scalar or [[stdlib_string_type(module):string_type(type)]].
+  This argument is intent(in).
+- `substring`: Character scalar or [[stdlib_string_type(module):string_type(type)]].
+  This argument is intent(in).
+
+#### Result value
+
+The result is of scalar logical type.
+
+#### Example
+
+```fortran
+program demo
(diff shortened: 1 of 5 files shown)


LKedward (approved):
Yep this is a simple but useful one to have. Implementation and spec look good to me.

jvdp1 (approved):

ivan-pi:
The Python versions of these functions also support some optional arguments (but I guess they are redundant since the same thing can be achieved with slicing):

> `str.startswith(prefix[, start[, end]])`
>
>    Return True if string starts with the prefix, otherwise return False. prefix can also be a tuple of prefixes to look for. With optional start, test string beginning at that position. With optional end, stop comparing string at that position.

> `str.endswith(suffix[, start[, end]])`
>
>    Return True if the string ends with the specified suffix, otherwise return False. suffix can also be a tuple of suffixes to look for. With optional start, test beginning at that position. With optional end, stop comparing at that position.

Would there be any interest to have these functions also or only as operators? The Python functions are methods of the string clas which makes the logic flow nicer:
```
>>> a = "Hello, World"
>>> a.startswith("Hello")
True
>>> a[2:].startswith("Hello")
False
```

In Fortran this could become `a .startswith. "Hello"`. Apart from the clunky import syntax ( `use stdlib_strings, only: operator(.startswith.), operator(.endswith.) I think I would prefer this vs the function call. 

Julia also has [`startswith`](https://docs.julialang.org/en/v1/base/strings/#Base.startswith) and [`endswith`](https://docs.julialang.org/en/v1/base/strings/#Base.endswith). I might also prefer to go without the underscore like Python and Julia, even though I know we've added it elsewhere.

milancurcic (approved):
Useful addition, thank you!

milancurcic:
:+

certik (approved):
That looks good. I think I needed these in `fpm`. I use these in Python and also I had to implement in C++ for LFortran. So they are very useful.

awvwgk:
Thanks everybody for the feedback. With four approvals I'll go ahead and merge.

PR #388, opened 2021-04-12
Changed filenames

wclodius2:
Changed filenames in test_stdlib_bitsets_64.f90 and test_stdlib_bitsets_large.f90 so that they were no longer the same. Also removed spurious blanks in error stops.

[ticket: X]


awvwgk (commented):
Would open with `status='scratch'` and `rewind(unit)` instead of closing and reopening a named file here work better?

wclodius2:
I have never before had a need for scratch files, but this does sound like an appropriate use for them.

awvwgk (approved):
This looks good to me.

LKedward (approved):
LGTM

PR #407, opened 2021-05-07
changed to_title to to_sentence and implemented correct to_title

Aman-Godara:
Status: Code is open for review
Concerning PR: #372 

Tasks:
- [x] implemented correct to_title & changed to_title to to_sentence
- [x] written/updated tests for both to_title & to_sentence
- [x] Documentation
- [x] Added module dependency of stdlib_math on stdlib_kinds in src/Makefile.manual

Closes https://github.com/fortran-lang/stdlib/pull/372

### DIFF
--- a/doc/specs/stdlib_string_type.md
+++ b/doc/specs/stdlib_string_type.md
@@ -1130,7 +1130,8 @@ end program demo
 
 #### Description
 
-Returns a new string_type instance which holds the lowercase version of the character sequence hold by the input string.
+Returns a new string_type instance which holds the lowercase version of the 
+character sequence hold by the input string.
 
 #### Syntax
 
@@ -1150,7 +1151,7 @@ Elemental function.
 
 #### Result Value
 
-The Result is a scalar `string_type` value.
+The result is a scalar `string_type` value.
 
 #### Example
 
@@ -1175,7 +1176,8 @@ end program demo
 
 #### Description
 
-Returns a new string_type instance which holds the uppercase version of the character sequence hold by the input string.
+Returns a new string_type instance which holds the uppercase version of the 
+character sequence hold by the input string.
 
 #### Syntax
 
@@ -1195,7 +1197,7 @@ Elemental function.
 
 #### Result Value
 
-The Result is a scalar `string_type` value.
+The result is a scalar `string_type` value.
 
 #### Example
 
@@ -1220,9 +1222,13 @@ end program demo
 
 #### Description
 
-Returns a new string_type instance which holds the titlecase (or capitalized) version of the character sequence hold by the input string.
-Capitalized version: The first alphabetical character of the input character sequence is transformed to uppercase unless it 
-follows a numeral and the rest of the characters in the sequence are transformed to lowercase.
+Returns a new string_type instance which holds the titlecase version 
+of the character sequence hold by the input string.  
+Title case: First character of every word in the sentence is converted to 
+uppercase and the rest of the characters are converted to lowercase.  
+A word is a contiguous sequence of character(s) which consists of alphabetical 
+character(s) and numeral(s) only and doesn't exclude any alphabetical character 
+or numeral present next to either of its 2 ends.
 
 #### Syntax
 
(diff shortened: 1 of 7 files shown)


src/stdlib_ascii.fypp
```diff
@@ -284,31 +291,58 @@ contains
...
     !> Version: experimental
     pure function to_title(string) result(title_string)
         character(len=*), intent(in) :: string
         character(len=len(string)) :: title_string
+        integer :: i
+        logical :: capitalize_switch
+
+        capitalize_switch = .true.
+        do i = 1, len(string)
+            if (is_alpha(string(i:i))) then
```
awvwgk, line 306: I would prefer to use alphanumerical characters to define words.
```suggestion
            if (is_alphanum(string(i:i))) then
```

The result of ``to_title("1st, 2nd, 3rd")`` with `"1St, 2Nd, 3Rd"` seems surprising to me. Given that `to_sentence currently uses alphanumeric characters to determine the beginning of the first word.

awvwgk (changes requested):

Aman-Godara:
Please rerun the workflow once. I don't think there is any error in the code.

LKedward:
> Please rerun the workflow once. I don't think there is any error in the code.

Yes, this is unrelated to your changes in this PR. It looks like the Makefile is missing a dependency for `stdlib_math`. Adding the following to `Makefile.manual` under `# Fortran module dependencies` (line 57) should fix it:

```makefile
stdlib_math.o: stdlib_kinds.o
```

Aman-Godara:
> ```makefile
> stdlib_math.o: stdlib_kinds.o
> ```
Ok!, I will make the necessary changes.

src/stdlib_string_type.fypp
```diff
@@ -122,6 +122,14 @@ module stdlib_string_type
         module procedure :: to_title_string
     end interface to_title
 
+    !> Returns the sentencecase version of the character sequence hold by the input string
+    !>
+    !> This method is Elemental and returns a new string_type instance which holds this
```
jvdp1, line 127:
```suggestion
    !> This method is elemental and returns a new string_type instance which holds this
```

src/stdlib_string_type.fypp
```diff
@@ -122,6 +122,14 @@ module stdlib_string_type
...
+    !> sentencecase character sequence
+    interface to_sentence
+        module procedure :: to_sentence_string
+    end interface to_sentence
+
     !> Reverses the character sequence hold by the input string
     !> 
     !> This method is Elemental and returns a new string_type instance which holds this
```
jvdp1, line 135:
```suggestion
    !> This method is elemental and returns a new string_type instance which holds this
```

src/tests/string/test_string_functions.f90
```diff
@@ -27,12 +27,21 @@ end subroutine test_to_upper_string
 
     subroutine test_to_title_string
         type(string_type) :: test_string, compare_string
-        test_string = "_#To tiTlE !$%-az09AZ"
-        compare_string = "_#To title !$%-az09az"
+        test_string = "tO t!TL3 7h1S p#ra$e"
+        compare_string = "To T!Tl3 7h1s P#Ra$E"
```
jvdp1, line 31: Why are the characters "_" and "#" removed?

jvdp1 (commented):
It sounds good to me.
The specs for `to_title` must be adapted.
The specs for `to_sentece` must be added.

src/tests/string/test_string_functions.f90
Aman-Godara: I removed `_#` from the start of the sentence to not give an impression, on someone who is reading the test case, that it is the `_#` which is controlling capitalization of `T`. First alphanum character of the sentence will always be capitalized whether is it is preceded by a non-alphanum character or not. So in short I wanted to improve test case quality because of which I chose to do so, otherwise there is no specific reason for doing so. Please let me know if there are any other ways in which I can further improve this test case.

jvdp1: Maybe you could add the test case with "_#". So someone who would read the tests would see the behaviour for both cases.

Aman-Godara: added `_#` in the test case for to_title.

awvwgk (approved):

doc/specs/stdlib_string_type.md
```diff
@@ -1242,31 +1248,80 @@ Elemental function.
...
 
-The Result is a scalar `string_type` value.
+The result is a scalar `string_type` value.
 
 #### Example
 
 ```fortran
-program demo
+program demo_to_title
   use stdlib_string_type
```
jvdp1, line 1257:
```suggestion
  use stdlib_string_type, only: string_type, to_title
```

doc/specs/stdlib_string_type.md
```diff
@@ -1242,31 +1248,80 @@ Elemental function.
...
+
+#### Result Value
+
+The result is a scalar `string_type` value.
+
+#### Example
+
+```fortran
+program demo_to_sentence
+  use stdlib_string_type
```
jvdp1, line 1305:
```suggestion
  use stdlib_string_type, only: string_type, to_sentence
```

jvdp1 (approved):
Thanks! Approved pending small suggestions

awvwgk:
Thanks for this patch. I'll go ahead and merge.

PR #408, opened 2021-05-12
Sorting2

wclodius2:
Merged inconsistent changes in doc/specs/stdlib_sorting.mb and src/stdlib_sorting.fypp.


gareth-nx (approved):
Looks great -- noting I already reviewed almost the same code at [in the related pull request](https://github.com/fortran-lang/stdlib/pull/386). I hope we get some more reviewers and this can be merged.

src/stdlib_sorting.fypp
```diff
@@ -0,0 +1,573 @@
...
+
+    use stdlib_string_type, only: string_type, assignment(=), operator(>), &
+        operator(>=), operator(<), operator(<=)
+
+    implicit none
+    private
+
+    public :: int_size
+
+    integer, parameter :: int_size = int64 !! Integer kind for indexing
```
jvdp1, line 120:
```suggestion
      integer, parameter, public :: int_size = int64 !! Integer kind for indexing
```

Matter of taste, I guess ;)

src/stdlib_sorting.fypp
```diff
@@ -0,0 +1,573 @@
...
+!!    ! Find the indices to sort the component
+!!        call sort_index(a(1:size(a_data)), index(1:size(a_data)),&
+!!                        work(1:size(a_data)/2), iwork(1:size(a_data)/2))
+!!    ! Sort a_data based on the sorting of that component
+!!        a_data(:) = a_data( index(1:size(a_data)) )
+!!    end subroutine sort_a_data
+!!```
+
+#:for k1 in INT_KINDS
+    public ${k1}$_ord_sort
```
jvdp1, line 326: Why are all these specific procedures `public`? Is the `public` `interface` `ord_sort` not enough?

src/stdlib_sorting_ord_sort.fypp
```diff
@@ -0,0 +1,1419 @@
+#! Integer kinds to be considered during templating
+#:set INT_KINDS = ["int8", "int16", "int32", "int64"]
+
+#! Real kinds to be considered during templating
+#:set REAL_KINDS = ["sp", "dp", "qp"]
```
jvdp1, line 5:
```suggestion
#:include "common.fypp"
```
These variables are already defined in `common.fypp`

src/stdlib_sorting_ord_sort.fypp
```diff
@@ -0,0 +1,1419 @@
...
+
+submodule(stdlib_sorting) stdlib_sorting_ord_sort
+
+    implicit none
+
+contains
+
+#:for k1 in INT_KINDS
+
+    module subroutine ${k1}$_ord_sort( array, work )
```
jvdp1, line 64: Could this subroutine be made `pure`? Why `merge_sort` cannot be pure?

src/stdlib_sorting_ord_sort.fypp
```diff
@@ -0,0 +1,1419 @@
...
+        integer :: stat
+
+        if ( present(work) ) then
+! Use the work array as scratch memory
+            call merge_sort( array, work )
+        else
+! Allocate a buffer to use as scratch memory.
+            array_size = size( array, kind=int_size )
+            allocate( buf(0:array_size/2-1), stat=stat )
+            if ( stat /= 0 ) error stop "Allocation of buffer failed."
```
jvdp1, line 98:
```suggestion
            if ( stat /= 0 ) error stop "${k1}$_ord_sort: Allocation of buffer failed."
```
To be a bit more explicit

src/stdlib_sorting_ord_sort.fypp
```diff
@@ -0,0 +1,1419 @@
...
+            num = n
+            r = 0_int_size
+
+            do while( num >= 64 )
+                r = ior( r, iand(num, 1_int_size) )
+                num = ishft(num, -1_int_size)
+            end do
+            min_run = num + r
+
+        end function calc_min_run
```
jvdp1, line 122: The function `calc_min_run` appears multiple times, while it seems to be always the same. What is the reason to replicate it multiple times? Could it be kept outside the scope of a procedure?

src/stdlib_sorting_ord_sort.fypp
jvdp1, line 64: The `integer` and `real` versions for `ord_sort` seem to be the same code (except for the type of the array). Therefore, I think that there could be only one copy of the code, and `fypp` could generate both codes. This could simplify maintainance.

I can submit a proposition with `fypp` instructions, if desired.

src/stdlib_sorting_ord_sort.fypp
```diff
@@ -0,0 +1,1419 @@
...
+! it drops both the use of 'galloping' to identify bounds of regions to be
+! sorted and the estimation of the optimal `run size`. However it remains
+! a hybrid sorting algorithm combining an iterative Merge sort controlled
+! by a stack of `RUNS` identified by regions of uniformly decreasing or
+! non-decreasing sequences that may be expanded to a minimum run size and
+! initially processed by an insertion sort.
+!
+! Note the Fortran implementation simplifies the logic as it only has to
+! deal with Fortran arrays of intrinsic types and not the full generality
+! of Rust's arrays and lists for arbitrary types. It also adds the
```
jvdp1, line 1098: This section is not really true in the context of `string_type`, right?

jvdp1 (commented):
Thank you for these implementations. Here are some general comments. Based on your answers , I will continue the review.

w6ws:
A few API suggestions:

1.) In subroutine sort_index - the index, work, and iwork arrays should probably be intent(out), not intent(inout).  Same with the work array in ord_sort.

2.) All three subroutines should have an optional return status that indicates success or failure.  Right now there are some hard stops in the code - specifically when allocate is called.  It would be good to have the ability to avoid these.

3.) Shouldn't all three have a 'reverse' argument so the caller can specify either ascending or decending sort order?

FWIW, in some code I've worked on, we use private types with public parameters for such things.  E.g.,:

module sorting_mod
  implicit none
  private
  ...
  type :: sortFlag_t  ! private type
    logical :: descending
  end type

  type(sortFlag_t), parameter, public ::  &
      SORTFLAG_ASCENDING  = sortFlag_t (.false.),  &
      SORTFLAG_DESCENDING = sortFlag_t (.true.)

While in the case of sort_index there are only two options, so not strictly needed, the technique is general to other areas of a library where many more options for an argument might be offered.  It ensures that the compiler can enforce correct option selections at compile time.  And it provides clear indication of what is being specified - rather than some "magic number".  It also removes some need for run-time validation checks.

wclodius2:
I am on travel and it is awkward for me to find the time to do the changes, but I will see what I can do.

src/stdlib_sorting.fypp
wclodius2: I agree with your taste. Done!

src/stdlib_sorting.fypp
wclodius2: My original `test_sorting.f90` passed  the sorting procedure as an argument to a testing procedure. That require that the specific procedures, at least for `int32`, be available. As the API changed the sorting procedure argument to the testing procedure became impractical, and I no longer needed the specific procedures. I have removed them from the public interface in both `stdlib_sorting.fypp` and `stdlib_sorting.md`.

src/stdlib_sorting_ord_sort.fypp
wclodius2: Done!

src/stdlib_sorting_ord_sort.fypp
wclodius2: If we just let tha allocation fail on its own then the statement would be eliminated. For the moment i have made the change you suggested.

src/stdlib_sorting_ord_sort.fypp
wclodius2: I thought `calc_min_run` was more likely to be inlined if made an internal procedure, but as it is invoked only once at the start of the main procedure, inlining only has a very minor impact on performance.

src/stdlib_sorting_ord_sort.fypp
wclodius2: I am skeptical that they can be combined.  The use of `FYPP` to generate but the names of procedures and the types of arguments to the procedures seemed to me to make that impractical If you have a concrete proposition show it and I will see what I can do.

wclodius2: "...to generate both the names..."

src/stdlib_sorting_ord_sort.fypp
wclodius2: Its still true, but not as strongly for the string type. The full generality of the `Rust` code is awfully complicated.

wclodius2:
I have committed and pushed changes suggested by Jeremie.

src/stdlib_sorting_ord_sort.fypp
jvdp1: If it is in the same module, I would expect that the compiler would inline it (if possile of course), even if it is not an internal procedure. I will leave this comment open for now.

src/stdlib_sorting_ord_sort.fypp
jvdp1: I will look at it and make a proposition if satisfied.

jvdp1:
>I am skeptical that they can be combined. The use of FYPP to generate but the names of procedures and the types of arguments to the procedures seemed to me to make that impractical If you have a concrete proposition show it and I will see what I can do.


@wclodius2
I modified your modules and submodules related to sorting algorithms, such that only one peace of code is used to generate the procedures for `integer`, `real`, and `string_type`. The source code is quite similar to what you proposed and much shorter.  
The generated code is the same as as the original generated code.
I opened a PR with my modifications against your own branch. See [here](https://github.com/wclodius2/stdlib/pull/1) for more details. If you agree with these changes, you can merge the PR, such that the changes appear to this PR too.

Note:

* Only one `int64` was modified to `int_size` (but I believe it should have been like that).
* The name of the procedures realted to `string_type` was change from `string_xxx`to `string_type_xxx`.

With this approach, it will also allow to easily extend the sorting procedures to any new derived types that are associated with operations ">", and so on.

Please let me know if this approach sounds good for you.

src/stdlib_sorting_ord_sort.fypp
jvdp1: See proposition in this [PR](https://github.com/wclodius2/stdlib/pull/1)

wclodius2:
I have pulled the PR into the current branch.

jvdp1:
> I have pulled the PR into the current branch.

Good. I will continue the review of the code!

src/stdlib_sorting_ord_sort.fypp
jvdp1: Not relevant anymore due to addition for Fypp pre-processing for generating code for integer, real, and string_type

src/stdlib_sorting_ord_sort.fypp
```diff
@@ -0,0 +1,738 @@
...
+!! https://svn.python.org/projects/python/trunk/Objects/listsort.txt
+            integer(int_size)             :: min_run
+            integer(int_size), intent(in) :: n
+
+            integer(int_size) :: num, r
+
+            num = n
+            r = 0_int_size
+
+            do while( num >= 64 )
```
jvdp1, line 453: Is the value `64` depending on the kind of `r` (i.e., `int_size`)? What happen is `int_size` become `int32`?

src/stdlib_sorting_sort.fypp
```diff
@@ -0,0 +1,401 @@
...
+! `quicksort` is converging too slowly the algorithm resorts
+! to `heapsort`. The algorithm is of order O(N Ln(N)) for all inputs.
+! Because it relies on `quicksort`, the coefficient of the O(N Ln(N))
+! behavior is typically small compared to other sorting algorithms.
+
+        character(len=*), intent(inout) :: array(0:)
+
+        integer(int32) :: depth_limit
+
+        depth_limit = 2 * int( floor( log( real( size( array, kind=int64 ),  &
```
jvdp1, line 251:
```suggestion
        depth_limit = 2 * int( floor( log( real( size( array, kind=int_size ),  &
```

jvdp1 (commented):
Still a few additional comments. Otherwise LGTM. Thank you.

src/stdlib_sorting_sort.fypp
wclodius2: I want to fetch and merge (pull) the most recent changes into my local repository. what should I use for the <repository> in the `git pull` command?

wclodius2:
@w6ws sorry for my late response, but I have been busy on my trip.

1. I am going to defer to @jvdp1 as to the intents of the arguments.
2. The errors can only be "caught" for a memory allocation problem. With lazy allocation on Linux. and virtual memory on all systems, it is not clear what can cause an allocation error, or that such an error is recoverable. I am right now inclined  to remove the `stat` argument to the `allocate` statements and simply let a memory error halt processing and let the processor report the problem if it occurs. This will allow my procedures to be pure.
3. The reverse argument is only needed for the sort_index procedure, because it is intended to be a stable sort. A stable reverse sort requires that the array to be sorted be reversed on entry and then the sorted `index` array be reversed on return. The other two procedures can be handled by simply reversing the sorted arrays. I mam willing to consider including a `reverse` procedure in the `stdlib_sorting` module for that purpose.

I assume that the rest of the message is informative only, and doesn't require a response.

src/stdlib_sorting_sort.fypp
jvdp1: in your local directory, assuming that `origin` refers to `https://github.com/wclodius2/stdlib`, run:
```shell
git pull origin sorting2
```
to get the latest changes done in the remote repo.

jvdp1: Note that you can first commit my suggestions, directly by blicking on the button "commit suggestion".

w6ws:
> @w6ws sorry for my late response, but I have been busy on my trip.
> 
> 1. I am going to defer to @jvdp1 as to the intents of the arguments.

In the case of sort_index, the index array should be intent(out).  Intent(out) is also appropriate for 'throw away' scratch arrays.    If the Standards committee had followed through with intent(scratch), it would have been even better.

> 2. The errors can only be "caught" for a memory allocation problem. With lazy allocation on Linux. and virtual memory on all systems, it is not clear what can cause an allocation error, or that such an error is recoverable. I am right now inclined  to remove the `stat` argument to the `allocate` statements and simply let a memory error halt processing and let the processor report the problem if it occurs. This will allow my procedures to be pure.

In general I like library routines to return control to their caller when they detect errors - then let the caller decide what to do.  For example in the case of sort_ord and sort_index, besides allocation errors, checks could also be made to ensure the scratch and output arrays are large enough.  Just makes things more user friendly.  But good point on enabling the procedures to be pure.

> 3. The reverse argument is only needed for the sort_index procedure, because it is intended to be a stable sort. A stable reverse sort requires that the array to be sorted be reversed on entry and then the sorted `index` array be reversed on return. The other two procedures can be handled by simply reversing the sorted arrays. I mam willing to consider including a `reverse` procedure in the `stdlib_sorting` module for that purpose.
> 
> I assume that the rest of the message is informative only, and doesn't require a response.

Seems like a pain to require the manual steps of reversing inputs and outputs to get a descending order.  Especially since, perhaps stability aside, it is only a matter of which comparison operator to use within the sorting algorithms.  Again, user-friendliness.

My comments on the 'reverse' argument is with an eye to future expansion.  For example, if it were desired to enhance the sort routines to accept sort order choices like ASCENDING, ASCENDING_STABLE, DESCENDING, DESCENDING_STABLE.  Maybe even specify alternative algorithms like ASCENDING_RADIX and DESCENDING_RADIX, etc.  A simple argument of type logical doesn't handle more than two choices.  And an integer argument would need to be checked at run-time.  Doing it as I described allows compile time enforcement of the only supported choices.

jvdp1:
@wclodius2 @w6ws 

> > @w6ws sorry for my late response, but I have been busy on my trip.
> > 
> > 1. I am going to defer to @jvdp1 as to the intents of the arguments.
> 
> In the case of sort_index, the index array should be intent(out). Intent(out) is also appropriate for 'throw away' scratch arrays. If the Standards committee had followed through with intent(scratch), it would have been even better.

I agree with @w6ws . It would make sense that `index` and all work arrays are `intent(out)`.

> > 2. The errors can only be "caught" for a memory allocation problem. With lazy allocation on Linux. and virtual memory on all systems, it is not clear what can cause an allocation error, or that such an error is recoverable. I am right now inclined  to remove the `stat` argument to the `allocate` statements and simply let a memory error halt processing and let the processor report the problem if it occurs. This will allow my procedures to be pure.
> 
> In general I like library routines to return control to their caller when they detect errors - then let the caller decide what to do. For example in the case of sort_ord and sort_index, besides allocation errors, checks could also be made to ensure the scratch and output arrays are large enough. Just makes things more user friendly. But good point on enabling the procedures to be pure.

The topic of handling errors come often in `stdlib`. Would it be an idea to add an `intent(out)` integer (e.g., called `info`, as in BLAS/LAPACK), that would return a value different than 0, with allocation erros, and on checks on scratch and output arrays? With such an approach, the `stat` argument in the `allocate` statement could be kept, while `error stop` could be removed, and the procedure be `pure`.

src/stdlib_sorting_ord_sort.fypp
wclodius2: No the 64 is independent of the kind of integer. It is an upper limit on the block size to be used for insertion sorts prior to using the heap sort. It depends on the relative processing speed of insertion sort versus heap sort. Ideally you divide the input array into a power of two number of blocks, but it is better to have sightly less than a power of two than slightly more. Tim Peters did extensive testing with `num >= 16`, `num >= 32`, `num >= 64`, and `num >= 128`, and found that the performance was fairly flat for the different cutoffs. Of course his testing was with Python, and the ideal cutoff may be language dependent. Once we have the API fixed we could run our own tests for different cutoffs and see which cutoff is best.

wclodius2:
I have pulled the changes of @jvdp1 and edited the results to incorporate the changes in intent suggested by @w6ws, committed the changes and pushed them with the command

    git push william

The push seemed to go well, but it did not result in an update check on this page. I assume that means the the update did not occur in sorting2. What can I do to update sorting2?

src/stdlib_sorting_ord_sort.fypp
jvdp1: Thank you for the explanation.

jvdp1:
@wclodius2 I can see your changes regarding `intent` (in addition to 3 other commits based on my suggestions). It can be found [here](https://github.com/fortran-lang/stdlib/pull/408/commits/8e59d2f30bb21e90a896d01a831c04291cbf0a38).

wclodius2:
@jvdp1 can you approve the changes I submitted [here](https://github.com/fortran-lang/stdlib/pull/408/commits/8e59d2f30bb21e90a896d01a831c04291cbf0a38)?

jvdp1:
> @jvdp1 can you approve the changes I submitted [here](https://github.com/fortran-lang/stdlib/pull/408/commits/8e59d2f30bb21e90a896d01a831c04291cbf0a38)?

@wclodius2 the changes suggested by @w6ws regarding `intent(out)` are available in this PR. LGTM. Thank you.

wclodius2:
> @wclodius2 @w6ws
> 
> > > @w6ws sorry for my late response, but I have been busy on my trip.
> > > 
> > > 1. I am going to defer to @jvdp1 as to the intents of the arguments.
> > 
> > 
> > In the case of sort_index, the index array should be intent(out). Intent(out) is also appropriate for 'throw away' scratch arrays. If the Standards committee had followed through with intent(scratch), it would have been even better.
> 
> I agree with @w6ws . It would make sense that `index` and all work arrays are `intent(out)`.

Done.

> 
> > > 1. The errors can only be "caught" for a memory allocation problem. With lazy allocation on Linux. and virtual memory on all systems, it is not clear what can cause an allocation error, or that such an error is recoverable. I am right now inclined  to remove the `stat` argument to the `allocate` statements and simply let a memory error halt processing and let the processor report the problem if it occurs. This will allow my procedures to be pure.
> > 
> > 
> > In general I like library routines to return control to their caller when they detect errors - then let the caller decide what to do. For example in the case of sort_ord and sort_index, besides allocation errors, checks could also be made to ensure the scratch and output arrays are large enough. Just makes things more user friendly. But good point on enabling the procedures to be pure.
> 
> The topic of handling errors come often in `stdlib`. Would it be an idea to add an `intent(out)` integer (e.g., called `info`, as in BLAS/LAPACK), that would return a value different than 0, with allocation erros, and on checks on scratch and output arrays? With such an approach, the `stat` argument in the `allocate` statement could be kept, while `error stop` could be removed, and the procedure be `pure`.

I had the impression that @w6ws wanted the argument to be optional, in which case I would want the `ERROR STOP` to be present and we would not gain purity.

Did you have any opinion on the @w6ws proposal for a `reverse` optional argument? I am concerned at the impact on the complexity of the  API, and also on its impaction my time. I want to get this PR done, though the   error  handling is the  bigger  potential  project.

wclodius2:
> > @w6ws sorry for my late response, but I have been busy on my trip.
> > 
> > <snip>
> Seems like a pain to require the manual steps of reversing inputs and outputs to get a descending order. Especially since, perhaps stability aside, it is only a matter of which comparison operator to use within the sorting algorithms. Again, user-friendliness.
> <snip>

While logically the only difference is in the comparison operators in practice it wouldn't be implemented that way as it  would result in a large duplication of code. It would instead be implemented using a reverse procedure.

jvdp1:
> > > > 1. The errors can only be "caught" for a memory allocation problem. With lazy allocation on Linux. and virtual memory on all systems, it is not clear what can cause an allocation error, or that such an error is recoverable. I am right now inclined  to remove the `stat` argument to the `allocate` statements and simply let a memory error halt processing and let the processor report the problem if it occurs. This will allow my procedures to be pure.
> > > 
> > > 
> > > In general I like library routines to return control to their caller when they detect errors - then let the caller decide what to do. For example in the case of sort_ord and sort_index, besides allocation errors, checks could also be made to ensure the scratch and output arrays are large enough. Just makes things more user friendly. But good point on enabling the procedures to be pure.
> > 
> > 
> > The topic of handling errors come often in `stdlib`. Would it be an idea to add an `intent(out)` integer (e.g., called `info`, as in BLAS/LAPACK), that would return a value different than 0, with allocation erros, and on checks on scratch and output arrays? With such an approach, the `stat` argument in the `allocate` statement could be kept, while `error stop` could be removed, and the procedure be `pure`.
> 
> I had the impression that @w6ws wanted the argument to be optional, in which case I would want the `ERROR STOP` to be present and we would not gain purity.

I agree with you: if the argument is optional, then an `error stop` is likely to be needed. However, I would not make such an argument `optional` (similarly to the `info` argument in many procedures of LAPACK, or other `error` arguments in other libraries): up to the user to use/check it or not.

> Did you have any opinion on the @w6ws proposal for a `reverse` optional argument? I am concerned at the impact on the complexity of the API, and also on its impaction my time. I want to get this PR done, though the error handling is the bigger potential project.

I would suggest to keep this implementation of `reverse` for another issue/PR. I also like the approach proposed by @w6ws in [this comment](https://github.com/fortran-lang/stdlib/pull/408#issuecomment-841908209). This would allow different options in the future, like sorting in increasing or decreasing order, and in a stable or unstable approach,... However, this requires more discussions on the API and the approach before implementation, because it could influence future procedures.

@awvwgk @milancurcic any opinions on this PR?

w6ws:
First, thank you for accepting my suggestion on intent(out) stuff!

Second, on the reverse capability - I'd think that if some of the sorts only support sorting in one direction, you'll have a lot of users scratching their heads wondering why.  (Like me.)

Years ago I added a sorting routine to the ESMF library - https://github.com/esmf-org/esmf/blob/develop/src/Infrastructure/Util/src/ESMF_UtilSort.cppF90.  This factored out a bunch of little sorting routines that had been sprinkled throughout the library.  It is based on the LAPACK SLASRT sort.  I templatized it for various data types using the C preprocessor (same technique is also used in many other parts of the ESMF library.)  Taking a look at it, the IF ELSE ENDIF blocks for direction are done at an outermost level.  The loops and interiors of the blocks are identical - except for comparison direction.  A characteristic of the SLASRT-based code is that it doesn't use much temporary storage space.  So passing in, or creating local, work arrays are not needed.

I've also, for personal use, implemented simplified versions of the HPF SORT_UP, SORT_DOWN, GRADE_UP, and GRADE_DOWN functions.  In these cases, the direction of the sort is embedded in the entry point name.  So no conditional code is needed for sort order.  FWIW, I implemented SORT_UP and SORT_DOWN again using SLASRT-based code.  Basically copied the input array to the functions result area - then did the sort on the result area.  I based GRADE_UP and GRADE_DOWN on the Numerical Recipies INDEXX routine.  INDEXX also uses minimal scratch space.  So again, large scratch arrays weren't needed.

Third, on error processing - yes I would suggest making an error return argument optional.  Referring back to the ESMF library again, the coding style we used was to make an optional Return Code (rc=) argument as the last argument of every library routine.  We also strongly encouraged users to check the RC.  If you look at ESMF_UtilSort, it checks for a couple of things and if it finds an error it will (optionally) set the RC to something other than ESMF_SUCCESS, and then return.  So it returns bad results and (optionally) an error RC, rather than a hard stop.  The ESMF philosophy, for better or worse, has been to encourage programmers to always check the RC for success.  I should note that ESMF normally operates in a MPI environment.  Some users routinely run climate and weather models with hundreds or thousands of MPI processes.  So if one thread abends, it can cause a calamity.  We really wanted the users to handle their own errors.

jvdp1:
> While logically the only difference is in the comparison operators in practice it wouldn't be implemented that way as it would result in a large duplication of code. It would instead be implemented using a reverse procedure.

I though a bit about this issue. If the main implementation is to replace all comparison operators (e.g., `>`) by its opposite (e.g., `<`), then I think it could be easily done by adding an additional `fypp` loop. In this case, no code should be duplicated. @wclodius2  I can give a try if it is a possible solution for you.

wclodius2:
> > While logically the only difference is in the comparison operators in practice it wouldn't be implemented that way as it would result in a large duplication of code. It would instead be implemented using a reverse procedure.
> 
> I though a bit about this issue. If the main implementation is to replace all comparison operators (e.g., `>`) by its opposite (e.g., `<`), then I think it could be easily done by adding an additional `fypp` loop. In this case, no code should be duplicated. @wclodius2 I can give a try if it is a possible solution for you.

At least for `stdlib_sorting_ord_sort` more than one form of comparison operator is used, i.e., `>` and `>=`. Other than that you are the expert on `FYPP`and if you think it is straightforward go for it.

jvdp1:
> > > While logically the only difference is in the comparison operators in practice it wouldn't be implemented that way as it would result in a large duplication of code. It would instead be implemented using a reverse procedure.
> > 
> > 
> > I though a bit about this issue. If the main implementation is to replace all comparison operators (e.g., `>`) by its opposite (e.g., `<`), then I think it could be easily done by adding an additional `fypp` loop. In this case, no code should be duplicated. @wclodius2 I can give a try if it is a possible solution for you.
> 
> At least for `stdlib_sorting_ord_sort` more than one form of comparison operator is used, i.e., `>` and `>=`. Other than that you are the expert on `FYPP`and if you think it is straightforward go for it.

@wclodius2 I opened a [PR](https://github.com/wclodius2/stdlib/pull/2) against your branch [sorting2](https://github.com/wclodius2/stdlib/tree/sorting2).
I added:

* an additional option in sort and ord_sort for reverse sorting
* check (error stop) to be sure that it is failing when tests are not succesful

Note: I didn't update the specs, in case the proposed approach is not correct/appropriate.

Also, it could be cleaner to create a new branch (e.g., "sorting3")  from "sorting2", to keep the lattter a bit clean.

jvdp1:
@wclodius2 I opened [this PR](https://github.com/wclodius2/stdlib/pull/3) with some updates in the specs regarding the `reverse` optional arg.

src/stdlib_sorting_ord_sort.fypp
```diff
@@ -0,0 +1,785 @@
...
+        ${t1}$, intent(out), optional :: work(0:)
+        logical, intent(in), optional :: reverse
+
+        logical :: reverse_
+
+        reverse_ = .false.
+        if(present(reverse)) reverse_ = reverse
+
+        if (reverse_) then
+            call ${k1}$_decrease_ord_sort(array, work)
```
jvdp1, line 76: Should we add a check on the size of `work` (to be sure it is larger than `size(array)/2`? Or is there no issues if it is not the case?

src/stdlib_sorting_ord_sort.fypp
```diff
@@ -0,0 +1,785 @@
...
+        integer :: stat
+
+        if ( present(work) ) then
+! Use the work array as scratch memory
+            call merge_sort( array, work )
+        else
+! Allocate a buffer to use as scratch memory.
+            array_size = size( array, kind=int_size )
+            allocate( buf(0:array_size/2-1), stat=stat )
+            if ( stat /= 0 ) error stop "${k1}$_${sname}$_ord_sort: Allocation of buffer failed."
```
jvdp1, line 121: This could be moved to the caller, such that this procedure become pure.

src/stdlib_sorting_sort_index.fypp
```diff
@@ -0,0 +1,873 @@
...
+        end do
+
+        if ( present(reverse) ) then
+            if ( reverse ) then
+                call reverse_segment( array, index )
+            end if
+        end if
+
+! If necessary allocate buffers to serve as scratch memory.
+        if ( present(work) ) then
```
jvdp1, line 109: Should there be a check on the size of `work`?

jvdp1 (commented):
Some additional comments on some checks. If you agree, I can have a look to implement them.

src/stdlib_sorting_ord_sort.fypp
wclodius2: We should ensure that work is greater than or equal to `size(array)/2`.

wclodius2: There is also a `work` and an `iwork` in `stdlib_sorting_index`.

src/stdlib_sorting_ord_sort.fypp
wclodius2: I don't understand this comment. How can the allocation be moved to the caller? `but` is allocated only if the user doesn't provide a `work` array.

src/stdlib_sorting_sort_index.fypp
wclodius2: Yes.

jvdp1 (approved):
With all the changes made, it is ready to be merged IMO

Thank you @wclodius2 for this module

jvdp1:
Could someone with a write access review this too? @milancurcic? @awvwgk? @ivan-pi?

awvwgk (approved):
This is not a full review, I followed the discussion in this thread a bit and started looking over the specs, but won't be able to give an in-depth review in the next days. Since we already have two approving reviews here, I'll approve based on @gareth-nx's review to make it count to towards the merge requirements. Sorry for being a bit missing in action here.

wclodius2:
This branch now has two approvals but  still needs someone to  merge the  PR.

Issue #399, opened 2021-04-25
Should reverse be in stdlib_string?


ivan-pi:
#310 added the `reverse` function to the `stdlib_ascii` module. Since the behavior of this function does not  depend on the character set, the `stdlib_string` module would be a better home for it.

On the other hand the overloaded functions `to_lower`, `to_upper`, and `to_title` for the `string_type`(currently in `stdlib_string_type`) do implement behavior depending on the character set, and should be placed in `stdlib_ascii`.

cc @awvwgk

awvwgk:
There is an issue with cyclic dependencies here. Moving reverse from `stdlib_ascii` to `stdlib_strings` requires to move the implementation in `stdlib_string_type` as well, same of `to_lower`, `to_upper` and `to_title`. Now that we have a `stdlib_strings` module we could think about reshuffling the procedures, leaving only inquiry functions in `stdlib_ascii`, the type definition + overloaded intrinsics in `stdlib_string_type` and everything related to string manipulation is moved to `stdlib_strings`.

aniket2405:
Hello, I would like to work on this issue, if available. Thanks!

awvwgk:
Feel free to submit a patch for this issue. If you have any questions regarding stdlib and the workflow, do not hesitate to ask, we are always happy to help.

Pranavchiku:
I feel it is implemented here: https://github.com/fortran-lang/stdlib/blob/master/src/stdlib_ascii.fypp#L336-L346, is there anything else we need for this issue?

Merothiya:
I'd like to work on this issue if it is still available. Thanks.

jvdp1:
@Merothiya Thank you for your interest. AFAIK this issue is still available. Some work have been done (e.g., #745 ), but not finished.

Issue #319, opened 2021-02-12
Proposal for clamp


milancurcic:
## Intro

`clamp(x, xmin, xmax)` is a function that clamps (or, clips) an input scalar or array `x` so that the result is within given bounds `xmin` and `xmax`. For example:

```
res = clamp(0.9. 0.2, 0.8) ! res is 0.8
res = clamp(0.7. 0.2, 0.8) ! res is 0.7
res = clamp(0.1, 0.2, 0.8) ! res is 0.2
```

This function is common in standard libraries. I've used it occasionally in Fortran, and I use it often in Python. It provides an easy way to ensure that a variable with a finite (non-NaN, non-Inf) value stays within bounds.

## Name

Python has `numpy.clip`, while C++, Julia, and Rust all have `clamp`. So `clamp` is more common. Both `clamp` and `clip` are equally meaningful to me. I have slight preference for `clip` because it's shorter and has a "sharper" sound to it.

* [numpy.clip](https://numpy.org/doc/stable/reference/generated/numpy.clip.html)
* [C++ clamp](https://en.cppreference.com/w/cpp/algorithm/clamp)
* [Julia clamp](https://docs.julialang.org/en/v1/base/math/#Base.Math.clamp)
* [Rust clamp](https://doc.rust-lang.org/std/primitive.f32.html#method.clamp)

What inspired me to open this proposal is reading that `clamp` was [stabilized in Rust 1.50.0](https://blog.rust-lang.org/2021/02/11/Rust-1.50.0.html).

## API

```
elemental real function clamp(x, xmin, xmax) result(res)
  real, intent(in) :: x ! input value to clip
  real, intent(in) :: xmin ! lower bound
  real, intent(in) :: xmax ! higher bound
```

Alternative names for the bounds could be `low` and `high`.

## Implementation

There are a few options that I know of, each of which produce different assembly instructions, and performance depending on optimization levels, array sizes, and whether the clamping is in a loop or over a whole array in a single call.

### A min/max clamp

This is the simplest implementation I could think of:

```
elemental real function clamp(x, xmin, xmax) result(res)
  real, intent(in) :: x, xmin, xmax
  res = min(max(x, xmin), xmax)
end function clamp
```

There are two intrinsic function calls here, but maybe a good compiler will optimize one away. I don't know.

### An if/then/else clamp

```
elemental real function clamp(x, xmin, xmax) result(res)
  real, intent(in) :: x, xmin, xmax
  if (x < xmin) then
[...]

PR #414, opened 2021-05-23
implemented intelligent slice functionality

### LINKED ISSUE #413: intelligent slice functionality for strings
**Description**

Name of functionality: **Slice**
**Signature**: slice(string, start(optional)= 1 or len(string), end(optional)=len(string) or 1, stride(optional)= 1, include_start(optional)= .true.)
**Output**: a new string_type object

Traverses the input _string_ from _start_ index to _end_ index taking a stride of _stride_ indexes to return a new string. _start_ can be greater than _end_ as well giving function an added functionality of reversing input string. _start_ index will always be included in the output substring unless _include_start_ is set to _.false._ where the _end_ index will be included. So either _start_ index or _end_ index will be included in the output string (or both).
This function is an intelligent one, if the user doesn’t provide any one or more of the 3 optional arguments (_start_, _end_ or _stride_) it figures them out automatically using optional arguments which are given (see examples).
But if the user provides arguments he/she is expected to be responsible for that (see example 3).

**Examples**:
1. slice(`‘12345’`, stride=-1) should give `'54321'`; start = 5, end = 1
2. slice(`‘abcd’`, stride=-2, include_start=.false.) should give `'ca'`; start = 4, end = 1
3. slice(`‘abcde’`, 5, 2, 1) will give `''` (empty string); user gave 1 as the stride argument
4. slice(`‘abcde’`, 5, 2) will give `'edcb'`; stride = -1
5. slice(`‘abcde’`, end = 1, stride = -2) will give `'eca'`; start = 5

**Prior Art**

Python has it, but not exactly in the manner proposed above. In python one can do
```python
str = 'spiderman'
print(str[1:9:2])
```
which will return `'pdra'`.

Please review the functionality and let me know your thoughts on this.

Aman-Godara:
Status: code is open for review
resolves #413

Tasks:
- [x] Implement `slice` interface in `stdlib_strings.f90`: add two procedures under `slice` interface
- [x] Write test cases for `slice` in `stdlib_string_functions.f90`
- [ ] Discuss about `include_last` feature after getting the basic `slice` function merged
- [x] Document `slice` function and correct documentation of `to_title` and `to_sentence`

### DIFF
--- a/doc/specs/stdlib_string_type.md
+++ b/doc/specs/stdlib_string_type.md
@@ -1254,7 +1254,7 @@ The result is a scalar `string_type` value.
 
 ```fortran
 program demo_to_title
-  use stdlib_string_type, only: string_type, to_title
+  use stdlib_string_type
   implicit none
   type(string_type) :: string, titlecase_string
 
@@ -1302,7 +1302,7 @@ The result is a scalar `string_type` value.
 
 ```fortran
 program demo_to_sentence
-  use stdlib_string_type, only: string_type, to_sentence
+  use stdlib_string_type
   implicit none
   type(string_type) :: string, sentencecase_string
 
--- a/doc/specs/stdlib_strings.md
+++ b/doc/specs/stdlib_strings.md
@@ -192,3 +192,64 @@ program demo
   print'(a)', ends_with("pattern", "pat")  ! F
 end program demo
 ```
+
+
+<!-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -->
+### `slice`
+
+#### Description
+
+Extracts the characters from the defined region of the input string.  
+Argument `first` and `last` defines the region for the function `slice` to operate.  
+Extraction starts from the index `first` and takes stride of length `stride`.  
+Argument `stride` cannot take the value 0.
+
+#### Syntax
+
+`string = [[stdlib_strings(module):slice(interface)]] (string, first, last, stride)`
+
+#### Status
+
+Experimental
+
+#### Class
+
+Pure function.
+
+#### Argument
+
+- `string`: Character scalar or [[stdlib_string_type(module):string_type(type)]]
+  This argument is intent(in).
+- `first`: integer
+  This argument is intent(in) and optional.
+- `last`: integer
+  This argument is intent(in) and optional.
+- `stride`: integer
+  This argument is intent(in) and optional.
+
+#### Result value
+
+The result is of the same type as `string`.
+
+#### Example
+
+```fortran
+program demo_slice
+  use stdlib_string_type
+  use stdlib_strings, only : slice
+  implicit none
+  type(string_type) :: string
+  character(len=10) :: char
+
+  string = "abcdefghij"
+  ! string <-- "abcdefghij"
+
+  char = "abcdefghij"
+  ! char <-- "abcdefghij"
+
+  print'(a)', slice("abcdefghij", 2, 6, 2)   ! "bdf"
+  print'(a)', slice(string, 2, 6, 2)         ! "bdf"
+  print'(a)', slice(char, 2, 6, 2)           ! "bdf"
+  
+end program demo_slice
+```
--- a/doc/specs/stdlib_strings.md
+++ b/doc/specs/stdlib_strings.md
@@ -201,8 +201,17 @@ end program demo
 
 Extracts the characters from the defined region of the input string.  
 Argument `first` and `last` defines the region for the function `slice` to operate.  
-Extraction starts from the index `first` and takes stride of length `stride`.  
-Argument `stride` cannot take the value 0.
+If the defined region is invalid (user provides atleast one invalid index), `first` and 
+`last` indexes are converted to first and last valid indexes in this defined region respectively, 
+if no valid index exists in this region an empty string is returned.  
+
+Extraction starts from `first` index and takes stride of length `stride`.  
+Extraction is active until `last` index is crossed.  
+Extraction starts only if `last` index is crossable from `first` index taking 
+stride of length `stride`.  
+`stride` can attain both negative or positive values but when the invalid value 
+0 is given, it is converted to 1.  
+Function automatically deduces the arguments which are not provided by the user.
 
 #### Syntax
 
@@ -248,8 +257,10 @@ program demo_slice
   ! char <-- "abcdefghij"
 
   print'(a)', slice("abcdefghij", 2, 6, 2)   ! "bdf"
-  print'(a)', slice(string, 2, 6, 2)         ! "bdf"
   print'(a)', slice(char, 2, 6, 2)           ! "bdf"
   
+  string = slice(string, 2, 6, 2)
+  ! string <-- "bdf"
+
 end program demo_slice
 ```
--- a/doc/specs/stdlib_strings.md
+++ b/doc/specs/stdlib_strings.md
@@ -205,13 +205,12 @@ If the defined region is invalid (user provides atleast one invalid index), `fir
 `last` indexes are converted to first and last valid indexes in this defined region respectively, 
 if no valid index exists in this region an empty string is returned.  
 
-Extraction starts from `first` index and takes stride of length `stride`.  
-Extraction is active until `last` index is crossed.  
-Extraction starts only if `last` index is crossable from `first` index taking 
-stride of length `stride`.  
 `stride` can attain both negative or positive values but when the invalid value 
 0 is given, it is converted to 1.  
-Function automatically deduces the arguments which are not provided by the user.
+Extraction starts from `first` index and takes stride of length `stride`.  
+Extraction starts only if `last` index is crossable from `first` index by taking 
+stride of length `stride`and is active until `last` index is crossed.  
+Function automatically deduces the optional arguments that are not provided by the user.
 
 #### Syntax
 
--- a/doc/specs/stdlib_strings.md
+++ b/doc/specs/stdlib_strings.md
@@ -199,18 +199,25 @@ end program demo
 
 #### Description
 
-Extracts the characters from the defined region of the input string.  
-Argument `first` and `last` defines the region for the function `slice` to operate.  
-If the defined region is invalid (user provides atleast one invalid index), `first` and 
-`last` indexes are converted to first and last valid indexes in this defined region respectively, 
-if no valid index exists in this region an empty string is returned.  
+Extracts the characters from the defined region of the input string by taking strides.
 
-`stride` can attain both negative or positive values but when the invalid value 
-0 is given, it is converted to 1.  
-Extraction starts from `first` index and takes stride of length `stride`.  
-Extraction starts only if `last` index is crossable from `first` index by taking 
-stride of length `stride`and is active until `last` index is crossed.  
-Function automatically deduces the optional arguments that are not provided by the user.
+Deduction Process:
+Function first automatically deduces the optional arguments that are not provided by the user.  
+This process is independent of both input `string` and permitted indexes of Fortran.  
+Deduced `first` and `last` argument take +infinity or -infinity value whereas deduced `stride` argument takes +1 or -1 value.
+
+Validation Process:
+Argument `first` and `last` defines this region for extraction by function `slice`.  
+If the defined region is invalid i.e. region contains atleast one invalid index, `first` and 
+`last` are converted to first and last valid indexes in this defined region respectively, 
+if no valid index exists in this region an empty string is returned.  
+`stride` can attain both negative or positive values but when the only invalid value 
+0 is given, it is converted to 1.
+
+Extraction Process:
+After all this, extraction starts from `first` index and takes stride of length `stride`.  
+Extraction starts only if `last` index is crossable from `first` index with stride `stride` 
+and remains active until `last` index is crossed.  
 
 #### Syntax
 
--- a/src/stdlib_strings.f90
+++ b/src/stdlib_strings.f90
@@ -58,9 +58,11 @@ module stdlib_strings
         module procedure :: ends_with_string_char
         module procedure :: ends_with_char_string
         module procedure :: ends_with_char_char
-    end interface 
+    end interface ends_with
     
+    !> Slices the input string to return a new string
     !> 
+    !> Version: experimental
     interface slice
         module procedure :: slice_string
         module procedure :: slice_char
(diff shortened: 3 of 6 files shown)


awvwgk (commented):
I would prefer to avoid keywords as (dummy) variable names, how about `end` -> `last`?

Aman-Godara:
> I would prefer to avoid keywords as (dummy) variable names, how about `end` -> `last`?

Using `start` with `last` might be non-intuitive for some. How about `first` and `last` or `from` and `to`?

awvwgk:
I like `first` and `last`.

ivan-pi:
Is their any reason to keep this in `stdlib_ascii` module? See https://github.com/fortran-lang/stdlib/issues/399.

Aman-Godara:
> Is their any reason to keep this in `stdlib_ascii` module? See #399.

So we will have a `slice` interface which will have 2 procedures in it like this:
```fortran
interface slice
    module procedure :: slice_character_sequence
    module procedure :: slice_string
end interface slice
```
and implementation of `slice_string` will call `slice_character_sequence`?

ivan-pi:
I'm guessing you have the right idea in mind. You can drop the `_sequence` postfix, IMO.

```
use stdlib_string, only: slice
use stdlib_string_type, only: string => string_type
character(len=16) :: cstr
type(string) :: sstr
cstr = "Hello, World"
sstr = string(slice(cstr,last=6)) ! sstr contains "Hello", call to slice_character
print *, slice(sstr,first=3)      ! prints "llo", call to slice_string
```

Aman-Godara:
Yeah, I got your point.

Your above comments start a new discussion here:
> sstr = string(slice(cstr,last=6)) ! sstr contains "Hello", call to slice_character
> print *, slice(sstr,first=3)      ! prints "llo", call to slice_string

Currently the function slice is `first` index and `last` index inclusive but some other languages like Java as well as Python keep `first` index inclusive and `last` index exclusive. So output of `slice(cstr, last=6)` will be `'Hello,'` for Fortran (output included index 6 as well)
whereas for Python index 6 will not be included:
```python
str = 'Hello, World'
print(str[:6])
# prints 'Hello,' (index 0 inclusive but index 6 exclusive)
```

This might be linked with the fact that Java & Python uses 0-based indexing whereas Fortran doesn't.

ivan-pi:
I reckon the first and last should behave inclusively in agreement with the
intrinsic slice syntax.

src/tests/string/test_string_functions.f90
```diff
@@ -52,6 +53,46 @@ subroutine test_reverse_string
...
     end subroutine test_reverse_string
 
+    subroutine test_slice_string
+        type(string_type) :: test_string
+        character(len=:), allocatable :: test_char
+        test_string = "abcdefghijklmnopqrstuvwxyz"
+        test_char = "abcdefghijklmnopqrstuvwxyz"
+
+        call check(slice(test_string, 2, 16, 3) == "behkn", &
+                    'function slice failed', warn=.true.)
```
awvwgk, line 63: Be careful when using `warn=.true.` in the tests, as a failing test will only produce output but don't change the exit status of the test driver. A failing test wouldn't show up in the CI with `warn=.true.`

awvwgk (approved):
Thanks. This is a very useful addition to stdlib.

Aman-Godara:
`slice` function has some overlap with [char_string_range](https://github.com/fortran-lang/stdlib/blob/1822e3f964492143ffde8819bfd69affa185fd87/src/stdlib_string_type.fypp#L461-L470).

ivan-pi:
> 
> 
> `slice` function has some overlap with [char_string_range](https://github.com/fortran-lang/stdlib/blob/1822e3f964492143ffde8819bfd69affa185fd87/src/stdlib_string_type.fypp#L461-L470).

Indeed, they perform roughly the same operation. But `char_string_range` is only accessed through it's public interface `char`, which has an entirely different purpose, that is transform a `string_type` into the intrinsic character type.

On the other hand `slice` is more of an alternative to the slice syntax `(1:n:s)` in form of a function (since the language is missing the the capability to overload the access operator).

doc/specs/stdlib_strings.md
```diff
@@ -192,3 +192,64 @@ program demo
...
+
+<!-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -->
+### `slice`
+
+#### Description
+
+Extracts the characters from the defined region of the input string.  
+Argument `first` and `last` defines the region for the function `slice` to operate.  
+Extraction starts from the index `first` and takes stride of length `stride`.  
+Argument `stride` cannot take the value 0.
```
ivan-pi, line 205: Might be worth stating explicitly, that in case stride is 0, the function uses the default value 1 instead.

doc/specs/stdlib_string_type.md
```diff
@@ -1254,7 +1254,7 @@ The result is a scalar `string_type` value.
 
 ```fortran
 program demo_to_title
-  use stdlib_string_type, only: string_type, to_title
+  use stdlib_string_type
```
ivan-pi, line 1257: Was fine to leave the only specifier. Explicit is better than implicit.

doc/specs/stdlib_strings.md
```diff
@@ -192,3 +192,64 @@ program demo
...
 ```
+
+
+### `slice`
+
+#### Description
+
+Extracts the characters from the defined region of the input string.  
+Argument `first` and `last` defines the region for the function `slice` to operate.
```
ivan-pi, line 203: We should also state that in case of invalid inputs, i.e. `first < 1` or `last > len(str)`, the function assumes the default limits.

ivan-pi (changes requested):
LGTM.

Only thing I was confused about was what happens in case of input like:

```
result = slice(str,8,2,stride=-2)
```
Are the characters now taken in reverse reverse? I would need to perform some tests with the function, to make sure all "invalid" input cases still produce sensible results.

doc/specs/stdlib_string_type.md
Aman-Godara: > ```
> result = slice(str,8,2,stride=-2)
> ```
> 
> Are the characters now taken in reverse reverse?

Behaviour of the function is similar to this fortran loop:

j = 1
do i = `start`, `end`, `stride`

<< insert the character present at index `i` of `input_string` to index `j` of `output_string` >>

j = j + 1
end do


let's say `input_string` = "abcdefgh"
`start` = 8
`end` = 2
`stride` = -2

I am showing few iterations as the function progresses here:
<loop starts>
function is at index = 8
`output_string` = "h"

function is at index = 6
`output_string` = "hf"

function is at index = 4
`output_string` = "hfd"

function is at index = 2
`output_string` = "hfdb"
<loop ends>

so we get a reversed output "hfdb"

Had the stride been positive 2 (`stride` = 2, `start` = 8, `end` = 2): the above mentioned loop would have executed for 0 number of times i.e. function will never reach any index.

Aman-Godara:
> ```
> result = slice(str,8,2,stride=-2)
> ```
> 
> Are the characters now taken in reverse reverse?

Behaviour of the function is similar to this fortran loop:

j = 1
do i = `start`, `end`, `stride`

<< insert the character present at index `i` of `input_string` to index `j` of `output_string` >>

j = j + 1
end do


let's say `input_string` = "abcdefgh"
`start` = 8
`end` = 2
`stride` = -2

I am showing few iterations as the function progresses here:
<loop starts>
function is at index = 8
`output_string` = "h"

function is at index = 6
`output_string` = "hf"

function is at index = 4
`output_string` = "hfd"

function is at index = 2
`output_string` = "hfdb"
<loop ends>

so we get a reversed output "hfdb"

Had the stride been positive 2 (`stride` = 2, `start` = 8, `end` = 2): the above mentioned loop would have executed for 0 number of times i.e. function will never reach any index.

doc/specs/stdlib_string_type.md
Aman-Godara: would require adding `assignment operator(=)` along with `string_type` and `to_title` because the demo code uses this operator (`=`) in line 1261.

Aman-Godara:
If
```fortran
input_string = 'abcdefgh'
output_string = slice(input_string, 12, 16)
```
`output_string` : `'h'`
This output is non-intuitive to me. Shouldn't the function better return `''` in this case?

ivan-pi:
I've written a new comment in https://github.com/fortran-lang/stdlib/issues/413#issuecomment-849445164

The implementation of `extract` by  @everythingfunctional can be found [here](https://gitlab.com/everythingfunctional/iso_varying_string/-/blob/main/src/iso_varying_string.f90#L1046).

In the case above where the `last` is larger than the length of the string, the call `slice(input_string,12,16)` should be transliterated to `input_string(12:min(len(input_string),finish))`, or more specifically `input_string(12:8)` which should automatically return a zero-length string.

Aman-Godara:
Things are different here in `slice` function.
even if I give 8 as the `last` argument, output will not empty:
```fortran
input_string = 'abcdefgh'
output_string = slice(input_string, 12, 8)
```
output will be `'h'`.
In extract function stride is fixed as +1 whereas this is not the case with slice function.

Now I have changed the code, so as to get `""` for the same code. This was not the case before my last commit (fa88905).

src/stdlib_strings.f90
```diff
@@ -290,5 +300,80 @@ elemental function ends_with_string_string(string, substring) result(match)
...
+            end if
+
+            if (present(first)) then
+                first_index =  first
+            end if
+            if (present(last)) then
+                last_index = last
+            end if
+
+            strides_taken = floor( real(last_index - first_index) / real(stride_vector) )
```
awvwgk, line 353: This should be safe to do with integer devision
```suggestion
            strides_taken = (last_index - first_index) / stride_vector
```

Aman-Godara: if last_index = 27
first_index = 26
stride_vector =-2
I want the output -1
that's why I added floor

awvwgk: Should give -1 with integer division right away, I don't think there is need for floating point arithmetic here.

ivan-pi:
The output of
```
print *,(27 - 26)/(-2)
end
```
is zero.

And the output of
```
print *,(27 - 24)/(-2)
end
```
is -1.

It seems like in the negative half, integer division truncates towards the
"higher" value.

ivan-pi:
In the zen of python it is stated that, 

> There should be one-- and preferably only one --obvious way to do it.

What is the value in allowing a negative stride, if we could also achieve this using the nested calls, e.g. 

```
result = reverse(slice(str,first,last,stride))
```

I'm happy to hear what are some possible counter-arguments. One is of course shorter call syntax. But it comes at the price of the complicated logic inside `slice`.

Aman-Godara:
> What is the value in allowing a negative stride, if we could also achieve this using the nested calls, e.g.

I can try implementing slice which takes positive strides only but I think it will still look more or less equally complicated.

ivan-pi:
I'd stay with the current behavior for now. If we can get the corner cases right, I have nothing against the current behavior. It is just a matter of preference in the end.

awvwgk:
I tried to break the implementation and haven't found any deviating behavior from the array slice so far (see 81f028a for a hacky `slice` implementation using the transfer intrinsic).

Aman-Godara:
What would this return?
```fortran
array = [1, 2, 3, 4, 5]
print *, array(-2 : 10 : 2)
```
in my case it returns `2  4  `.
Is it right or should it return output equivalent to `array(1 : 5 : 2)` which would be `1  3  5  `.

ivan-pi:
> What would this return?

You can think of the slice syntax `-2:10:2` as the integer array `[-2,0,2,4,6,8,10]`. The 2nd and 4th elements are then printed. However, I don't think we want to follow the behavior of the compiler in this case because the example violates good programming practice. gfortran raises the warning

```
test_slice_index.f90:5:15:

    5 | print *, array(-2 : 10 : 2)
      |               1
Warning: Lower array reference at (1) is out of bounds (-2 < 1) in dimension 1
```
If you add the `-fcheck=all` flag the program will produce: `Fortran runtime error: Index '-2' of dimension 1 of array 'array' outside of expected range (1:5)`.

awvwgk:
We need a couple of more reviewers here to move this patch forward.

src/tests/string/test_string_functions.f90
```diff
@@ -52,6 +53,58 @@ subroutine test_reverse_string
...
+                    'function slice failed', warn=.false.)
+        call check(slice(test_string, -500, 500) == "abcdefghijklmnopqrstuvwxyz", &
+                    'function slice failed', warn=.false.)
+
+        test_string = ""
+        test_char = ""
+        call check(slice(test_string, 2, 16, 3) == "", &
+                    'function slice failed', warn=.false.)
+        call check(slice(test_char, 2, 16, 3) == "", &
+                    'function slice failed', warn=.false.)
```
everythingfunctional, line 104: Try and come up with better descriptions for the aspect that each test is checking.

src/stdlib_strings.f90
```diff
@@ -290,5 +299,72 @@ elemental function ends_with_string_string(string, substring) result(match)
 
     end function ends_with_string_string
 
+    !> Extract the characters from the region between 'first' and 'last' index (both inclusive)
+    !> of the input 'string' by taking strides of length 'stride'
+    !> Returns a new string
+    elemental function slice_string(string, first, last, stride) result(sliced_string)
```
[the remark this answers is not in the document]
ivan-pi: See https://github.com/fortran-lang/stdlib/issues/413#issuecomment-846933463 for the discussion why elemental can't work here.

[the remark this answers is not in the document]
everythingfunctional: Yeah, totally understand why it can't be elemental for `character`. I was more referring to the discussions in #319 that would also apply to why we may not want `slice _string` to be elemental. But I think the resolution was that we should allow it to be elemental, and I agree with that. Just thought I'd bring it up in case sentiments may have changed or there was some nuance here that would sway the decision the other way.

src/tests/string/test_string_functions.f90
awvwgk: I would highly recommend to keep the specific tests, because they should serve as example usage to help users understand the functionality.

everythingfunctional: I agree with @awvwgk

src/tests/string/test_string_functions.f90
```diff
@@ -52,6 +56,182 @@ subroutine test_reverse_string
...
+        end do
+
+        do i = 1 - offset, len(test) + offset
+            do j = 1 - offset, len(test) + offset
+                do k = -len(test) - offset, len(test) + offset
+                    call check_slicer(test, first=i, last=j, stride=k)
+                end do
+            end do
+        end do
+    end subroutine test_slice_gen
```
everythingfunctional, line 154: This makes it even harder to see what the expected behavior is supposed to be. The goal of a good test suite is to serve as an example of how to use the code, and a definition of it's expected behavior. You should go back to the specific tests you had and just give them more meaningful descriptions. The message from a failing test should be a hint as to what aspect of the code is not working correctly. Something as vague as "it failed" doesn't accomplish that. Ideally, I should be able to read the test suite alone and understand the expected behavior.

src/tests/string/test_string_functions.f90
Aman-Godara: arranged test cases and added some descriptive comments so as to help a reader build the intuition for `slice`. Let me know if comments are good enough or not.

everythingfunctional: You're headed in the right direction, now take the comments you added, and combine them into the description that gets printed if the check fails. I.e.

```
! Only one argument is given
          ! Valid
          call check(slice(test_string, first=10) == "jklmnopqrstuvwxyz") ! last=+inf
```
becomes
```
call check(slice(test_string, first=10) == "jklmnopqrstuvwxyz", "Only a valid argument for first is provided")
```

everythingfunctional: The new versions are better. I'm good with it, but I suggest working on this area for future PRs. Think along the lines of "If the only thing a developer got was the output from failing tests, would they be able to understand what the code is supposed to do and then be able write a correct implementation?" Writing a good test suite is a skill and it takes practice, but it's worth it.

everythingfunctional (approved):

awvwgk:
Thanks everybody for the discussion on this patch. To move https://github.com/fortran-lang/stdlib/pull/433 forward I will go ahead and merge this PR.

PR #417, opened 2021-05-24
Add GCC-11 to workflow

### LINKED ISSUE #416: Building with GCC 11 fails
GCC 11 is now available and we already have packages for [Ubuntu 21.04](https://packages.ubuntu.com/hirsute/gfortran-11) and [homebrew](https://formulae.brew.sh/formula/gcc), would be nice if we could start testing this here as well. Since our Linux runner is using Ubuntu 20.04 it might be a bit more involved to install GCC 11 there.

awvwgk:
Related #416

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -21,7 +21,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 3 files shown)


LKedward (approved):

jvdp1 (approved):
LGTM. Thanks

PR #418, opened 2021-05-28
Improved support for NAG

sakamoti:
Because of nagfor's strict error checking characteristics, execution is halted by default when a floating point exception occurs. Some of the runtime errors can be resolved by adding the option '-ieee=full'.
Other errors have been corrected to conform more strictly to the fortran standard.
These commit still leaves the one runtime error 'string_derivedtype_io'.
#108

### DIFF
--- a/src/stdlib_logger.f90
+++ b/src/stdlib_logger.f90
@@ -352,8 +352,8 @@ contains
             end if
 
 ! Check that unit is opened
-            inquire( unit, opened=question, iostat=istat, iomsg=msgtxt)
-            if(istat/=0) question=.false.
+            inquire( unit, opened=question, iostat=istat, iomsg=msgtxt )
+            if(istat /= 0) question = .false.
             if ( .not. question ) then
                 if ( present(stat) ) then
                     stat = unopened_in_error
--- a/src/stdlib_logger.f90
+++ b/src/stdlib_logger.f90
@@ -309,7 +309,6 @@ contains
         character(12) :: specifier
         logical :: question
         integer :: istat
-        character(len=80) :: msgtxt
 
         call validate_unit()
         if ( present(stat) ) then
@@ -352,7 +351,7 @@ contains
             end if
 
 ! Check that unit is opened
-            inquire( unit, opened=question, iostat=istat, iomsg=msgtxt )
+            inquire( unit, opened=question, iostat=istat )
             if(istat /= 0) question = .false.
             if ( .not. question ) then
                 if ( present(stat) ) then
--- a/src/stdlib_logger.f90
+++ b/src/stdlib_logger.f90
@@ -308,6 +308,8 @@ contains
         integer :: lun
         character(12) :: specifier
         logical :: question
+        integer :: istat
+        character(len=80) :: msgtxt
 
         call validate_unit()
         if ( present(stat) ) then
@@ -350,7 +352,8 @@ contains
             end if
 
 ! Check that unit is opened
-            inquire( unit, opened=question )
+            inquire( unit, opened=question, iostat=istat, iomsg=msgtxt)
+            if(istat/=0) question=.false.
             if ( .not. question ) then
                 if ( present(stat) ) then
                     stat = unopened_in_error
--- a/src/stdlib_string_type.fypp
+++ b/src/stdlib_string_type.fypp
@@ -571,11 +571,11 @@ contains
         logical, intent(in), optional :: back
         integer :: pos
 
-        if(present(back))then
-          pos = index(maybe(string), maybe(substring),back)
+        if (present(back)) then
+          pos = index(maybe(string), maybe(substring), back)
         else
-          pos = index(maybe(string), maybe(substring),.false.)
-        endif
+          pos = index(maybe(string), maybe(substring), .false.)
+        end if
 
     end function index_string_string
 
@@ -587,11 +587,11 @@ contains
         logical, intent(in), optional :: back
         integer :: pos
 
-        if(present(back))then
-          pos = index(maybe(string), substring,back)
+        if (present(back)) then
+          pos = index(maybe(string), substring, back)
         else
-          pos = index(maybe(string), substring,.false.)
-        endif
+          pos = index(maybe(string), substring, .false.)
+        end if
 
     end function index_string_char
 
@@ -603,11 +603,11 @@ contains
         logical, intent(in), optional :: back
         integer :: pos
 
-        if(present(back))then
-          pos = index(string, maybe(substring),back)
+        if (present(back)) then
+          pos = index(string, maybe(substring), back)
         else
-          pos = index(string, maybe(substring),.false.)
-        endif
+          pos = index(string, maybe(substring), .false.)
+        end if
 
     end function index_char_string
 
@@ -621,11 +621,11 @@ contains
         logical, intent(in), optional :: back
         integer :: pos
 
-        if(present(back))then
-          pos = scan(maybe(string), maybe(set),back)
+        if (present(back)) then
+          pos = scan(maybe(string), maybe(set), back)
         else
-          pos = scan(maybe(string), maybe(set),.false.)
-        endif
+          pos = scan(maybe(string), maybe(set), .false.)
+        end if
 
     end function scan_string_string
 
@@ -637,11 +637,11 @@ contains
         logical, intent(in), optional :: back
         integer :: pos
 
-        if(present(back))then
-          pos = scan(maybe(string), set,back)
+        if (present(back)) then
+          pos = scan(maybe(string), set, back)
         else
-          pos = scan(maybe(string), set,.false.)
-        endif
+          pos = scan(maybe(string), set, .false.)
+        end if
 
     end function scan_string_char
 
@@ -653,11 +653,11 @@ contains
         logical, intent(in), optional :: back
         integer :: pos
 
-        if(present(back))then
-          pos = scan(string, maybe(set),back)
+        if (present(back)) then
+          pos = scan(string, maybe(set), back)
         else
-          pos = scan(string, maybe(set),.false.)
-        endif
+          pos = scan(string, maybe(set), .false.)
+        end if
 
     end function scan_char_string
 
@@ -671,11 +671,11 @@ contains
         logical, intent(in), optional :: back
         integer :: pos
 
-        if(present(back))then
-          pos = verify(maybe(string), maybe(set),back)
+        if (present(back)) then
+          pos = verify(maybe(string), maybe(set), back)
         else
-          pos = verify(maybe(string), maybe(set),.false.)
-        endif
+          pos = verify(maybe(string), maybe(set), .false.)
+        end if
 
     end function verify_string_string
 
@@ -688,11 +688,11 @@ contains
         logical, intent(in), optional :: back
         integer :: pos
 
-        if(present(back))then
-          pos = verify(maybe(string), set,back)
+        if (present(back)) then
+          pos = verify(maybe(string), set, back)
         else
-          pos = verify(maybe(string), set,.false.)
-        endif
+          pos = verify(maybe(string), set, .false.)
+        end if
 
     end function verify_string_char
 
@@ -705,11 +705,11 @@ contains
         logical, intent(in), optional :: back
         integer :: pos
 
-        if(present(back))then
-          pos = verify(string, maybe(set),back)
+        if (present(back)) then
+          pos = verify(string, maybe(set), back)
         else
-          pos = verify(string, maybe(set),.false.)
-        endif
+          pos = verify(string, maybe(set), .false.)
+        end if
 
     end function verify_char_string
 
--- a/src/stdlib_string_type.fypp
+++ b/src/stdlib_string_type.fypp
@@ -421,7 +421,7 @@ contains
         type(string_type), intent(in) :: string
         integer :: ich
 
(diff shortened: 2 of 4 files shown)


src/stdlib_logger.f90
```diff
@@ -308,6 +308,8 @@ subroutine add_log_unit( self, unit, stat )
         integer :: lun
         character(12) :: specifier
         logical :: question
+        integer :: istat
+        character(len=80) :: msgtxt
```
awvwgk, line 312: 80 characters are a bit short, maybe have bit bigger buffer here, even if it is not used?

src/stdlib_string_type.fypp
```diff
@@ -421,7 +421,7 @@ contains
         type(string_type), intent(in) :: string
         integer :: ich
 
-        ich = merge(ichar(string%raw), 0, allocated(string%raw))
+        ich = merge(ichar(string%raw(1:1)), 0, allocated(string%raw))
```
awvwgk, line 424: Will this work for an empty string (len(raw) == 0)?

src/stdlib_string_type.fypp
```diff
@@ -431,7 +431,7 @@ contains
         type(string_type), intent(in) :: string
         integer :: ich
 
-        ich = merge(iachar(string%raw), 0, allocated(string%raw))
+        ich = merge(iachar(string%raw(1:1)), 0, allocated(string%raw))
```
awvwgk, line 434: Will this work for an empty string?

src/stdlib_string_type.fypp
```diff
@@ -571,8 +571,11 @@ contains
         logical, intent(in), optional :: back
         integer :: pos
 
-        pos = index(maybe(string), maybe(substring), &
-            merge(back, .false., present(back)))
+        if(present(back))then
+          pos = index(maybe(string), maybe(substring),back)
+        else
+          pos = index(maybe(string), maybe(substring),.false.)
+        endif
```
awvwgk, line 578:
```suggestion
        if (present(back)) then
          pos = index(maybe(string), maybe(substring), back)
        else
          pos = index(maybe(string), maybe(substring), .false.)
        end if
```

src/stdlib_string_type.fypp
```diff
@@ -584,8 +587,11 @@ contains
         logical, intent(in), optional :: back
         integer :: pos
 
-        pos = index(maybe(string), substring, &
-            merge(back, .false., present(back)))
+        if(present(back))then
+          pos = index(maybe(string), substring,back)
+        else
+          pos = index(maybe(string), substring,.false.)
+        endif
```
awvwgk, line 594:
```suggestion
        if (present(back)) then
          pos = index(maybe(string), substring, back)
        else
          pos = index(maybe(string), substring, .false.)
        end if
```

src/stdlib_string_type.fypp
```diff
@@ -597,8 +603,11 @@ contains
         logical, intent(in), optional :: back
         integer :: pos
 
-        pos = index(string, maybe(substring), &
-            merge(back, .false., present(back)))
+        if(present(back))then
+          pos = index(string, maybe(substring),back)
+        else
+          pos = index(string, maybe(substring),.false.)
+        endif
```
awvwgk, line 610:
```suggestion
        if (present(back)) then
          pos = index(string, maybe(substring), back)
        else
          pos = index(string, maybe(substring), .false.)
        end if
```

src/stdlib_string_type.fypp
```diff
@@ -612,8 +621,11 @@ contains
         logical, intent(in), optional :: back
         integer :: pos
 
-        pos = scan(maybe(string), maybe(set), &
-            merge(back, .false., present(back)))
+        if(present(back))then
+          pos = scan(maybe(string), maybe(set),back)
+        else
+          pos = scan(maybe(string), maybe(set),.false.)
+        endif
```
awvwgk, line 628:
```suggestion
        if (present(back)) then
          pos = scan(maybe(string), maybe(set), back)
        else
          pos = scan(maybe(string), maybe(set), .false.)
        end if
```

src/stdlib_string_type.fypp
```diff
@@ -625,8 +637,11 @@ contains
         logical, intent(in), optional :: back
         integer :: pos
 
-        pos = scan(maybe(string), set, &
-            merge(back, .false., present(back)))
+        if(present(back))then
+          pos = scan(maybe(string), set,back)
+        else
+          pos = scan(maybe(string), set,.false.)
+        endif
```
awvwgk, line 644:
```suggestion
        if (present(back)) then
          pos = scan(maybe(string), set, back)
        else
          pos = scan(maybe(string), set, .false.)
        end if
```

src/stdlib_string_type.fypp
```diff
@@ -638,8 +653,11 @@ contains
         logical, intent(in), optional :: back
         integer :: pos
 
-        pos = scan(string, maybe(set), &
-            merge(back, .false., present(back)))
+        if(present(back))then
+          pos = scan(string, maybe(set),back)
+        else
+          pos = scan(string, maybe(set),.false.)
+        endif
```
awvwgk, line 660:
```suggestion
        if (present(back)) then
          pos = scan(string, maybe(set), back)
        else
          pos = scan(string, maybe(set), .false.)
        end if
```

src/stdlib_string_type.fypp
```diff
@@ -653,8 +671,11 @@ contains
         logical, intent(in), optional :: back
         integer :: pos
 
-        pos = verify(maybe(string), maybe(set), &
-            merge(back, .false., present(back)))
+        if(present(back))then
+          pos = verify(maybe(string), maybe(set),back)
+        else
+          pos = verify(maybe(string), maybe(set),.false.)
+        endif
```
awvwgk, line 678:
```suggestion
        if (present(back)) then
          pos = verify(maybe(string), maybe(set), back)
        else
          pos = verify(maybe(string), maybe(set), .false.)
        end if
```

src/stdlib_string_type.fypp
```diff
@@ -667,8 +688,11 @@ contains
         logical, intent(in), optional :: back
         integer :: pos
 
-        pos = verify(maybe(string), set, &
-            merge(back, .false., present(back)))
+        if(present(back))then
+          pos = verify(maybe(string), set,back)
+        else
+          pos = verify(maybe(string), set,.false.)
+        endif
```
awvwgk, line 695:
```suggestion
        if (present(back)) then
          pos = verify(maybe(string), set, back)
        else
          pos = verify(maybe(string), set, .false.)
        end if
```

src/stdlib_string_type.fypp
```diff
@@ -681,8 +705,11 @@ contains
         logical, intent(in), optional :: back
         integer :: pos
 
-        pos = verify(string, maybe(set), &
-            merge(back, .false., present(back)))
+        if(present(back))then
+          pos = verify(string, maybe(set),back)
+        else
+          pos = verify(string, maybe(set),.false.)
+        endif
```
awvwgk, line 712:
```suggestion
        if (present(back)) then
          pos = verify(string, maybe(set), back)
        else
          pos = verify(string, maybe(set), .false.)
        end if
```

src/tests/logger/test_stdlib_logger.f90
```diff
@@ -417,6 +417,8 @@ end subroutine test_adding_log_files
     subroutine test_removing_log_units()
 
         logical :: opened
+        integer :: istat
+        character(len=80) :: msgtxt
```
awvwgk, line 421: 80 characters might be a bit short.

src/tests/stats/test_mean_f03.f90
```diff
@@ -20,7 +20,7 @@ program test_mean
 call check( sum( abs( mean(d,2) - sum(d,2)/real(size(d,2), dp) )) < dptol)
 
 !dp rank 8
-allocate(d8(size(d,1), size(d,2), 3, 4, 5, 6, 7, 8))
+allocate(d8(size(d,1), size(d,2), 3, 4, 5, 6, 7, 8),source=0.0_dp)
```
awvwgk, line 23:
```suggestion
allocate(d8(size(d,1), size(d,2), 3, 4, 5, 6, 7, 8), source=0.0_dp)
```

src/tests/logger/test_stdlib_logger.f90
```diff
@@ -462,7 +464,8 @@ subroutine test_removing_log_units()
 
         end if
 
-        inquire( unit4, opened=opened )
+        inquire( unit4, opened=opened ,iostat=istat,iomsg=msgtxt)
+        if(istat/=0) opened=.false.
```
awvwgk, line 468:
```suggestion
        inquire( unit4, opened=opened, iostat=istat, iomsg=msgtxt )
        if(istat /= 0) opened = .false.
```

src/stdlib_logger.f90
```diff
@@ -350,7 +352,8 @@ subroutine validate_unit()
             end if
 
 ! Check that unit is opened
-            inquire( unit, opened=question )
+            inquire( unit, opened=question, iostat=istat, iomsg=msgtxt)
+            if(istat/=0) question=.false.
```
awvwgk, line 356:
```suggestion
            inquire( unit, opened=question, iostat=istat, iomsg=msgtxt )
            if(istat /= 0) question = .false.
```

awvwgk (approved):
Thanks for sharing. This looks good to me, pending a some minor comments and a few whitespace nitpicks.

src/stdlib_logger.f90
sakamoti: Delete unused msgtxt.

src/tests/logger/test_stdlib_logger.f90
sakamoti: Delete unused msgtxt.

src/stdlib_string_type.fypp
sakamoti: Thank you for pointing this out.
Since "iachar" only accepts a character type of size 1 as an argument, I checked it in the if statement at "git commit 1a87696".

src/stdlib_string_type.fypp
```diff
@@ -421,7 +421,11 @@ contains
         type(string_type), intent(in) :: string
         integer :: ich
 
-        ich = merge(ichar(string%raw(1:1)), 0, allocated(string%raw))
+        if (allocated(string%raw) .and. len_trim(string%raw) > 0) then
```
awvwgk, line 424: `ichar(" ")` should be allowed, therefore it's better to use `len()` here. Also for safe access:
```suggestion
        if (allocated(string%raw) .and. len(string) > 0) then
```

src/stdlib_string_type.fypp
sakamoti: Thank you for pointing this out.
Since "iachar" only accepts a character type of size 1 as an argument, I checked it in the if statement at "git commit 1a87696".

src/stdlib_string_type.fypp
```diff
@@ -431,7 +435,11 @@ contains
         type(string_type), intent(in) :: string
         integer :: ich
 
-        ich = merge(iachar(string%raw(1:1)), 0, allocated(string%raw))
+        if (allocated(string%raw) .and. len_trim(string%raw) > 0) then
```
awvwgk, line 438:
```suggestion
        if (allocated(string%raw) .and. len(string) > 0) then
```

sakamoti: Thank you. Apply suggestions.

src/stdlib_string_type.fypp
sakamoti: Thank you. Apply suggestions.

src/stdlib_string_type.fypp
```diff
@@ -421,7 +421,11 @@ contains
         type(string_type), intent(in) :: string
         integer :: ich
 
-        ich = merge(ichar(string%raw), 0, allocated(string%raw))
+        if (allocated(string%raw) .and. len(string) > 0) then
+          ich = ichar(string%raw(1:1))
+        else
+          ich = 0
+        end if
```
jvdp1, line 428: May I ask what was the issue with the original code?

jvdp1 (approved):
LGTM, pending a question.
Thank you.

src/stdlib_string_type.fypp
sakamoti: Accordings to the book  '_Modern Fortran explained (Incorporating Fortran 2018)'_,  argument  of '_ichar_' is a character which length equal one. 
'string%raw' can be more than length one. In such case, runtime error occurred when use nagfor. On the other hands, gfortran seems to accept string argument.

jvdp1: Thank you @sakamoti for your answer. But I was wondering why the `merge` option was removed.

awvwgk: `merge` doesn't necessarily short-circuit the evaluation of its arguments like a ternary operator in C would. This could lead to an access of the uninitialized raw payload of the `string_type` which is an error.

jvdp1: Thanks @awvwgk for the explanation.

sakamoti: I havn't know  which one is more efficiently to select 'merge' or 'if construct'. I just felt a bit complex to write with 'merge'. And I want to acces 'string%raw' only when 'conditions' becoms true. So, I select to write with 'if construct'.
Thank you @awvwgk for the explanation.

PR #426, opened 2021-06-04
Addition of a subroutine to compute the median of array elements

jvdp1:
API can be already reviewed 

Still to do:
* [x] Update hyperlinks in specs
* [x] Add comments in common.fypp
* [x] Add some tests (e.g., for int64 and real(dp))
* [x] Change `sort` to `ord_sort` when #428 will be merged


gareth-nx:
Untested suggestion: 

I wonder if the code duplication for the integer/real cases can be reduced by including the output type `o1` in the code generation loop:
`#:for k1, t1, o1 in ( zip(REAL_KINDS, REAL_TYPES, REAL_TYPES) + zip(INT_KINDS, INT_TYPES, ['real(dp)']*len(INT_KINDS)) )`

So here 'o1' is the type of `res` (which is always `real(dp)` in the integer case). 

The calculations seem to rely on constants (1. and 0.5) that have the right precision, but it looks like they could be defined as local parameters of type `o1`, making the integer/real cases have identical code (?).

gareth-nx:
Another suggestion: From the sorting review, I understood that `ord_sort` is much faster than `sort` on data with significant pre-sorted chunks, while it is just a fraction slower for random data. 


Considering that for median calculation both cases should be pretty common, I wonder if `ord_sort` is a better general choice? Although the downside is that more scratch memory will be required. 

Perhaps the routine could have an optional argument `sort_method` that allows switching between the two?

jvdp1:
Thank you @gareth-nx for your suggestions.

>I wonder if the code duplication for the integer/real cases can be reduced by including the output type o1 in the code generation loop:
#:for k1, t1, o1 in ( zip(REAL_KINDS, REAL_TYPES, REAL_TYPES) + zip(INT_KINDS, INT_TYPES, ['real(dp)']*len(INT_KINDS)) )
>
>So here 'o1' is the type of res (which is always real(dp) in the integer case).

Great idea. I did it a bit differently that what you proposed. But it reduced the code a lot.

>Considering that for median calculation both cases should be pretty common, I wonder if ord_sort is a better general choice? >Although the downside is that more scratch memory will be required.
>

Indeed, I also think that `ord_sort` is a better choice. I didn't use it because it needs more scratch memory. While I was trying `ord_sort` in this code, I found a bug in it (see #428 for details). When #428 will be merged, I will change `sort` for `ord_sort` in this procedure too.

>Perhaps the routine could have an optional argument sort_method that allows switching between the two?

The API of `median` is currently the same as the API of `mean`. IMO it would be good to keep it like that.

gareth-nx:
In the case with `mask` as an argument, consider checking whether the size of the `mask` is equal to the size of `x`. Maybe this is not desired (to avoid throwing errors etc). But results like the following could easily be a user-error.
```fortran
program median_local
    use stdlib_stats, only : median
    use iso_c_binding, only : dp => C_DOUBLE
    implicit none

    real(dp) :: x(10), y(11)
    integer :: i

    x = (/(i*1.0_dp, i = 1, size(x))/)
    y = (/(i*1.0_dp, i = 1, size(y))/)
    
    ! Should this be an error, because size(mask) != size(y)? 
    print*, median(y, mask= (x > 5.5_dp))
end program
```

jvdp1:
> In the case with `mask` as an argument, consider checking whether the size of the `mask` is equal to the size of `x`. Maybe this is not desired (to avoid throwing errors etc). But results like the following could easily be a user-error.
> 
> ```fortran
> program median_local
>     use stdlib_stats, only : median
>     use iso_c_binding, only : dp => C_DOUBLE
>     implicit none
> 
>     real(dp) :: x(10), y(11)
>     integer :: i
> 
>     x = (/(i*1.0_dp, i = 1, size(x))/)
>     y = (/(i*1.0_dp, i = 1, size(y))/)
>     
>     ! Should this be an error, because size(mask) != size(y)? 
>     print*, median(y, mask= (x > 5.5_dp))
> end program
> ```

Good suggestion. I wonder what the intrinsic `sum` reports (and what the standard says) in such a case.

jvdp1:
> > In the case with `mask` as an argument, consider checking whether the size of the `mask` is equal to the size of `x`. Maybe this is not desired (to avoid throwing errors etc). But results like the following could easily be a user-error.
> > ```fortran
> > program median_local
> >     use stdlib_stats, only : median
> >     use iso_c_binding, only : dp => C_DOUBLE
> >     implicit none
> > 
> >     real(dp) :: x(10), y(11)
> >     integer :: i
> > 
> >     x = (/(i*1.0_dp, i = 1, size(x))/)
> >     y = (/(i*1.0_dp, i = 1, size(y))/)
> >     
> >     ! Should this be an error, because size(mask) != size(y)? 
> >     print*, median(y, mask= (x > 5.5_dp))
> > end program
> > ```
> 
> Good suggestion. I wonder what the intrinsic `sum` reports (and what the standard says) in such a case.

With GFortran, it seems that no checks are done with `sum` in a release mode. In a debug mode, a runtime error is provided and mentioned that a mismatch was found. I am in favor to keep the same behavior as with the intrinsic `sum` implemented in gfortran (which is the case currently). However, I open to implemeent a check if desired by the community.

arjenmarkus:
I would say such a check is light-weight and potentially saves a lot of
problems that would otherwise not be easily noticed. For the runtime
library of a compiler the situation may be different: the compiler knows
that extra checks are called for (debug, checks on array bounds and such),
but our code does not have that luxury. In debug builds you would see an
array bounds problem reported, not a mismatch in array arguments.

My preference would be to add the check ;), but others may disagree,

Op di 8 jun. 2021 om 19:44 schreef Jeremie Vandenplas <
***@***.***>:

jvdp1:
@gareth-nx @arjenmarkus I added a check on the shapes of `x` and `mask` ;) Question: how to test that it works properly, since it used an `error stop`?

src/stdlib_stats_median.fypp
```diff
@@ -0,0 +1,219 @@
...
+        ${t1}$, intent(in) :: x${ranksuffix(rank)}$
+        logical, intent(in) :: mask${ranksuffix(rank)}$
+        real(${o1}$) :: res
+
+        integer(kind = int64) :: c, n
+        ${t1}$, allocatable   :: x_tmp(:)
+
+        if (any(shape(x) .ne. shape(mask))) then
+            call error_stop("ERROR (median): shapes of x and mask are different")
+            error stop
```
jvdp1, line 128:
```suggestion
```

src/stdlib_stats_median.fypp
```diff
@@ -0,0 +1,219 @@
...
+        #:if rank > 1
+        #:for fj in range(1, rank+1)
+        integer :: j${"_" * fj}$
+        #:endfor
+        #:endif
+        ${t1}$, allocatable :: x_tmp(:)
+
+        if (any(shape(x) .ne. shape(mask))) then
+            call error_stop("ERROR (median): shapes of x and mask are different")
+            error stop
```
jvdp1, line 173:
```suggestion
```

jvdp1 (commented):
remove error stop

gareth-nx:
Hi @jvdp1 

Is this ready for review? I will be happy to do that once it is (but I note the very top of the thread suggests you still need to update some hyperlinks).

jvdp1:
Hi @gareth-nx,

Thjank you. you may start the review. I will try to update the hyperlinks
for FORD later...

src/stdlib_stats_median.fypp
```diff
@@ -0,0 +1,217 @@
...
+        end if
+
+        n = size(x, dim)
+        c = floor( (n + 1) / 2._${o1}$ )
+
+        allocate(x_tmp(n))
+
+        select case(dim)
+          #:for fi in range(1, rank+1)
+          case(${fi}$)
```
gareth-nx, line 82: Here I suggest to add a comment for clarity
```
! Loop over every dimension of the array except "dim"
```

Also, to make it simpler, can you reduce the 2 levels of nested loop construction into 1, with something like the following (untested)?:
```
#: for fj in [range(1, fi), range(fi+1, rank+1)]
```

src/stdlib_stats_median.fypp
```diff
@@ -0,0 +1,217 @@
...
+        #:endif
+        ${t1}$, allocatable :: x_tmp(:)
+
+        if (any(shape(x) .ne. shape(mask))) then
+            call error_stop("ERROR (median): shapes of x and mask are different")
+        end if
+
+        select case(dim)
+          #:for fi in range(1, rank+1)
+          case(${fi}$)
```
gareth-nx, line 176: As per the comment above, suggest to add a comment `! Loop over every dimension of the array except dim`, and also, consider if you can collapse the 2 fypp loops to 1.

src/stdlib_stats_median.fypp
```diff
@@ -0,0 +1,217 @@
+#:include "common.fypp"
+#:set RANKS = range(1, MAXRANK + 1)
+#:set IR_KINDS_TYPES_OUTPUT = list(zip(REAL_KINDS, REAL_TYPES, REAL_KINDS)) + list(zip(INT_KINDS,INT_TYPES, ['dp']*len(INT_KINDS)))
+
+
+submodule (stdlib_stats) stdlib_stats_median
+
+  use, intrinsic:: ieee_arithmetic, only: ieee_value, ieee_quiet_nan
+  use stdlib_error, only: error_stop
+  use stdlib_optval, only: optval
```
gareth-nx, line 10: Here I suggest to add a comment explaining why you are using `ord_sort` rather than `sort`, e.g. 
```
! Use "ord_sort" rather than "sort" because the former can be much faster for arrays that are already partly sorted. While it is slightly slower for random arrays, ord_sort seems a better overall choice. 
```

src/stdlib_stats_median.fypp
gareth-nx: The above should be a 'review' comment, blame my poor github skills...

doc/specs/stdlib_stats.md
```diff
@@ -166,6 +178,75 @@ program demo_mean
...
+### Status
+
+Experimental
+
+### Description
+
+Returns the median of all the elements of `array`, or of the elements of `array` along dimension `dim` if provided, and if the corresponding element in `mask` is `true`.
+
+After that the elements are sorted in an increasing order, e.g. `array_sorted =
+sort(array)`, the median of the elements of `array` are defined as, if
```
gareth-nx, line 192: suggest to edit as: ... the median of the elements of `array` are defined in the usual way. If 
`n = size(array)` is an even number the median is:
... equation here
and if `n` is an odd number, the median is:
... equation here

doc/specs/stdlib_stats.md
```diff
@@ -166,6 +178,75 @@ program demo_mean
...
+
+`result = [[stdlib_stats(module):median(interface)]](array, dim [, mask])`
+
+### Class
+
+Generic subroutine
+
+### Arguments
+
+`array`: Shall be an array of type `integer` or  `real`.
```
gareth-nx, line 220: Here could you add a comment about the range of supported kinds for each?

doc/specs/stdlib_stats.md
```diff
@@ -166,6 +178,75 @@ program demo_mean
...
+
+`dim`: Shall be a scalar of type `integer` with a value in the range from 1 to `n`, where `n` is the rank of `array`.
+
+`mask` (optional): Shall be of type `logical` and either a scalar or an array of the same shape as `array`.
+
+### Return value
+
+If `array` is of type `real`, the result is of the same type as `array`.
```
gareth-nx, line 228: Should this say  "... , the result is `real` with the same kind as the entries of `array`"?

And similarly for the integer case below?

src/common.fypp
```diff
@@ -156,4 +156,59 @@ ${prefix + joinstr.join([line.strip() for line in txt.split("\n")]) + suffix}$
...
+#!   rank (int): Rank of the original variable
+#!   dim (int): Dimension of the variable
+#!
+#! Returns:
+#!   Array rank suffix string enclosed in braces
+#!
+#! E.g.,
+#!  select_subvector(5, 2)
+#!   -> (j_, :, j___, j____, j_____)
+#!
```
gareth-nx, line 172: Nice additions here (albeit I hope in future Fortran can offer a better way to do this!).

gareth-nx: Is there any value in providing an optional argument to make `j` some other letter? Not sure if situations would arise where that was useful? Although in any case, if you don't need it now, this does not have to be done.

gareth-nx (changes requested):
Very nice -- a couple of minor comments for your consideration, but they should be easy to address.

doc/specs/stdlib_stats.md
jvdp1: Not sure what you  mean. I copied the other descriptions in these specs.

doc/specs/stdlib_stats.md
jvdp1: Agreed for the `real`.

src/common.fypp
jvdp1: Thank you for the suggestion. done.

src/stdlib_stats_median.fypp
jvdp1: Thank you @gareth-nx for your review. Both done.

src/stdlib_stats_median.fypp
jvdp1: Thank you. I copied-pasted your comment.

src/stdlib_stats_median.fypp
jvdp1: done

jvdp1:
Thank you @gareth-nx for your review. I think I answered all your comments.

doc/specs/stdlib_stats.md
```diff
@@ -166,6 +178,75 @@ program demo_mean
...
+median(array) = array_sorted( floor( (n + 1) / 2.))
+```
+
+or as if `n` is an odd number:
+
+```
+median(array) = mean( array_sorted( floor( (n + 1) / 2.):floor( (n + 1) / 2.) + 1 ) )
+```
+
+The array is sorted using the subroutine `sort` provided by the `stdlib_sorting`
```
gareth-nx: Needs to be updated to `ord_sort`

doc/specs/stdlib_stats.md
gareth-nx: Oh I mean like single precision, double precision, etc. If that is already implied by the statement above, then no problem. I get confused because in Fortran "real" is single precision, but we can also speak of the "real kinds".

gareth-nx (approved):
I added a few more trivial comments, but approve, as I'm sure you can address them OK. Great work -- I really like this family of stats functions you have added.

doc/specs/stdlib_stats.md
jvdp1: Done. Thank you for noting it

jvdp1:
@gareth-nx 
> Oh I mean like single precision, double precision, etc. If that is already implied by the statement above, then no problem. I get confused because in Fortran "real" is single precision, but we can also speak of the "real kinds".

The sentence in the specs is similar to the sentence used to decribed the `array` in `sum` intrinsic in [the gcc page](https://gcc.gnu.org/onlinedocs/gfortran/SUM.html).  If you have another proposition, please provide it.

jvdp1:
@gareth-nx reviewed this PR. Could some others with write-access review it too, please? @ivan-pi @awvwgk @milancurcic ? others?

awvwgk (approved):

msz59:
Median is just a special case of [per]centile, namely the 50th percentile or second quartile. I wonder whether it might be worth adding an optional parameter to specify the percentile (default=50). It would change the code just in one final place, to compute the index into the sorted array.
Another approach could be naming the (extended) function 'percentile' or 'centile' and add median as an 'alias' with that optional parameter set to 50.

jvdp1:
> Median is just a special case of [per]centile, namely the 50th percentile or second quartile. I wonder whether it might be worth adding an optional parameter to specify the percentile (default=50). It would change the code just in one final place, to compute the index into the sorted array.

Thank you @msz59 for your comments. I agree that median is just a special case of (per)centile and quantile. However, I don't think that the change of the proposed code to a more general code like for computing quantiles is as easy as proposed. Indeed, several methods are available to compute quantiles, and some of them rely on the computation of the median. However, any propositions to improve the API/code for generalisation are welcome.

src/common.fypp
```diff
@@ -156,4 +156,61 @@ ${prefix + joinstr.join([line.strip() for line in txt.split("\n")]) + suffix}$
   #:endcall
 #:enddef
 
+#!
+#! Generates an array rank suffix for subarrays along a dimension
+#!
```
ivan-pi, line 161: Thanks for writing these nice fypp functions. Adding a usage example or at least a reference to one of the modules where they are used would be helpful for future developers to get a better idea how to use them.

src/common.fypp
```diff
@@ -156,4 +156,61 @@ ${prefix + joinstr.join([line.strip() for line in txt.split("\n")]) + suffix}$
...
+#! E.g.,
+#!  select_subvector('j', 5, 2)
+#!   -> (j_, :, j___, j____, j_____)
+#!
+#:def select_subvector(varname, rank, idim)
+  #:assert rank > 0
+  #:call join_lines(joinstr=", ", prefix="(", suffix=")")
+    #:for i in range(1, idim)
+      ${varname}$${ "_" * (i) }$
```
ivan-pi, line 178: I recall in one of the previous stats modules, we commented that using integer indexes, i.e. (`j1`, `j2`, `j3`, and so on), seemed more legible than the use of multiple underscores.

Is this applicable here too?

src/stdlib_stats.fypp
```diff
@@ -2,6 +2,7 @@
 #:set RANKS = range(1, MAXRANK + 1)
 #:set REDRANKS = range(2, MAXRANK + 1)
 #:set RC_KINDS_TYPES = REAL_KINDS_TYPES + CMPLX_KINDS_TYPES
+#:set IR_KINDS_TYPES_OUTPUT = list(zip(REAL_KINDS, REAL_TYPES, REAL_KINDS)) + list(zip(INT_KINDS,INT_TYPES, ['dp']*len(INT_KINDS)))
```
ivan-pi, line 5:
```suggestion
#:set RI_KINDS_TYPES_OUTPUT = list(zip(REAL_KINDS, REAL_TYPES, REAL_KINDS)) + list(zip(INT_KINDS,INT_TYPES, ['dp']*len(INT_KINDS)))
```

This is just for consistency with the order of the list elements.

src/stdlib_stats.fypp
```diff
@@ -305,11 +306,65 @@ module stdlib_stats
...
 
+  interface median
+    !! version: experimental
+    !!
+    !! Median of array elements
+    !! ([Specification](../page/specs/stdlib_stats.html#median-median-of-array-elements))
+    #:for k1, t1, o1 in IR_KINDS_TYPES_OUTPUT
+      #:for rank in RANKS
+        #:set RName = rname("median_all",rank, t1, k1, o1)
+        module function ${RName}$ (x, mask) result(res)
```
ivan-pi, line 317: This is just a nitpick, but since Fortran is case insensitive, I wonder if having consistent "styles" for both the preprocessor and Fortran code would mean avoiding having variables/functions with the same name but different capitalization?

src/stdlib_stats_median.fypp
```diff
@@ -0,0 +1,216 @@
...
+  ! Use "ord_sort" rather than "sort" because the former can be much faster for arrays
+  ! that are already partly sorted. While it is slightly slower for random arrays,
+  ! ord_sort seems a better overall choice.
+  use stdlib_sorting, only:sort => ord_sort
```
ivan-pi, line 14:
```suggestion
  use stdlib_sorting, only: sort => ord_sort
```

src/tests/stats/test_median.fypp
```diff
@@ -0,0 +1,196 @@
...
+
+#:for k1,t1 in REAL_KINDS_TYPES
+subroutine test_median_${k1}$()
+    ${t1}$, allocatable :: d0(:), d1(:), d2(:,:), d3(:,:,:)
+
+    allocate(d0(0))
+    call check(size(d0) == 0, 'test_median_${k1}$: should be of size 0')
+
+    d1 = [10., 2., -3., -4., 6., -6., 7., -8., 9., 0., 1., 20.]
+    call check(mod(size(d1), 2) == 0, 'stest_median_${k1}$: hould be an even number')
```
ivan-pi, line 115:
```suggestion
    call check(mod(size(d1), 2) == 0, 'stest_median_${k1}$: should be an even number')
```

src/tests/stats/test_median.fypp
```diff
@@ -0,0 +1,196 @@
+#:include "common.fypp"
+#:set IR_KINDS_TYPES = INT_KINDS_TYPES + REAL_KINDS_TYPES
+program test_median
+    use stdlib_error, only: check
+    use stdlib_kinds, only: sp, dp, qp, &
+        int8, int16, int32, int64
+    use stdlib_stats, only: median, mean
+    use,intrinsic :: ieee_arithmetic, only : ieee_is_nan
```
ivan-pi, line 8:
```suggestion
    use, intrinsic :: ieee_arithmetic, only : ieee_is_nan
```

doc/specs/stdlib_stats.md
```diff
@@ -166,6 +178,75 @@ program demo_mean
...
+
+Experimental
+
+### Description
+
+Returns the median of all the elements of `array`, or of the elements of `array` along dimension `dim` if provided, and if the corresponding element in `mask` is `true`.
+
+After that the elements are sorted in an increasing order, e.g. `array_sorted =
+sort(array)`, the median of the elements of `array` is defined as the "middle"
+element. If `n = size(array)` is an even number, the median is:
```
ivan-pi, line 193:
```suggestion
The median of the elements of `array` is defined as the "middle"
element, after that the elements are sorted in an increasing order, e.g. `array_sorted =
sort(array)`. If `n = size(array)` is an even number, the median is:
```

src/stdlib_stats.fypp
```diff
@@ -305,11 +306,65 @@ module stdlib_stats
...
+          ${t1}$, intent(in) :: x${ranksuffix(rank)}$
+          logical, intent(in), optional :: mask
```
ivan-pi, line 319: This is just a question. I couldn't really grasp the purpose of median with the scalar optional mask? The false case will always give NaN, and the true case is the same as having no mask, right?

Is the use case supposed to be something like:

```
x = [...]
y = median(x, all(x > 0))
```

src/tests/stats/test_median.fypp
```diff
@@ -0,0 +1,196 @@
...
+
+end subroutine
+#:endfor
+
+#:for k1,t1 in REAL_KINDS_TYPES
+subroutine test_median_${k1}$()
+    ${t1}$, allocatable :: d0(:), d1(:), d2(:,:), d3(:,:,:)
+
+    allocate(d0(0))
+    call check(size(d0) == 0, 'test_median_${k1}$: should be of size 0')
```
ivan-pi, line 112: I don't understand the purpose of this test.

ivan-pi (changes requested):
Hi @jvdp1, could you point us to some resources about the median computation? I am not familiar enough with the median computation to evaluate the implementation in `stdlib_stats_median.fypp`.

However all the support infrastructure looks good, apart from small stylistic issues.

leonfoks:
I'm not sure where comments about the implementation of this should go, please let me know if I should submit something elsewhere etc.

The calculation of the median would benefit from using a "quick select" algorithm preferably with random pivots, it picks the kth value in an unsorted array by partial sorting, rather than a full sort and choose.  [This](https://stackoverflow.com/questions/56940793/quickselect-time-complexity-explained) was a nice explanation of the difference.  Using the full sort, we are bound by the complexity of the sorting algorithm, where as a quick select is O(n).
Looking at [this comment](https://github.com/fortran-lang/stdlib/pull/386#issuecomment-824092859) there could be considerable gains in time.

An implementation of quick select will be applicable to both a median function and percentile function as @msz59  suggested, median(array) could call a function "percentile(array, 0.5)" but, in the special case of median, it would be faster to call "quickselect(array, N/2)" (I'm omitting the case where we need the average of two central values here).
A fast median algorithm will be required if we implement things like KdTrees, where splitting can occur at the median value along the dimension with the highest variance.

jvdp1:
> Hi @jvdp1, could you point us to some resources about the median computation? I am not familiar enough with the median computation to evaluate the implementation in `stdlib_stats_median.fypp`.

Thank you @ivan-pi for your review. To answer your question, here is some documentions:

* [wikipedia](https://en.wikipedia.org/wiki/Median)
* [Octave](https://octave.org/doc/v4.0.3/Descriptive-Statistics.html#XREFmedian)
* [Julia](https://docs.julialang.org/en/v1/stdlib/Statistics/#Statistics.median)
* [Numpy](https://numpy.org/doc/stable/reference/generated/numpy.median.html)

As mentioned by @leonfoks I agree that the implemented approach for computing the median is most likely not the most efficient. However, my idea was to first propose this function with what was available in `stdlib`, such that users can test its API. Improvement of the implementation could come later. However, this approach might not be appropriate.

jvdp1:
> I'm not sure where comments about the implementation of this should go, please let me know if I should submit something elsewhere etc.

Thank you @leonfoks for you comment. I think it is the right place to discuss the API and implementation of this function.

> The calculation of the median would benefit from using a "quick select" algorithm preferably with random pivots, it picks the kth value in an unsorted array by partial sorting, rather than a full sort and choose. [This](https://stackoverflow.com/questions/56940793/quickselect-time-complexity-explained) was a nice explanation of the difference. Using the full sort, we are bound by the complexity of the sorting algorithm, where as a quick select is O(n).
> Looking at [this comment](https://github.com/fortran-lang/stdlib/pull/386#issuecomment-824092859) there could be considerable gains in time.

I totally agree with you that the implemented approach for computing the median is most likely not the most efficient. However, my idea was to first propose this function with what was available in `stdlib_sorting`, such that users can test its API. Improvement of the implementation could come later. However, this approach might not be appropriate.

> An implementation of quick select will be applicable to both a median function and percentile function as @msz59 suggested, median(array) could call a function "percentile(array, 0.5)" but, in the special case of median, it would be faster to call "quickselect(array, N/2)" (I'm omitting the case where we need the average of two central values here).
> A fast median algorithm will be required if we implement things like KdTrees, where splitting can occur at the median value along the dimension with the highest variance.

This would require the implementation of a quick select algorithm in `stdlib`, which is not available currently. Do you have any references to propose for this?

Maybe @wclodius2 who develop `stdlib_sorting` have some ideas on this topic.

gareth-nx:
Regarding quickselect, there is a matlab routine here (licence seems suitably permissive for a translation): https://www.mathworks.com/matlabcentral/fileexchange/68947-qselect

ivan-pi:
Since we need a "naive" median anyway to verify correctness/speed improvement of better algorithms my suggestion would be to 

1) confirm through review the implementation and tests are consistent
2) approve this PR and merge, thereby stabilizing the interface suggested by @jvdp1 
3) implement the quick-select or whatever other median algorithm in a new pull request

gareth-nx:
I agree that quickselect can be added later on. 

That said I just translated the matlab version to Fortran. It's nowhere near ready for stdlib, but copied below with a couple of tests in case anyone wants to try and/or modify.

```Fortran

module qselect_mod
!
! This code is a translation of a matlab implementation "qselect" by Manolis Lourakis
!   https://www.mathworks.com/matlabcentral/fileexchange/68947-qselect 
! Below is the licence of qselect
!
!
! Copyright (c) 2018, Manolis Lourakis
! All rights reserved.
! 
! Redistribution and use in source and binary forms, with or without
! modification, are permitted provided that the following conditions are met:
! 
! * Redistributions of source code must retain the above copyright notice, this
!   list of conditions and the following disclaimer.
! 
! * Redistributions in binary form must reproduce the above copyright notice,
!   this list of conditions and the following disclaimer in the documentation
!   and/or other materials provided with the distribution
! * Neither the name of Foundation for Research and Technology - Hellas nor the names of its
!   contributors may be used to endorse or promote products derived from this
!   software without specific prior written permission.
! THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
! AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
! IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
! DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE
! FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
! DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
! SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
! CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
! OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
! OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
!

    use iso_c_binding, only: C_DOUBLE, C_INT
    implicit none

    integer, parameter :: dp = C_DOUBLE, ip = C_INT

    contains

    function qselect(a, k) result(kth_largest)
! qselect - select the k-th smallest out of n numbers
! Implements Hoare s Quickselect algorithm (https://en.wikipedia.org/wiki/Quickselect)
! with the median-of-3 pivot strategy. Operates in-place, avoiding sorting and recursion
!
!    kth = qselect(a, k)
!
!    a      - array of n elements
!    k      - specifies the desired k-th smallest element
!             the k-th *largest* element can be found by passing length(a)+1-k
!
! Returns
!
!    kth    - the sought element
! Manolis Lourakis 2007-18
! Institute of Computer Science, Foundation for Research & Technology - Hellas
! Heraklion, Crete, Greece
! Sep  2018  - Original version. (v. 1.0)
!
        real(dp), intent(inout) :: a(:)
        integer(ip), intent(in) :: k
        real(dp) :: kth_largest

        integer(ip) :: l, r, s, i, j, k_local
        real(dp) :: temp, pivot

        l = 1_ip
        r = size(a)
        if(k < 1 .or. k > r) stop "in qselect, k must be between 1 and size(a)";
        k_local = k

        do while(.true.)
            s = (l+r)/2_ip ! Deliberate integer division
            if(a(s) < a(r)) then
                temp = a(s); a(s) = a(r); a(r) = temp
            end if
            if(a(s) < a(l)) then
                temp = a(s); a(s) = a(l); a(l) = temp
            end if
            if(a(r) < a(l)) then
                temp = a(r); a(r) = a(l); a(l) = temp; 
            end if
            pivot = a(r) ! Median
            i = l

            do j = l, r-1
                if(a(j) <= pivot) then
                    temp = a(i); a(i) = a(j); a(j) = temp
                    i = i+1_ip
                end if
            end do
            temp = a(r); a(r) = a(i); a(i) = temp
            s = i-l+1
            if(k_local < s) then
                r = i-1_ip
            else if(k_local > s) then
                l=i+1_ip; k_local=k_local-s;
            else
                kth_largest = a(i)
                return
            end if
        end do
    end function

end module

program test_quickselect
    use qselect_mod
    implicit none

    real(dp) :: x(10), x_copy(10), mat(8), mat_copy(8), len1(1), len2(2)
    integer(ip) :: k, i

    ! Make x contain the numbers 1**2, 2**2, .... 10**2,
    x = (/( i**2, i=1, size(x) )/)
    ! but mix up the order 
    x(5:2:-1) = x(2:5)
    x(10:8:-1) = x(8:10)

    ! Check that qselect(x, i) == i**2
    do i = 1, size(x)
        x_copy = x
        if(qselect(x_copy, i) == i**2) then
            print*, 'PASS'
        else
            print*, 'FAIL'
        end if
    end do

    !
    ! Test that came with the matlab documentation 
    !
    mat = 1.0_dp * [3, 2, 7, 4, 5, 1, 4, -1]

    mat_copy = mat
    print*, merge('PASS', 'FAIL', qselect(mat_copy, 2_ip) == 1)

    mat_copy = mat
    print*, merge('PASS', 'FAIL', qselect(mat_copy, size(mat)+1_ip-4_ip) == 4)

    ! Check it works for size(a) == 1
    len1(1) = -1.0_dp
    print*, merge('PASS', 'FAIL', qselect(len1, 1_ip) == -1.0_dp)

    ! size(a) == 2
    len2 = [-3.2_dp, -5.1_dp]
    print*, merge('PASS', 'FAIL', qselect(len2, 2_ip) == -3.2_dp)
    len2 = [-3.2_dp, -5.1_dp]
    print*, merge('PASS', 'FAIL', qselect(len2, 1_ip) == -5.1_dp)

end program
```

leonfoks:
I think at this stage, the current implementation is fine and should definitely be incorporated. Thank you for doing the work and having this ready!

I implemented a quick-select in coretran [here](https://github.com/leonfoks/coretran/blob/master/src/sorting/sm_quickSelect.f90). The partition submodule that I use within quick select is [here](https://github.com/leonfoks/coretran/blob/master/src/core/sm_partition.f90). I switch to an insertion sort at short lengths, I seem to remember doing some timing and found that to be faster than without the insertion sort.

It would be far less work to translate the matlab implementation above, rather than incorporating what I did in coretran.  I pulled out in the inner section into a partition subroutine since I needed it for my introspection sort.  I also had a simple swap function (I think this just got requested in an issue). From what I gather it's faster to just do the swap rather than call a function to do it.

leonfoks:
I tested the matlab version using the same arrays and compile flags as ive been using in coretran. Here are the results!
I ran each timing 10x and divided the total time by 10.  The matlab code is marginally faster, I put this down to coretran calling functions to do simple things.  So I would say it's better to incorporate the matlab translation into stdlib!

```
==========================
Testing : Selection
==========================
Double precision quick select
16777216  1.815E-01
8388608  1.086E-01
4194304  5.130E-02
2097152  2.430E-02
1048576  1.280E-02
524288  6.000E-03
Already sorted double precision quick select
16777216  2.950E-02
8388608  1.440E-02
4194304  7.700E-03
2097152  3.300E-03
1048576  1.500E-03
524288  7.000E-04
==========================
Testing : Matlab Selection
==========================
Double precision quick select
16777216  1.524E-01
8388608  7.870E-02
4194304  4.100E-02
2097152  1.880E-02
1048576  1.120E-02
524288  5.500E-03
Already sorted double precision quick select
16777216  1.900E-02
8388608  9.400E-03
4194304  4.100E-03
2097152  2.200E-03
1048576  8.000E-04
524288  4.000E-04
```

src/stdlib_stats.fypp
milancurcic: Same, I'm confused about this as well.

milancurcic (approved):
Pending remaining questions and reviews, I think this is good to go. Thank you!

leonfoks:
I just tested the matlab code using the same compile flags and test arrays that I used in Coretran.  Timings are below, interestingly i took out the insertion sort in coretran and times are better...  The times are very close with coretran being surprisingly faster for sorted arrays but maybe this is just timing fluctuations.  

It might be easiest at this point to simply translate the matlab code! Thanks for hearing me out about this!

Left column is array length, right is time in seconds.  Timing was done around a loop of 10 quick selections and the average taken.
```
==========================
Testing : Selection
==========================
Double precision quick select
16777216  1.694E-01
8388608  9.460E-02
4194304  5.000E-02
2097152  2.530E-02
1048576  1.030E-02
524288  3.500E-03

Already sorted double precision quick select
16777216  1.030E-02
8388608  4.500E-03
4194304  2.800E-03
2097152  1.000E-03
1048576  6.000E-04
524288  2.000E-04

==========================
Testing : Matlab Selection
==========================
Double precision quick select
16777216  1.579E-01
8388608  8.530E-02
4194304  4.260E-02
2097152  2.130E-02
1048576  9.300E-03
524288  4.100E-03
Already sorted double precision quick select
16777216  1.310E-02
8388608  6.200E-03
4194304  3.600E-03
2097152  1.500E-03
1048576  1.200E-03
524288  4.000E-04
```

src/stdlib_stats.fypp
jvdp1: This is to be in agreement with the API of the intrinsic `sum`, and the already implemented `mean`, `var`,...
Such API will indeed cover cases as proposed by @ivan-pi

src/stdlib_stats.fypp
jvdp1: I am not sure to understand your comment. Do you suggest to use another name than "RName" to avoid capitalization?

src/common.fypp
jvdp1: You are right. I didn't remember this discussion. I'll change that.

src/common.fypp
jvdp1: good suggestion. I'll do it.

src/stdlib_stats.fypp
jvdp1: modified for consistency. Thanks.

src/tests/stats/test_median.fypp
jvdp1: Yes, it is just to make sure the setup of d0 is correct. I will add a comment.

jvdp1:
Thank you @gareth-nx @leonfoks @ivan-pi @milancurcic for your review and comments.
I believe I answered all of them.

This whole discussion about selection algorithms led me to add some rules for cases where `array` contains `NaN`. Now if it is the case, the result will contain `NaN` too. In the previous implementation, the result was undetermined in such cases.

milancurcic:
@ivan-pi can this PR be merged?

src/common.fypp
```diff
@@ -156,4 +156,67 @@ ${prefix + joinstr.join([line.strip() for line in txt.split("\n")]) + suffix}$
...
+  #:endcall
+#:enddef
+
+#!
+#! Generates an array rank suffix for arrays
+#!
+#! Args:
+#!   varname (str): Name of the variable to be used as origin
+#!   rank (int): Rank of the original variable
+#!   idim (int): Dimension of the variable
```
ivan-pi, line 196:
```suggestion
#!   idim (int): Dimension of the variable dropped
```
Maybe "reduced" instead of "dropped" is more suitable.

src/common.fypp
```diff
@@ -156,4 +156,67 @@ ${prefix + joinstr.join([line.strip() for line in txt.split("\n")]) + suffix}$
...
+    #:endfor
+  #:endcall
+#:enddef
+
+#!
+#! Generates an array rank suffix for arrays
+#!
+#! Args:
+#!   varname (str): Name of the variable to be used as origin
+#!   rank (int): Rank of the original variable
```
ivan-pi, line 195:
```suggestion
#!   rank (int): Rank of the original array variable
```

src/stdlib_stats.fypp
ivan-pi: Yes. In Fortran code
```
real :: RName
integer :: rname
RName = rname
```
would not be allowed, because Fortran does not recognize capitalization. 

This was merely a stylistic suggestion, that it might be good to follow the same rule in fypp code, and reduce the mental burden of switching between capitalization sensitive and capitalization insensitive modes.

src/stdlib_stats.fypp
ivan-pi: Yes, this is indeed the case for [SUM](https://gcc.gnu.org/onlinedocs/gfortran/SUM.html). It would be interesting to learn the output of `sum(array,.false.)`. Is it also NaN?

src/stdlib_stats.fypp
```diff
@@ -305,11 +306,65 @@ module stdlib_stats
...
+          integer, intent(in) :: dim
+          logical, intent(in), optional :: mask
+          real(${o1}$) :: res${reduced_shape('x', rank, 'dim')}$
+        end function ${RName}$
+      #:endfor
+    #:endfor
+  
+    #:for k1, t1, o1 in IR_KINDS_TYPES_OUTPUT
+      #:for rank in RANKS
+        #:set RName = rname('median_mask_all',rank, t1, k1, o1)
```
ivan-pi, line 339: This is also a nitpick, but could this be renamed to `median_all_mask`? IMO it feels more consistent to have the pairs:
* `median_all` and `median_all_mask`
* `median` and `median_mask`

src/stdlib_stats_median.fypp
```diff
@@ -0,0 +1,255 @@
...
+          if (any(ieee_is_nan(x))) then
+            res = ieee_value(1._${o1}$, ieee_quiet_nan)
+            return
+          end if
+        #:endif
+
+        n = size(x, kind=int64)
+        c = floor( (n + 1) / 2._${o1}$, kind=int64 )
+
+        x_tmp = reshape(x, [n])
```
ivan-pi, line 44: Is the reshape necessary here?

src/tests/stats/test_median.fypp
```diff
@@ -0,0 +1,234 @@
...
+    call check( ieee_is_nan(median(d0, 1, .false.)), '${k1}$ median(d0, 1, .false.)' )
+    call check( ieee_is_nan(median(d1, 1, .false.)), '${k1}$ median(d1, 1, .false.)' )
+    call check( any(ieee_is_nan(median(d2, 1, .false.))), '${k1}$ median(d2, 1, .false.)' )
+    call check( any(ieee_is_nan(median(d2, 2, .false.))), '${k1}$ median(d2, 2, .false.)' )
+    call check( any(ieee_is_nan(median(d3, 1, .false.))), '${k1}$ median(d3, 1, .false.)' )
+    call check( any(ieee_is_nan(median(d3, 2, .false.))), '${k1}$ median(d3, 2, .false.)' )
+    call check( any(ieee_is_nan(median(d3, 3, .false.))), '${k1}$ median(d3, 3, .false.)' )
+
+    call check( abs(median(d1, 1) - 1.5_${k1}$) < ${k1}$tol, '${k1}$ median(d1, 1), even')
+    call check( sum(abs(median(d2, 1) - [2._${k1}$, -4._${k1}$, 7._${k1}$, 1._${k1}$])) < ${k1}$tol, &
```
ivan-pi, line 148: Would using the array kind specifier help reduce the preprocessor noise?

```
[real(${k1}$) :: 2.0, -4.0, 7.0, 1.0]
```

src/stdlib_stats.fypp
jvdp1: `sum(array, .false.)` will return `0`, as following the standard.

The functions `mean`, `var`, .... will return `NaN`, because of `mean(array, .false.) = sum(array, .false.) / 0`

For `median`, if  all elements correspond to `.false.`, then there are no elements for `median`. Hence, it makes sense to return `NaN` IMO.

ivan-pi (changes requested):
Following the links you provided the implementation looks good to me. The preprocessing work necessary to manage the different ranks is admirable. 

The only comment I have left is how does the behavior compare to other languages when NaN values are present? The check at the beginning `any(ieee_is_nan(x))` seems quite expensive, assuming that the majority of cases NaNs won't be there. But perhaps I am wrong and this is not a big issue.

src/stdlib_stats_median.fypp
ivan-pi: Okay, I see this is to flatten the array. I guess `pack` could also be used in this case?

src/stdlib_stats.fypp
jvdp1: Would `genname` be fine, instead of `RName`?

gareth-nx:
> Following the links you provided the implementation looks good to me. The preprocessing work necessary to manage the different ranks is admirable.
> 
> The only comment I have left is how does the behavior compare to other languages when NaN values are present? The check at the beginning `any(ieee_is_nan(x))` seems quite expensive, assuming that the majority of cases NaNs won't be there. But perhaps I am wrong and this is not a big issue.

The R interpreter gives NA if there are NA or NaN values present in the input vector:
```
> median(c(1,2,3,NA))
[1] NA

> median(c(1,2,3,NaN))
[1] NA
```

src/stdlib_stats.fypp
ivan-pi: How about just `name`?

src/stdlib_stats_median.fypp
jvdp1: The subroutine `sort` only accepts rank-1 array, while `median` support arrays of all ranks.

Do you have another suggestion to avoid the `reshape`?

src/tests/stats/test_median.fypp
jvdp1: With `gfortran`, this implies a conversion from `real(4)` to `real(8)`, with a warning triggered by `-Wconversion-extra`.
For this reason, I am inclined to keep as it is now.

src/stdlib_stats.fypp
jvdp1: done.

src/stdlib_stats.fypp
jvdp1: Thank you @ivan-pi for the suggestion. It is now renamed `name`.

src/stdlib_stats_median.fypp
jvdp1: Indeed, `x_tmp = pack(x, .true.)` should work too. Do you think that `pack` is more efficient than `reshape` in this case?

jvdp1:
> The only comment I have left is how does the behavior compare to other languages when NaN values are present? The check at the beginning `any(ieee_is_nan(x))` seems quite expensive, assuming that the majority of cases NaNs won't be there. But perhaps I am wrong and this is not a big issue.

Julia [median](https://docs.julialang.org/en/v1/stdlib/Statistics/#Statistics.median) returns NaN when NaN are present in the array (and as mentioned in @gareth-nx [comment](https://github.com/fortran-lang/stdlib/pull/426#issuecomment-885449070), it seems to be case for R too).

My main issue was that the result of `sort` is undetermined in prensence of NaN. Therefore, it would be also the case for `median` if there were no checks for NaN, and modifying its implementation (as proposed with quickselect) may result in a different behaviour. Checking for NaN allows to avoid possible future changes in the behavior of `median` if its implementation is modified.

@ivan-pi I answered all your questions. However, a couple of comments remain opened.

src/stdlib_stats_median.fypp
ivan-pi: I imagine that a good compiler would do the same thing. So it can remain as is.

ivan-pi (approved):

ivan-pi:
The open comments don't affect the behaviour, so with three approvals this can be merged. 

Thanks for your work.🙏

milancurcic:
Thank you all, I'll merge.

PR #428, opened 2021-06-05
Issue with stdlib_sorting

jvdp1:
The option `-fcheck=all` of gfortran triggered a problem with the suboutine `insertion_sort`, when `i` becomes equal to `0`, and tried to access `array( i-1 )` (while the lower bound of `array` is `0`.

The solution copied what was already implemented in `stdlib_sorting_sort.fypp`.

### DIFF
--- a/src/tests/sorting/test_sorting.f90
+++ b/src/tests/sorting/test_sorting.f90
@@ -14,7 +14,7 @@ program test_sorting
     integer(int32), parameter :: char_size = 26**4
     integer(int32), parameter :: string_size = 26**3
     integer(int32), parameter :: block_size = test_size/6
-    integer, parameter        :: repeat = 1
+    integer, parameter        :: repeat = 8
 
     integer(int32) ::             &
         blocks(0:test_size-1),    &
--- a/src/tests/sorting/test_sorting.f90
+++ b/src/tests/sorting/test_sorting.f90
@@ -179,7 +179,9 @@ contains
     subroutine test_int_ord_sorts( ltest )
         logical, intent(out) :: ltest
 
-        logical :: ldummy
+        integer(int64)       :: i
+        integer, allocatable :: d1(:)
+        logical              :: ldummy
 
         ltest = .true.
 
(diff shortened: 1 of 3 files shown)


jvdp1:
To review: @wclodius2

src/tests/sorting/test_sorting.f90
```diff
@@ -14,7 +14,7 @@ program test_sorting
     integer(int32), parameter :: char_size = 26**4
     integer(int32), parameter :: string_size = 26**3
     integer(int32), parameter :: block_size = test_size/6
-    integer, parameter        :: repeat = 8
+    integer, parameter        :: repeat = 1
```
gareth-nx, line 17: Is this change deliberate?

gareth-nx (commented):
Looks good to me, except (possibly) for the change in `repeat`, which I don't understand.

src/tests/sorting/test_sorting.f90
wclodius2: FWIW this decreases the number of times the sorts are repeated, making the test complete faster, but  potentially resulting in more noise in the comparison of the different tests.

wclodius2 (approved):
The changes to `stdlib_sorting_ord_sort.fypp`and `stdlib_sorting_sort_index.fypp` are all appropriate. The change to `repeat` in `test_sorting.f90` is  useful if the runtime is too long. The rest of the changes to test the changes the other codes are appropriate.

src/tests/sorting/test_sorting.f90
jvdp1: It was to speed up a it the tests. I forgot to revert it. Thank you for mentioning it. I wil revert it.

jvdp1, line 17:
```suggestion
    integer, parameter        :: repeat = 8
```

awvwgk (approved):
Looks good to me

jvdp1:
Could someone with  write access review this PR that fixes a bug in stdlib_sort?

LKedward (approved):
Cheers, LGTM 👍

milancurcic:
Reviews are positive, so I'll merge. Thanks!

PR #420, opened 2021-05-31
First implementation of real-valued linspace.

ejovo13:
Hi all,

This is my first time contributing on any project, ever, so bear with me if I make any missteps. As per issue #17, there seems to be a demand to implement some simple functions that return a range of values. 

So, I implemented `linspace` using MATLAB's syntax as a guide. Currently there is no support for passing complex values into the function but if this pull request gets approved I will implement it ASAP. This pull request is a gentle attempt to make my first contribution and to understand the ropes before I start trying to tackle other open issues.

I didn't know how to implement a robust test for checking all of the elements in the returned rank 1 array, so I instead check the first and last values, and the length of the returned array in the `test_linspace.f90` test.

Nevertheless, here is a quick demo when using the `linspace` function from another project which imports the fortran_stdlib:

image

Issue #296, opened 2021-01-10
test_loadtxt_qp error with gfortran-7.4.0 (32-bit) on Windows


brocolis:
I got an error testing test_loadtxt_qp.f90 with gfortran-7.4.0 (32-bit) on Windows.

Here are the debugging sessions.

GDB-8.2
![test_loadtxt_qp-gdb-8 2](https://user-images.githubusercontent.com/5913872/104114205-6f656500-52e0-11eb-8941-7f174c1f13df.png)

GDB-9.1
![test_loadtxt_qp-gdb-9 1](https://user-images.githubusercontent.com/5913872/104114207-70969200-52e0-11eb-8829-f81485d04743.png)

LKedward:
Hi @brocolis, I believe this test is not supported on Windows (due to quad precision) and is hence ignored when running the tests with cmake ctest.

If you're running these tests with the fpm package that [I recommended to you](https://github.com/fortran-lang/stdlib/issues/281#issuecomment-751440233) then unfortunately there is no way to ignore these tests on Windows (except for deleting them).

brocolis:
Hi @LKedward. After a few more testing I found that stdlib passes all tests when compiled by gfortran-9.3.0 (32 bit) or gfortran-8.4.0 (64 bit) or gfortran-10.2.0 (64 bit). When stdlib is compiled by gfortran-7.4.0 (32 bit), I have to remove tests _tests\io\test_loadtxt_qp.f90_, _tests\io\test_savetxt_qp.f90_ and _tests\stats\test_var.f90_ from the Makefile, then stdlib passes the test command.

| Compiler       |  Platform     | Build | Test |
| -------------- | ------------- |------ |------|
| gfortran 7.4.0 | mingw 32-bit  | ✅   | ❌   |
| gfortran 8.4.0 | mingw 64-bit  | ✅   | ✅   |
| gfortran 9.3.0 | mingw 32-bit  | ✅   | ✅   |
| gfortran 10.2.0| mingw 64-bit  | ✅   | ✅   |

### Build command
make -f Makefile.manual FYPPFLAGS=-DMAXRANK=4

### Test command
make -f Makefile.manual FYPPFLAGS=-DMAXRANK=4 test

LKedward:
Thanks for the comprehensive testing @brocolis, that's helpful in pinpointing this as a compiler bug.
I was testing with gfortran 8.1.0 on mingw 64-bit where the same tests build but fail to run.
[...]

PR #430, opened 2021-06-06
Remove support for GCC 7 and 8

awvwgk:
- remove GCC 8.1 from Windows workflow
- remove GCC 7.5 and GCC 8.4 from Ubuntu/MacOS workflow
- remove GCC workaround from stdlib_string_type
- move GCC <9 from supported to failing compilers
- create error in CMake for too old GCC version


ivan-pi (approved):
All looks good to me.

README.md
```diff
@@ -106,7 +105,7 @@ A list of tested compilers which are currently not working and the respective is
 
 Name | Version | Platform | Architecture | Status
 --- | --- | --- | --- | ---
-GCC Fortran | 7.4 | Windows 10 | i686 | [#296](https://github.com/fortran-lang/stdlib/issues/296)
+GCC Fortran | <9 | any | any | [#430](https://github.com/fortran-lang/stdlib/pull/430)
```
jvdp1, line 108: I think it would be good to keep the reference to #298, in case that someone is still using GCC Fortran 7.4 on Windows.

jvdp1 (approved):
LGTM, pending one minor comment

README.md
awvwgk:
```suggestion
GCC Fortran | <9 | any | any |  [#296](https://github.com/fortran-lang/stdlib/issues/296), [#430](https://github.com/fortran-lang/stdlib/pull/430)
```

LKedward:
I'm reading the README and it's not clear which issues were causing problems with GCC<9 on Ubuntu/MacOS - I though #296 only affected Windows? (Apologies if I missed something)

Romendakil:
In general, I think ropping older compiler versions is fine, however, I don't know what the vision is for stdlib. I'd assume that the goal would be to deliver stdlib as a precompiled library on OS in a similar fashion as libstdc++. This would mean that LTS OS versions which still have gcc 7/8 as their main compiler (Ubuntu 18 e.g.) could support stdlib only as an addition to their special gcc-9 and newer packages, but not for their system compiler.

awvwgk:
> I'm reading the README and it's not clear which issues were causing problems with GCC<9 on Ubuntu/MacOS

We are using one feature implemented in GCC 9 and later for `stdlib_string_type`, which was using a workaround before:
https://github.com/fortran-lang/stdlib/blob/1c8681299510d90f10245a4dddcaf3dd0ad802a5/src/stdlib_string_type.fypp#L1244-L1246

Also, for #420 we hit an issue where GCC 9 and later were required for usage with the real and imaginary component selectors
https://github.com/fortran-lang/stdlib/blob/e62f3fcc05f3367fe01df04902441d78a5a7e8c8/src/stdlib_math_linspace.fypp#L68-L69

> In general, I think ropping older compiler versions is fine, however, I don't know what the vision is for stdlib. I'd assume that the goal would be to deliver stdlib as a precompiled library on OS in a similar fashion as libstdc++. This would mean that LTS OS versions which still have gcc 7/8 as their main compiler (Ubuntu 18 e.g.) could support stdlib only as an addition to their special gcc-9 and newer packages, but not for their system compiler.

I have mixed feelings about LTS support of compilers, some project require me to write code compatible with GCC 4.9, the conda-forge Windows toolchain is running on GCC 5.3 and won't get updated any time soon. Using old GCC versions really limits the available features and the known bugs and issues can become tedious to workaround. Sticking with the supported compilers by GCC sounds like the easiest solution for me.

There are currently two lines of code that break GCC 7/8 support, which can be easily reverted, #420 could also be adapted to be compatible with GCC 7/8 without much problems. If we want to keep supporting those versions, this is still easily possible right now. And I'm open to reverting this patch.

LKedward:
Thanks for the explanation @awvwgk - I had missed the discussion in #420 so this PR was a little out of the blue. I've updated the PR description for anyone else who ends up here from the README.

> Sticking with the supported compilers by GCC sounds like the easiest solution for me.

This is probably the key point here, particularly regarding bugs and the future development/maintainability of `stdlib`, so this seems fine to me :+1:

Issue #427, opened 2021-06-05
Rank one (outer) product


ghbrown:
**Description**

A simple function to compute the rank one (a.k.a. outer) product of two vectors, which would return a matrix.

I believe the current way to do this (or at least the way I do) involves explicit looping or `reshape`ing and then `matmul` (even worse than looping probably). I just suggest the mathematics be hidden away from a user inside the function, so they can pass two true vectors in and it will return a matrix.

Rank one updates are sufficiently common that I think we should include a specific function to accommodate this. Here is my crack at a simple "column-major-optimized" version (I'd do speed tests comparing the various methods if we want this function):
```
function rop(u,v) result(mat)
    implicit none
    real, intent(in) :: u(:), v(:)
    real :: mat(size(u),size(v))
    integer :: col
    do col=1,size(v)
        mat(:,col) = v(col) * u
    end do
end function
```

Options/alternatives to a dedicated function:
- this functionality may eventually be provided by a more general tensor product routine (like those below), so perhaps this would just be a holdover
- something like this likely exists in BLAS somewhere in which case this function would be a wrapper for that if desired
- there could also be a good reason (that I don't) know why this feature doesn't exist, hasn't been proposed, and shouldn't be implemented (perhaps like the fact it requires quadratic storage for a linear amount of data in which case it might be preferable to update all elements in a loop instead, who knows?)

**Prior Art**

Numpy does not provide a rank-one-product-specific function, but provides this functionality through multiple functions, including:
- `numpy.outer(u,v)`
- `numpy.tensordot(u,v,axes=0)`
- `numpy.einsum('i,j',u,v)` or `numpy.einsum('i,j->ij',a,a)`
given vectors `u`, `v`.

**Relevant but separate discussion**
Perhaps this is just a result of my field of interest, but I'd like to bring up the discussion of terminology for the number of array dimensions (only since it's relevant here):
- I believe the current Fortran terminology is "rank", this is not optimal in my eyes because rank has a conflicting linear algebra definition (especially problematic since Fortran is commonly used for linear algebra); Python/Numpy also uses rank, but that doesn't make it "right"
-"order",  "mode", or "way" are my suggestions and what many people use as an alternative to rank in literature on multi-way arrays beyond matrices
- "dimension" seems to make sense but is almost ubiquitously used to describe the length of a mode, so I would say it's out of the question

ivan-pi:
Here is an implementation I have used previously:

```
module outer_prod_mod
  implicit none
[...]

Issue #429, opened 2021-06-06
Concurrencies in stdlib (do concurrent)


awvwgk:
*Should we use `do concurrent` in stdlib as parallel concurrency?*

There have been a lot of discussion about the `do concurrent` construct in Fortran ([j3-fortran], [discourse]), especially with respect to the question whether a concurrency should imply parallel execution, which some compilers ([intel], [nvidia]) already support. I don't want to open a discussion about the `do concurrent` construct here, instead I want to discuss how we can make best use of `do concurrent` in stdlib.

From my experience concurrent but not parallel constructs inside `do concurrent` can cause issues with compiler enabling aggressive parallelization for concurrencies. I therefore suggest to only use `do concurrent` for parallel concurrencies, unless the locality specification is explicitly given (Fortran 2018 feature).

It is important that we test the parallel concurrencies in our continuous integration workflows, this means we actually have to compile a parallel version of stdlib and enable the compiler support for parallelization of `do concurrent` in our build files, which we are currently not doing.

[j3-fortran]: https://github.com/j3-fortran/fortran_proposals/issues/62
[discourse]: https://fortran-lang.discourse.group/t/do-concurrent-issues-documented/277
[intel]: https://software.intel.com/content/www/us/en/develop/documentation/fortran-compiler-oneapi-dev-guide-and-reference/top/language-reference/a-to-z-reference/c-to-d/do-concurrent.html
[nvidia]: https://developer.nvidia.com/blog/accelerating-fortran-do-concurrent-with-gpus-and-the-nvidia-hpc-sdk/

LKedward:
Would this be mutually exclusive with #213 (OpenMP)? If so I would think OpenMP is preferable due to better compiler support no? For those with less experience with `do concurrent` (me), are you able to elaborate on what you mean by parallel and non-parallel concurrencies in `do concurrent` with examples and the issues you mentioned with the latter? Another basic question (sorry): is `do concurrent` deterministic, since this will affect whether we can effectively test it in our CI?

awvwgk:
I think for our purpose concurrencies and OpenMP are in principle similar and therefore more or less mutually exclusive.

[...]

PR #432, opened 2021-06-10
Outer product

ghbrown:
Hi, all! I have implemented an outer product function (for vectors only) `outer_product` as discussed in #427.

This is my first pull request, so I wanted to check in early before I got going too far in the wrong direction. I have done speed tests on different implementations, implemented the most performant version, added the submodule files to the Makefiles, and made sure the project built. I did not see any other tests I could run locally, but perhaps there are additional tools for this.

Known deficiencies of pull request (to be addressed after feedback):
- no tests
- no documentation

Speed test results
```
--- for two vectors of size n= 15000 ---
outer_product_spread() :    4.30 s  (spread vectors into matrices of proper size and multiply elementswise)
outer_product_rowwise():    3.33 s  (assign result rowwise)
outer_product_matmul():     0.49 s  (reshape input vectors into mx1 and 1xn matrices, then matmul())
outer_product_columnwise(): 0.33 s  (assign result columnwise, THIS IS THE IMPLEMENTATION USED)
```

Thank you!


src/stdlib_linalg.fypp
```diff
@@ -63,6 +65,22 @@ module stdlib_linalg
...
+  interface outer_product
+    !! version: experimental
+    !!
+    !! Computes outer product of two vectors, returning a matrix
+    !! ([Specification](../page/specs/stdlib_linalg.html#description_3))
+    #:for k1, t1 in RCI_KINDS_TYPES
+     module function outer_product_${t1[0]}$${k1}$(u,v) result(res)
+       ${t1}$, intent(in) :: u(:), v(:)
+       ${t1}$ :: res(size(u),size(v))
+       integer :: col
```
ivan-pi, line 79: I think it's best to leave `col` in the submodule part with the implementation.

src/tests/linalg/test_linalg.f90
```diff
@@ -56,6 +56,24 @@ program test_linalg
   call test_trace_int32
   call test_trace_int64
 
+  !
+  ! outer product
+  !
```
ivan-pi, line 61: You haven't implemented the tests yet right? In either case, the overall structure is fine. I would suggest some simple cases with vectors of length 2, 3, and 4:

```
u = [1.,2.]
v = [1.,3.]
expected = reshape([1.,3.,2.,6.],[2,2])
abs_diff = abs(expected - outer_product(u,v)) 
call check(all(abs_diff < sptol), ...)
```

src/stdlib_linalg_outer_product.fypp
```diff
@@ -0,0 +1,20 @@
...
+  implicit none
+
+contains
+
+    #:for k1, t1 in RCI_KINDS_TYPES
+     module function outer_product_${t1[0]}$${k1}$(u,v) result(res)
+       ${t1}$, intent(in) :: u(:), v(:)
+       ${t1}$ :: res(size(u),size(v))
+       integer :: col
+       do col=1,size(v)
```
ivan-pi, line 14:
```suggestion
       do col = 1, size(v)
```

ivan-pi (commented):
Overall, I think it's a great start!

Thanks for doing the benchmark. It would be interesting to add the results of using `do_concurrent` (take a look at https://github.com/fortran-lang/stdlib/issues/429). As far as I can tell the columns of the result matrix are independent.

src/stdlib_linalg.fypp
GHBrown: Definitely, I just copy and pasted indiscriminately from the submodule.

awvwgk:
I don't think this is a good example to test `do concurrent`. First, we can't actually test this within stdlib, because the compiler specific optimization for concurrencies is disabled by default. If you test them outside stdlib, make sure to enable vectorization and parallelization of concurrencies in your compiler (in `ifort` you have to add `-qopenmp` for this purpose).

I attempted a test on 8 cores of an AMD EPYC 7742 using `ifort -Ofast -qopenmp -march=native -fma` (dim=20000, 20 repeats):

implementation | average runtime
--- | ---
omp parallel do simd | 7.2194s
do concurrent | 7.9281s
normal do loop | 9.5253s

My conclusion from those numbers here is that parallel execution will only burn more CPU time than doing any good for this kind of problem.

GHBrown:
@awvwgk Thank you for the more detailed tests and suggestions. I believe this means my tests probably didn't even enable parallelization based on your comment. I will leave it as a standard `do` loop.

I'm working on getting the tests done now, but I am a bit worried that `git status` informs me of a horde of new files that will be committed. It seems like mainly `.f90` files and files inside `tests/` with no extensions (executables probably?); I believe these files are from building the project.

How do you all commit after confirming a change builds properly? Do you `make clean` to clear all the artifacts of the build, or have I just configured something wrong? My assumption was that such files would be excluded by `.gitignore`s, but that doesn't seem to be the case.

awvwgk:
@GHBrown I highly recommend to build out-of-tree (`cmake -B build` instead of just `cmake`, described in https://github.com/fortran-lang/stdlib#build-with-cmake) as this separates the build artifacts from the actual source.

milancurcic:
Thanks @GHBrown! Pending docs and simple tests as suggested by @ivan-pi, I think this is good to go.

GHBrown:
The tests should now be completed and committed. I will work on the documentation now.

@awvwgk Thanks. Yes, I had been using that out-of-tree make command. Even after using `cmake --build build --target clean` I still have the following new files:
```
	src/tests/logger/fifth_log_file.txt
	src/tests/logger/first_log_file.txt
	src/tests/logger/fourth_log_file.txt
	src/tests/logger/second_log_file.txt
	src/tests/logger/sixth_log_file.txt
	src/tests/logger/third_log_file.txt
	src/tests/sorting/test_sorting.txt
	src/tests/string/scratch.txt
```
Not a big deal, just curious. Perhaps I hit return too early after entering `cmake`, but even so, I guess I expect `clean` and `.gitignore`s to cover everything.

GHBrown:
I believe these last three commits should cover the documentation. I made a few assumptions about how the documentation works in `stdlib_linalg.fypp`, but I think it should all be okay.

Please let me know if there are any outstanding issues to address. Thanks.

doc/specs/stdlib_linalg.md
```diff
@@ -168,3 +168,41 @@ program demo_trace
...
+
+## `outer_product` - Computes the outer product of two vectors
+
+### Status
+
+Experimental
+
+### Description
+
+Create a diagonal array or extract the diagonal elements of an array
```
ivan-pi, line 180: Mind fixing the description?

doc/specs/stdlib_linalg.md
```diff
@@ -168,3 +168,41 @@ program demo_trace
...
+
+### Arguments
+
+`u`: Shall be a rank-1 array
+
+`v`: Shall be a rank-1 array
+
+### Return value
+
+Returns a rank-2 array equal to `u v^T` (assuming `u, v` are column vectors). The dimensions of the returned array are `[size(u), size(v)]`
```
ivan-pi, line 194: Instead of "The dimensions" I would say "The shape". Fortran has an intrinsic function `shape`  which will return a rank-1 array of sizes.

doc/specs/stdlib_linalg.md
```diff
@@ -168,3 +168,41 @@ program demo_trace
...
+
+### Example
+
+```fortran
+program demo_outer_product
+    use stdlib_linalg, only: outer_product
+    implicit none
+    real, allocatable :: A(:,:), u(:), v(:)
+    u = [1.,2.,3.]
```
ivan-pi, line 203: For consistency with the rest of the example it's best to add spaces following commas.

doc/specs/stdlib_linalg.md
```diff
@@ -168,3 +168,41 @@ program demo_trace
...
+    v = [3., 4.]
+    A=outer_product(u,v)
+    !A = transpose(reshape([3., 4., 6., 8., 9., 12.], [3,2]))
```
ivan-pi, line 206: The transpose of a 3×2 array will be a 2×3 array, which disagrees with the description of `outer_product`. I think it's best to just use a reshape, and specify column by column:

 ```
! A = reshape([3., 6., 9., 4., 8., 12.], [3, 2])
```

File: doc/specs/stdlib_linalg.md [hunk 8]
```diff
@@ -168,3 +168,41 @@ program demo_trace
...
+### Example
+
+```fortran
+program demo_outer_product
+    use stdlib_linalg, only: outer_product
+    implicit none
+    real, allocatable :: A(:,:), u(:), v(:)
+    u = [1.,2.,3.]
+    v = [3., 4.]
+    A=outer_product(u,v)
```
ivan-pi, line 205: Missing spaces before and after equal sign.

awvwgk (approved):
Thanks for contributing here, this is a useful addition to stdlib.

doc/specs/stdlib_linalg.md
```diff
@@ -168,3 +168,41 @@ program demo_trace
...
+
+## `outer_product` - Computes the outer product of two vectors
+
+### Status
+
+Experimental
+
+### Description
+
+Computes the outer product of two vectors
```
jvdp1, line 180: I suggest to swap the sections `Description` and `Status`

doc/specs/stdlib_linalg.md
```diff
@@ -168,3 +168,41 @@ program demo_trace
...
+
+### Arguments
+
+`u`: Shall be a rank-1 array
+
+`v`: Shall be a rank-1 array
+
+### Return value
+
+Returns a rank-2 array equal to `u v^T` (assuming `u, v` are column vectors). The shape of the returned array is `[size(u), size(v)]`.
```
jvdp1, line 194:
```suggestion
Returns a rank-2 array equal to `u v^T` (where `u, v` are considered as column vectors). The shape of the returned array is `[size(u), size(v)]`.
```

doc/specs/stdlib_linalg.md
```diff
@@ -168,3 +168,41 @@ program demo_trace
...
+
+### Example
+
+```fortran
+program demo_outer_product
+    use stdlib_linalg, only: outer_product
+    implicit none
+    real, allocatable :: A(:,:), u(:), v(:)
+    u = [1. ,2. ,3. ]
```
jvdp1, line 203:
```suggestion
    u = [1., 2., 3. ]
```

src/stdlib_linalg.fypp
```diff
@@ -63,6 +65,21 @@ module stdlib_linalg
     #:endfor
   end interface
 
+
+  ! Outer product (of two vectors)
+  interface outer_product
+    !! version: experimental
+    !!
+    !! Computes the outer product of two vectors, returning a matrix
```
jvdp1, line 73:
```suggestion
    !! Computes the outer product of two vectors, returning a rank-2 array
```

src/tests/linalg/test_linalg.f90
```diff
@@ -424,7 +440,129 @@ subroutine test_trace_int64
...
+  subroutine test_outer_product_rsp
+    integer, parameter :: n = 2
+    real(sp) :: u(n), v(n), expected(n,n), diff(n,n)
+    write(*,*) "test_outer_product_rsp" 
+    u = [1.,2.]
+    v = [1.,3.]
+    expected = reshape([1.,2.,3.,6.],[n,n])
+    diff = expected - outer_product(u,v)
+    call check(all(abs(diff) < sptol), &
+         msg="all(abs(diff) < sptol) failed.",warn=warn)
```
jvdp1, line 455: several tests have the same name, which can be tricky if an issue appear. I suggest to at least add the name of the function and the kind/type, e.g., 
```suggestion
         msg="outer rsp all(abs(diff) < sptol) failed.",warn=warn)
```

src/stdlib_linalg_outer_product.fypp
```diff
@@ -0,0 +1,20 @@
+#:include "common.fypp"
+#:set RCI_KINDS_TYPES = REAL_KINDS_TYPES + CMPLX_KINDS_TYPES + INT_KINDS_TYPES
+submodule (stdlib_linalg) stdlib_linalg_outer_product
+
+  implicit none
+
+contains
+
+  #:for k1, t1 in RCI_KINDS_TYPES
+    module function outer_product_${t1[0]}$${k1}$(u, v) result(res)
```
jvdp1, line 10: Could it be `pure`

jvdp1 (commented):
Thank you for this nice addition. Here are some comments/suggestions.

doc/specs/stdlib_linalg.md
GHBrown: Thank you for the comments.
This one puzzles me though, partly because I'm not sure exactly what you are suggesting. Are you suggesting we swap the contents of these sections, the order of these two sections, or something else?
In any case, it's not clear to me why we would want to do this. Thanks.

jvdp1: Sorry for the confusion. What I tried to suggest is to have the following order:
```
### Description

Comment (jvdp1) [reply doc/specs/stdlib_linalg.md:180]: >  Perhaps if we decide to switch the order for these sections in the docs we would want to make a separate issue and pull rather than tucking it away in this PR? If I were to change just my function it wouldn't match the rest of the docs here.

I agree with you. Sorry, and ignore this comment.

File: src/tests/linalg/test_linalg.f90 (hunk 13 above)
Comment (ghbrown) [reply src/tests/linalg/test_linalg.f90:455]: I believe this is the only outstanding comment/change left.

I'm confused what the point of the `write (*,*), ...` lines are in all of the tests. These contain the test name, so perhaps we wouldn't need to put it in the `msg`? However, I've purposefully failed the `linalg` test and `grep`ed for the contents of the `write (*,*)` statement text and the `msg` text. I'm sure I've missed something, but to me it seems inconclusive whether there is enough info for a failed test to be traced back to the function.

Again, I tried to follow the previous examples in `test_linalg.f90`, which did not contain names in the `msg`s, but I'm not exactly sure how the testing and failure logging is done.

Comment (jvdp1) [reply src/tests/linalg/test_linalg.f90:455]: If I am not wrong, the `msg` of the other tests are uniq. Therefore, it is easy to retrieve which tests failed. In your cases, the same `msg` is used by multiple tests, and it would need a bit more thinking to retrieve which test failed (by combining with the `write(*,*)`). I think it would be easier to trace a failed test if the `msg` and the `write(*,*)` would be combined into `msg`. However, I might be wrong.

Comment (ghbrown) [reply src/tests/linalg/test_linalg.f90:455]: Yes, you are correct, they are all technically unique. The `msg`s don't necessarily have the function name in them (what I was expecting and looking for) but are unique enough to be `grep`ed, which I suppose is good enough for a quick traceback.

I'll make something like your suggested change for each test. I'm still not exactly clear on the point of `write` and `msg`, since I couldn't find either text even after a failed test, but this change will put `outer_product` in line with the others...

Anyway, thanks for catching this.

Comment (ghbrown) [reply src/tests/linalg/test_linalg.f90:455]: I dug back into `check` and looked at `stdlib_error.f90` and still can't quite figure out what's happening with logs and error handling when testing. I'd like to better understand the error handling and logging for tests before I make this change.

From what I can deduce, this is the chain:
- start "execution" of `test_linalg.f90`
- `write(*,*), test_function_typeprecision` to terminal (or default out, not sure where that would be set?), though I'm not seeing this text in terminal when testing
- test fails in `check` statement
- since `warn=.false.` the program should stop via a call to `error_stop` (whose function definition I cannot find) with `msg`, which passes `msg` to `error stop` stopping execution

I am not sure if the above is quite right, but  the important source of confusion for me is: if a test fails, what is the proper way to determine this and trace back to the offending function? All I can see when testing is `stdlib_linalg........**Failed` and then
```
The following tests FAILED:
	 12 - linalg (Failed)
Errors while running CTest
make: *** [Makefile:130: test] Error 8
```
I don't even see info about a log file when running `cmake --build build --target test`, and even after `grep`ing for the `msg` (which I only know because I wrote the tests) I still turn up nothing related to why this test failed (note: I'm artificially making it fail). Perhaps I'm missing something, but with the provided info in terminal, it seems like it would be quite difficult to accurately trace this error, regardless of the `msg` text.

@ivan-pi Would you mind weighing in on this message and this comment thread?

Comment (jvdp1) [reply src/tests/linalg/test_linalg.f90:455]: I am not sure what you mean, but you may want to look in `Testing/Temporary/LastTest.log`

Comment (ghbrown) [reply src/tests/linalg/test_linalg.f90:455]: Crazy, not sure why that didn't come up via `grep`, but thanks.

Anyway, here's what I see when I purposefully fail `test_outer_product_rsp`:
```
...
 test_trace_int32
 test_trace_int64
 test_outer_product_rsp
 all(abs(diff) < sptol) failed.
ERROR STOP 1
```

The function name prints right before the `msg`, then execution of that test collection stops. Since the function name is right there, it should be just as easy to trace as any of the other tests.

However, one change I think might make sense is to pad `msg` with spaces on the left, so the failure message looks like it belongs to the failing test above it (for example):
```
...
 test_trace_int32
 test_trace_int64
 test_outer_product_rsp
    all(abs(diff) < sptol) failed.
ERROR STOP 1
```
but I think the log is clear enough as it stands, and this would probably be another PR itself.

jvdp1: I always prefer to have the whole information in one line. But it could be matter of taste. If it is ok for others like that, it will be also ok for me

milancurcic (approved):
Thank you @ghbrown and reviewers! I will wait one more day before merging in case there are any more requests for changes.

jvdp1 (approved):

PR #433, opened 2021-06-10
implemented low level find function for string matching

### LINKED ISSUE #421: find function for substring matching
**Description**

Signature: find(string, pattern, start(Optional) = `1`, end(Optional) = `len(string)`, occurrence(Optional) = `1`, consider_overlapping(Optional)= `.true.`)

Output: an integer (index)

Returns the index at which `occurrence`'th occurrence of substring `pattern` is present in the input string `string`
between the index `start` and `end`, if not found returns `0`.
`consider_overlapping`: [comment #1](https://github.com/fortran-lang/stdlib/issues/366#issuecomment-808897041) & [comment #2](https://github.com/fortran-lang/stdlib/issues/366#issuecomment-810893962)

**Prior Art**

Fortran has an intrinsic function `index` ([description](http://www.lahey.com/docs/lfpro73help/F95ARINDEXFn.htm)) which does the same thing but has limited features.

PS: Some of the features proposed above may not be best to add to the stdlib at this moment, hence there is also a need to discuss the importance of these features. They are added in the issue to bring them into the discussion.

Aman-Godara:
closes #421

Tasks:

- [x] Implemented `find` function
- [x] Added unit tests in `test_strings_functions.f90`
- [x] Document the newly created function
- [x] Function find: Pure or Elemental? -> Elemental

### DIFF
--- a/doc/specs/stdlib_strings.md
+++ b/doc/specs/stdlib_strings.md
@@ -270,3 +270,62 @@ program demo_slice
 
 end program demo_slice
 ```
+
+
+<!-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -->
+### `find`
+
+#### Description
+
+Returns the starting index of the `occurrence`th occurrence of the substring `pattern` 
+in the input string `string`.  
+Default value of `occurrence` is set to `1`. 
+If `consider_overlapping` is not provided or is set to `.true.` the function counts two overlapping occurences of substring as two different occurrences.  
+If `occurrence`th occurrence is not found, function returns `0`.
+
+
+#### Syntax
+
+`string = [[stdlib_strings(module):find(interface)]] (string, pattern, occurrence, consider_overlapping)`
+
+#### Status
+
+Experimental
+
+#### Class
+
+Pure function
+
+#### Argument
+
+- `string`: Character scalar or [[stdlib_string_type(module):string_type(type)]]
+  This argument is intent(in).
+- `pattern`: Character scalar or [[stdlib_string_type(module):string_type(type)]]
+  This argument is intent(in).
+- `occurrence`: integer
+  This argument is intent(in) and optional.
+- `consider_overlapping`: logical
+  This argument is intent(in) and optional.
+
+#### Result value
+
+The result is of scalar integer type.
+
+#### Example
+
+```fortran
+program demo_find
+  use stdlib_string_type
+  use stdlib_strings, only : find
+  implicit none
+  string_type :: string
+
+  string = "needle in this character-stack"
+
+  print *, find(string, "needle")   ! 1
+  print *, find(string, "a", 3)     ! 28
+  print *, find("qwqwqwq", "qwq", 3, .false.)  ! 0
+  print *, find("qwqwqwq", "qwq", 3, .true.)   ! 5
+
+end program demo_find
+```
--- a/doc/specs/stdlib_strings.md
+++ b/doc/specs/stdlib_strings.md
@@ -315,7 +315,7 @@ The result is a scalar of integer type or integer array of rank equal to the hig
 
 ```fortran
 program demo_find
-  use stdlib_string_type
+  use stdlib_string_type, only: string_type, assignment(=)
   use stdlib_strings, only : find
   implicit none
   string_type :: string
--- a/doc/specs/stdlib_strings.md
+++ b/doc/specs/stdlib_strings.md
@@ -294,7 +294,7 @@ Experimental
 
 #### Class
 
-Pure function
+Elemental function
 
 #### Argument
 
@@ -309,7 +309,7 @@ Pure function
 
 #### Result value
 
-The result is of scalar integer type.
+The result is a scalar of integer type or integer array of rank equal to the highest rank among all dummy arguments.
 
 #### Example
 
@@ -320,12 +320,11 @@ program demo_find
   implicit none
   string_type :: string
 
-  string = "needle in this character-stack"
+  string = "needle in the character-stack"
 
-  print *, find(string, "needle")   ! 1
-  print *, find(string, "a", 3)     ! 28
-  print *, find("qwqwqwq", "qwq", 3, .false.)  ! 0
-  print *, find("qwqwqwq", "qwq", 3, .true.)   ! 5
+  print *, find(string, "needle")                       ! 1
+  print *, find(string, ["a", "c"], [3, 2])             ! [27, 20]
+  print *, find("qwqwqwq", "qwq", 3, [.false., .true.]) ! [0, 5]
 
 end program demo_find
 ```
--- a/doc/specs/stdlib_strings.md
+++ b/doc/specs/stdlib_strings.md
@@ -233,13 +233,13 @@ Pure function.
 
 #### Argument
 
-- `string`: Character scalar or [[stdlib_string_type(module):string_type(type)]]
+- `string`: Character scalar or [[stdlib_string_type(module):string_type(type)]].
   This argument is intent(in).
-- `first`: integer
+- `first`: integer.
   This argument is intent(in) and optional.
-- `last`: integer
+- `last`: integer.
   This argument is intent(in) and optional.
-- `stride`: integer
+- `stride`: integer.
   This argument is intent(in) and optional.
 
 #### Result value
@@ -298,13 +298,13 @@ Pure function
 
 #### Argument
 
-- `string`: Character scalar or [[stdlib_string_type(module):string_type(type)]]
+- `string`: Character scalar or [[stdlib_string_type(module):string_type(type)]].
   This argument is intent(in).
-- `pattern`: Character scalar or [[stdlib_string_type(module):string_type(type)]]
+- `pattern`: Character scalar or [[stdlib_string_type(module):string_type(type)]].
   This argument is intent(in).
-- `occurrence`: integer
+- `occurrence`: integer.
   This argument is intent(in) and optional.
-- `consider_overlapping`: logical
+- `consider_overlapping`: logical.
   This argument is intent(in) and optional.
 
 #### Result value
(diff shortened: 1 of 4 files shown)


src/stdlib_strings.f90
```diff
@@ -4,13 +4,15 @@
...
 !> The specification of this module is available [here](../page/specs/stdlib_strings.html).
 module stdlib_strings
-    use stdlib_ascii, only : whitespace
-    use stdlib_string_type, only : string_type, char, verify
+    use stdlib_ascii, only: whitespace
+    use stdlib_string_type, only: string_type, char, verify
+    use stdlib_optval, only: optval
     implicit none
-    private
+    private :: compute_lps
```
awvwgk, line 11: Just leave the private statement here, it sets the default for all routines not explicitly declared
```suggestion
    private
```

src/stdlib_strings.f90
```diff
@@ -290,5 +309,193 @@ elemental function ends_with_string_string(string, substring) result(match)
...
+        logical :: consider_overlapping_
+
+        consider_overlapping_ = optval(consider_overlapping, .true.)
+        occurrence_ = optval(occurrence, 1)
+        res = 0
+        length_string = len(string)
+        length_pattern = len(pattern)
+
+        if (length_pattern > 0 .and. length_pattern <= length_string & 
+        .and. occurrence_ > 0) then
```
awvwgk, line 440: Keep leading continuation character for readability
```suggestion
            & .and. occurrence_ > 0) then
```

awvwgk (commented):
Thanks, this looks great so far.

Have you checked how robust your implementation works for an empty string input to the string and/or pattern?

milancurcic:
Thanks, great progress so far. A few tests segfault, but I assume you're working on that.

What do you think about making `find` an `elemental` function instead of just `pure`?

Aman-Godara:
I checked for empty `string`/`pattern` input: code was working fine. But I couldn't find the segfault @milancurcic mentioned. Can you please send me that particular test case which gave this error?

milancurcic:
> But I couldn't find the segfault @milancurcic mentioned.

This may be specific to my setup. I'll look at it later today and report back.

awvwgk:
> I checked for empty `string`/`pattern` input: code was working fine.

Sounds good, could you add those cases to the testsuite as well?

Aman-Godara:
> I think this branch should be rebased against the latest changes on the default branch.

Are you talking about the pull request 419 and 428?
I have already merged my `develop_slice` branch in this `develop_find` branch through f0b9a0d.

awvwgk:
> > I think this branch should be rebased against the latest changes on the default branch.
> 
> Are you talking about the pull request 419 and 428?
> I have already merged my `develop_slice` branch in this `develop_find` branch through [f0b9a0d](https://github.com/fortran-lang/stdlib/commit/f0b9a0de8238d35ee92991ae862a7a4ebbd75a2b).

Looks like it is currently not picked up correctly, there are conflicts in `stdlib_strings` and commits already on the default branch appear again here. Can you try to rebase your branch?

awvwgk:
Thanks, the tests look complete now. Let us know once this PR is ready for review.

Aman-Godara:
> What do you think about making `find` an `elemental` function instead of just `pure`?

Currently this function is `pure`, I will gather some information and write back to you.

> Sounds good, could you add those cases to the testsuite as well?

Done

> Can you try to rebase your branch?

Thanks for this, done. That's exactly what I was looking for.

I didn't get any errors from `cmake` testing on my local machine but here on GitHub `CI / Build (ubuntu-latest, 10) (pull_request)` says there is some problem with `optval`.

awvwgk:
> I didn't get any errors from `cmake` testing on my local machine but here on GitHub `CI / Build (ubuntu-latest, 10) (pull_request)` says there is some problem with `optval`.

There is something broken with the manual makefile. Not sure what in particular is broken there, will have a look later.

Aman-Godara:
There is a possible optimization that I missed here when `consider_overlapping` is `.true.`. Give me some time to add it before you leave your final feedback.

Aman-Godara:
> What do you think about making `find` an `elemental` function instead of just `pure`?

I gave some thought to it, it can be made `elemental`.

Argument `string` and `pattern` cannot be passed in an array unless they both are of type `string_type` because, if otherwise, they might have varying lengths causing a problem to pass in the same array. Thus, I will have to remove `find_char_char` function.

src/Makefile.manual
```diff
@@ -126,7 +126,8 @@ stdlib_stats_distribution_PRNG.o: \
         stdlib_kinds.o \
 	stdlib_error.o
 stdlib_string_type.o: stdlib_ascii.o \
-	stdlib_kinds.o
+    stdlib_kinds.o
 stdlib_strings.o: stdlib_ascii.o \
-	stdlib_string_type.o
+    stdlib_string_type.o \
+	stdlib_optval
```
awvwgk, line 132:
```suggestion
	stdlib_optval.o
```

awvwgk:
> > What do you think about making `find` an `elemental` function instead of just `pure`?
> 
> I gave some thought to it, it can be made `elemental`.
> 
> Argument `string` and `pattern` cannot be passed in an array unless they both are of type `string_type` because, if otherwise, they might have varying lengths causing a problem to pass in the same array. Thus, I will have to remove `find_char_char` function.

`elemental` sounds fine to me. I would vote for making all of them `elemental`. Searching a whole array of fixed length characters still seems seems like a valid application for find, even if slightly inconvenient. We don't run into a situation of returning strings of different length here, therefore I think `elemental` will be fine.

awvwgk (approved):
Thanks a lot, this looks good now.

doc/specs/stdlib_strings.md
```diff
@@ -270,3 +270,61 @@ program demo_slice
...
+Returns the starting index of the `occurrence`th occurrence of the substring `pattern` 
+in the input string `string`.  
+Default value of `occurrence` is set to `1`. 
+If `consider_overlapping` is not provided or is set to `.true.` the function counts two overlapping occurrences of substring as two different occurrences.  
+If `occurrence`th occurrence is not found, function returns `0`.
+
+
+#### Syntax
+
+`string = [[stdlib_strings(module):find(interface)]] (string, pattern, occurrence, consider_overlapping)`
```
jvdp1, line 289:
```suggestion
`string = [[stdlib_strings(module):find(interface)]] (string, pattern [, occurrence, consider_overlapping])`
```

To highlight that these 2 arguments are `optional`

doc/specs/stdlib_strings.md
```diff
@@ -270,3 +270,61 @@ program demo_slice
...
+
+#### Result value
+
+The result is a scalar of integer type or integer array of rank equal to the highest rank among all dummy arguments.
+
+#### Example
+
+```fortran
+program demo_find
+  use stdlib_string_type
```
jvdp1, line 318:
```suggestion
  use stdlib_string_type, only: string_type
```

src/stdlib_strings.f90
```diff
@@ -67,6 +68,15 @@ module stdlib_strings
         module procedure :: slice_char
     end interface slice
 
+    !> Finds the starting index of substring 'pattern' in the input 'string'
+    !>
```
jvdp1, line 72:
```suggestion
    !> [Specifications](link to the specs - to be completed)
    !> 
```

jvdp1 (approved):
LGTM. Thank you @Aman-Godara . IMO It can be merged, pending minor changes.

doc/specs/stdlib_strings.md
Aman-Godara: If `only` specifier is used here, `assignment(=)` should also be added after `string_type` because it is used in the line 323.

Aman-Godara:
I added brackets ( `[` and `]`) to specify that arguments are optional for `slice` function (#414) in this PR.

everythingfunctional (commented):
Often there is a `back=.false.` argument to functions like this (i.e. the intrinsic function `index`, which is very much like this one). I'm about to leave on vacation so I won't have time for a much more in depth review. At a quick glance it otherwise looks ok.

@Aman-Godara (and anybody else), I recommend reading the book The Art of Unit Testing by Roy Osherove to improve your unit test skills, if you can find the time.

Aman-Godara:
> Often there is a `back=.false.` argument to functions like this (i.e. the intrinsic function `index`, which is very much like this one).

A possible way IMO of doing this is to make a new function `find_low_level` and cut & paste all the code written in `find_char_char` to this new function. I will keep this new function `private` and out of the interface `find`. Then I can add `back` argument to all the 4 functions that are in `find` interface. So, these 4 functions will call the `find_low` in their implementation. `find_low` will not take any `back` argument, rather these 4 high level functions will take care of the `back` argument and modify the input given by user accordingly before calling `find_low_level`.

everythingfunctional:
> > Often there is a `back=.false.` argument to functions like this (i.e. the intrinsic function `index`, which is very much like this one).
> 
> A possible way IMO of doing this is to make a new function `find_low_level` and cut & paste all the code written in `find_char_char` to this new function. I will keep this new function `private` and out of the interface `find`. Then I can add `back` argument to all the 4 functions that are in `find` interface. So, these 4 functions will call the `find_low_level` in their implementation. `find_low_level` will not take any `back` argument, rather these 4 high level functions will take care of the `back` argument and modify the input given by user accordingly before calling `find_low_level`.

Sounds reasonable to me.

jvdp1:
> > Often there is a `back=.false.` argument to functions like this (i.e. the intrinsic function `index`, which is very much like this one).
> 
> A possible way IMO of doing this is to make a new function `find_low_level` and cut & paste all the code written in `find_char_char` to this new function. I will keep this new function `private` and out of the interface `find`. Then I can add `back` argument to all the 4 functions that are in `find` interface. So, these 4 functions will call the `find_low_level` in their implementation. `find_low_level` will not take any `back` argument, rather these 4 high level functions will take care of the `back` argument and modify the input given by user accordingly before calling `find_low_level`.

@Aman-Godara @awvwgk Do you still plan to investigate/implement this option, suggested by @everythingfunctional ?

awvwgk:
Let's open a new issue on this and continue the discussion there.

Issue #366, opened 2021-03-27
Replace substring in character/string_type


ivan-pi:
I also propose to add a few other functions that can be commonly used.

- Replace function, this can be implemented in several ways:
   - First implementation: string.replace(pos , len, new_string)
     - pos = starting position of the sub-string to be replaced.
     - len = length of the substring
     - new_string = the string that will replace the substring
   - Second implementation: string.replace(pos, old_string, new_string)
      - old_string = substring to be replaced.
      - pos = position to start the search of old_string
      - new_string = string that will replace old_substring
      - This function will replace only the first occurrence of the old_string.
   - Third implementation: string.replaceall(pos, old_string, new_string)
     - This meaning of variables is similar to the second implementation.
     - The difference is it will find and replace all the occurrences of old_string after pos and not just one.

- We already have an operator to Concat a string with another but I feel we should also implement a function for the same. 

The community can have a discussion over the need for such functions and if needed also suggest some other implementation of Replace function.

_Originally posted by @ChetanKarwa in https://github.com/fortran-lang/stdlib/issues/364#issuecomment-808800508_

ivan-pi:
This has also been requested a few times before.

@zbeekman wrote in https://github.com/fortran-lang/stdlib/issues/69#issuecomment-570690516:

> sub: from Ruby, replaces the first occurance of a substring with a substitution, optional argument to replace the last string
> gsub: Replaces all occurrences with a substring

awvwgk:
There is also a proposal from the string thread:

> ## Gsub
>
> * Global substitution. Same as @jacobwilliams' [`string_replace`](https://github.com/j3-fortran/fortran_proposals/issues/96#issuecomment-557896465) as far as I can tell. Eventually it would be nice to support regular expressions [like Ruby](https://ruby-doc.org/core-2.6/String.html#method-i-gsub), but, for now, we should start smaller and just make it behave like:
  `sed 's/<pattern>/<replacement>/g'`
> * Kinda [Ruby](https://ruby-doc.org/core-2.6/String.html#method-i-gsub), [Python](https://docs.python.org/3.8/library/stdtypes.html#str.replace) and `sed`. Behave like Python without the optional `count` argument
> * Fortran:
>  ``` fortran
>  gsub("hello", "l", "L") ! "heLLo"
>  gsub("the quick brown fox jumped over the lazy dog", "the", "a")
>      ! "a quick brown fox jumped over a lazy dog"
>  !! Or OO approach
[...]

PR #436, opened 2021-06-15
implemented low-level replace_all function

Aman-Godara:
#366 

Current implement provides with `replace_overlapping` optional dummy argument which is `logical`.
This argument helps provide functionality of both OPTION: 1 and OPTION: 2 of [overlapping problem](https://github.com/fortran-lang/stdlib/issues/366#issuecomment-808897041) depending upon whether it is set to `.true.` or `.false.`.

[EDIT]: `replace_overlapping` has been removed from `replace_all` now.

Tasks:

- [x] Implemented `replace_all` function
- [x] Added some test cases for `replace_all` to show the behaviour of the function
- [x] Added more test cases for `replace_all`
- [x] Document the function
- [x] Pure or Elemental -> Pure.

### DIFF
--- a/src/stdlib_strings.f90
+++ b/src/stdlib_strings.f90
@@ -612,7 +612,7 @@ contains
         character(len=*), intent(in) :: string
         character(len=*), intent(in) :: pattern
         character(len=*), intent(in) :: replacement
-        character(:), allocatable :: res
+        character(len=:), allocatable :: res
         integer :: lps_array(len(pattern))
         integer :: s_i, p_i, last, length_string, length_pattern
 
@@ -626,7 +626,7 @@ contains
 
             s_i = 1
             p_i = 1
-            do while(s_i <= length_string)
+            do while (s_i <= length_string)
                 if (string(s_i:s_i) == pattern(p_i:p_i)) then
                     if (p_i == length_pattern) then
                         res = res // &
--- a/src/stdlib_strings.f90
+++ b/src/stdlib_strings.f90
@@ -367,7 +367,7 @@ contains
         end if
 
         if (present(first)) then
-            first_index =  first
+            first_index = first
         end if
         if (present(last)) then
             last_index = last
@@ -522,7 +522,7 @@ contains
         type(string_type), intent(in) :: replacement
         type(string_type) :: res
 
-        res = string_type(replace_all_char_char_char(char(string), & 
+        res = string_type(replace_all(char(string), & 
                 & char(pattern), char(replacement)))
 
     end function replace_all_string_string_string
@@ -536,7 +536,7 @@ contains
         character(len=*), intent(in) :: replacement
         type(string_type) :: res
 
-        res = string_type(replace_all_char_char_char(char(string), char(pattern), replacement))
+        res = string_type(replace_all(char(string), char(pattern), replacement))
 
     end function replace_all_string_string_char
 
@@ -549,7 +549,7 @@ contains
         type(string_type), intent(in) :: replacement
         type(string_type) :: res
 
-        res = string_type(replace_all_char_char_char(char(string), pattern, char(replacement)))
+        res = string_type(replace_all(char(string), pattern, char(replacement)))
 
     end function replace_all_string_char_string
 
@@ -562,7 +562,7 @@ contains
         type(string_type), intent(in) :: replacement
         character(len=:), allocatable :: res
 
-        res = replace_all_char_char_char(string, char(pattern), char(replacement))
+        res = replace_all(string, char(pattern), char(replacement))
 
     end function replace_all_char_string_string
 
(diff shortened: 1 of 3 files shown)


Aman-Godara:
Although I have added `replace_overlapping` feature to the function but I would like to know the opinion of the community on what behaviour do they find more intuitive from the function in case of overlapping substrings `pattern`.

Aman-Godara:
This comment is regarding commit e325c2b.
replace_all(`string`, `pattern`, `replacement`, `replace_overlapping`)

Please note that for input: `replace_all("abcc", "abc", "ab")` 
commit e325c2b returns `"abc"` even if `replace_overlapping` is set to `.true.` because we are returning a new string as the output. Function is using the input string just to read from it and not to write in it.

***
If we would have designed `replace_all` function differently (i.e. by keeping `replace_all` as subroutine and making few other modifications in the algorithm), function could have been made to give `"ab"` as the output. Answer `"ab"` occurs as such:
**1st replacement**: `"abcc"` -> `"abc"`      (replaced `"abc"` with `"ab"`)
**2nd replacement**: `"abc"` -> `"ab"`        (replaced `"abc"` with `"ab"`)
**No further replacements.**

Both these replacements are happening in the one single call of `replace_all`. Here after **1st replacement**, `"ab"` fitted very well with `"c"` acting as the `pattern` for the **2nd replacement**.

Moreover, this behaviour cannot be achieved by running `replace_all` (as implement in commit e325c2b even with `replace_overlapping` as `.true.`) several times, as:
running `replace_all` 2 times on `"acaaca"` to replace **pattern**: `"aca"` with **replacement**: `"ac"` returns `"acc"` whereas correct output should be `"acca"` as such:
**1st replacement**: `"acaaca"` -> `"acaca"`
**2nd replacement**: `"acaca"` -> `"acca"`
**No further replacements.**

@arjenmarkus too has highlighted a similar behaviour through this sentence in [his comment](https://github.com/fortran-lang/stdlib/issues/366#issuecomment-810893962):
> If the replacing string is qwqwq you would get an infinitely long string.


Whole aim of the comment was to highlight the problems with subroutine kind of `replace_all` function and to bring forward what implementation e325c2b cannot do.

src/stdlib_strings.f90
```diff
@@ -499,5 +513,140 @@ pure function compute_lps(string) result(lps_array)
...
+                    p_i = p_i + 1
+                else if (p_i > 1) then
+                    p_i = lps_array(p_i - 1) + 1
+                else
+                    s_i = s_i + 1
+                end if
+            end do
+        end if
+        
+        res = res // slice(string, first=last)
```
Aman-Godara, line 648: line 648 is equivalent to: `res = res // string(last : length_string)`

src/stdlib_strings.f90
```diff
@@ -499,5 +513,140 @@ pure function compute_lps(string) result(lps_array)
...
+    !> Replaces all occurrences of substring 'pattern' in the input 'string'
+    !> with the replacement 'replacement'
+    !> Returns a new string
+    pure function replace_all_string_string_string(string, pattern, replacement) result(res)
+        type(string_type), intent(in) :: string
+        type(string_type), intent(in) :: pattern
+        type(string_type), intent(in) :: replacement
+        type(string_type) :: res
+
+        res = string_type(replace_all_char_char_char(char(string), &
```
Aman-Godara, line 525: Should I call `replace_all` instead of `replace_all_char_char_char` in line 525?

awvwgk: It's preferable to call the generic interface here

awvwgk (approved):
Looks good to me.

Aman-Godara:
On page 124 of https://j3-fortran.org/doc/year/18/18-007r1.pdf, It is mentioned that output of character slicing is of length `max(l - f + 1, 0)` which is exactly what is desired in here by the function `replace_all`. Thus, I am making use of intrinsic character slicing instead of `slice` in line 633 and 648. This will slightly improve the performance of the function.

image

awvwgk:
@fortran-lang/push-access Can anyone help out with reviewing this patch to allow @Aman-Godara to move forward with his GSoC project?

milancurcic (approved):
Thank you @Aman-Godara, I think this is a very nice addition.

I can't find the explanation in this thread or in the related issue: Why did you choose `pure` over `elemental`?

milancurcic:
> I can't find the explanation in this thread or in the related issue: Why did you choose pure over elemental?

All good, I see why, please disregard :+1:

Aman-Godara:
> I can't find the explanation in this thread or in the related issue: Why did you choose `pure` over `elemental`?

Let me still mention it for future reference,
if the length of `replacement` string is not equal to the length of `pattern` string, then different outputs might have varying length which become problematic to be stored in a single output array.

for example:
`replace_all (string, pattern, replacement)`

`replace_all(["abcabcT", "bcdabcT"], ["abc", "bcd"], ["X", "Y"])`
output array in this case will be: `["XXT", "YabcT"]` which cannot be stored in a single output array because of varying lengths of different outputs `"XXT"` and `"YabcT`.

PR #437, opened 2021-06-16
[FPM] add fpm support

zoziha:
Initial support `fpm`:
I don't know if this is feasible, but I want `stdlib` to support `fpm` as soon as possible so that the two can feed back to each other.
My method:
1. Use `make dev -f Makefile.manual` to generate `f90` files from the `src` folder to the `src/fpm` folder;
2. Use `fpm.toml` to build the `fpm` version of `stdlib.`      

I added a `src/fpm` folder and subordinate files, and added `stdlib`'s initial support for `fpm`. I don't know if this is very feasible, but it has passed the `fpm build`.


fpm.toml
```diff
@@ -0,0 +1,18 @@
...
+author = "stdlib contributors"
+maintainer = "https://github.com/fortran-lang/stdlib"
+copyright = "2019-2021 stdlib contributors"
+description = "Fortran Standard Library"
+categories = ["numerical"]
+keywords = ["numerical", "stdlib"]
+
+# cd stdlib && make dev -f Makefile.manual && fpm build
+[library]
+source-dir="src/fpm"
```
zoziha, line 13: attention: src/fpm

jvdp1:
Thank you @zoziha. Nice and highly needed proposition. After a quick look, I wonder if `src/fpm` should already contain the files (since they will be generated by the Makefile).
Question: why not generating files inside `src`, instead of `src/fpm`?

Note that @LKedward proposed another (but similar) [solution](https://github.com/LKedward/stdlib-fpm) for `fpm` supporting `stdlib`. However, from a high level, I would prefer yours (i.e. everything is contained in one single place).

zoziha:
> Thank you @zoziha. Nice and highly needed proposition. After a quick look, I wonder if `src/fpm` should already contain the files (since they will be generated by the Makefile).

Yes, `src/fpm` already contains the generated files, so that other people can directly compile these files through `fpm.toml` without the need for `make dev -f Makefile.manual`.

> Question: why not generating files inside `src`, instead of `src/fpm`?

Because I don’t really know whether it’s wise to modify the files directly under `src`. After all, the `src` folder files are **the source code files of meta-programming**, so I chose to generate a separate space just to facilitate `fpm` compilation. Of course, we can discuss how to do more reasonable.

> Note that @LKedward proposed another (but similar) [solution](https://github.com/LKedward/stdlib-fpm) for `fpm` supporting `stdlib`. However, from a high level, I would prefer yours (i.e. everything is contained in one single place).

I’m sorry I don’t know that [solution](https://github.com/LKedward/stdlib-fpm) also has an implementation of `stdlib-fpm`, maybe this is why I want to add fpm support directly in the `stdlib` library website, because in general, users prefer to get the `stdlib-fpm` package through the **official** stdlib library link.

LKedward:
Thanks @zoziha - yes as Jeremie pointed out, my solution is to generate the source files automatically in the CI. This is generally more robust than checking the preprocessed files into git. As a compromise, I'd like to implement the same workflow here to deploy the generated files to a separate branch. For the time being, you can use my stdlib fpm package linked above for testing the two together - I'm using it myself more and more as stdlib grows and it's a really quick way to spin up an stdlib project and test its functionality.

zoziha:
> To have everything in one place, I'd like to implement the same workflow here to deploy the generated files to a separate branch. 

Thanks @LKedward , I agree with you. I think `stdlib` can set up a `stdlib-fpm` branch, automatically generate source code files through `github-ci`, and declare the reference of this master branch in README.md, like:
```bash
[dependencies]
stdlib = { git = "https://github.com/fortran-lang/stdlib", branch = "stdlib-fpm" }
```

jvdp1:
> Thanks @LKedward , I agree with you. I think `stdlib` can set up a `stdlib-fpm` branch, automatically generate source code files through `github-ci`, and declare the reference of this master branch in README.md, like:

I likes this plan! 
Supporting `stdlib` through `fpm` is quite a recurrent topic, and IMO, highly important for the development and popularity of both `stdlib` and `fpm`.
@LKedward could `fpm` already support such an approach (e.g., with the branch option)? What would be needed to be done? Maybe @zoziha could help you in this?

LKedward:
Yes, fpm already allows you to specify a branch for git dependencies. I believe I can just copy the CI workflow from my stdlib-fpm repo with a few small changes - hopefully I can find some time soon to get it working for a PR.

ivan-pi:
I imagine in many cases, users will only need one or two modules from `stdlib` and not the entire library. Some users are facing problems with the compilation time (#438) in connection with the large number generic routines in `stdlib_stats`. Having `fpm` clone and build the entire `stdlib` is sub-optimal if the module is not used at all.

I am not sure what would be the best way to address this long term. Maybe add more control over stdlib directly to fpm?

zoziha:
### User efficiency
In order to improve the efficiency of users of `stdlib`, we can set the `stdlib-fpm` branch for `stdlib`, allowing users to add the following references in their projects:
```bash
[dependencies]
stdlib = { git = "https://github.com/fortran-lang/stdlib", branch = "stdlib-fpm" }
```

### Developer efficiency
However, the efficiency of developers has not been improved. `stdlib` now uses `make` and `cmake` tools to build code, which means that when we add functions to `stdlib`, we not only need to take care of the function implementation, but also need to maintain more `make` and `cmake` configurations files (In the development process, it is easy to cause `git` conflicts with other developers, and it is easy to forget or write the configuration files incorrectly, which increases the development cost.)  can we consider using `fpm` to empower developers?
Although the solution of the `src/fpm` folder is a clumsy way, it allows developers to use the easiest `fpm` to develop `stdlib` (In the future, `stdlib` will also need to support `fpm`). Developers only need to consider function implementation, function testing, `fpm.toml` and use a simple `makefile` to convert `fypp` code to `f90` code.
I don’t know if there will be any technical difficulties.

### Possible solution
#### For developers: 
Gradually use `fpm` instead of `make/cmake`, allow developers to use `makefile` in `stdlib` to convert `*.fypp` to `*.f90`, allow developers to use `fpm.toml` in `stdlib` to develop and test code, and **developers only submit meta-programming source codes to `stdlib-master`.**
Set the `stdlib-fpm` branch in `stdlib` for users, and convert `*.fypp` in `stdlib-master` into `*.f90` files through `github-ci`.
#### For users: 
Users can use the following references to use `stdlib`,
```bash
[dependencies]
stdlib = { git = "https://github.com/fortran-lang/stdlib", branch = "stdlib-fpm" }
```

If `fpm` supports `fypp` in the future, the `stdlib-fpm` branch can be retired.

zoziha:
Hello, I am very happy to announce that I have implemented `stdlib-fpm` using the @LKedward 's  solution(github-ci) under my branch:
1. [fortrans-fans/add_master_fpm_ci](https://github.com/fortran-fans/stdlib/tree/add_master_fpm_ci) + [fortran-fans/stdlib-fpm](https://github.com/fortran-fans/stdlib/tree/stdlib-fpm)
2. [fortran-fans/add_fpm_ci2](https://github.com/fortran-fans/stdlib/tree/add_fpm_ci2)(To PR) + fortran-lang/stdlib-fpm(needed, todo.)

That we can use `stdlib` as follows:
```fortran
[dependencies]
stdlib = { git = "https://github.com/fortran-lang/stdlib", branch = "stdlib-fpm" }
```
Before that, I need an administrator with permission for fortran-lang to set up an empty `stdlib-fpm` branch in `fortran-lang/stdlib` repo, so that I can submit a PR.

PR #439, opened 2021-06-20
Sorting4

wclodius2:
Changed `src/tests/sorting/test_sorting.f90` so that it should run in under one second, by reducing the length of the arrays being sorted, and setting the number of repeats to 1. Changed `doc/specs/stdlib_sorting.md` so its discussion of the output of `test_sorting.f90` reflected the changes in the code. Changed two comments in `src/stdlib_sorting.fypp` so they were less than 80 columns wide.

### DIFF
--- a/src/tests/sorting/test_sorting.f90
+++ b/src/tests/sorting/test_sorting.f90
@@ -253,7 +253,8 @@ contains
         call verify_reverse_sort( dummy, valid, i )
         ltest = (ltest .and. valid)
         if ( .not. valid ) then
-            write( *, * ) "reverse + work ORD_SORT did not sort " // a_name // "."
+            write( *, * ) "reverse + work ORD_SORT did not sort " // a_name // &
+                "."
             write(*,*) 'i = ', i
             write(*,'(a12, 2i7)') 'dummy(i-1:i) = ', dummy(i-1:i)
         end if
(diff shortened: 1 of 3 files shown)


wclodius2:
I have made the test code, `src/tests/sorting/test_sorting.f90`, run considerably faster by shortening  the arrays and setting repeat to 1. I would appreciate reviews by two reviewers with write access. At least three such reviewers have expressed interest in this code @milancurcic , @jvdp1 , @ivan-pi and also @awvwgk who may not have write access.

src/tests/sorting/test_sorting.f90
```diff
@@ -4,17 +4,17 @@ program test_sorting
...
+    use stdlib_string_type, only: string_type, assignment(=), operator(>), &
+        operator(<), write(formatted)
     use stdlib_error, only: check
 
     implicit none
 
-    integer(int32), parameter :: test_size = 2_int32**20
-    integer(int32), parameter :: char_size = 26**4
-    integer(int32), parameter :: string_size = 26**3
+    integer(int32), parameter :: test_size = 2_int32**16
```
jvdp1, line 13: For flexibility, could the `16` become a parameter?

jvdp1 (approved):
Thank you. LGTM. Following the CI, the test for `sort` takes now < 1 second. I just have a minor comment.

src/tests/sorting/test_sorting.f90
wclodius2: There are two kinds of things that now have a value of 16:
1. The number of elements in the integer array are now `2**16`.
2. The number of characters in the character set used to generate the strings.
Both could be parameters, but they should be separate parameters.

jvdp1: Thank you @wclodius2 . Then I think that at least one of the two values could be a parameter to clarify the code

wclodius2: I have made both values parameters, committed the change and pushed it to the PR.

jvdp1: Great. Thank you.

awvwgk (approved):
Looks good to me

Issue #394, opened 2021-04-23
Function to pad string with zeros


ivan-pi:
**Description**

A function to pad a string with zeros or another symbol to a given width could be helpful in many cases, especially together with the `to_string()` function for integer to string conversion.

Currently available methods to achieve this are using concatenation
```
width = 32
str = "hello"
padded_str = repeat('0',width-len(str)')//str
```
and perhaps also internal file I/O
```
write(buffer,'(A32)', blank='zero') "hello"   ! doesn't work for some reason
```

I propose to add the following functions
```
! left - padding
function padl(str,width,symbol) result(padded_str)
  character(len=*), intent(in) :: str
  integer, intent(in) :: width
  character(len=1), intent(in), optional :: symbol
  character(len=max(len(str),width) :: padded_str
end function
! right - padding
function padr(str,width,symbol) result(padded_str)
  character(len=*), intent(in) :: str
  integer, intent(in) :: width
  character(len=1), intent(in), optional :: symbol
  character(len=max(len(str),width) :: padded_str
end function
```

The original string is returned if `width < len(str)`. The default padding symbol could be either the blank symbol `' '` or `'0'`. Special treatment of a leading sign symbol `'+'`/`'-'` might be also desirable.

**Prior art**

* Python: [zfill](https://docs.python.org/3/library/stdtypes.html?highlight=zfill#str.zfill)
  (Another method to achieve padding in Python is using the string class method `.format()`)
* Julia: [lpad](https://docs.julialang.org/en/v1/base/strings/#Base.lpad), [rpad](https://docs.julialang.org/en/v1/base/strings/#Base.rpad)
* MATLAB: [pad](https://uk.mathworks.com/help/matlab/ref/pad.html)

milancurcic:
I use this often to left-pad integers with zeros when enumerating filenames, but using the edit descriptors in format.

Do you think a variant of `padl` and `padr` that accepts integers would be useful in addition to the ones you proposed?

```
res = padl(23, 4) ! returns "0023"
```

Compare it to:

```
res = padl(to_string(23), to_string(4)) ! returns "0023"
```

I like the proposed names. Rarely there's an opportunity to introduce such short but clear names.

ivan-pi:
> I use this often to left-pad integers with zeros when enumerating filenames, but using the edit descriptors in format.

That's my use case too. Currently I resort to the pattern:
```
integer :: istep,unit
character(len=32) :: step
write(step,'(I0.8)') istep
[...]

PR #441, opened 2021-06-22
implemented pad function

Aman-Godara:
#394 

Task:

- [x] Implemented pad function
- [x] add test cases
- [x] document the function
- [x] Elemental or Pure? -> Pure

### DIFF
--- a/src/stdlib_strings.f90
+++ b/src/stdlib_strings.f90
@@ -12,7 +12,7 @@ module stdlib_strings
 
     public :: strip, chomp
     public :: starts_with, ends_with
-    public :: slice, find, replace_all
+    public :: slice, find, replace_all, padl, padr
 
 
     !> Remove leading and trailing whitespace characters.
@@ -669,7 +669,7 @@ contains
 
     end function replace_all_char_char_char
 
-    !> Left pad the input string with the 'pad_with' string
+    !> Left pad the input string with " " (1 whitespace)
     !>
     !> Returns a new string
     pure function padl_string_default(string, output_length) result(res)
@@ -677,10 +677,11 @@ contains
         integer, intent(in) :: output_length
         type(string_type) :: res
 
-        res = string_type(padl_char_char(char(string), output_length, " "))
+        res = string_type(padl(char(string), output_length, " "))
+
     end function padl_string_default
 
-    !> Left pad the input string with the 'pad_with' string
+    !> Left pad the input string with the 'pad_with' character
     !>
     !> Returns a new string
     pure function padl_string_pad_with(string, output_length, pad_with) result(res)
@@ -689,10 +690,11 @@ contains
         character(len=1), intent(in) :: pad_with
         type(string_type) :: res
 
-        res = string_type(padl_char_char(char(string), output_length, pad_with))
+        res = string_type(padl(char(string), output_length, pad_with))
+
     end function padl_string_pad_with
 
-    !> Left pad the input string with the 'pad_with' string
+    !> Left pad the input string with " " (1 whitespace)
     !>
     !> Returns a new string
     pure function padl_char_default(string, output_length) result(res)
@@ -700,10 +702,11 @@ contains
         integer, intent(in) :: output_length
         character(len=max(len(string), output_length)) :: res
 
-        res = padl_char_char(string, output_length, " ")
+        res = padl(string, output_length, " ")
+
     end function padl_char_default
 
-    !> Left pad the input string with the 'pad_with' string
+    !> Left pad the input string with the 'pad_with' character
     !>
     !> Returns a new string
     pure function padl_char_pad_with(string, output_length, pad_with) result(res)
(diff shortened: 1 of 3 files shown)


Aman-Godara:
While I was writing code I got **ambiguous interface error** to which I preferred to keep `pad_with` argument as non-optional (to support backward compatibility) over other possible solutions.

ivan-pi:
Nice work @Aman-Godara!

Some comments:
* What happens when the `pad_with` dummy variable is of `type(string_type)` and the length is larger than 1? Wouldn't it be easier to limit `pad_with` to `character(len=1)` for now?
* In most cases I imagine padding will be performed with the whitespace character `' '`. For such cases it is inconvenient to have to specify `pad_with = ' '`.

What I propose is to have the following versions (and analogous for `padr`):

```
    interface padl
        module procedure :: padl_string_default      ! whitespace
        module procedure :: padl_string_pad_with     ! caller provided padding symbol
        module procedure :: padl_char_default
        module procedure :: padl_char_pad_with
    end interface padl
```

If we decide to allow `len(pad_with) > 1` in the future, I believe it remains possible to do it in a backward compatible way for both string_type and character. Even if we were to allow `len(pad_with) > 1` I still believe most usage will be with character literals like `pad_with = 'aabbcc'`. The case where you would want to pass a variable of `type(string_type)` seems unlikely to me, but can be always added later if someone requests it.

So I guess my bottom line is that having `pad_with` of `type(string_type)` is not worth the implementation/maintenance effort at this early stage.

Aman-Godara:
> What happens when the pad_with dummy variable is of type(string_type) and the length is larger than 1? Wouldn't it be easier to limit pad_with to character(len=1) for now?

Yes, `pad_with` is limited to character(len=1) as of now. If a user gives `pad_with` as `type(string_type)` with length > 1, function will only consider first character for padding.
Also, when user gives `""` (empty string) as `pad_with` padding is performed but with default `" "` (1 space).

I also liked the interface suggested by you let me make the changes.

***
Also, how about making the function **elemental**?
The only problem in doing so is that if a user gives an input like this:
`padl(["abcde", "bcd", "abc"], 6, "#")` output will be: 
`["#abcde", "#bcd  ", "#abc  "]` which actually looks as if the input got its right side padded with `" "` (1 space) and left with `"#"`.

Except this case in all other cases having an **elemental** `pad` function will be quite useful as length of outputs is any way going to be same for all the possible cases (including the one mentioned above) so there is no problem in returning the output.
OR
we can also keep `padl_char_pad_with` and `padl_char_default` as `pure` as of now, and make `padl_string_pad_with` and `padl_string_default` as **elemental** because they don't have any of these problems highlighted above and are quite useful if a user is interested in padding many strings at once.

Aman-Godara:
I didn't consider the case where a user gives an array in `output_length` argument as such:
`pad_with(["abc", "bcd"], [5, 6], '#')`
where returning the output `["abc##", "bcd###"]` in an array won't be possible.
Thus, I am keeping all functions as `pure` for now.

src/stdlib_strings.f90
```diff
@@ -644,38 +654,48 @@ end function padl_char_char
...
-    pure function padr_char_string(string, output_length, pad_with) result(res)
+    pure function padr_char_default(string, output_length) result(res)
         character(len=*), intent(in) :: string
         integer, intent(in) :: output_length
-        type(string_type), intent(in) :: pad_with
         character(len=max(len(string), output_length)) :: res
 
-        res = padr_char_char(string, output_length, char(pad_with))
-    end function padr_char_string
+        res = padr_char_char(string, output_length, " ")
```
ivan-pi, line 685: For the default it should be possible to do simply
```
res = string
```
getting rid of one function and branch (if) call.

But I would need to consult the Fortran standard to guarantee the padding symbol will be whitespace under all possible circumstances.

Aman-Godara: on page 162 of https://j3-fortran.org/doc/year/18/18-007r1.pdf I have got this:
image

ivan-pi: The second bullet point sounds like what we want. Thanks for finding and referencing this.

src/stdlib_strings.f90
```diff
@@ -79,6 +79,40 @@ module stdlib_strings
...
+        !module procedure :: replace_all_string_char_string
+        !module procedure :: replace_all_char_string_string
+        !module procedure :: replace_all_string_char_char
+        !module procedure :: replace_all_char_string_char
+        !module procedure :: replace_all_char_char_string
+        module procedure :: replace_all_char_char_char
+    end interface replace_all
+
+    !> Left pad the input string
+    !> [Specifications](link to the specs - to be completed)
```
awvwgk, line 97:
```suggestion
    !> [Specifications](../page/specs/stdlib_strings.html#padl)
```

src/stdlib_strings.f90
```diff
@@ -79,6 +79,40 @@ module stdlib_strings
...
+    !> Version: experimental
+    interface padl
+        module procedure :: padl_string_default
+        module procedure :: padl_string_pad_with
+        module procedure :: padl_char_default
+        module procedure :: padl_char_pad_with
+    end interface padl
+
+    !> Right pad the input string
+    !> [Specifications](link to the specs - to be completed)
```
awvwgk, line 107:
```suggestion
    !> [Specifications](../page/specs/stdlib_strings.html#padr)
```

awvwgk (approved):
Only two minor comments, otherwise looks good to me.

src/stdlib_strings.f90
```diff
@@ -649,4 +669,122 @@ pure function replace_all_char_char_char(string, pattern, replacement) result(re
...
+        character(len=1), intent(in) :: pad_with
+        character(len=max(len(string), output_length)) :: res
+        integer :: string_length
+
+        string_length = len(string)
+
+        res = string
+        if (string_length < output_length) then
+            res(string_length + 1 : output_length) = repeat(pad_with, &
+            & output_length - string_length)
```
milancurcic, line 784: I find the line break in the middle of function arguments list somewhat awkward, what do you think about breaking it at the assignment operator, i.e. LHS on one line and RHS on the other?

```suggestion
            res(string_length + 1 : output_length) = &
              repeat(pad_with, output_length - string_length)
```

I noticed that you use the continuation symbol `&` both at the end of the continuing line and at the beginning of the continued line. It's not necessary, you can have it only on the continuing line. But maybe you use both as a visual aid, which I think is fine.

milancurcic (approved):
Thank you @Aman-Godara, great addition, I left two comments.

awvwgk:
Sorry for the merge conflict, I probably merged the PRs in the wrong order.

Issue #435, opened 2021-06-13
Function to format variable as string


awvwgk:
We can already format integer and logical variables to strings, but we don't have routines to write floating point values yet. Maybe we can implement a simple solution by wrapping the internal IO in a function as a separate module. The size of the buffer might be problematic since it can overflow in case a format specifier like `'(10000x, g0)'` is passed.

The actual implementation is pretty straight-forward:

```f90
! SPDX-Identifier: MIT
#:include "common.fypp"
#:set KINDS_TYPES = REAL_KINDS_TYPES + INT_KINDS_TYPES + LOG_KINDS_TYPES

!> Routines for dealing with format specifiers
module stdlib_format
    use stdlib_kinds, only : sp, dp, qp, int8, int16, int32, int64, lk, c_bool
    implicit none
    private
    public :: format_string

    !> Create a character string representing the value of the provided variable.
    interface format_string
    #:for type, kind in KINDS_TYPES
        module procedure :: format_string_${type[0]}$${kind}$
    #:endfor
    end interface format_string

contains

    #:for type, kind in KINDS_TYPES
    !> Format ${type}$ variable as character sequence
    pure function format_string_${type[0]}$${kind}$(val, format) result(string)
        integer, parameter :: ik = ${kind}$
        ${type}$, intent(in) :: val
        character(len=*), intent(in) :: format
        character(len=:), allocatable :: string
        integer, parameter :: buffer_len = 512
        character(len=buffer_len) :: buffer

        write(buffer, format, iostat=stat) val
        if (stat == 0) then
            string = trim(buffer)
        else
            string = "*"
        end if
    end function format_string_${type[0]}${kind}$
    #:endfor

end module stdlib_format
```

zoziha:
Hello,  if my [num2str](https://github.com/zoziha/forlab/blob/master/src/forlab_num2str.fypp) can be combined with your code, especially the `complex` type `num2str`:
https://github.com/fortran-lang/stdlib/issues/337#issuecomment-867821014
```fortran
 forlab_num2str.fypp :
 ---------------------- 
 num2str(c):
 (1.00000000,1.00000000)
 (1.00,1.00)
 num2str(i):
 2
 2
 num2str(r):
 1.00000000
 1.00
 forlab_disp.fypp :
 ----------------------
 disp(ones(c)):
          (1.000,1.000)           (1.000,1.000)
          (1.000,1.000)           (1.000,1.000)
 disp(ones(i)):
          1           1
          1           1
[...]

PR #444, opened 2021-06-25
Add format_string routine to format other types to strings

St-Maxwell:
1. src/stdlib_strings_format_string.fypp
2. src/tests/string/test_strings_format_string.f90
3. stdlib_strings.fypp%format_string interface
modify:
1. doc/spec/stdlib_strings.md%format_string(doc)
2. stdlib_strings.f90 -> stdlib_strings.fypp(rename)
3. src/Makefile.manual%format_string(make)
4. src/tests/string/Makefile.manual%format_string(make)
5. src/CMakelists.txt%format_string(cmake)
6. src/CMakelists.txt%format_string(cmake)
note:
make test passed.
cmake test passed


src/stdlib_strings_format_string.fypp
```diff
@@ -0,0 +1,52 @@
...
+    integer, parameter :: buffer_len = 512
+
+contains
+
+    #:for kind, type in RIL_KINDS_TYPES
+    module procedure format_string_${type[0]}$${kind}$
+        !! Format ${type}$ variable as character sequence
+        character(len=buffer_len) :: buffer
+        integer :: stat
+        if(present(fmt)) then
```
St-Maxwell, line 15: maybe optval is preferred, i.e. `write(buffer, optval(fmt, default_fmt), iostat=stat) val`

doc/specs/stdlib_strings.md
```diff
@@ -328,3 +328,63 @@ program demo_find
...
+### `format_string`
+
+#### Description
+
+Format or transfer a integer/real/complex/logical variable as a character sequence.
+
+
+#### Syntax
+
+`format_string = [[stdlib_strings(module):format_string(interface)]] (value, [, format])`
```
awvwgk, line 342:
```suggestion
`format_string = [[stdlib_strings(module):format_string(interface)]] (value[, format])`
```

src/stdlib_strings_format_string.fypp
```diff
@@ -0,0 +1,52 @@
...
+            write(buffer, fmt, iostat=stat) val
+            if(stat == 0) then
+                string = trim(adjustl(buffer))
```
awvwgk, line 18: I would not left adjust the buffer but only trim trailing whitespace from the buffer:

For a calls like this I would expect that the width of the format is honored:

```
print '(a)', format_string(1, '(i0)'), format_string(1, '(i5)'), format_string(-1, '(i5)')
```

Resulting in an output like:
```
1
    1
   -1
```
While the current behaviour is more like
```
1
1
-1
```

src/stdlib_strings_format_string.fypp
```diff
@@ -0,0 +1,52 @@
...
+            else
+                string = '*'
+                    !!\TODO: *?
```
awvwgk, line 21: For a meaningful number of asterisks in the output we must be able to parse format strings and determine the width of the field we are writing into. Just writing a single asterisk until we are able to do so in stdlib sounds like a good compromise for now. What do you think?

src/stdlib_strings_format_string.fypp
awvwgk: `optval` might work if we pass `(g0)` as `default_fmt`, which should be pretty much what we get from list-directed IO.

src/stdlib_strings_format_string.fypp
zoziha: It is indeed better to remove `adjustl`, thank you~

awvwgk:
I added a suggestion as PR against your (see https://github.com/St-Maxwell/stdlib/pull/1) for expanding the unit testing and providing better error reporting in case of a mismatch in the formatted strings.

awvwgk:
Looks like our choice of default formatter with `(g0)` is not well defined across compilers. This is a bit surprising to me to be honest, but maybe someone more well-versed in the standard can shed some light on this issue.

zoziha:
# Links
1. [Intel '(g0)'](https://software.intel.com/content/www/us/en/develop/documentation/fortran-compiler-oneapi-dev-guide-and-reference/top/language-reference/i-o-formatting/data-edit-descriptors/real-and-complex-editing/g-editing.html)
   > If w is 0, the field width is selected by the processor. If w is zero, you can only specify forms G0 or G0.d.
2. fortran18 standard, page270
image

---
It seems that the standard says that the width is determined by the **processor**, but it is actually determined by the compiler?

awvwgk:
> It seems that the standard says that the width is determined by the processor

Processor is standard-speak for compiler.

src/tests/string/test_strings_format_string.f90
```diff
@@ -0,0 +1,82 @@
+program test_strings_format_string
+    use stdlib_strings, only: format_string, starts_with
+    use stdlib_error, only: check
+    use stdlib_optval, only: optval
+    implicit none
+    print *, 'format_string(complex) : '
+    call check_formatter(format_string((1, 1)), "(1.00000000,1.00000000)", &
+        & "Default formatter for complex number")
```
awvwgk, line 8:
```suggestion
    call check_formatter(format_string((1, 1)), "(1.0", &
        & "Default formatter for complex number", partial=.true.)
```

src/tests/string/test_strings_format_string.f90
```diff
@@ -0,0 +1,82 @@
...
+        & "Multiple formatters for integers")
+    call check_formatter(format_string(34, '(B8)'), "  100010", &
+        & "Binary formatter for integer number")
+    call check_formatter(format_string(34, '(O0.3)'), "042", &
+        & "Octal formatter with zero padding for integer number")
+    call check_formatter(format_string(34, '(Z3)'), " 22", &
+        & "Hexadecimal formatter for integer number")
+    print *, 'format_string(real) : '
+    call check_formatter(format_string(100.), "100.000000", &
+        & "Default formatter for real number")
```
awvwgk, line 36:
```suggestion
    call check_formatter(format_string(100.), "100.0", &
        & "Default formatter for real number", partial=.true.)
```

src/tests/string/test_strings_format_string.f90
```diff
@@ -0,0 +1,82 @@
...
+        & "Exponential formatter with rounding for real number")
+    call check_formatter(format_string(128., '(ES8.2)'), "1.28E+02", &
+        & "Exponential formatter for real number")
+    ! Wrong demonstration
+    call check_formatter(format_string(-100., '(F6.2)'), "*", &
+        & "Too narrow formatter for signed real number", partial=.true.)
+    call check_formatter(format_string(1000., '(F6.3)'), "*", &
+        & "Too narrow formatter for real number", partial=.true.)
+    call check_formatter(format_string(1000, '(F7.3)'), "*", &
+        & "Real formatter for integer number", partial=.true.)
```
awvwgk, line 49: `ifort` seems to implicitly cast the integer into real before printing, which is smart but not consistent with `gfortran`:

```
 Real formatter for integer number
Expected: '*' but got '  0.000'
```

I'm not sure how to make a test that consistently fails here, maybe by passing an invalid format descriptor like here:

```suggestion
    call check_formatter(format_string(1000., '(7.3)'), "*", &
        & "Invalid formatter for real number", partial=.true.)
```

zoziha: Sorry i just saw this comment.

St-Maxwell:
🤔I'm wondering if there are some compiler options of Intel compiler that can prevent applying edit descriptor on mismatching type of values. And if exist, we can add them into `CMakeLists.txt`.

awvwgk:
I opened #446 to discuss handling smart behaviour in compilers. I think we want consistent results which are independent of the compiler, but we also don't want to set too many special flags when compiling, because some of them (like `-standard-semantics` for `ifort`) can cause problems when trying to reuse `stdlib` in other projects.

Eventually, not using the internal IO but have our own implementation of format descriptors would be the preferred solution, as it makes as most independent of the actual details of compiler extensions.

St-Maxwell:
> Eventually, not using the internal IO but have our own implementation of format descriptors would be the preferred solution, as it makes as most independent of the actual details of compiler extensions.

I support implementing an procedure converting floating-point numbers to strings in stdlib in the future. I found [Grisu](https://www.cs.tufts.edu/~nr/cs257/archive/florian-loitsch/printf.pdf) or [Ryū](https://dl.acm.org/doi/pdf/10.1145/3192366.3192369) algorithm might be the choice. Unfortunately I'm not able to implement it. This needs the help from the community.

src/tests/string/test_strings_format_string.f90
```diff
@@ -0,0 +1,82 @@
...
+    call check_formatter(format_string(.true.), "T", &
+        & "Default formatter for logcal value")
+    call check_formatter(format_string(.true., '(L2)'), " T", &
+        & "Formatter for logical value")
+    call check_formatter(format_string(.false., '(L2)')//format_string(.true., '(L5)'), &
+        & " F    T", &
+        & "Multiple formatters for logical values")
+    ! Wrong demonstration
+    call check_formatter(format_string(.false., '(I5)'), "*", &
+        & "Integer formatter for logical value", partial=.true.)
```
awvwgk, line 60: Let's try some easy failing case instead for now:
```suggestion
    call check_formatter(format_string(.false., '(1x)'), "*", &
        & "Invalid formatter for logical value", partial=.true.)
```

Just tried using a floating point descriptor with `ifort`, this would yield an `NaN` as result.

doc/specs/stdlib_strings.md
```diff
@@ -328,3 +328,175 @@ program demo_find
...
+  This argument is intent(in) and optional.
+
+#### Result value
+
+The result is a allocatable length Character scalar.
+
+#### Example
+
+```fortran
+program test_strings_format_string
```
awvwgk, line 366: I liked the previous example more as it actually demonstrates how the `format_string` function would be used in practice.

zoziha: I went back to the previous example, and made more detailed comments to make it easier to read.

src/tests/string/test_strings_format_string.f90
zoziha:
```fortran
read(str, '(7.3)', iostat=stat) 1000
```
The above statement is on `gfortran`, will product `stat\=0`, but on `ifort`, will product `stat=0`, and converts `1000` to `    0`(see [here](https://github.com/fortran-lang/stdlib/runs/2920528642)). This situation also occurs in the following conditions.
image

I don't think this is the flexibility or smart of `ifort`(see https://github.com/fortran-lang/stdlib/issues/446). In these cases, `ifort` should have reported errors to users, but it has been transformed, and the transformed results seem to lack practical significance.

awvwgk (approved):
Looks good to me, thanks a lot.

Aman-Godara:
@St-Maxwell raised a [PR](https://github.com/St-Maxwell/stdlib/pull/2) concerning aesthetics of the code. Please have a look into it.

zoziha:
@Aman-Godara Thank you very much for your work! I think it is good. My English is not very good and I am personally careless. There are indeed some problems in the writing. Thank you for your corrections.😘
I was influenced by the `ford` document, so I used the way of writing comments recommended by `ford`. (See [ford wiki](https://github.com/Fortran-FOSS-Programmers/ford/wiki/Writing-Documentation))

src/stdlib_strings_format_string.fypp
zoziha: image
I think the `*` sign here should be distinguished a little bit more, for example, written as `[*]`, `!*!` and the like can be distinguished from the `*****` sign (too narrow formatter for variables) **before and after it**.

src/Makefile.manual
```diff
@@ -35,8 +37,7 @@ SRC = f18estop.f90 \
       stdlib_io.f90 \
       stdlib_kinds.f90 \
       stdlib_logger.f90 \
-      stdlib_quadrature_gauss.f90 \
-      stdlib_strings.f90 \
+      stdlib_quadrature_gauss.f90
```
awvwgk, line 40:
```suggestion
      stdlib_quadrature_gauss.f90 \
```

src/stdlib_strings_format_string.fypp
awvwgk: Checking a failed formatter should be simple, currently I'm using `starts_with(format_string(var, "(...)"), "*")`. This is not fail-safe as it would match a formatter like `'("*", g0)'` as well. Using `format_string` should imply that we trade some of the versatility of the internal IO approach for having a function.

How about returning an empty string on failure? This would be clearly distinct from a too narrow formatter?

milancurcic:
@St-Maxwell @zoziha @awvwgk thank you for working on this. I took a look at the implementation and it looks good to me. 

I have a concern about the name `format_string`. To me, this sounds like it formats an input string into something. But it actually formats other stuff into a string. So a more accurate (and less confusing IMO) name would be `format_to_string`.

However, this being a wrapper around `write(string, fmt='...')`, I suggest calling it `format` just like the statement. We have a precedent for this in `stdlib_io` [`open`](https://stdlib.fortran-lang.org/page/specs/stdlib_io.html#open-open-a-file) function, and we envision having similar `read` and `write` wrappers that will provide similar functionality to existing statements but with a simpler API. 

@certik can you chime in on this considering that we discussed this a lot at the beginning of stdlib?

src/stdlib_strings_format_string.fypp
zoziha: It is mainly to remind **the users**: an empty string may not be effectively fed back to the user, this is a `format_string` failure.

1. We need to notify the user **more clearly that this has failed**;
2. The string output after failure is **simple enough**;
3. **Does not hinder the effective operation** of the main process of our `Fortran` program.

milancurcic:
I think the name `format_to_string` is too long for a function that's likely to be used often. What do you think about `format`? Or `to_string`?

awvwgk:
I'd be fine with `format`. Using `to_string` would make it mutually exclusive with the existing `to_string` routines due to the optional `format` argument. Unless we make the `format` argument mandatory we can't have `to_string` as name for this procedure.

zoziha:
Call it `num2str` directly, and don't care about the `optional` `format` when naming it? It was called `num2str` in [forlab/num2str](https://github.com/keurfonluu/Forlab/blob/62b80c031fe5f283e6696be0353c13f4ba5886de/src/lib/forlab.f90#L6329). (also see https://github.com/fortran-lang/stdlib/issues/435#issuecomment-868309435)

milancurcic:
@awvwgk Ah, yes, I forgot about the `to_string` in the `ascii` module. Let's not call it that then.

urbanjost:
I have found that it is useful for such a function to also take CHARACTER types and so consider trailing space significant and for the leading and trailing parenthesis in the format to be optional.  It is not a "perfect" solution, but I insert ",a" before the closing parenthesis and then use INDEX() and search for a null character from the right end of the buffer to still return an allocatable string but leave trailing whitespace. All the writes then have the argument and achar(0) as values to print. Food for thought, but is it useful for creating strings of specific lengths for lining up output if formats like "g0,t20" are used, and although you could do it with concatenation, things like  
  write(*,*)FORMAT('hello',brackets)
is nice where
  character(len=*),parameter :: brackets='("[",a,"]")'
and so on as a case for support of CHARACTER.  As far as the name goes, I use FMT(), but used to use FORMAT() but short is nice in this case, as it can often be called multiple times in a single statement.

zoziha:
> write(_,_)FORMAT('hello',brackets)
> is nice where
> character(len=*),parameter :: brackets='("[",a,"]")'
> and so on as a case for support of CHARACTER. As far as the name goes, I use FMT(), but used to use FORMAT() but short is nice in this case, as it can often be called multiple times in a single statement.

Do you mean to let `format_string` support the formatting of character? I think this is a good proposal. There are indeed needs in this area. Its function looks like this process:
```fortran
write(string_after, format) string_before
=>
string_after = format_string( string_before, format )
```

jvdp1:
@St-Maxwell Should we review this PR before #445? If yes, could the conflicts be resolved, please?

zoziha:
E.g:
```
string = format_to_string((1000, 1),'(SP,F6.3)')
```
Is the name `format_to_string` too long, do we need to think about it again?

milancurcic:
Yes, I think `format_to_string` is too long. Are you opposed to `format`?

zoziha:
I think the main contradiction of the difficulty of naming is that there is already a function `to_string`, and `to_string` and `format_string` overlap to a certain extent :

`to_string` is suitable for **unformatted** `integer` and `logical` variables.  
`format_string` is suitable for **unformatted and formatted** variables except for `character` type and `string_type` type variables.

I tried to simply test the time performance of the two converted to strings, see [test_to_string.f90](https://github.com/St-Maxwell/stdlib/blob/7a1d7054745545a443d389f0a959bde03acf501f/src/tests/string/test_to_string.f90#L77~L85).
```fortran
!! Test for array of dimension 10000
Item                                                       Time /ms
`to_string` time for integer `i=1` :                       1.1416999999999999
`format_to_string` time for integer `i=1` :                63.093699999999998
`to_string` time for integer `i=huge(i(1))` :              1.5802000000000000
`format_to_string` time for integer `i=huge(i(1))` :       60.947099999999999
`to_string` time for logical `l=.true.` :                  0.92299999999999993E-1
`format_to_string` time for logical `l=.true.` :           59.209699999999998
```

I think it may be feasible to merge `to_string` and `format_string` and put them in `stdlib_string`, and the name is still taken as **`to_string`**.
Do you have any other good opinions?

zoziha:
After thinking about it again, the last idea is a bit unrealistic, so it is better to separate them.
#### How about `to_chars(value, format)`? 
1. It is more similar to `to_string` in semantics, but `to_string` belongs to the `stdlib_ascii` module, and `to_chars` will belongs to the `stdlib_string` module. The main functions of the two modules have a certain distinction. 
2. The name of `to_chars` is **shorter**, and it is more suitable for the following style with the optional `format` argument :
```fortran
string = format_to_string((1000, 1),'(SP,F6.3)')
string = to_chars((1000, 1),'(SP,F6.3)')         # `to_chars` is 8 characters less than `format_to_string`.
```
```fortran
1. `string` - 6 characters;
2. `num2str` - 7 characters;
3. `to_chars` - 8 characters; 
4. `to_string` - 9 characters; 
5. `format_string` - 13 characters;
6. `format_to_string` - 16 characters;
```

milancurcic:
@zoziha did you see [my comment](https://github.com/fortran-lang/stdlib/pull/444#issuecomment-884296293)?

> Yes, I think format_to_string is too long. Are you opposed to format?

zoziha:
@milancurcic I think `format` is OK, even though I think the name of `format` is a bit broad and special.

The name `format` expresses the meaning of `format` in the name `format_to_string`, but fails to express the meaning of `to_string`, and the `format` arguments in `format_to_string` is `optional`.
But I also think it is ok, `stdlib_string(module):format(function)` and `stdlib_ascii(module):to_string(function)` have a strong distinction, users will not mistake them, and `format` indicates the intention of format conversion.

milancurcic:
Let's get some more preferences on the naming; @certik, @LKedward @nshaffer @ivan-pi @jvdp1 which names do you like best?

awvwgk:
`format` would be an option for me. `to_string` would also work but requires to drop the optional format argument from the interface.

jvdp1:
My preference would go for `to_chars`, or`num2char`. Because the result is a `character` scalar, I don't really like a name which containts `string` (the user could think that the result is a DT `string`.
I think the name `format` is not self explainable, but I prefer it over something with `string`.

nshaffer:
I definitely prefer `format` over `to_*`. In my own codes, I call it `str`, but that is too terse for stdlib. If we go with `format`, I suggest renaming the format-string argument either to `fmt` or `as` (as in: "format X as Y"). 

Reading the source, I have some comments on the buffer. 

First, Is a buffer size of 512 necessary? For formatting scalars, I think we can use much less. What's the widest reasonable value someone might ask to format? Maybe a complex number with quad-precision real and imaginary components? IIRC, the IEEE quad format has up to 36 significant digits in the fractional part, and up to 4 digits in the exponent. So each part of a complex number would need at most 45 characters, if I'm counting right (1 for sign, 1 for first digit, 1 for the dot, 36 for the fraction, 1 to denote the start of the exponent (e.g., 'E', 'D', or 'Q'), 1 for exponent's sign, 4 for exponent). Plus 3 for the parens and comma. That's 93 characters. So I think we can drop the buffer size down to 256, or maybe even 128. 

2. Trimming the buffer means that trailing whitespace will not be respected. That is, the result of `format_to_string(1.0, "(g0,5x)")` will not include the five spaces at the end, which the user presumably wanted to have. For now, I think it is good enough to mention this behavior in the documentation. We can consider fancier approaches later, if we decide to start parsing the input format string.

Besides those technical points, I think this PR is good.

awvwgk:
>     1. Is a buffer size of 512 necessary? For formatting scalars, I think we can use much less. What's the widest reasonable value someone might ask to format? Maybe a complex number with quad-precision real and imaginary components? IIRC, the IEEE quad format has up to 36 significant digits in the fractional part, and up to 4 digits in the exponent. So each part of a complex number would need at most 45 characters, if I'm counting right (1 for sign, 1 for first digit, 1 for the dot, 36 for the fraction, 1 to denote the start of the exponent (e.g., 'E', 'D', or 'Q'), 1 for exponent's sign, 4 for exponent). Plus 3 for the parens and comma. That's 93 characters. So I think we can drop the buffer size down to 256, or maybe even 128.

A format like `(1000x, g0)` will still overflow with the current buffersize, what we need in the long run is a format parser which can tell us exactly how long the buffer string at least has to be.

>     2\. Trimming the buffer means that trailing whitespace will not be respected. That is, the result of `format_to_string(1.0, "(g0,5x)")` will not include the five spaces at the end, which the user presumably wanted to have. For now, I think it is good enough to mention this behavior in the documentation. We can consider fancier approaches later, if we decide to start parsing the input format string.

This is actually the correct behavior since I would consider the formatter routines as advancing IO, if you write with `print '(g0, 5x)', 0.0` to stdout there will be no trailing spaces as well, since trailing skip formatters are only active in non-advancing IO. Also, we will always exhaust only the part of the formatter to write one element in this context.

ivan-pi:
> 
> 
> Let's get some more preferences on the naming; @certik, @LKedward @nshaffer @ivan-pi @jvdp1 which names do you like best?

How about `print` like in the [proposals for WG5 from Espen Myklebust](https://wg5-fortran.org/N1951-N2000/N1972.pdf) (see Section 7.1.3)?

```text
function print(expr,fmt)
  <numeric-type>, intent(in) :: expr
  character(len=*), intent(in), optional :: fmt
  character(len=:), allocatable :: print
end function
```

ivan-pi:
@zoziha, concerning the following point

>     1. It is more similar to `to_string` in semantics, but `to_string` belongs to the `stdlib_ascii` module, and `to_chars` will belongs to the `stdlib_string` module. The main functions of the two modules have a certain distinction.

see https://github.com/fortran-lang/stdlib/issues/399. IMO, also `to_string` should be moved from `stdlib_ascii` to `stdlib_string`.

St-Maxwell:
Sorry for my absence in recent discussion because I was busy in my project. I like `format`, and `to_chars` is also OK for me.

zoziha:
> So I think we can drop the buffer size down to 256, or maybe even 128.  

Yes, 512 is a bit long, **we can decide a certain length, it can be 128.** **And it is stated in the docs (stdlib_strings.md).**
> Trimming the buffer means that trailing whitespace will not be respected.  

If I try to cancel the `trim`, more complicated things will happen, or we have to spend time trying to write routines to parse the `"(g0, 5x)"` format, which will become more complicated. At this stage, **it is better for us to annotate it in the document**. "We can consider fancier approaches later."
> A format like (1000x, g0) will still overflow with the current buffersize, what we need in the long run is a format parser which can tell us exactly how long the buffer string at least has to be.

If `format_to_string` is renamed to `format`, then it should be more like a formatter, and you need to consider `"(1000X, g0)"`;
```
1. Data format edit descriptor: I, B, O, Z, F, E, EN, ES, D, G, L, A
2. Control format edit descriptor: T, TL, TR, X, S, SP, SS,:, /
3. Character string format edit descriptor: H, ``, ""
```
If it is renamed `num2str`, its main function is to convert other variables into characters, mainly involving **data format edit descriptors**, and `SP` is not necessary. 

> IMO, also to_string should be moved from stdlib_ascii to stdlib_string.

> I don't really like a name which containts string (the user could think that the result is a DT string.

If `stdlib_ascii(module):to_string(function)` is to be migrated to `stdlib_strings(module)` in the future, I would prefer to name `format_to_string` as **`num2str`, `to_chars` or `format`.**

zoziha:
Because of the API name, we seem to be stuck in the rhythm, why not decisively, directly `num2str`, taken from the `Matlab` API name `num2str`.

awvwgk:
We rejected `to_char` in the original discussion about the naming of `to_string`, because we are dealing with strings rather than single characters here. Similar considerations are relevant for `num2char`. An issue with `num2str` is that it can format a logical value, which is not strictly a number. Therefore, I think while short this name can be misleading about the provided functionality.

I'm still fine with naming it `format_string`, also `format` or `print` work fine with me. We can also overload the existing interface `to_string` and make the format specified mandatory. Those suggestions have been made, now it is up to the OP to make a decision and adopt the patch accordingly.

Whatever name we arrive with in the end, I'm still in favor of moving this patch forward.

zoziha:
Thanks, I canceled the idea (see [comment](https://github.com/fortran-lang/stdlib/pull/444#issuecomment-884595978)) merge `to_string` to `format_to_string` because I think this task is too large and complicated.
I heard that you also intend to migrate `to_string` to `stdlib_strings`. I think it is the best way to merge `to_string` and `format_to_string` into `stdlib_strings`. Their combined name can be called `to_string`. 

I completed this task in another branch (see [implementation](https://github.com/St-Maxwell/stdlib/blob/merge_tostring/src/stdlib_string_to_string.fypp)). It's true that the task is a bit heavy.
If you think there is no problem with this `merge` idea, I will merge that branch ([merge_tostring](https://github.com/St-Maxwell/stdlib/tree/merge_tostring)) into this PR tomorrow.

zoziha:
I merged `to_string` and `format_to_string` to `stdlib_string(module):to_string(interface)`, which looks like `GitHub-CI` has no error report, nice~
Please review it, thanks @awvwgk @St-Maxwell 😘.

src/tests/string/test_string_to_string.f90
```diff
@@ -0,0 +1,112 @@
...
+
+contains
+
+    subroutine check_formatter(actual, expected, description, partial)
+        character(len=*), intent(in) :: actual, expected, description
+        logical, intent(in), optional :: partial
+        logical :: stat
+        character(len=:), allocatable :: msg
+
+        if (merge(partial, .false., present(partial))) then
```
zoziha, line 16: I will replace `merge` with `optval` later.

milancurcic:
In a nutshell, these are now specific functions that go in addition to the previously existing `to_string` for integers, and they're now all under one generic name `to_string`. That sounds good to me. I built and ran tests with CMake and manual Makefile locally without issues.

milancurcic (approved):

milancurcic:
If there are no objections, I'll merge this tomorrow.

milancurcic:
Thanks a lot, @zoziha!

src/stdlib_strings_to_string.fypp
```diff
@@ -0,0 +1,123 @@
...
+    !> Format or transfer a ${t1}$ scalar as a string.
+    pure module function to_string_${t1[0]}$_${k1}$(value, format) result(string)
+        ${t1}$, intent(in) :: value
+        character(len=*), intent(in), optional :: format
+        character(len=:), allocatable :: string
+        
+        character(len=buffer_len) :: buffer
+        integer :: stat
+
+        write(buffer, optval(format, "(g0)"), iostat=stat) value
```
milancurcic, line 20: @St-Maxwell @zoziha @awvwgk 

I apologize for the last minute suggestion. What do you think about the following:

```suggestion
        write(buffer, "(" // optval(format, "g0") // ")", iostat=stat) value
```

Whether the format dummy is present or not, this encloses the format string in an extra set of parentheses. This makes the function slightly more flexible. For example, you can now do:

```
print *, to_string(1.2345, 'f4.2') ! prints '1.23'
```

which would previously not work. I don't know why Fortran requires parentheses around format strings, but it's always been one of those little annoying things to me. And if you already have an outer set of parentheses, the semantics don't change. I think this allows for a better UI.

awvwgk: I have to admit, a format string without parenthesis looks slightly off to me. But I have no strong feelings about this either way.

The only issue I could see, is that it might be surprising for a new user if the intrinsic transfer routines require a different syntax than the standard library. But maybe this is a good thing as well and could be used to suggest a simplification of format strings in the next Fortran standard.

milancurcic: Yes, if we decide to do this we should document it clearly.

zoziha: Without parentheses, it looks more concise. I think it's good, as long as we explain it in the document: We use parentheses by default inside `to_string` to help users write less.
I will confirm the feasibility of this writing later.

doc/specs/stdlib_strings.md
```diff
@@ -538,3 +538,69 @@ program demo_count
...
+
+#### Class
+
+Pure function.
+
+#### Argument
+
+- `value`: Shall be an `integer/real/complex/logical` scalar.
+  This is an `intent(in)` argument.
+- `format`: Shall be a `character(len=*)` scalar like `'(F6.2)'`.
```
milancurcic, line 566: @zoziha @awvwgk Suggestion for docs improvement, to go together with https://github.com/fortran-lang/stdlib/pull/444#discussion_r693108150.

```suggestion
- `format`: Shall be a `character(len=*)` scalar that contains the edit descriptor to format `value` into a string,
  for example `'(F6.2)'` or `'(f6.2)'`. `to_string` will automatically enclose `format` in a set of parentheses,
  so passing `F6.2` or `f6.2` as `format` is possible as well.
```

src/stdlib_strings_to_string.fypp
milancurcic: I just made a suggestion in https://github.com/fortran-lang/stdlib/pull/444#discussion_r693386425.

zoziha:
I tried to change the corresponding content of `format` argument to support `f6.2` format that **does not require parentheses**. 
Yes, it is also compatible with format like `(f6.2)`!
This is user-friendly and a practical change. @milancurcic Thanks a lot!

milancurcic:
Indeed, Fortran allows formats like this: `(((((f6.2)))))` :)

PR #445, opened 2021-06-25
[feature] `disp`(display your data) & `format_string`(format other type to string)

zoziha:
This branch develops subroutines `disp` and `format_string` for `stdlib` based on the `disp` and `num2str` functions in the [forlab](https://github.com/zoziha/forlab) library.
### `disp` subroutine
Quickly display strings, scalars and low-dimensional arrays to the screen (or the default output location).
### `format_string` function
Format or transfer a integer/real/complex/logical variable as a character sequence.
### Links
1. https://github.com/fortran-lang/stdlib/issues/435  🔰
2. [keurfonluu/Forlab](https://github.com/keurfonluu/Forlab)
3. [zoziha/forlab](https://github.com/zoziha/forlab)
4. https://github.com/j3-fortran/fortran_proposals/issues/212
#### 1. src/stdlib_string_format_string.fypp
❓[print *, '*'](https://github.com/St-Maxwell/stdlib/blob/fec8c688994d92e80641f8abc213e867bb6b4df8/src/stdlib_strings_format_string.fypp#L21)
When an error occurs, an asterisk is output. But the fortran language standard is to output multiple asterisks. How do we deal with this problem?
#### 2. src/stdlib_disp.fypp
❓[format disp](https://github.com/St-Maxwell/stdlib/blob/zoziha/feature/disp/doc/specs/stdlib_io.md#example-3)
The format looks a bit messy, should we align all the data to the left? Or keep `Fortran`'s default output style (like `print *,`)?
#### 3. format of `complex` type
❓https://github.com/j3-fortran/fortran_proposals/issues/212
Which style is better? Do we need to make its output style for `complex` numbers, just like above proposal. Now `disp` outputs the style `(1.00, 1.00)` by default.

### Effects
```fortran
 disp(string):
 string
 It is a note.

 disp(r):
 0.1000E-10   1.000       1.000
  1.000       1.000       1.000
 disp(c):
          (1.000,0.000)           (1.000,0.000)           (1.000,0.000)
          (1.000,0.000)      (0.1000E+07,0.000)           (1.000,0.000)
 disp(i):
          1           1           1
          1           1           1
 disp(l):
          T           T           T
          T           T           T
 disp(c_3d, 3):
 Slice (:,:,1):
          (2.000,0.000)           (2.000,0.000)      (1000.,0.1000E-02)
          (2.000,0.000)           (2.000,0.000)           (2.000,0.000)
 Slice (:,:,2):
          (2.000,0.000)           (2.000,0.000)           (2.000,0.000)
          (2.000,0.000)           (2.000,0.000)           (2.000,0.000)
 disp(c_3d, 2):
 Slice (:,1,:):
          (2.000,0.000)           (2.000,0.000)
          (2.000,0.000)           (2.000,0.000)
 Slice (:,2,:):
          (2.000,0.000)           (2.000,0.000)
          (2.000,0.000)           (2.000,0.000)
 Slice (:,3,:):
     (1000.,0.1000E-02)           (2.000,0.000)
          (2.000,0.000)           (2.000,0.000)
```


src/stdlib_io.fypp
```diff
@@ -7,18 +7,45 @@ module stdlib_io
...
   private
   ! Public API
-  public :: loadtxt, savetxt, open
+  public :: loadtxt, savetxt, open, disp
 
   ! Private API that is exposed so that we can test it in tests
   public :: parse_mode
 
+  interface disp
+    !!
```
zoziha, line 23: PS. I forgot to write a interface comment here, I will continue it next time

awvwgk:
Thanks @zoziha, @St-Maxwell for opening those pull requests. I think it is fine to have two PRs here, we will first discuss on #444 about the string formatting routines and once we agree on the API we will continue with the review of the display function. How does the plan sound to you?

src/stdlib_io_disp.fypp
```diff
@@ -0,0 +1,125 @@
...
+
+    module procedure disp_3_${type[0]}$${kind}$
+        integer :: i, dim1, dim2, dim3
+        dim1 = size(val, 1)
+        dim2 = size(val, 2)
+        dim3 = size(val, 3)
+        if(present(string)) print *, trim(string)
+        if(dim == 1) then
+            do i = 1, dim1
+                print *, 'Slice ('//format_string(i)//',:,:):'
```
awvwgk, line 44: You can use `to_string` from `stdlib_ascii` here, it usually faster than internal IO and yields the same result

src/stdlib_io_disp.fypp
```diff
@@ -0,0 +1,125 @@
...
+    implicit none
+    character(len=*), parameter :: fmt_r = '(*(g11.4, 1x))'
+    character(len=*), parameter :: fmt_c = '(*(g23.4, 1x))'
+    
+contains
+    
+    #:for kind, type in RIL_KINDS_TYPES
+    module procedure disp_0_${type[0]}$${kind}$
+        if(present(string)) print *, trim(string)
+        print fmt_r, val
```
awvwgk, line 16: What about writing to a formatted unit using `write` instead of `print`? The unit could be passed optionally and defaults to `output_unit`.

src/stdlib_io.fypp
```diff
@@ -7,18 +7,45 @@ module stdlib_io
...
+    #:set DISP_KINDS_TYPES = REAL_KINDS_TYPES + INT_KINDS_TYPES &
+        & + CMPLX_KINDS_TYPES + LOG_KINDS_TYPES
+    #:set DISP_RANKS = range(0, 4)
+    #:for kind, type in DISP_KINDS_TYPES
+    #:for rank in DISP_RANKS
+    #:if rank != 3
```
awvwgk, line 29: The preprocessor is quite complex here, since we are dealing with a range of 0, 1, 2, 3 where only rank 3 is special, maybe we can create two separate interface generation loops.

src/stdlib_io.fypp
```diff
@@ -7,18 +7,45 @@ module stdlib_io
...
+    #:else
+    impure module subroutine disp_${rank}$_${type[0]}$${kind}$(val, dim, string)
+        ${type}$, intent(in) :: val${ranksuffix(rank)}$
+        integer, intent(in) :: dim
+        character(len=*), intent(in), optional :: string
+    end subroutine disp_${rank}$_${type[0]}$${kind}$
+    #:endif
+    #:endfor
+    #:endfor
+    impure module subroutine disp_str(val, string)
```
awvwgk, line 43: `impure` should be the default, is there a reason to be extra verbose here?

src/stdlib_io_disp.fypp
```diff
@@ -0,0 +1,125 @@
+#:include "common.fypp"
+#:set RIL_KINDS_TYPES = REAL_KINDS_TYPES + INT_KINDS_TYPES + LOG_KINDS_TYPES
+
+submodule (stdlib_io) stdlib_io_disp
+
+    use, non_intrinsic :: stdlib_strings, only: format_string
```
awvwgk, line 6: `non_intrinsic` is the default, is there a reason for the extra verbosity?

src/stdlib_io.fypp
zoziha: Yes, because I consider that the screen and ASCII code text space are two-dimensional, and use `disp` to print up to three-dimensional arrays (one of the dimensions is realized by `do` loop), so the default highest dimension here is **three,** which has an upper limit . 
The existing `RANK` of `stdlib` seems to mainly serve the `stdlib_stats` routines. I took a look at it, and the dimensions inside generally need to be **three-dimensional or more**. 
But in fact, three-dimensional is not necessary, we can only leave this interface implementation:
```fortran
call disp( A(i, j, k, :, :) [, string] )    !! `i,j,k...` can be determined by `do` loop
```

src/stdlib_io.fypp
zoziha: Originally I wrote `pure` here, and the compiler prompted me to use the `error stop` routine inside this routine, which is `impure`, so I filled in the `im` character. Yes, I think **it should be removed**, haha🤪.

src/stdlib_io_disp.fypp
zoziha: This may be my own bad habit, sorry, **I will remove it， too**.

src/stdlib_io_disp.fypp
zoziha: Great suggestion! I am not so familiar with the code inside `stdlib`, thank you for your reminder. This is also the reason why I recommend `stdlib` to quickly support `fpm`(see https://github.com/fortran-lang/stdlib/pull/437), so that users like me can use `stdlib` directly in `fpm` and become familiar with it.

src/stdlib_io_disp.fypp
zoziha: Originally, the main purpose of `disp` was to quickly print a simple low-dimensional array to our users **on the screen**, and it was simple and easy to use (user-friendly). Users can choose the redirection function of `bash` to switch the screen to other default output files, like:
```fortran
a.out <1.txt >2.txt
```
For outputting arrays to files, there are routines like `savetxt`, which are mainly for outputting two-dimensional arrays to ascii text files, and their functions are quite fixed. 
Yes, we can figure out how to keep `disp` simple and easy to use, and enable it to output files.

src/stdlib_io_disp.fypp
```diff
@@ -0,0 +1,125 @@
...
+    implicit none
+    character(len=*), parameter :: fmt_r = '(*(g11.4, 1x))'
+    character(len=*), parameter :: fmt_c = '(*(g23.4, 1x))'
+    
+contains
+    
+    #:for kind, type in RIL_KINDS_TYPES
+    module procedure disp_0_${type[0]}$${kind}$
+        if(present(string)) print *, trim(string)
```
St-Maxwell, line 15: I would prefer `header` instead of `string` as the name of dummy argument.

src/stdlib_io_disp.fypp
```diff
@@ -0,0 +1,125 @@
...
+        print fmt_r, (val(i), i=1,m)
+    end procedure disp_1_${type[0]}$${kind}$
+
+    module procedure disp_2_${type[0]}$${kind}$
+        integer :: i, j, m, n
+        m = size(val, 1)
+        n = size(val, 2)
+        if(present(string)) print *, trim(string)
+        do i = 1, m
+            print fmt_r, (val(i,j), j=1, n)
```
St-Maxwell, line 32: I think `print fmt_r, val(i,:)` just works, since an asterisk is already in `fmt_r = '(*(g11.4, 1x))'`.

src/stdlib_io_disp.fypp
```diff
@@ -0,0 +1,125 @@
...
+        end do
+    end procedure disp_2_${type[0]}$${kind}$
+
+    module procedure disp_3_${type[0]}$${kind}$
+        integer :: i, dim1, dim2, dim3
+        dim1 = size(val, 1)
+        dim2 = size(val, 2)
+        dim3 = size(val, 3)
+        if(present(string)) print *, trim(string)
+        if(dim == 1) then
```
St-Maxwell, line 42: I prefer `select case` block here.

src/stdlib_io_disp.fypp
```diff
@@ -0,0 +1,125 @@
...
+        else
+            call error_stop('Error(disp): wrong dimension')
+        end if
+    end procedure disp_3_${type[0]}$${kind}$
+    #:endfor
+
+    #:for kind, type in CMPLX_KINDS_TYPES
+    module procedure disp_0_${type[0]}$${kind}$
+        if(present(string)) print *, trim(string)
+        print fmt_c, format_string(cmplx(real(val), &
```
St-Maxwell, line 66: Just `format_string(val, '(g11,4)')`.

zoziha (commented):
Strange discovery

src/stdlib_io.fypp
zoziha: ✔It is OK, I misunderstood what you meant at first, now I understand.👍

src/stdlib_io_disp.fypp
```diff
@@ -3,55 +3,57 @@
...
 submodule (stdlib_io) stdlib_io_disp
 
-    use, non_intrinsic :: stdlib_strings, only: format_string
+    use stdlib_ascii, only: to_string
+    use stdlib_strings, only: format_string
     implicit none
-    character(len=*), parameter :: fmt_r = '(*(g11.4, 1x))'
-    character(len=*), parameter :: fmt_c = '(*(g23.4, 1x))'
+    character(len=*), parameter :: fmt_r = '(*(g12.4, 1x))'
+    character(len=*), parameter :: fmt_c = '(*(g25.0, 1x))'
```
zoziha, line 10: I don't know how to control the formatter here. What I want is to keep the output type with 4 significant digits, but the consequence is that when outputting `-1.e10`, the output result is `-0.1000E+ 11`, here will occupy 11 characters width.

src/stdlib_io_disp.fypp
jvdp1: Like @awvwgk I am also in favor of an `optional` argument `output_unit`

src/stdlib_io_disp.fypp
jvdp1: I suggest to use `output_unit` provided by `iso_fortran_env`

src/stdlib_strings_format_string.fypp
```diff
@@ -0,0 +1,37 @@
+#:include "common.fypp"
+#:set RIL_KINDS_TYPES = REAL_KINDS_TYPES + INT_KINDS_TYPES + LOG_KINDS_TYPES
+submodule (stdlib_strings) stdlib_strings_format_string
+
+    implicit none
+    integer, parameter :: buffer_len = 512
```
jvdp1, line 6: The `buffer_len` should bre mentioned in the specs (note that I may have miss it).

src/tests/io/test_io_disp.f90
```diff
@@ -0,0 +1,26 @@
+program test_io_disp
```
jvdp1, line 1: This is not a test. If there is an issue, the test might still pass, right?

jvdp1 (commented):
Thank you for this PR.
It might be good to submit 2 PRs, instead of one. It could help the reviewers.

src/stdlib_io.fypp
```diff
@@ -7,18 +7,48 @@ module stdlib_io
...
+        ${type}$, intent(in) :: val${ranksuffix(rank)}$
+        character(len=*), intent(in), optional :: string
+    end subroutine disp_${rank}$_${type[0]}$${kind}$
+    #:else
+    module subroutine disp_${rank}$_${type[0]}$${kind}$(val, dim, string)
+        ${type}$, intent(in) :: val${ranksuffix(rank)}$
+        integer, intent(in) :: dim
+        character(len=*), intent(in), optional :: string
+    end subroutine disp_${rank}$_${type[0]}$${kind}$
+    #:endif
```
jvdp1, line 43:
```suggestion
    module subroutine disp_${rank}$_${type[0]}$${kind}$(val, #{if rank == 3}#dim, #{endif}# string)
        ${type}$, intent(in) :: val${ranksuffix(rank)}$
        character(len=*), intent(in), optional :: string
    end subroutine disp_${rank}$_${type[0]}$${kind}$
```

A proposition to avoid the duplictation of the code. If accepted, the previous line (31) should be deleted too

doc/specs/stdlib_strings.md
```diff
@@ -328,3 +328,63 @@ program demo_find
...
+Pure function
+
+#### Argument
+
+- `value`: Integer/real/complex/logical scalar.
+  This argument is intent(in).
+- `format`: Character scalar like `'(F6.2)'`.
+  This argument is intent(in) and optional.
+
+#### Result value
```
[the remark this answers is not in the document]
zoziha: Thank you very much for checking. I will actively correct some of the problems you mentioned. There are indeed many doubts and errors. `format_string` function is used by the `disp` subroutine as an upstream function. The invalidation of `format_string` is discussed here #444 (see https://github.com/fortran-lang/stdlib/pull/444#discussion_r659025616).

zoziha:
I may have some questions about the scheme of setting `output_unit`:
1. A single `*` Already means the default output location, which is the same as `[iso_fortran_env(module):output_unit(integer)]` and belongs to the standard output, or is it that I understand it wrong, they are different?
2. Or your idea is to **add a input argument** for setting the `output_unit` number to the `disp` subroutine.
If it is the latter, you will encounter such a problem, and the polymorphic interface with two optional arguments will encounter such a problem (see [fortran interface `optional`](https://stackoverflow.com/questions/3121954/fortran-90-presence-of-optional-arguments)). At this time, it is better to consider rewriting a set of functions disp(value [, string]) + disp(value, unit [, string]). (I can't express what I mean very well)

The existing test ([comment](https://github.com/fortran-lang/stdlib/pull/445#discussion_r663538439)) does not report an error if there is any issue, because it simply outputs data to the screen (`output_unit`) and cannot be tested well. If the `output_unit` scheme is improved, further testing may be possible.

src/stdlib_io_disp.fypp
zoziha: This format here is set by default and is not shown to users. Do we allow users to customize the output format? This will increase the complexity of `disp`.

St-Maxwell:
> I may have some questions about the scheme of setting `output_unit`:
> 
> 1. A single `*` Already means the default output location, which is the same as `[iso_fortran_env(module):output_unit(integer)]` and belongs to the standard output, or is it that I understand it wrong, they are different?
> 2. Or your idea is to **add a input argument** for setting the `output_unit` number to the `disp` subroutine.
>    If it is the latter, you will encounter such a problem, and the polymorphic interface with two optional arguments will encounter such a problem (see [fortran interface `optional`](https://stackoverflow.com/questions/3121954/fortran-90-presence-of-optional-arguments)). At this time, it is better to consider rewriting a set of functions `disp(value [, string]) + disp(value, unit [, string])`. (I can't express what I mean very well)
> 
> The existing test ([comment](https://github.com/fortran-lang/stdlib/pull/445#discussion_r663538439)) does not report an error if there is any issue, because it simply outputs data to the screen (`output_unit`) and cannot be tested well. If the `output_unit` scheme is improved, further testing may be possible.

I think the interface of `disp` would be something like `disp(value [, unit=output_unit] [, string])`. The dummy arguments `unit` and `string` are optional, and `unit` has a default value `output_unit`.
In fact, `disp` reminds me of  [`raw_print`](http://arma.sourceforge.net/docs.html#raw_print) function in Armadillo. That's why I proposed a similar interface for `disp`.

arjenmarkus:
The parameter output_unit is equivalent to the * in "write(*,'(a)')
'Hello'". Historically the value was 6, so that you may encounter
"write(6,*) ..." as well, but there is NO guarantee that LU-number 6
corresponds to the console. Before the introduction of output_unit and its
brethren it was not really possible to refer to "standard output" etc in a
standard way.

Op di 6 jul. 2021 om 05:50 schreef zoziha ***@***.***>:

zoziha:
Generally, in the example below, calling the `disp` subroutine needs to pass the parameters in the order of `unit` and `string`. If the parameters are not passed in the order of this order, we **need to** use the variable name to explicitly achieve this. I am worried that this is more demanding for users (not easy-to-use).
```fortran
module itf
contains
    subroutine disp(value, unit, string)
        use, intrinsic :: iso_fortran_env, only: output_unit
        real :: value
        integer, optional :: unit
        character(*), optional :: string
        integer :: unit_
        unit_ = merge(unit, output_unit, present(unit))
        if (present(string)) write (unit_, *), string
        write (unit_, *), value
    end subroutine disp
end module itf

program test_itf
    use itf, only: disp
    integer :: unit

    call disp(1.0)
    call disp(1.0, 'disp(1.0) :')           !! ❌
    call disp(1.0, string='disp(1.0) : ')   !! ✔

    open (newunit=unit, file='temp.out')
    call disp(1.0, unit)
    call disp(1.0, unit, 'disp(1.0) :')
end program test_itf

```

zoziha:
Hello, I got a suggestion from @St-Maxwell  and tried to implement it simply. I created a new branch [disp_unit_brief](https://github.com/St-Maxwell/stdlib/tree/feature/disp_unit_brief) in which I implemented `optional` `unit` and `brief`. If all of you feel good, you can merge it to this PR.
### Example
(see [here](https://github.com/St-Maxwell/stdlib/blob/feature/disp_unit_brief/doc/specs/stdlib_io.md#example-3))
```fortran
 disp(string):
 string
 It is a note.

 disp(r):
 [matrix size: 2×3]
 -0.1000E-10  -0.1000E+11    1.000
   1.000        1.000        1.000
 disp(c):
 [matrix size: 2×3]
            (1.000,0.000)             (1.000,0.000)             (1.000,0.000)
            (1.000,0.000) (-0.1000E+11,-0.1000E+11)             (1.000,0.000)
 disp(i):
 [matrix size: 2×3]
           1            1            1
           1            1            1
 disp(l):
 [matrix size: 10×10]
           T            T            T          ...            T
           T            T            T          ...            T
           T            T            T          ...            T
           :            :            :            :            :
           T            T            T          ...            T
 disp(c_3d, 3):
 [matrix size: 2×100]
            (2.000,0.000)             (2.000,0.000)             (2.000,0.000)                       ...             (2.000,0.000)  
            (2.000,0.000)             (2.000,0.000)             (2.000,0.000)                       ...             (2.000,0.000)  
 disp(c_3d, 2):
 [matrix size: 100×20]
            (2.000,0.000)             (2.000,0.000)             (2.000,0.000)                       ...             (2.000,0.000)  
            (2.000,0.000)             (2.000,0.000)             (2.000,0.000)                       ...             (2.000,0.000)  
            (2.000,0.000)             (2.000,0.000)             (2.000,0.000)                       ...             (2.000,0.000)  
                        :                         :                         :                         :                         :  
            (2.000,0.000)             (2.000,0.000)             (2.000,0.000)                       ...             (2.000,0.000) 
```
### Reference link
+ [armadillo/brief_print](http://arma.sourceforge.net/docs.html#brief_print)
image

+ [ctsa/mdisplay](https://github.com/zoziha/fortsa/blob/256001363bd250dc81c5a20090049052033a9fe8/src/ctsa/matrix.c#L368)

zoziha:
I found two problems during my coding process:
1. Other data types, such as `integer`, `complex`, and `logical` types, are all right-aligned, but the default printing format of `real`r types is neither left-aligned nor right-aligned. This looks weird in `disp_brief`, of course, we can ignore it, this problem is not fatal.
image
2. I saw `stdlib` set the derived string type `string_type`, I wonder if `disp` needs to support it?

src/stdlib_io_disp.fypp
```diff
@@ -0,0 +1,181 @@
...
+        write(unit_, *) value_
+    end procedure disp_character
+
+    module procedure disp_string_type
+        !! Display a `string_type` scalar
+        integer :: unit_
+        unit_ = merge(unit, output_unit, present(unit))
+        if(present(header)) write(unit_, *) header
+        write(unit_, *) char(value)
+            !!\TODO: Need to improve ?
```
zoziha, line 178: The companion function `char` of `string_type` type is used here, which is convenient for the use of `write(unit, *)` in multiple scenarios, and may reduce the speed a little.

zoziha (commented):
Add support for `string_type` type initially, and ready to get your feedback and update docs.
This PR is for `stdlib_io_disp` routines, #444 PR is for `stdlib_string_format_string` routines.
Thank you all~

src/stdlib_io_disp.fypp
```diff
@@ -0,0 +1,181 @@
...
+        write(unit_, fmt_c) format_to_string(value, format_)
+    end procedure disp_0_${type[0]}$${kind}$
+
+    module procedure disp_1_${type[0]}$${kind}$
+        !! Display a ${type}$ vector.
+        integer :: unit_
+        logical :: brief_
+        integer :: i, m
+        unit_  = merge(unit, output_unit, present(unit))
+        brief_ = merge(brief, .false., present(brief))
```
zoziha, line 102: In order to prevent users from accidentally using `disp` to print very long-dimensional arrays, causing unnecessary IO blockage, can we set brief to `.true.` by default? 
Previously, brief was set to `.false.` by default.

zoziha: I thought about the following implementation of the `brief` argument:
1. `brief_ = .true. .and. present(brief)`
5*5 (width * length elements)
2. `.not.present(brief)`
10*50
3. `brief_ = .false. .and. present(brief)`
all

Purpose: To prevent users from accidentally passing a very large array to disp, causing unnecessary io blockage.

zoziha:
Because `to_string` has been merged and `disp` depends on `to_string`, I updated `disp` this time.
Come to think of it, `disp` seems to have enough possibilities. We welcome your suggestions for extension and modification. The purpose of `disp` is to help users quickly understand the matrix data and print statements.
```fortran
call disp("It is a note.")
call disp(A(1:10, 2:4), "Let me see the array:")
```

PR #452, opened 2021-07-04
Implementation of a map data type

awvwgk:
This is a very basic implementation of a container data type and a map data type. The main purpose of this PR is to explore a possible API for a container and map data type, which can later get expanded to become more general

*Additions*

- add placeholder `container_type`
- add abstract base class for map data type (`map_class`)
- add implementation of unsorted map using cuckoo hashing (`cuckoo_hash`, pure allocatable implementation)
- map data structure owns data containers to avoid possibilities of memory leaks
- add Fortran portation of waterhash (Unlicense, *i.e.* public domain)

*Short comings*

- map only defined between `character` scalar keys and `container_type` values
- no possibility to retrieve a list of all keys or iterate over the map (will add this later)
- ~~requires C component until we have a hash function implemented in Fortran~~
- ~~CMake build system is not adding flags correctly for the build therefore build of C component currently fails with Intel~~


doc/specs/stdlib_hash.md
```diff
@@ -0,0 +1,103 @@
+---
+title: Hashing algorithms
```
jvdp1, line 2: @wclodius2 was working on an implementation of several hash procedures. It is maybe worthwhile to check with him what is the status of this project.

awvwgk: I was actually planning to just use the C implementation until we are able to replace it with @wclodius2 implementation. But than I had some issues with CMake and the manual makefile and decided it would be easier to port the C implementation directly to Fortran rather than waste my time on build systems.

If you think we should split this PR to discuss the hash implementation and the map base class separately, I close this one and create two separate PRs.

Also, the waterhash implementation is now available as standalone fpm project to ease the code review: https://github.com/awvwgk/waterhash.

gareth-nx:
Currently the container has a value of type `class(*)`. Another option is to have a value of `type(c_ptr)`. It would also be possible to support both, say by providing `container%val_cptr` in the container type. 

To my understanding these have different strengths and weaknesses when it comes to holding arbitrary data. 

I believe the `class(*)` approach retains the type information of its value, which is good, but a downside is we cannot hold a array variable. Those cases can be treated by the user making a new derived-type containing the array -- but different derived types are required for each rank, kind, etc. I also understand the language does not permit setting a `class(*)` variable to be of `type(c_ptr)`. 

OTOH I understand `type(c_ptr)` can point to anything, including arrays of arbitrary rank. But the downside is that we lose the type information, and so the programmer has to "know" from the context what the type is, and then use it by providing a fortran pointer of the right type and calling `c_f_pointer`. 

Is there value to supporting both?

doc/specs/stdlib_hash.md
wclodius2: I ran into the problem that there are at least three general API's for hash functions:
1. A single object to be hashed as the single argument.
2. The hash object and a single salt (seed) value.
3. The hash object and two salt values. 
These different APIs reflect different goals for the hash functions. Those with the single argument are typically simple functions intended to hash small objects in situations where denial of service attacks are a negligible concern. I suspect they are appropriate for the majority of hash table applications. Those with a salt/seed are typically complicated functions optimized for hashing large objects, and use the seeds to deal with denial of service attacks. This is useful for internet applications, but I suspect less useful for the applications Fortran would be used for. Then there is the difference between 32 bit and 64 biT hashes. The Waterhash I know is aa 32 bit hash with aa single seed.

The design of the hash table will reflect the the type of hash function it is intended to use. If the hash function uses seeds the hash table will need to set aside storage for the seed, and have an argument representing the seed. Between the three APIs and 32 vs. 64 bits to address all combinations requires 6 different hash table implementations. I can't see doing that as it would confuse the users. I can see implementing two hash tables, so a decision should be made as to .

awvwgk:
One drawback for `type(c_ptr)` I see is that the higher-level data structure cannot own the data anymore. With `class(*), allocatable` we should be memory safe automatically as long as the map data type doesn't use `pointer` attributes (requires `final` procedure).

Having a `type(c_ptr)` requires some extra attention when deleting from a data structure, *i.e.* we cannot delete the data that is behind the pointer, because we don't know nothing about it. One way around it is to require the `drop` or `delete` procedure in data structures to provide the removed container as output and the user has to finalize the contained data manually. I don't see an easy way to use a `final` procedure in this case, maybe with a user provided procedure pointer to, which is invoked on a container on finalization?

gareth-nx:
> One drawback for type(c_ptr) I see is that the higher-level data structure cannot own the data anymore. 

That's a good point, and IMO seems reason enough to keep it as is. The options for managing the deletion would seem to complicate things quite a bit, as compared to the alternative approach of working with arrays -- where users make a custom derived type to hold whatever they need.

doc/specs/stdlib_hash.md
awvwgk: I think we can and should provide several implementation for the map base class. We will probably need at least two specific implementations to cover unordered and ordered maps. In some applications I also have the need to preserve the order in which the elements are added to the map, even if it adds a runtime penality.

For very specific requirement of the map an external implementation of stdlib's base class could be used seamlessly as well if we design the base class well.

jvdp1: I would propose that we start to implement something that would cover e.g., 90-95% of the cases we would meet in Fortran. IMO, it would be 64bit + single seed (32bit might be enough for most users but I usually need a 64bit  ;) ).
@wclodius2 I believe you already implemented and tested different hash functions. Could it be a start for such a module? @awvwgk could possibly complement it with is waterhash implementation. I also have a Fortran version of Lookup3. What do you think both?

awvwgk: Agree, I see some value to start with a `stdlib_hash` module on its own featuring different hashing algorithms. I can start a separate PR to create such a module and we can start collecting hashing algorithms in Fortran there, which are used to implement the unordered maps later.

awvwgk: I opened #455 to start the `stdlib_hash` module.

wclodius2: I am going on a trip tomorrow. Preparing for the trip has kept me busy, and the first few days of the trip will be very busy. It will be difficult for me to do much until this weekend and maybe not until sometime next week. FWIW I have about a dozen hash code implementations, including one for Waterhash. I have a couple of hash table implementations that could be the basis of an unordered map. I had been fiddling with the hash tables as I tried out new hash codes so they may need some polishing before I share them. I also have a red-black tree implementation that could be the basis of an ordered map. Typically in my implementations the key and data are passed as `INT8` vectors, that are presumably generated by using transfer on the original data. The ordering in the red-black tree is therefore not what might be expected from the original data. This raises the following questions:
1. Is there interest in seeing the codes, and which codes have the highest interest?
2. What is the best way of distributing the code? As a proposed addition to `stdlib`, as a separate git project, or as an fpm project?
3. If it is to be distributed as a fpm project how do I go about doing that?

jvdp1: Thank you @wclodius2 for your answer.

> I am going on a trip tomorrow. Preparing for the trip has kept me busy, and the first few days of the trip will be very busy. It will be difficult for me to do much until this weekend and maybe not until sometime next week. 

No problem. Enjoy your trip ;)

> 1. Is there interest in seeing the codes, and which codes have the highest interest?
Yes, sure. See #455 for example. 

> 2. What is the best way of distributing the code? As a proposed addition to `stdlib`, as a separate git project, or as an fpm project?

Since @awvwgk started #455, I guess that the best way is to combine our effort in #455, to discuss there API and implementations.

wclodius2: So should I branch #455 and add my 32 bit hash codes as a PR to it, or should I start my own FPM project, or something else?

jvdp1: I suggest that you start from #455. @awvwgk is ok for you?

awvwgk: Whatever works best for you.

awvwgk:
Closing this in favor of #479 for further discussions.

PR #456, opened 2021-07-09
slice function's documentation made user friendly

Aman-Godara:
Previously, the documentation was bit lengthy and complex to understand especially for a person learning a programming language for the first time.
Although conceptually documentation is still same but it is simpler and shorter now.

Please let me know which one did you find easier to understand.

### DIFF
--- a/doc/specs/stdlib_strings.md
+++ b/doc/specs/stdlib_strings.md
@@ -210,10 +210,10 @@ Deduced `first` and `last` argument take +infinity or -infinity value and deduce
 takes value +1 or -1 depending upon the actual argument(s) provided by the user.  
 
 Extraction Process:  
+Extraction starts only if `last` is crossable from `first` with stride of `stride`.  
 Extraction starts from the first valid index in the defined region to take stride of `stride` 
 and ends when the last valid index in the defined region is crossed.  
 If no valid index exists in the defined region, empty string is returned.
-Extraction starts only if `last` is crossable from `first` with stride of `stride`.
 
 #### Syntax
 
--- a/doc/specs/stdlib_strings.md
+++ b/doc/specs/stdlib_strings.md
@@ -199,25 +199,21 @@ end program demo
 
 #### Description
 
-Extracts the characters from the defined region of the input string by taking strides.
-
-Deduction Process:
-Function first automatically deduces the optional arguments that are not provided by the user.  
-This process is independent of both input `string` and permitted indexes of Fortran.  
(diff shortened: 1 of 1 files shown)


milancurcic (approved):
Thanks, looks good. I'd remove L207 (Deduction process:) and L212 (Extraction process:), I think they're redundant because the paragraphs that follow each already say what process is happening.

Aman-Godara:
> I'd remove L207 (Deduction process:) and L212 (Extraction process:), I think they're redundant because the paragraphs that follow each already say what process is happening.

I added them so that a person can categorise a line under either of the 2 different processes. Explicitly grouping them makes it easier to get a bird view of the overall function and memorise what function does, otherwise there will be too many lines for a user under a single large umbrella to draw a map (in brain) of what function `slice` does. That was my reason to add these 2 lines, if others think the same then I will remove them.

milancurcic:
Sounds good, I don't think it's too important to remove them.

doc/specs/stdlib_strings.md
```diff
@@ -199,25 +199,21 @@ end program demo
...
 #### Description
 
-Extracts the characters from the defined region of the input string by taking strides.
+Extracts the characters from the defined region of the input string by taking strides.  
+Argument `first` and `last` defines this region for extraction by function `slice`.  
+Argument `stride` defines the magnitude and direction (+/-) of stride to be taken while extraction. 
+`stride` when given invalid value 0, is converted to +1.
 
-Deduction Process:
+Deduction Process:
```
jvdp1, line 207:
```suggestion
Deduction Process:
```

jvdp1 (approved):
LGTM. Thank you.

doc/specs/stdlib_strings.md
Aman-Godara: When viewed on GitHub the modified `stdlib_strings.md` file looks like [this](https://github.com/Aman-Godara/stdlib/blob/replace_all/doc/specs/stdlib_strings.md#slice).

awvwgk (approved):
Looks good to me, thanks.

doc/specs/stdlib_strings.md
Aman-Godara: Should I consider this as resolved or needing changes? @jvdp1

jvdp1: It was to remove trailing spaces. If there are needed, consider it as resolved. Otherwise, remove them, and it will be resolved too ;)

jvdp1:
Ok. 3 approvals, I will merge it.

PR #459, opened 2021-07-11
Fix CMake variable usage

wyphan:
`CMAKE_MAXIMUM_RANK` is now properly defined as a CMake cache variable of type string (even though it functions as an int). Thus, at configure time, it is passed on as `cmake -DCMAKE_MAXIMUM_RANK:String=7`, for instance.
I only changed the semantics of defining the variable -- use `set` instead of `option` -- and updated the README file and CI workflows accordingly.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -62,7 +62,7 @@ jobs:
       run: >-
         cmake -Wdev
         -DCMAKE_BUILD_TYPE=Release
-        -DCMAKE_MAXIMUM_RANK=4
+        -DCMAKE_MAXIMUM_RANK:String=4
         -DCMAKE_INSTALL_PREFIX=$PWD/_dist
         -S . -B build
 
@@ -179,7 +179,7 @@ jobs:
       run: >-
          cmake -Wdev
          -DCMAKE_BUILD_TYPE=Release
-         -DCMAKE_MAXIMUM_RANK=4
+         -DCMAKE_MAXIMUM_RANK:String=4
          -DCMAKE_INSTALL_PREFIX=$PWD/_dist
          -S . -B build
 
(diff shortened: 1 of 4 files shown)


awvwgk (approved):
Thanks for sharing. I'm learning something new about CMake every day ;).

I'm intrigued, what is the difference between cached variables and options in CMake? Is there a reason to prefer cached variables over option?

wyphan:
@awvwgk The [online documentation][1] for CMake on `option` seems to suggest that it is used mainly for boolean `ON`/`OFF` values, not for integer data. It has its uses though, as you can see in this StackOverflow [page][2]. Don't worry, I'm still learning CMake too!

[1]: https://cmake.org/cmake/help/latest/command/option.html
[2]: https://stackoverflow.com/questions/36358217/what-is-the-difference-between-option-and-set-cache-bool-for-a-cmake-variabl#36359003

wyphan:
Also, the thing that made me recognize the issue in the first place is when I ran `cmake-gui`. If declared as an `option`, it shows up a checkmark, instead of a text field where I can input a number.

jvdp1:
Thank you for this PR. I am learning something too. However, I think that `cmake -DCMAKE_MAXIMUM_RANK:String=4` is not really intuitive, and more "complex" than `cmake -DCMAKE_MAXIMUM_RANK=4`.

Are there any other alternatives?

wyphan:
The old syntax `-DCMAKE_MAXIMUM_RANK=4` should still work, albeit not
"standard-conforming".

jvdp1 (approved):
Thank you for your explanations @wyphan . LGTM.

certik:
Fine with me also.

jvdp1:
Fine for @certik + 2 approvals, I'll merge.

PR #467, opened 2021-07-16
implemented move_alloc for string_type

### LINKED ISSUE #449: move_alloc for strings
**Description**

Since the underlying `raw` variable (`character(:), allocatable`) of `string_type` is allocatable, we can have a `move_alloc` for `string_type`.

**Prior Art**

**fortran** has `move_alloc` for allocatables.

Aman-Godara:
Closes #449 

Tasks:

- [x] Implemented move
- [x] added test cases for move
- [x] documented newly created function
- [ ] Pure vs Elemental?
- [ ] Name of the function?: as of now it is named as `move`
- [ ] `move` if both arguments are of type `character(len=:)`?: `move_alloc` does the same thing **OR** should we consider renaming it from `move` to `move_alloc`.

### DIFF
--- a/src/stdlib_string_type.fypp
+++ b/src/stdlib_string_type.fypp
@@ -206,6 +206,16 @@ module stdlib_string_type
         module procedure :: verify_char_string
     end interface verify
 
+    !> Version: experimental
+    !>
+    !> Moves the allocated character scalar from 'from' to 'to'
+    !> [Specifications](../page/specs/stdlib_string_type.html#move)
+    interface move
+        module procedure :: move_string_string
+        module procedure :: move_string_char
+        module procedure :: move_char_string
+    end interface move
+
     !> Lexically compare the order of two character sequences being greater,
     !> The left-hand side, the right-hand side or both character sequences can
     !> be represented by a string.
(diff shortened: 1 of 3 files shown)


awvwgk:
> Pure vs Elemental?

It can't be `elemental` since it might handle an `allocatable` data type in case of deferred length characters. You could swap two `string_type` arrays with an `elemental` subroutine, though. Not sure if this is a common application of this procedure.

> `move` if both arguments are of type `character(len=:)`?: `move_alloc` does the same thing **OR** should we consider renaming it from `move` to `move_alloc`.

It can't overload `move_alloc` since we won't be able to use the procedure in the context of `type(string_type), allocatable` anymore, like in the example below:

```f90
use stdlib_string_type
type(string_type), allocatable :: str1, str2
str1 = string_type("Moveable string value")
call move_alloc(str1, str2)
end
```

awvwgk (approved):
This is a useful addition, thanks.

Aman-Godara:
Requesting Clarification:

As I understand, the function can't be made `elemental` because a user cannot create an array in which each element is of type `character(len=:), allocatable`? Have I understood it correctly?

Aman-Godara:
How about instead of overloading the `move_alloc` we provide the functionality of `move_alloc` in `move` as well. To do this we will just call `move_alloc` inside `move` and that will do the job.

```fortran
character(len=:), allocatable :: char_1, char_2
char_1 = "moveable char"
move(char_1, char_2)      ! char_2 <-- "moveable char"
```
above expected behaviour can be carried out using `move_alloc` as well. Thus, both `move_alloc` and `move` can do this.

Aman-Godara:
@awvwgk I have added a commit 73ec8c5 corresponding to comment mentioned just above by me.

I added `move` for both arguments as character allocatable `(character(len=:), allocatable)` so that a user need not to use a boilerplate if-else condition before using `move` to use `move_alloc` if both arguments are character allocatable.
If there are no repercussions, then in my opinion this will be good to have. Let me know your thoughts on this.

src/stdlib_string_type.fypp
```diff
@@ -721,6 +732,45 @@ contains
 
     end function verify_char_string
 
+    !> Moves the allocated character scalar from 'from' to 'to'
+    !> No output
+    subroutine move_string_string(from, to)
+        type(string_type), intent(inout) :: from
+        type(string_type), intent(out) :: to
+
+        call move_alloc(from%raw, to%raw)
```
zoziha, line 741: `move_alloc` requires its arguments to be `allocatable`. 
The `move` function arguments here are `type(string_type)`, which is not `allocatable`. 
If we overload `move_alloc`, it may cause users to misunderstand that `move_alloc` is different from the Fortran standard at `allocatable` attribute . 
From this perspective, I think the name can be `move` or `move_alloc_`, better not `move_alloc`?

Aman-Godara: I found `move` to be intuitive as well. As if it cut and pastes (not exactly) a string from one location to other. Similar to moving a file from one folder to other.

milancurcic (approved):
Very nice, thanks @Aman-Godara. We'll merge it tomorrow if there are no objections.

Aman-Godara:
But wait! I have objections here. Have a look at commit 73ec8c5 once. I will also improve test cases and documentation for this.

PR #470, opened 2021-07-21
Revival string list 

Aman-Godara:
This PR is based upon #311 by Arjen. Since 311 is very long, I decided to break it into parts and making it easier for others to review and for me to concentrate on one thing at a time.
[How to Start Reviewing](https://fortran-lang.discourse.group/t/reviewers-and-maintainers-needed-for-stdlib/1506/4?u=aman) by gareth.

**Time Line**:

- [ ] 1. Arjen's kickstart
- [ ] 2. Refactored code to use `stdlib`'s `string_type`
- [ ] 3. Corrected a bug in `append operator(//)`
- [ ] 4. Used concept of forward backward indexes
- [ ] 5. Commented lots of code
- [ ] 6.
- [ ] 7.
- [ ] 8.

**Current Status**: Need to get a review of forward backward indexes type. Soon I will provide an image based explanation as well. This PR provides a reference to discuss the issues with design in detail.

**Note** for reviewers: Only review `stringlist_type.f90` file for now. Ignore commented code it is not the focus at this moment and hence it is commented.

**Pre-requisites**: basic knowledge of Fortran would be enough. I will try to explain lines which I might find requiring explanation to make things easy. If I miss anything, please feel free to ask for clarifications. I will continue to mention updates of PR here in this heading comment, so make sure to look for changes to know the current status of PR.

**Inviting**: @zoziha, @St-Maxwell, others are welcomed as well.


src/stdlib_stringlist.f90
```diff
@@ -0,0 +1,1063 @@
...
+!
+!     Not implemented yet:
+!     insert a list or an array of character strings
+!     replace a string, list or an array of character strings
+!     concatenate a list with another list or an array
+!
+!     Limited to implemented routines
+!
+module stdlib_stringlist
+    use stdlib_string_type, only: string_type, move, assignment(=) !, char, operator(==)
```
Aman-Godara, line 19: move: `move_alloc` for `string_type` find it [here](https://github.com/fortran-lang/stdlib/pull/467)

src/stdlib_stringlist.f90
```diff
@@ -0,0 +1,1063 @@
...
+    use stdlib_math, only: clip
```
Aman-Godara, line 20: clip: find it [here](https://stdlib.fortran-lang.org/page/specs/stdlib_math.html#clip-function)

Aman-Godara:
reason for using special indexes as explained by Arjen:

> The idea of the special indices is that that way you can be very precise about the effects - does “insert at the head” mean insert AFTER the first element or BEFORE the first element? Things like that.

Aman-Godara:
Explanation of forward and backward indexes:
**forward index**: `fidx`
**backward index**: `bidx`


After inserting an element at `fidx( x )`, the added element will appear at index `x` when counted from left to right.
After inserting an element at `bidx( y )`, the added element will appear at index `y` when counted from right to left.
inserting at `fidx( 1 )` is equivalent to inserting at `bidx( len + 1 )`
inserting at `bidx( 1 )` is equivalent to inserting at `fidx( len + 1 )`

Inserting an element at **HEAD** is like inserting at `fidx( 1 )` and inserting an element at **TAIL** is like inserting at `bidx( 1 )`

**backward index** is like the **forward index** of the reversed form of the list.


***

<html>
<body>


--- | A | B | ... | Y | Z | ---
-- | -- | -- | -- | -- | -- | --
fidx( 0 ) | fidx( 1 ) | fidx( 2 ) | ... | fidx( len - 1 ) | fidx( len ) | fidx( len + 1 )
bidx( len + 1 ) | bidx( len ) | bidx( len - 1 ) | ... | bidx( 2 ) | bidx( 1 ) | bidx( 0 )


</body>
</html>

zoziha:
In fact, my programming ability is average. It may be more difficult for me to read your code, but I will actively participate in your code review. 
Thank you for your work, it looks great!

Aman-Godara:
> In fact, my programming ability is average.

I only have 4-5 months of experience with Fortran. I too am learning Fortran through these Pull Requests, making mistakes, asking for clarifications and crawling internet. So, if you need to any clarification please feel free to ask, I would be happy to share whatever I have learned and meanwhile get to learn from you.

Aman-Godara:
Apart from reviewing code, you can also leave your opinion and feedback on the design. Coming up with a good design is the major challenge in this PR. 

The design that I have proposed has problems (I agree) one of them is: no one would prefer to use these forward and backward indexes over integer indexes. Having these `stringlist_index_type` indexes just makes code more verbose by calling `fidx` and `bidx` function again and again (I wish I had a better solution to propose).

I could have used positive `integer indexes` for `forward indexes` and negative `integer indexes` for `backward indexes` but the problem was with the zero. Take help of [this](https://github.com/fortran-lang/stdlib/pull/311#issuecomment-779417445) comment to understand it better. While reviewing #311 I got the idea of using special indexes: `forward` and `backward`. Now, I need community's help in figuring out other problems. So feel free to leave your opinion.

src/stdlib_stringlist.f90
```diff
@@ -0,0 +1,1063 @@
...
+        procedure         :: insert_stringlist_int  => insert_stringlist_int_impl
+        procedure         :: insert_stringarray_idx => insert_stringarray_idx_wrap
+        procedure         :: insert_stringarray_int => insert_stringarray_int_impl
+        generic, public   :: insert                 => insert_string_int,      insert_string_idx,     &
+                                                       insert_stringlist_int,  insert_stringlist_idx, &
+                                                       insert_stringarray_int, insert_stringarray_idx
+        procedure         :: get_string_int         => get_string_int_impl
+        procedure         :: get_string_idx         => get_string_idx_wrap
+        generic, public   :: get                    => get_string_int, get_string_idx
+        procedure, public :: len                    => length_list
```
zoziha, line 76: Do you want to bind the `len` function to `stdlib_stringlist(module):stringlist_type(type)`?
Such usage seems to be inconsistent with Fortran's native function `len` and `stdlib_string_type(module):len(interface)`.

src/stdlib_stringlist.f90
```diff
@@ -0,0 +1,1063 @@
...
+    type(string_type)                  :: get_string_int_impl
+
+    integer                            :: idxnew
+
+    !
+    ! Examine the actual index:
+    ! - if the index is out of bounds, return a string_type equivalent to empty string
+    !
+    if ( 1 <= idxn .and. idxn <= len( list ) ) then
+        get_string_int_impl = list%stringarray(idx)
```
zoziha, line 514: There seems to be a typo here: `idx` -> `idxn` or `idxnew`

src/stdlib_stringlist.f90
```diff
@@ -0,0 +1,1063 @@
...
+    if ( allocated( list%stringarray ) ) then
+        capacity = size( list%stringarray )
+    end if
+
+end function capacity
+
+! Convert fidx and bidx to the equivalent integer
+pure integer function to_idxn( idx )
+    type(stringlist_index_type), intent(in) :: idx
+    integer                                 :: to_idxn
```
zoziha, line 292: Symbol 'to_idxn' already has basic type of INTEGER:
```fortran
pure integer function to_idxn( idx )
    integer :: to_idxn
```

src/stdlib_stringlist.f90
```diff
@@ -0,0 +1,1063 @@
...
+
+    to_idxn = merge( idx%offset, list%size + 2 - idx%offset, idx%head )
```
zoziha, line 294: "Symbol 'list' has no IMPLICIT type"

zoziha (commented):
It seems that `CMakeLists.txt` and `Makefile.manual` have not been set up to compile your code, so maybe I should focus on understanding the issues you want to discuss, not grammatical typos.

src/stdlib_stringlist.f90
Aman-Godara: You mean `size` should have been a better choice to be consistent with Fortran arrays?

src/stdlib_stringlist.f90
Aman-Godara: my bad!, if you wish you can suggest these changes using github to get the contribution to your name.

src/stdlib_stringlist.f90
zoziha: I am a bit confused, I saw that you bind `len` procedure to `stringlist_type` type (see [stdlib_stringlist.f90#L76](https://github.com/Aman-Godara/stdlib/blob/bc9eda7a9381dbde6c22c8216b37ce447d045e3a/src/stdlib_stringlist.f90#L76)), but then use the `len` function like this (see [stdlib_stringlist.f90#L362](https://github.com/Aman-Godara/stdlib/blob/bc9eda7a9381dbde6c22c8216b37ce447d045e3a/src/stdlib_stringlist.f90#L362)). My point is that `len` should be like this (see [stdlib_string_type.fypp#L60](https://github.com/Aman-Godara/stdlib/blob/bc9eda7a9381dbde6c22c8216b37ce447d045e3a/src/stdlib_string_type.fypp#L60)), it is an independent interface.
The following two syntaxes do not seem to be the same function :
```fortran
length = len(stringlist_type :: list)

---

type(stringlist_type) :: list
length = list%len()
```
So I want to confirm with you : 
> Do you want to bind the `len` function to `stdlib_stringlist(module):stringlist_type(type)`?

Aman-Godara: I should have used `list%len()`. I guess this is object oriented programming stuff where `list` is an object of type `stringlist_type` and `len()` is a method. I apparently have confused this with functional programming throughout the PR. Need to make a lot of changes now.

Aman-Godara:
Need help with ubuntu-latest-10 error.

src/Makefile.manual
```diff
@@ -40,6 +40,7 @@ SRC = f18estop.f90 \
       stdlib_logger.f90 \
       stdlib_quadrature_gauss.f90 \
       stdlib_strings.f90 \
+	  stdlib_stringlist.f90
```
awvwgk, line 43:
```suggestion
      stdlib_stringlist.f90 \
```

Aman-Godara:
There is one thing that I am really interested in discussing: Should we or Should we not include the functionality of extending a string list when a out of bounds index is provided.
Take help of this example, to know more.

string_list = [ `'a'`, `'b'`, `'c'` ] (consider this as a string list with 3 elements)
Now a user does, `string_list%insert_at( 10, 'j' )` what should be state of the string_list after that?
Option 1: [ `'a'`, `'b'`, `'c'`, `''`, `''`, `''`, `''`, `''`, `''`, `'j'` ]
Option 2: [ `'a'`, `'b'`, `'c'`, `'j'` ]

Currently this functionality is not provided and the `string_list` is behaving as **Option 2**.

awvwgk (approved):
Thanks a lot for your work here, Aman @Aman-Godara.

Arjen @arjenmarkus, could you have another look over this patch.

src/stdlib_stringlist.f90
```diff
@@ -0,0 +1,726 @@
...
+    subroutine destroy_list( list )
+        !> TODO: needs a better name?? like clear_list or reset_list
+        class(stringlist_type), intent(out) :: list
+
+        list%size = 0
+        if ( allocated( list%stringarray ) ) then
+            deallocate( list%stringarray )
+        end if
+
+    end subroutine destroy_list
```
everythingfunctional, line 421: The `intent(out)` here means that `list` will always have `list%size = 0` and `.not. allocated(list%stringarray)` at the start of the procedure.

src/stdlib_stringlist.f90
```diff
@@ -0,0 +1,726 @@
...
+    end function backward_index
+
+    !> Returns a deep copy of the stringlist 'original'
+    pure function create_copy( original )
+        class(stringlist_type), intent(in)  :: original
+        type(stringlist_type)               :: create_copy
+
+        create_copy = original
+
+    end function create_copy
```
everythingfunctional, line 168: Is there a reason to have an explicit `copy` procedure if intrinsic assignment already makes a copy? Do you envision changes at some point in the future that would cause assignment to no longer make a full, deep copy?

src/stdlib_stringlist.f90
```diff
@@ -0,0 +1,726 @@
...
+
+    !> Version: experimental
+    !> 
+    !> Returns an instance of type 'stringlist_index_type' representing backward index
+    !> [Specifications](../page/specs/stdlib_stringlist.html#bidx)
+    interface bidx
+        module procedure backward_index
+    end interface
+
+    type stringlist_type
```
everythingfunctional, line 53: I don't see any "constructors" for this type (i.e. overload `stringlist_type()`). Must one always build up a list imperatively and mutably? I tend to prefer a declarative, immutable style (i.e. once I construct an object, I don't mutate it).

src/stdlib_stringlist.f90
```diff
@@ -0,0 +1,726 @@
...
+    
+    contains
+        private
+        procedure         :: copy                           =>  create_copy
+
+        procedure, public :: destroy                        =>  destroy_list
+
+        procedure, public :: len                            =>  length_list
+
+        procedure         :: capacity                       =>  capacity_list
```
everythingfunctional, line 66: `len` and `capacity` seem like maybe they should be identical, but `len` doesn't check if the list is allocated.

src/stdlib_stringlist.f90
Aman-Godara: I thought `intent(out)` mean that I won't read anything from the input argument `list` and only write in it.

everythingfunctional: `intent(out)` does mean that, but it also means that the actual argument (e.g. on the calling side) is finalized (either an explicit finalizer is called on the object, or it is returned to a state "as if" no operations had ever been performed on it) before execution of the procedure begins.

everythingfunctional: From the standard:

> The INTENT (OUT) attribute for a nonpointer dummy argument specifies that the dummy argument becomes
undefined on invocation of the procedure, except for any subcomponents that are default-initialized (7.5.4.6). Any
actual argument that corresponds to such a dummy argument shall be definable. The INTENT (OUT) attribute
for a pointer dummy argument specifies that on invocation of the procedure the pointer association status of
the dummy argument becomes undefined. Any actual argument that corresponds to such a pointer dummy shall
be a pointer variable or a procedure pointer that is not the result of a function reference. Any undefinition or
definition implied by association of an actual argument with an INTENT (OUT) dummy argument shall not
affect any other entity within the statement that invokes the procedure.

src/stdlib_stringlist.f90
Aman-Godara: I added it to make it more explicit.
I preferred it over normal assignment operator, because it helped me mention the motive. Someone who is reading the code might miss this motive of the developer otherwise. You won't find me using an assignment operator to do deep copying anywhere in the PR.
Otherwise, there is no specific reason for doing so and even it makes the procedure to take more time.

If you wish I can remove it?

src/stdlib_stringlist.f90
Aman-Godara: `capacity` is only for private use of developer.
whereas `len` is for the user using the module as well as the developer.
`capacity` is the length of the underlying array whereas `len` is the number of elements in the list. At this moment `len` and `capacity` behaves the same way but later if we decide to double the size of underlying array at insertion (instead of increasing the size by 1) then they will behave differently.

src/stdlib_stringlist.f90
Aman-Godara: Yeah that's right I forgot to add that.
A constructor which can take an array as input and which can take no argument is what I am planning to add. I will make the changes.

src/stdlib_stringlist.f90
Aman-Godara: So, how should I solve this problem now:
1). remove line 417:419
2). change `intent(out)` to `intent(inout)`

awvwgk: Using `intent(inout)` seems preferable.

Aman-Godara: I just realised for **gfortran** that when I use `intent(inout)` it deallocated `stringarray`.

awvwgk: Isn't that the expected behaviour here and the same you are getting with `intent(out)`. Destroying the string list should reset it, which means that the list of stored strings is destroyed as well. The string list will have a capacity of zero afterwards and will get reallocated at the next insertion.

Aman-Godara: I meant that when I removed line 417:419 and used `intent(inout)` (instead of `intent(out)`). After doing this I expected failures in test cases but they all passed because I thought `list` will not get reset but it did for **gfortran**.

src/stdlib_stringlist.f90
awvwgk: `copy` is not part of the public API. But used internally to create deep copies when appending lists to lists. I think the motivation here is to avoid destroying the list in case of aliasing for an operation like:
```f90
list = list // list
```

src/stdlib_stringlist.f90
awvwgk: From my understanding of `stringlist_type` there is no need for initialization, you should be able to start right away with the default constructor and can just append using `operator(//)` to the instance.

The question is whether we want to additionally support `stringlist_type(["1", "2", "3"])` instead of requiring the usage of `stringlist_type() // ["1", "2", "3"]`?

src/stdlib_stringlist.f90
awvwgk: Sounds like a stale build artifact somewhere, try to run the clean target and compile again.

awvwgk: You still set size to zero, but keep the buffer allocated. This would keep the stale buffer around unless you reallocate it at an insertion. Since size is zero you can't access the stale buffer anymore.

Still, it is better to drop the buffer as well. Destroying the list shouldn't only loose the references to the entries but also free the memory they occupy.

everythingfunctional: I would just do 1. Leave it as `intent(out)`, and the compiler does all the work for you.

src/stdlib_stringlist.f90
everythingfunctional: The semantics of the language and your API's (i.e. `intent(in)`) already guarantee that you're doing a deep copy that things like `list = list // list` will work fine. You can see it in that the implementation of `create_copy` just uses intrinsic assignment, and that intrinsic assignment works for `new_list = list%copy()`. In principle that latter one is doing a copy twice (although in practice the compiler is probably smart enough to only do it once). Unless you switch the internal representation to using pointers - which would cause intrinsic assignment to not make a deep copy - I don't see any need to have an explicit copy procedure.

src/stdlib_stringlist.f90
everythingfunctional: At the time I wrote that comment, I thought the default structure constructor was private, since the components are private, but since all the components either 1) have default initialization, 2) are allocatable, they are optional in the default structure constructor, and so the default structure constructor with no arguments is still public. The nuances of some of this are hard. At any rate, I think it would still be nice to support `stringlist_type(["1", "2", "3"])`.

src/stdlib_stringlist.f90
everythingfunctional: In adherence to the [YAGNI principle](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it), I'd suggest only having one for now. At the very least, the public one should check the allocation status.

src/stdlib_stringlist.f90
awvwgk: At least Intel will be generating warnings for the `intent(out)` only case. Alternatively we could recommend to destroy a list to recreate it with:

```f90
list = stringlist_type()
```

Romendakil: Looking at the code above, I don't think that any such finalizer is necessary for allocatable subcomponents. In any case, such a routine (from the intention) needs to have intent(inout).

awvwgk: It's not meant as `final` procedure, but to provide a way to clear all entries from the list beside the default finalizer when leaving a scope. Maybe the name `destroy` is not the best choice here.

everythingfunctional: > such a routine (from the intention) needs to have intent(inout).

That makes sense. I agree.

Looking at the TODO there, I think `clear` is the most applicable name, and matches well with the `intent(inout)`.

src/stdlib_stringlist.f90
Aman-Godara: added the constructor which will take carray (char array) as well as sarray (string array) as input.

src/stdlib_stringlist.f90
Aman-Godara: Ok, I will remove the copy function and add some comments there instead to mention my motive.

After reading Sebastian's comment of `list // list`, I realised that inserting a list in the same list will cause errors.
```
list%insert_at( fidx(x), list )
```

Aman-Godara:
I have 2 questions:

1). Should the module be renamed to `stringlist_type` to be consistent with `string_type`?
2). Should `to_current_idxn` be made public?

src/stdlib_stringlist.f90
Aman-Godara: If public one is checking the allocation status then should I remove the component `size` from the whole module?

everythingfunctional: You probably could. I usually avoid keeping essentially duplicate information that needs to be kept consistent, as it's one more chance for bugs to creep in.

Aman-Godara:
> 1). Should the module be renamed to `stringlist_type` to be consistent with `string_type`?
> 2). Should `to_current_idxn` be made public?

Let me rename module from `stringlist` to `stringlist_type` and keep `to_current_idxn` as private.

src/stdlib_stringlist.f90
Aman-Godara: What are thoughts on commit 1584eeb. Preserved `len` function but didn't remove `size` component from the module.
Removed some other lines as well.

everythingfunctional: You should check the allocation status in the `len` function. I.e. the logic you had in the `capacity` function.

Aman-Godara:
All changes have been made.

everythingfunctional (approved):
I could probably nitpick some more, but honestly this is a pretty good first implementation. I also don't want to jeopardize you completing GSoC, because IMO, this is more than sufficient to pass. Thanks for the hard work you put in.

doc/specs/stdlib_stringlist_type.md
```diff
@@ -0,0 +1,379 @@
...
+
+- `idx`: [[stdlib_stringlist_type(module):stringlist_index_type(type)]].
+  This argument is intent(in).
+
+- `string`: Character scalar or [[stdlib_string_type(module):string_type(type)]].
+  This argument is intent(in).
+
+#### Result value
+
+No result.
```
jvdp1, line 185: I don't agree with `No result`. Something happens. What about:

```suggestion
The result is the instance of type `strdlinglist_type` with an additional string.
```

doc/specs/stdlib_stringlist_type.md
```diff
@@ -0,0 +1,379 @@
...
+
+Pure subroutine.
+
+#### Argument
+
+No arguments.
+
+#### Result value
+
+No result.
```
jvdp1, line 355: Same as before, this results in an empty stringlist.

src/stdlib_stringlist_type.f90
```diff
@@ -0,0 +1,726 @@
...
+!                     In other words, after insertion the element will be present at the asked index
+!                       for both forward and backward indexes                    
+!     insert BEFORE:  Inserts an element BEFORE the element present currently at the asked index
+!     insert AFTER:   Inserts an element AFTER the element present currently at the asked index
+!
+!     Note the distinction between AT and BEFORE in the module. Care has been taken to keep it consistent
+!     throughout the PR
+!
+module stdlib_stringlist_type
+    use stdlib_string_type, only: string_type, operator(/=) !, move
```
jvdp1, line 18:
```suggestion
    use stdlib_string_type, only: string_type, operator(/=)
```

src/stdlib_stringlist_type.f90
```diff
@@ -0,0 +1,726 @@
...
+    private
+
+    public :: stringlist_type, operator(//), operator(==), operator(/=)
+    public :: list_head, list_tail, fidx, bidx, stringlist_index_type
+
+    type stringlist_index_type
+        private
+        logical :: forward
+        integer :: offset
+
```
jvdp1, line 30:
```suggestion
```

src/stdlib_stringlist_type.f90
```diff
@@ -0,0 +1,726 @@
...
+        module procedure new_stringlist
+        module procedure new_stringlist_carray
+    end interface
+
+    !> Version: experimental
+    !> 
+    !> Concatenates stringlist with the input entity
+    !> Returns a new stringlist
+    !> [Specifications](../page/specs/stdlib_stringlist_type.html#append-operator)
+    interface operator(//)
```
jvdp1, line 107: Are these operators defined in the specs?

src/stdlib_stringlist_type.f90
```diff
@@ -0,0 +1,726 @@
...
+        type(stringlist_type)                           :: new_stringlist
+        type(string_type), dimension(0)                 :: sarray
+
+        new_stringlist = stringlist_type( sarray )
+
+    end function new_stringlist
+
+    !> Constructor to convert chararray to stringlist
+    !> Returns a new instance of type stringlist
+    pure function new_stringlist_carray( array )
```
jvdp1, line 161: Why is there not a constructor with array of type `string_type`?

jvdp1 (approved):
nice job. Thank you!
I approved it, pending some minor comments.

doc/specs/stdlib_stringlist_type.md
Aman-Godara: Here `No result` means that there is no output.
The subroutine here does not return anything as the output but it does what it is mentioned in the description of the functionality.

jvdp1: I don't agree with you: there is an output: an "updated" instance of type "stringlist_type" with an additional string.

The instance is not the same before and after the call of this subroutine. This output is also represented by the `intent(inout)`.

Aman-Godara: If we mention this can it likely cause a reader to misunderstand this subroutine as a function instead which is returning a new `stringlist`, and a user might end using a syntax like this:
`list = list%insert_at( fidx(10), "10th element")`

jvdp1:
Well, I don't completely agree with you.
But if you want to not mention a result, then I would suggest to remove the
section `Result value`, as done in
https://stdlib.fortran-lang.org/page/specs/stdlib_sorting.html#specifications-of-the-stdlib_sorting-procedures
. So there will be no confusion.

Le ven. 20 août 2021 à 15:53, Aman Godara ***@***.***> a
écrit :

src/stdlib_stringlist_type.f90
Aman-Godara: After size component was discarded it was causing ambiguous interfaces. Which prompted me to remove it but I have added it again. Please have look at the commit 4b69e3d

src/stdlib_stringlist_type.f90
Aman-Godara: Done

jvdp1 (approved):
Thank you @Aman-Godara for this work. Really useful.

Aman-Godara:
Looks like `to_string` has been deleted or something from `stdlib_ascii`. It was one of the dependencies for `stringlist`.

awvwgk:
`to_string` has been moved `stdlib_strings` with #444.

milancurcic (approved):
Thanks @Aman-Godara and reviewers! Unless there are objections let's merge this tomorrow.

Issue #472, opened 2021-07-22
Allocatable argument 'x' is not allocated


brocolis:
**Describe the issue**
Command `gnumake -f Makefile.manual FYPPFLAGS=-DMAXRANK=4 test` on Windows with GCC-11.1.0 x64, ends with

```
./test_linspace
 Made it through first round of tests
At line 111 of file test_linspace.f90
Fortran runtime error: Allocatable argument 'x' is not allocated
```

With GCC-10.3.0 x64, it works:
```
./test_linspace
 Made it through first round of tests
./test_logspace
gnumake[2]: Leaving directory 'C:/fortran/stdlib/src/tests/math'
gnumake[1]: Leaving directory 'C:/fortran/stdlib/src/tests'

All tests passed.
```

jvdp1:
thank you @brocolis for reporting this issue. I can reproduce it on my Fedora computer (gcc 11.1.1). I will have a check

jvdp1:
the issue happens at the [line 109](https://github.com/fortran-lang/stdlib/blob/88e221923f6724f4863a202cd7eb1dbaf0aec83e/src/tests/math/test_linspace.f90#L109) of [test_linspace.f90](https://github.com/fortran-lang/stdlib/blob/master/src/tests/math/test_linspace.f90#L109).

Issue: a negative argument (`-15`) is provided to the function `linspace`. In such a case,  it is assumed that the result would be an array of size 0 (see the code [here](https://github.com/fortran-lang/stdlib/blob/88e221923f6724f4863a202cd7eb1dbaf0aec83e/src/stdlib_math_linspace.fypp#L34)). However, with gcc11, the array remains dealloacted, which triggers the runtine error when evaluating `size(x)` at the following line.

Is it an issue with the code (violating the standard) or with gfortran?

Ping: @ejovo13 @certik @milancurcic @ivan-pi @awvwgk

ejovo13:
I'm on vacation camping but I will be back on the grid by this weekend. Feel free to leave any more information and I will investigate.

I know that other versions of gcc/gfortran were working. For example, on arch gcc version 11, the negative index was properly allocated to a size 0 array.

14NGiestas:
[this answer](https://stackoverflow.com/a/48155875) STD 2018, 12.6.2.2, pg4 states that:

> The characteristics (12.3.3) of the function result are those of the result variable. On completion of execution of the function, the value returned is that of its result variable. 

[...]

PR #474, opened 2021-07-23
Bug fix: Allocatable argument 'x' is not allocated #472

14NGiestas:
There is a unreliable behaviour when dealing with negative size at runtime in GCC 11, this bug fix is addressing this issue.
```f90
program main
    implicit none
    real, allocatable :: x(:)

    x = compile_time()
    print*, allocated(x) ! T
    print*, size(x)      ! 0

    x = runtime(100)
    print*, allocated(x) ! T
    print*, size(x)      ! 100

    x = runtime(-100)
    print*, allocated(x) ! F -> this is a problem!
    print*, size(x)      ! 0

    x = runtime(0)
    print*, allocated(x) ! T
    print*, size(x)      ! 0

    x = fix_runtime(-100)
    print*, allocated(x) ! T
    print*, size(x)      ! 0

contains

    function compile_time() result(res)
        real :: res(-100)
    end function

    function runtime(n) result(res)
        integer, intent(in) :: n
        real :: res(n)
    end function

    function fix_runtime(n) result(res)
        integer, intent(in) :: n
        real :: res(merge(n, 0, n > 0))
    end function

end program
```

### DIFF
--- a/src/stdlib_math.fypp
+++ b/src/stdlib_math.fypp
@@ -50,7 +50,7 @@ module stdlib_math
         ${t1}$, intent(in) :: end
         integer, intent(in) :: n
 
-        ${t1}$ :: res(merge(n, 0, n > 0))
+        ${t1}$ :: res(max(n, 0))
       end function ${RName}$
     #:endfor
 
@@ -77,7 +77,7 @@ module stdlib_math
         ${t1}$, intent(in) :: end
         integer, intent(in) :: n
 
-        real(dp) :: res(merge(n, 0, n > 0))
+        real(dp) :: res(max(n, 0))
       end function ${RName}$
     #:endfor
 
@@ -127,7 +127,7 @@ module stdlib_math
       ${t1}$, intent(in) :: end
       integer, intent(in) :: n
 
-      ${t1}$ :: res(merge(n, 0, n > 0))
+      ${t1}$ :: res(max(n, 0))
     end function ${RName}$
   #:endfor
   #! Integer support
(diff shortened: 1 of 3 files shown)


jvdp1 (commented):
Thank you. The declaration should be also changed in `stdlib_math.fypp`.

14NGiestas:
I believe I should guard the logspace as well, to avoid any further trouble.

jvdp1 (approved):
Thank you for these changes.

ejovo13 (approved):
Elegant solution. Thank you for reporting this and working to clean it up so quickly. I also appreciate the fact that through this bug's discussion I have learned about the intrinsic function `merge` and its use cases compared to the `stdlib` `optval`.

brocolis (approved):
Looks good. Thank you for the fix.

src/stdlib_math.fypp
```diff
@@ -50,7 +50,7 @@ module stdlib_math
         ${t1}$, intent(in) :: end
         integer, intent(in) :: n
 
-        ${t1}$ :: res(n)
+        ${t1}$ :: res(merge(n, 0, n > 0))
```
awvwgk, line 53: Maybe more intuitive than `merge`:
```suggestion
        ${t1}$ :: res(max(n, 0))
```

jvdp1: Indeed! Simpler, and cleaner. +1

14NGiestas: Agreed

awvwgk (approved):

14NGiestas:
Maybe [this two](https://github.com/fortran-lang/stdlib/blob/b66a7b3d74912a29ace7a28f8e22198323c11289/src/stdlib_math_linspace.fypp#L68-L72) should also be guarded.

brocolis (approved):

jvdp1:
Thank you @14NGiestas for this bug fix. I'll merge it.

Fixes #472

PR #475, opened 2021-07-23
Generating sorting subroutines specific to `character` type with fypp

jvdp1:
Previously all sorting subroutines for `integer`, `real`, and `string` were generated with `fypp`. `character` had its own subroutines.

I noticed that some checks were not present in `character`-based subroutines, while present for other types. Therefore, I modified the fypp instructions, such that the sorting subrotuines are generated from a same template for `integer`, `real`, `string`, and `character`.

With this strategy, if someone implements a derived type associated with `>`, `<`, `==`,...., the sorting subroutines can be easily implemented for this DT by modifying only the fypp instructions.

Additional reviewer: @wclodius2

### DIFF
--- a/src/stdlib_sorting.fypp
+++ b/src/stdlib_sorting.fypp
@@ -1,5 +1,10 @@
 #:include "common.fypp"
-#:set IRS_KINDS_TYPES = INT_KINDS_TYPES + REAL_KINDS_TYPES + STRING_KINDS_TYPES
+
+#:set INT_TYPES_ALT_NAME = list(zip(INT_TYPES, INT_TYPES, INT_KINDS))
+#:set REAL_TYPES_ALT_NAME = list(zip(REAL_TYPES, REAL_TYPES, REAL_KINDS))
+#:set STRING_TYPES_ALT_NAME = list(zip(STRING_TYPES, STRING_TYPES, STRING_KINDS))
+#:set CHAR_TYPES_ALT_NAME = list(zip(["character(len=*)"], ["character(len=len(array))"], ["char"]))
+#:set IRSC_TYPES_ALT_NAME = INT_TYPES_ALT_NAME + REAL_TYPES_ALT_NAME + STRING_TYPES_ALT_NAME + CHAR_TYPES_ALT_NAME
 
 !! Licensing:
 !!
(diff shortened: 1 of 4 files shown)


awvwgk (approved):
I'm always happy to negative line count patch. Looks good to me.

src/stdlib_sorting.fypp
```diff
@@ -1,5 +1,10 @@
 #:include "common.fypp"
-#:set IRS_KINDS_TYPES = INT_KINDS_TYPES + REAL_KINDS_TYPES + STRING_KINDS_TYPES
+
```
gareth-nx, line 2: Suggest you add a comment explaining these fypp definitions. Something along the lines of `For better code reuse in fypp, make lists that contain the input types, with each having output types and a separate name prefix for subroutines`.

gareth-nx: Also another sentence about why this makes sense would be good (noting it greatly reduces the code, yet is not entirely obvious).

src/stdlib_sorting_ord_sort.fypp
```diff
@@ -1,5 +1,9 @@
 #:include "common.fypp"
-#:set IRS_KINDS_TYPES = INT_KINDS_TYPES + REAL_KINDS_TYPES + STRING_KINDS_TYPES
```
gareth-nx, line 2: As above, please add a comment explaining the fypp definitions.

src/stdlib_sorting_ord_sort.fypp
```diff
@@ -105,23 +109,28 @@ contains
...
         if ( present(work) ) then
             if ( size( work, kind=int_size) < array_size/2 ) then
-                error stop "${k1}$_${sname}$_ord_sort: work array is too small."
+                error stop "${name1}$_${sname}$_ord_sort: work array is too small."
             endif
 ! Use the work array as scratch memory
             call merge_sort( array, work )
         else
 ! Allocate a buffer to use as scratch memory.
+            #:if t1[0] == "c"
```
gareth-nx, line 127: Here it might be better to check more characters (e.g.  "char" instead of "c") from a clarity perspective.

src/stdlib_sorting_sort.fypp
```diff
@@ -1,5 +1,9 @@
 #:include "common.fypp"
-#:set IRS_KINDS_TYPES = INT_KINDS_TYPES + REAL_KINDS_TYPES + STRING_KINDS_TYPES
+#:set INT_TYPES_ALT_NAME = list(zip(INT_TYPES, INT_TYPES, INT_KINDS))
```
gareth-nx, line 2: As noted earlier, please comment here.

src/stdlib_sorting_sort_index.fypp
```diff
@@ -1,5 +1,9 @@
 #:include "common.fypp"
-#:set IRS_KINDS_TYPES = INT_KINDS_TYPES + REAL_KINDS_TYPES + STRING_KINDS_TYPES
```
gareth-nx, line 2: Please add comment explaining these

src/stdlib_sorting_sort_index.fypp
```diff
@@ -121,7 +125,12 @@ contains
                 call merge_sort( array, index, work, ibuf )
             end if
         else
+            #:if t1[0] == "c"
```
gareth-nx, line 128: As noted previously -- consider comparing with "char" instead of "c"

gareth-nx (approved):
This is really nice -- in addition to the code simplification I think there are lessons on using fypp that would be nice to share more broadly.

src/stdlib_sorting_ord_sort.fypp
jvdp1: True. One additional advantage is that more considered letters will result in less potential conflicts in the futute (e.g., in the case that we want add a new DT that starts with "c")

src/stdlib_sorting.fypp
jvdp1: Done

src/stdlib_sorting_ord_sort.fypp
jvdp1: Done

src/stdlib_sorting_sort_index.fypp
jvdp1: Done

src/stdlib_sorting_sort_index.fypp
jvdp1: Done

jvdp1:
@gareth-nx Thank you for your review. I believed I answered all your comments.

gareth-nx:
@jvdp1 Thanks for the changes.

If anyone with commit rights has time to assist, note this patch seems very straightforward (it is only code simplification) and has been reviewed twice -- but we need another reviewer with commit rights to move it forward. @milancurcic @ivan-pi

jvdp1:
@milancurcic @ivan-pi @LKedward Could one of you review this PR, and merge it if you agree with it, please?

awvwgk:
With two approvals this patch is ready to go. I'll go ahead and merge it now.

Issue #476, opened 2021-07-30
[feature] Add more base routines for `stdlib_linalg`


zoziha:
**Description**

`Fortran` has its unique grammatical characteristics. I think it is impossible to learn from some routines of `Matlab` and `Python` completely. We should be based on the characteristics of `Fortran`.
Seeing the implementation of `linspace/logspace` in `stdlib`, I want to implement some routines in `stdlib`, mainly for `stdlib_linalg`.

#### Tasks
- [ ] `empty`
- [ ] `zeros`
- [ ] `ones`
- [ ] `ex`
- [ ] `arange`
- [ ] `seq`
- [ ] `eye`, for matrix is not squared
- [ ] `det`
- [ ] `diff`
---

- [ ] `is_squared`
- [ ] `is_symmetric`

**Prior Art**

Some routines has been implemented in [keurfonluu/Forlab](https://github.com/keurfonluu/Forlab).
I will rethink how to adapt to `Fortran` to write code.

ghbrown:
I was thinking about proposing and implementing some matrix property checks like
- `is_square` (not sure if it's the same is `is_squared`?)
- `is_symmetric`
- `is_triangular` (upper, lower, and diagonal can fit in here somewhere)
- `is_hermitian`
- `is_hessenberg` (upper, lower, etc.)
etc.

I think there is at least a little bit about API to discuss here, since for example I have found it very useful for `is_square` to also set the dimension `n`. Also, there should be a very clean way to set the API for checks that can result in upper/lower/diagonal/none, but we could discuss what is best for everyone.

I can open an issue to discuss this if that works for you.

PR #477, opened 2021-07-30
[stdlib_linalg] Add empty function. 

zoziha:
- [x] add `empty` function. (see [numpy/empty](https://numpy.org/devdocs/reference/generated/numpy.empty.html#numpy.empty))
#### Tasks
- [x] ready to `allocatable` solution. (see https://github.com/fortran-lang/stdlib/pull/477#issuecomment-890406002)
see #476.

### DIFF
--- a/doc/specs/stdlib_linalg.md
+++ b/doc/specs/stdlib_linalg.md
@@ -207,7 +207,7 @@ program demo_outer_product
 end program demo_outer_product
 ```
 
-## `empty` - `empty` creates a new vector or matrix of `integer` type and given shape, without initializing values.
+## `empty` - Create a new vector or matrix of `integer/real/complex` type and given shape, without initializing values.
 
 ### Status
 
@@ -219,7 +219,7 @@ Pure function.
 
 ### Description
 
-`empty` creates a new vector or matrix of `integer` type and given shape, without initializing values.
+`empty` creates a new vector or matrix of `integer/real/complex` type and given shape, without initializing values.
 
 `empty`, unlike `zeros`, does not set the array values to zero, and may therefore be marginally faster. On the other hand, it requires the user to manually set all the values in the array, and should be used with caution.
 
@@ -233,18 +233,27 @@ For matrix:
 
 ### Arguments
 
-`dim/dim1`: Shall be an `integer` type.
+`dim/dim1`: Shall be an `integer` scalar.
+This is an `intent(in)` argument.
 
-`dim2`: Shall be an `integer` type.
+`dim2`: Shall be an `integer` scalar.
+This is an `intent(in)` argument.
+
+#### Note
+
+Because of `huge(integer :: i) == 2147483647`, the dimensional maximum length of array created by the `empty` function is `2147483647`. 
 
 ### Return value
 
-Returns a new vector or matrix of `integer` type and given shape, without initializing values.
+Returns a new `vector` or `matrix` of `integer` type and given shape, without initializing values.
+
+#### Note
+If the receiving `array` of the return value of the `empty` function is of a `real/complex` type, conversion from `integer` type to `real/complex` type will occur. 
 
 ### Example
 
 ```fortran
-program demo_linlag_empty
+program demo_linlag_empty_1
     
     use stdlib_linlag, only: empty
     implicit none
@@ -252,11 +261,30 @@ program demo_linlag_empty
 
     print *, empty(2, 1)
     print *, 0.0*empty(2)                               !! 0.00000000       0.00000000
-    print *, empty(2) + empty(2)
+    print *, 0.0*empty(2) + 1.0*empty(2)
     print *, (0.1, 0.1)*empty(2) + (0.2, 0.2)*empty(2)
 
     i = empty(2,2)
     print *, i
 
-end program demo_linalg_empty
+end program demo_linalg_empty_1
+```
+
+```fortran
+program demo_linlag_empty_2
+    
+    use stdlib_linlag, only: empty
+    implicit none
+    integer, allocatable :: i(:,:)
+    real, allocatable :: r(:,:)
+    complex, allocatable :: c(:,:)
+    integer :: j(2)
+
+    i = empty(2,1)
+    r = empty(2,1)
+    c = empty(2,1)
+    j = empty(2)
+    print *, i, r, c, j
+
+end program demo_linalg_empty_2
 ```
--- a/doc/specs/stdlib_linalg.md
+++ b/doc/specs/stdlib_linalg.md
@@ -207,7 +207,13 @@ program demo_outer_product
 end program demo_outer_product
 ```
 
-## `empty` - Create a new vector or matrix of `integer/real/complex` type and given shape, without initializing values.
+## `empty`
+
+### Description
+
+`empty` creates a new rank-1 or rank-2 `array` of `integer/real/complex` type and given shape, without initializing values.
+
+`empty`, unlike `zeros`, does not set the array values to zero, and may therefore be marginally faster. On the other hand, it requires the user to manually set all the values in the array, and should be used with caution.
 
 ### Status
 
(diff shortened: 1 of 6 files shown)


src/tests/linalg/test_linalg_empty.f90
```diff
@@ -0,0 +1,59 @@
...
+
+    use test_linalg_empty
+
+    call test_linalg_empty_integer
+    call test_linalg_empty_real
+    call test_linalg_empty_cmplx
+
+    print *, empty(2) + empty(3), size(empty(2) + empty(3))
+        !!TODO: ??
+        !! 44           0           3? / 2?
```
zoziha, line 57: #### In msys2-gfortran
```fortran
size(integer :: i(2) + integer :: j(3)) = 3
size(real :: x(2) + real:: y(3)) = 2
```
It is really wiered.

awvwgk: I would rather expect a runtime error due to a mismatch in the shape. Also since we are dealing with unintialized stack arrays, nobody can tell what the result of the addition of the elements would be.

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,57 @@ program demo_outer_product
...
+    use stdlib_linlag, only: empty
+    implicit none
+    integer, allocatable :: i(:,:)
+
+    print *, empty(2, 1)
+    print *, 0.0*empty(2)                               !! 0.00000000       0.00000000
+    print *, empty(2) + empty(2)
+    print *, (0.1, 0.1)*empty(2) + (0.2, 0.2)*empty(2)
+
+    i = empty(2,2)
```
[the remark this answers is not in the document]
zoziha: Thanks, this is an interesting question. I think that for `empty`, it mainly serves `stdlib_linalg`, with a clearer purpose and simpler use.
It has the following advantages:
1. Can quickly create vectors and matrices;
2. You can `allocate` and repeat to `allocate` a certain `allocatable` `vector/matrix`.
3. It can directly participate in the calculation **as a function**, which has no application requirements in most cases.

The disadvantages of `allocate` are:
1. It is not a function.
2. Allocation cannot be repeated. (This may be an advantage in terms of rigor)
example:
```fortran
real, allocatable :: A(:, :)
A = empty(2,2)
A = empty(3,3) !! It is OK.
---
real, allocatable :: A(:, :)
allocate(A(2,2))
allocate(A(3,3)) !! It is an error: Attempting to allocate already allocated variable'A'.
````

In fact, `empty` is similar to `allocate`, with subtle differences. `empty` is only suitable for the creation of 1, 2 dimensional arrays, which meets the general needs of `stdlib_linalg`, and it can repeat `allocate` for `vector/matrix`, and the readability is different from `allocate`.
_In fact, to a certain extent, I am not clear about the practical applications of `empty`._
It is theoretically faster than creating `zeros` and `ones` (see #478) because there is no assignment process.

**Comment by awvwgk** [comment]
The main purpose of `numpy.empty` would be to create an allocation on the heap and return its reference. For a similar mechanism in Fortran we would have to use the pointer attribute, also we cannot assign in such case:

```f90
implicit none
real, pointer :: array(:)
integer :: n = 6000000

array => empty_ptr(n)
! ... do something with array
deallocate(array)  ! explicit free required
contains
   function empty_ptr(n) result(array)
      integer, intent(in) :: n
      real, pointer :: array(:)
      allocate(array(n))
   end function
end
```

The current implementation uses an automatic array to trigger an automatic LHS (re)allocation. This has several drawback, first the automatic array will most likely be created on the stack and cannot be moved to an allocatable array on the heap, but must be copied to the newly allocated array. This will be considerably worse than just using `allocate`. Also large stack arrays allocated by this mean are at risk of overflowing the stack, which leads to hard to debug stack overflows.

Preferably, to match the original intent of the functionality provided by `numpy.empty` to provide a functional access to the `allocate` procedure without introducing additional overhead, the implementation would be along the lines of:

```f90
function empty(n) result(array)
  integer, intent(in) :: n
  real, allocatable :: array(:)
  allocate(array(n))
end function empty
```

In the best case the compiler might move the allocation produced by the function to the LHS array instead of separately allocating and copying the uninitialized array.

doc/specs/stdlib_linalg.md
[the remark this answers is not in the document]
awvwgk: >  It is theoretically faster than creating zeros and ones (see #478) because there is no assignment process.

This is actually not correct, in the current implementation `zeros`, `ones` and `empty` are identical, as they all require to copy the automatic array to the newly allocated LHS array.

zoziha:
> In the best case the compiler might move the allocation produced by the function to the LHS array instead of separately allocating and copying the uninitialized array.

Take a look at these two examples @awvwgk , I see from the example here, the current code implementation is more efficient in gfortran (not ifort), and the code implementation method you mentioned can also be (re)allocated.
#### Example 1
In the case of small stack usage, `gfortran` is more efficient and `ifort` is less efficient for the current implementation. (Um..)
image
```fortran
program test

    real :: stime, etime
    real(kind=8), allocatable :: A(:,:), B(:,:)

    call cpu_time(stime)
    A = empty1(100,100)
    A = empty1(200,200)
    call cpu_time(etime)
    print *, "etime - stime (seconds) : ", etime - stime

    call cpu_time(stime)
    B = empty2(100,100)
    B = empty2(200,200)
    call cpu_time(etime)
    print *, "etime - stime (seconds) : ", etime - stime

contains

    pure function empty1(ndim1, ndim2) result(result)
        implicit none
        integer, intent(in) :: ndim1, ndim2
        real(kind=8), allocatable :: result(:,:)
        allocate(result(ndim1, ndim2))
    end function empty1

    pure function empty2(ndim1, ndim2) result(result)
        implicit none
        integer, intent(in) :: ndim1, ndim2
        real(kind=8) :: result(ndim1, ndim2)
    end function empty2

end program test
```
#### Example 2
In the case of large stack usage, `gfortran` is more efficient as well and `ifort` stack overflows for the current implementation. (Um.. +1)
image
```fortran
program test

    real :: stime, etime
    real(kind=8), allocatable :: A(:,:), B(:,:)

    call cpu_time(stime)
    A = empty1(1000,1000)
    A = empty1(2000,2000)
    call cpu_time(etime)
    print *, "etime - stime (seconds) : ", etime - stime

    call cpu_time(stime)
    B = empty2(1000,1000)
    B = empty2(2000,2000)
    call cpu_time(etime)
    print *, "etime - stime (seconds) : ", etime - stime

contains

    pure function empty1(ndim1, ndim2) result(result)
        implicit none
        integer, intent(in) :: ndim1, ndim2
        real(kind=8), allocatable :: result(:,:)
        allocate(result(ndim1, ndim2))
    end function empty1

    pure function empty2(ndim1, ndim2) result(result)
        implicit none
        integer, intent(in) :: ndim1, ndim2
        real(kind=8) :: result(ndim1, ndim2)
    end function empty2

end program test
```

zoziha:
I think based on the above examples, to be compatible with `ifort`'s compiler and maintain the robustness of `empty`, we use the `allocatable` solution seems to be a better choice.

src/tests/linalg/test_linalg_empty.f90
zoziha: I am tring to add a patch to this PR for `allocatable` solution. 
The topic discussed here does not seem to belong to the `empty` function. I just discovered this problem accidentally. Do I still need to add it to the `empty` test?

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,81 @@ program demo_outer_product
...
+
+#### Note
+If the receiving `array` of the return value of the `empty` function is of a `real/complex` type, conversion from `integer` type to `real/complex` type will occur. 
+
+### Example
+
+```fortran
+program demo_linlag_empty_1
+    
+    use stdlib_linlag, only: empty
```
milancurcic, line 254:
```suggestion
program demo_linalg_empty_1
    
    use stdlib_linalg, only: empty
```

milancurcic (approved):
Like with `eye`, I don't think that the `allocatable` result is a better choice here over the automatic array and the examples show it. However, the difference is probably low impact because this is likely to be used as a convenience function rather than a high-performance one, and in my view UX > performance. I think this can go forward. Let's merge it tomorrow if there are no objections.

awvwgk:
The only use case I see for this function is to trigger an automatic LHS allocation, which I think is rather limited compared to the intrinsic function, because it will always require a statement to make use of this.

```f90
allocate(array(n, n))
array = empty(n, n)
```

Using it in an expression would only work when multiplying by zero, because the initial entries are undefined, but for this purpose we could use `zeros` instead. Also, in this context the result could change depending on compiler options, *e.g.* when requesting the compiler to initialize all reals with signalling NaNs.

I can see why this function is required in numpy to return a pointer to a memory allocation, but I don't think there is need for it in Fortran.

zoziha:
I think the purpose of `empty` is indeed not clear at present, and I agree to cancel this PR.

PR #478, opened 2021-07-30
[stdlib_linalg] Add zeros, ones function.

zoziha:
- [x] add `zeros` function;
- [x] add `ones` function;

> add `expand` function. (the name `ex` needs to discussed) (aready to `expand`) (discard!😏)

see #476.


src/tests/linalg/test_linalg_exs.f90
```diff
@@ -0,0 +1,75 @@
...
+                   msg="all(ex(.true., 2) .eqv. [.true., .true.]) failed", warn=warn)
+        call check(all(ex(.true., 1, 2) .eqv. reshape([.true., .true.], [1, 2])), &
+                   msg="all(ex(.true., 1, 2) .eqv. reshape([.true., .true.],[1,2])) failed", warn=warn)
+    end subroutine test_linalg_ex_logical
+
+    subroutine test_linalg_ex_string_type
+        call check(all(ex(string_type("A"), 2) == [string_type("A"), string_type("A")]), &
+                   msg='all(ex(string_type("A"), 2) == [string_type("A"), &
+                   &string_type("A")]) failed', warn=warn)
+        call check(all(ex(string_type("A"), 1, 2) == reshape([string_type("A"), string_type("A")], [1, 2])), &
```
zoziha: `reshape` function is seem to not compatible with `string_type` type in some compilers.

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,107 @@ program demo_outer_product
...
+    real, allocatable :: A(:,:)
+    
+    A = zeros(2,2)          !! 0 0 0 0
+    A = ones(4,4)           !! 1 1 1 1; 1 1 1 1; 1 1 1 1; 1 1 1 1
+    A = 2.0*ones(2,2)       !! 2.00000000       2.00000000       2.00000000       2.00000000
+
+end program demo
+```
+
+## `ex` - creates an vector or matrix of `integer/logical/real/complex/string_type` type, given shape and `value` value.
```
awvwgk, line 260: I think we have an intrinsic which can deal with this already: [spread](https://gcc.gnu.org/onlinedocs/gfortran/SPREAD.html)

zoziha: My intention is that `zeros` and `ones` cannot directly create the following statement based on exist `Fortran` snytax:
```fortran
real(kind=4, 8, 16) :: x = ones(2, 2)
```
That's why I want to create an `ex` function. I think it serves `stdlib_linalg`, which is simpler. There is indeed a little overlap with the `spread` function.
```Fortran
x = ex(1.0, 2,2) !! Same as 1.0*ones(2,2)
y = ex(1.0d0, 2) !! Same as spread(1.0d0, 1, 2) and 1.0*ones(2)
z = ex(1.0d0, 2,2) !! Not same as spread?
```

awvwgk: The equivalent for `ex` for matrices is a combination of `spread` and `reshape`
```
z = reshape(spread(1.0d0, 2*2), [2, 2])
```

`ex` is rather short, how about `expand` instead.

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,107 @@ program demo_outer_product
...
+Experimental
+
+### Class
+
+Pure function.
+
+### Description
+
+`zeros/ones` creates an vector or matrix of `integer` type, given shape and `0/1` value.
+
```
jvdp1, line 223: What is the advantages of these functions over,e.g.,:

```fortran
real, allocatable :: a(:, :)
allocate( a(ndim1, ndim2), source = 0.)
```

zoziha: Like the `empty` comment (see https://github.com/fortran-lang/stdlib/pull/477#discussion_r680288115), `ones/zeros/ex` can directly participate in the calculation as a function, which is more convenient:
```fortran
A = A + 0.0*zeros(2,2)
---
real :: B(2,2)
B = 0.0
A = A + B
---
real, allocatable :: B(:, :)
allocate(B(2,2), source=0.)
A = A + B
```
They serve `stdlib_linalg`. Compared with `allocate`, they are at the business level, and `allocate` is at the grammar level?

zoziha:
The existing grammar of `Fortran` cannot reproduce the `ones` and `zeros` of `Matlab` and `Python` very well. I am looking for ways to adapt the `Fortran` grammar to write such routines, hoping that the `Fortran` committee will support features such as generics and look forward to changes. We can keep such routines in the **experimental version**, and if we know that `Fortran` has a better solution (maybe it has to wait for generics or other?), correct them!

doc/specs/stdlib_linalg.md
zoziha: That is cool:
```fortran
z = reshape(spread(1.0d0, 2*2), [2, 2]) !! It is cool, but more complicated
z = expand(1.0d0, 2, 2) !! It is more simpler.
```
Although I always prefer shorter names, I like `expand`, thanks!

doc/specs/stdlib_linalg.md
awvwgk: Would there be a difference since we declare an automatic array in `zeros` and `ones` rather than an allocatable array? With the former being most likely placed on the stack while the latter is placed on the heap?

It is not unknown that automatic arrays in combination with large array sizes lead to stack overflows, here is an example for testing the difference, `n` might have to be increased a bit to see the effect depending on your setup.

```f90
implicit none
real, allocatable :: array(:)
integer :: n = 6000000

array = ones_heap(n)
print *, size(array)
array = ones_stack(n)
print *, size(array)
contains
   function ones_heap(n) result(array)
      integer, intent(in) :: n
      real, allocatable :: array(:)
      allocate(array(n), source=1.0)
   end function
   function ones_stack(n) result(array)
      integer, intent(in) :: n
      real :: array(n)
      array = 1.0
   end function
end
```

This fails as expected on my machine with Intel Fortran:

```
❯ ulimit -s
8192
❯ ifort test.f90 && ./a.out
     6000000
forrtl: severe (174): SIGSEGV, segmentation fault occurred
a.out              0000000000404A8A  Unknown               Unknown  Unknown
libpthread-2.33.s  00007FDEC3495870  Unknown               Unknown  Unknown
a.out              0000000000403A48  Unknown               Unknown  Unknown
a.out              00000000004037A2  Unknown               Unknown  Unknown
libc-2.33.so       00007FDEC32DDB25  __libc_start_main     Unknown  Unknown
a.out              00000000004036AE  Unknown               Unknown  Unknown
❯ ulimit -s unlimited
❯ ifort test.f90 && ./a.out
     6000000
     6000000
```

Note that even as we are assigning the automatic array to a heap allocation, a stack overflow occurs here.

zoziha: It seems that `ifort`'s implementation of the return value of the `array` type is quite different from that of `gfortran`, and the efficiency is also different. (see https://github.com/fortran-lang/stdlib/pull/477#issuecomment-890441564)

zoziha: I think this may be to blame for the implementation of the `ifort` compiler?? They seem to be normal `fortran` syntax, whether it is the current implementation or an `allocatable` solution.
I see that [`stdlib_linalg(module):eye(function)`](https://github.com/fortran-lang/stdlib/blob/ce3a106f14afcc01b8d2e48437c94e7c5e7a19a1/src/stdlib_linalg.fypp#L85~L97) also uses `integer :: eye(n)` instead of `integer, allocatable :: eye(:)`.

But unfortunately, the current implementation has a very big efficiency advantage in `gfortran`:
image

awvwgk: I see, we should discuss the usage of automatic arrays as function results in a separate issue than. From the experience of a couple of my projects, we got almost half of our support requests due to stack overflows from the usage of automatic arrays, therefore I decided to avoid this feature because whenever possible.

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,140 @@ program demo_outer_product
...
+Experimental
+
+### Class
+
+Pure function.
+
+### Description
+
+`zeros/ones` creates an vector or matrix of `integer/real/complex` type, given shape and `0/1` value.
```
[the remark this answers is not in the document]
Aman-Godara: We need to improve an -> a at many places in the PR

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,140 @@ program demo_outer_product
...
+    compelx :: cA(2)
+    
+    A = zeros(2,2)          !! [0.0,0.0; 0.0,0.0]
+    A = ones(4,4)           !! [1.0,1.0,1.0,1.0; 1.0,1.0,1.0,1.0; 1.0,1.0,1.0,1.0; 1.0,1.0,1.0 1.0]
+    A = 2.0*ones(2,2)       !! [2.0,2.0; 2.0,2.0]
+
+    iA = ones(2)            !! [1,1]
+    cA = ones(2)            !! [(1.0,0.0),(1.0,0.0)]
+    cA = (1.0,1.0)*ones(2)  !! [(1.0,1.0),(1.0,1.0)]
+end program demo
```
Aman-Godara, line 272: It would be good to have a non-square matrix as one of the example.

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,140 @@ program demo_outer_product
...
+`dim2`: Shall be an `integer` type.
+This is an `intent(in)` argument.
+
+#### Note
+
+Because of `huge(integer :: i) == 2147483647`, the dimensional maximum length of array created by the `zeros/ones` function is `2147483647`. 
+
+### Return value
+
+Returns an `vector` or `matrix` of `integer` type, given shape and `0/1` value.
```
Aman-Godara, line 249:
```suggestion
Returns a `vector` or `matrix` of the given shape, filled completely with either `0` or `1` `integer` type values.
```
function doesn't take 0/1 as the input. Name of the function `zeros`/`ones` tells us which value should be used to fill.

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,140 @@ program demo_outer_product
...
+
+Experimental
+
+### Class
+
+Pure function.
+
+### Description
+
+`expand` creates an vector or matrix of `integer/logical/real/complex/string_type` type, given shape and given `value` value.
```
zoziha, line 287: sorry, `an` -> `a`
```suggestion
`expand` creates a vector or matrix of `integer/logical/real/complex/string_type` type, given shape and given `value` value.
```

doc/specs/stdlib_linalg.md
zoziha: Thank you very much 💖, I will correct such writing errors or ambiguities next time.

Jim-215-Fisher:
I will review this one.

src/stdlib_linalg_expand.fypp
```diff
@@ -0,0 +1,74 @@
...
+    pure module function expand_1_${t1[0]}$_${k1}$(value, dim) result(result)
+        
+        ${t1}$, intent(in) :: value
+        integer, intent(in) :: dim
+        ${t1}$, allocatable :: result(:)
+
+        allocate(result(dim))
+        result = value
+        
+    end function expand_1_${t1[0]}$_${k1}$
```
zoziha, line 59: I was thinking, is there a need for `expand` function?
I thought it existed to make up for the existing shortcomings of `zeros` and `ones`.

And it also manipulates `logical` and `string_type` arguments, which appear in `stdlib_linalg`, which seems out of place. (I would remove the support for `logical` and `string_type` arguments)

Aman-Godara: I was rather thinking why it is named as `expand` when it is not expanding anything.

zoziha: I think if the support of `logical` and `string_type` is removed. The functions of `expand` and `zeros/ones` overlap too much, and the difference is not too big. I may consider removing the `expand` function, focusing on the fair use  of the current stage  `zeros/ones` (experimental version).

zoziha: > I was rather thinking why it is named as `expand` when it is not expanding anything.

Haha😜, It expands the `value` value to a vector or a matrix. (the API name is discussed here, https://github.com/fortran-lang/stdlib/pull/478#discussion_r679893172)

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,155 @@ program demo_outer_product
...
+Pure function.
+
+### Description
+
+`expand` creates a `vector` or `matrix` of the given shape, filled completely with `value` `integer/logical/real/complex/string_type` type values.
+
+### Syntax
+
+For vector:  
+`result = [[stdlib_linalg(module):expand(interface)]](value, dim)`
```
Jim-215-Fisher, line 307: `expand` function for vector in vanilla form may not be necessary since we already have `spread` intrinsic function for the same results. However, it may be interesting to have a third optional logical vector mask with size of `dim`.

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,155 @@ program demo_outer_product
...
+
+`expand` creates a `vector` or `matrix` of the given shape, filled completely with `value` `integer/logical/real/complex/string_type` type values.
+
+### Syntax
+
+For vector:  
+`result = [[stdlib_linalg(module):expand(interface)]](value, dim)`  
+
+For matrix:  
+`result = [[stdlib_linalg(module):expand(interface)]](value, dim1, dim2)`
```
Jim-215-Fisher, line 310: It may be interesting to have a third optional logical matrix mask with dimension of `dim1` and `dim2`.

doc/specs/stdlib_linalg.md
zoziha: Yes, I was thinking of removing `expand`. The mask scheme you mentioned seems to be impossible to implement in `Fortran`'s existing syntax. We can't arbitrarily set the dimensions of the returned array in Fortran. (see https://github.com/j3-fortran/fortran_proposals/issues/157)
```fortran
x = reshepe(ones(2*3*4), [2,3,4])  !! This use style will be written in doc later.
```

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,155 @@ program demo_outer_product
...
+`dim2`: Shall be an `integer` type.
+This is an `intent(in)` argument.
+
+#### Note
+
+Because of `huge(integer :: i) == 2147483647`, the dimensional maximum length of array created by the `zeros/ones` function is `2147483647`. 
+
+### Return value
+
+Return a `vector` or `matrix` of the given shape, filled completely with either `0` or `1` `integer [/real/complex]` type values.
```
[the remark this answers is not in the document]
zoziha: Thanks, you are right, I will remove them.😉

doc/specs/stdlib_linalg.md
Jim-215-Fisher: My suggest is like the following:
`function expand(value, dim, mask) result(res)`
`integer, intent(in) :: value, dim`
`logical, intent(in) :: mask(dim)`
`integer, allocatable :: res(:)`

`allocate(res(dim))`
`where(mask)`
`res = value`
`elsewhere`
`res = 0`
`end where`
`end function expand`

zoziha: It seems that `merge` may achieve this effect, for example:
```fortran
x = merge(1, 0, mask) !! `x` is consistent with the dimension information of the `mask` array.
```

src/Makefile.manual
```diff
@@ -81,7 +81,8 @@ stdlib_io.o: \
 	stdlib_optval.o \
 	stdlib_kinds.o
 stdlib_linalg.o: \
-	stdlib_kinds.o
+	stdlib_kinds.o \
+    stdlib_string_type.o
```
zoziha, line 85: Later I will remove this dependency `stdlib_string_type.o` (There is no need here, it is a wrong).

src/stdlib_linalg.fypp
```diff
@@ -4,14 +4,16 @@ module stdlib_linalg
   !!Provides a support for various linear algebra procedures
   !! ([Specification](../page/specs/stdlib_linalg.html))
   use stdlib_kinds, only: sp, dp, qp, &
-    int8, int16, int32, int64
+    int8, int16, int32, int64, lk, c_bool
+  use stdlib_string_type, only: string_type
```
zoziha, line 8: Ditto

gareth-nx:
Hi @zoziha 

Thanks for the pull request. I just started looking at this, and think it might be useful to have more discussion about the use-cases of these routines in the context of Fortran.

To me there are a couple of situations to consider:
1. We want to initialize an array `A` to 0's or 1's. In Fortran there are 2 such situations:
    - If the array size is already known then we can just write `A=1` or `A=0` to achieve results equivalent to `zeros` and `ones`. That is simpler than a function call, and should be easier for a compiler to optimize.
    - If the array size is unknown (allocatable), anyway the rank is known (here Fortran differs from matlab/python). In that case the alternative `allocate(A(dim), source=0)` is only slightly more complex than `zeros` or `ones`. But `allocate` has the advantage of generalising to any rank, and any value. 
2. Using the functions within a right-hand-slide statement (e.g. `A = (20.0*ones(2,2) + 5.0)`, or whatever). In this situation we could simply write `A = 20.0 + 5.0`. Offhand I can't think of cases where the function call could not be straightforwardly eliminated like this -- what are examples? Another benefit of elimination is that there is less risk of users not noticing that `zeros` and `ones` always return integers (as noted in the documentation, it is easy to imagine some users not realising `ones(2,2)/2 == 0` -- whereas IMO they are more likely to notice that `1/2 == 0`).  

This is all I could think of -- and it did make me question whether `zeros` and `ones` would be helpful for stdlib -- however I may well be missing important use cases, so would appreciate some discussion.

zoziha:
Thank you @gareth-nx  and everyone who reviewed this PR , I think what you @gareth-nx  said makes sense, and I agree to **cancel this PR**.
My original intention was:
1. Replace `allocate(array(2, 2), source=0)` to `array = ones(2, 2)`, which is simpler for `stdlib_linalg` (only for matrix and vector operations). 
And I may want to adapt `matlab.ones and numpy.ones` too much, and ignore some differences between statically compiled language and dynamic scripting language.
I will follow the stdlib workflow and discussions about API more.
```fortran
real, allocatable :: array(:, :)

array = ones(2, 2)
array = ones(3, 4)
---
allocate(array(2, 2), source=0)
deallocate(array)
allocate(array(3, 4), source=0)
```
Now it seems that it is not necessary. Fortran itself can handle these well, and `zeros/ones` have some adaptive shortcomings.

Issue #448, opened 2021-06-28
[Feedback] Why not use `merge` but create another `optval`?


zoziha:
**Description**

`merge` is a standard function of `Fortran` that has been implemented, but `stdlib` has created a generic function `optval` (`optval` has a more fixed scope). I think the two functions overlap, and I think it may be `stdlib`'s obligation to use the standard function.
(See [202X feature: Conditional Expressions](https://fortran-lang.discourse.group/t/poll-fortran-202x-conditional-expressions-syntax/1425/39?u=zoziha))
**Prior Art**

```fortran
program main
    
    call check_merge(1.)
    call check_merge()

contains

    subroutine check_merge(opt)
        real, optional :: opt
        
        print *, merge(opt, 2., present(opt))
        !! print *, optval(opt, 2.)

    end subroutine check_merge

end program main
```

awvwgk:
This fails with at least one compiler (`nvfortran`) with a segmentation fault as `merge` is not required to short-circuit the evaluation.

Romendakil:
Does stdlib actually compile and run with `nvfortran`? My own experience in our projects is that `nvfortran` does compile a lot of code, so it understands the syntax, but many implementations are quite buggy, and I would even hesitate to call it a complete F2003 compiler.

awvwgk:
I have tried compiling it a few times because I'm using `nvfortran` for one of my projects, but so far without success (main issue is missing quadruple precision support). I totally agree that `nvfortran` has a very unstable front-end, which makes it a real challenge to work with. Same holds for the current `flang` compiler which uses basically the same front-end.

There is hope that `nvfortran` will at some point use the `f18` front-end instead of the current `pgfortran` based one, but this is still a long way to go.

zoziha:
In the situation I encountered today, `merge(character(6) :: format, character(4) :: "(g0)", present(character(6) :: format))` will report an error: "Unequal character lengths ( 6/4) in MERGE intrinsic". 
But the `optval` routine will not report such an error. 
This is because the function definition of `merge` is stricter. `RESULT = MERGE (TSOURCE, FSOURCE, MASK)` requires: `FSOURCE` shall be of the same type and type parameters as `TSOURCE`. (See https://gcc.gnu.org/onlinedocs/gfortran/MERGE.html)
This round, `optval` wins!

milancurcic:
On a related topic, conditional expression syntax was accepted today for F202X (hopefully F2023). You will then be able to do

```
[...]

PR #480, opened 2021-07-31
[stdlib_math] Add arange function.

zoziha:
#### Description
`seq` returns evenly spaced values within a given interval.
- [x] Add `seq` function. (see [numpy/arange](https://numpy.org/devdocs/reference/generated/numpy.arange.html#numpy.arange))  
(the API name `seq` needs to be discussed, `arange`? I think `seq` is better, it is shorter.) 
(aready to `arange`🔰)
#### Tasks
- [x] `by <= 0` condition. (see https://github.com/fortran-lang/stdlib/pull/480#discussion_r680382134)
- [x] put it in `stdlib_math` or `stdlib_linalg`: `stdlib_math` is better.
- [x] `floor` or `ceiling` function: I think `floor` is better. (see https://github.com/fortran-lang/stdlib/pull/480#discussion_r680382361)
- [x] `by` value is a very small amount. (Need to discuss) (see https://github.com/fortran-lang/stdlib/pull/480#discussion_r681494363) (let users judge😏)

see #476.


doc/specs/stdlib_math.md
```diff
@@ -275,3 +275,61 @@ program demo_logspace_rstart_cbase
...
+
+`end`: Shall be an `integer/real` scalar.
+This is an `intent(in)` and `optional` argument.
+
+`by`: Shall be an `integer/real` scalar and large than `0`.
+This is an `intent(in)` and `optional` argument.  
+
+Warning:  
+If `by = 0`, the `by` argument will be corrected to `1/1.0` by the internal process of the `seq` function.   
+If `by < 0`, the `by` argument will be corrected to `abs(by)` by the internal process of the `seq` function.
```
zoziha, line 309: `by <= 0` condition. Is this way ok?

src/stdlib_math_seq.fypp
```diff
@@ -0,0 +1,58 @@
...
+        integer :: i
+
+        start_ = merge(start, 1.0_${k1}$, present(end))
+        end_   = merge(end,   start,      present(end))
+        by_    = sign(merge(merge(by, 1.0_${k1}$, by /= 0.0_${k1}$), &
+                 1.0_${k1}$, present(by)), end_ - start_)
+
+        allocate(result(floor((end_ - start_)/by_) + 1))
+            #!TODO: ??
+            #! floor or ceiling, floor is better.
```
zoziha, line 27: When `ceiling`:
```fortran
seq(1.0, 3.0, 1.0) = [1.0, 2.0]
seq(1.0, 3.1, 1.0) = [1.0, 2.0, 3.0]
```
When `floor`:
```fortran
seq(1.0, 3.0, 1.0) = [1.0, 2.0, 3.0]
```
So `floor` is better!

awvwgk:
This looks like a useful addition, but the name `seq` seems rather short. Why not use `arange` here? `sequence` would also be an option, but this is already an attribute for derived types used for memory alignment or creating non-extendible types.

zoziha:
In fact, the name `seq` is derived from the `seq` function in the `R` language, taking into account the experience in `fortran`:
```fortran
v = seq(10.0_dp, 100.0_dp, 10.0_dp)    !! It is shorter.
v = (1.0_dp, 1.0_dp)*seq((10.0_dp, 100.0_dp, 10.0_dp)
---
v = arange(10.0_dp, 100.0_dp, 10.0_dp) !! It is OK.
v = (1.0_dp, 1.0_dp)*arange((10.0_dp, 100.0_dp, 10.0_dp)
```
`seq` has three arguments, I hope this API name can be shorter, but I also like `arange` very much, next time I will consider replacing `seq` with `arange`💖.

src/stdlib_math_arange.fypp
```diff
@@ -0,0 +1,56 @@
...
+        
+        ${t1}$ :: start_, end_, by_
+        integer :: i
+
+        start_ = merge(start, 1.0_${k1}$, present(end))
+        end_   = merge(end,   start,      present(end))
+        by_    = sign(merge(merge(by, 1.0_${k1}$, by /= 0.0_${k1}$), &
+                 1.0_${k1}$, present(by)), end_ - start_)
+
+        allocate(result(floor((end_ - start_)/by_) + 1))
```
zoziha, line 25:
```fortran
by /= 0.0_${k1}$
```
When the `by` value is a very small amount, here becomes dangerous. Do we need to set a limit for it?

src/stdlib_math_arange.fypp
```diff
@@ -0,0 +1,56 @@
...
+
+        result = [(start_ + (i - 1)*by_, i=1, size(result), 1)]
```
zoziha, line 27:
```fortran
i = size(result)    !! We know: huge(integer :: i) = 2147483647
```
So `arange` only can create a vector with a maximum length 2147483647.

Jim-215-Fisher:
I will review this one.

src/stdlib_math_arange.fypp
Jim-215-Fisher: This doesn't sound right for the real values. How about arange(1.0,6.0,2.0) which should give us [1.0, 3.5, 6.0]. Here it gives us [1.0, 3.0, 5.0].

zoziha:
> This doesn't sound right for the real values. How about arange(1.0,6.0,2.0) which should give us [1.0, 3.5, 6.0]. Here it gives us [1.0, 3.0, 5.0].

Thanks for your review! @Jim-215-Fisher  `arange` is different with `linspace`, I think you may be talking about the function of [`linspace`](https://github.com/fortran-lang/stdlib/blob/master/doc/specs/stdlib_math.md#linspace---create-a-linearly-spaced-rank-one-array).
(see numpy: [arange](https://numpy.org/devdocs/reference/generated/numpy.arange.html#numpy.arange) and [linspace](https://numpy.org/devdocs/reference/generated/numpy.linspace.html))

Jim-215-Fisher:
> > This doesn't sound right for the real values. How about arange(1.0,6.0,2.0) which should give us [1.0, 3.5, 6.0]. Here it gives us [1.0, 3.0, 5.0].
> 
> Thanks for your review! @Jim-215-Fisher `arange` is different with `linspace`, I think you may be talking about the function of [`stdlib_math(module):linspace(interface)`](https://github.com/fortran-lang/stdlib/blob/master/doc/specs/stdlib_math.md#linspace---create-a-linearly-spaced-rank-one-array).
> (see numpy: [arange](https://numpy.org/devdocs/reference/generated/numpy.arange.html#numpy.arange) and [linspace](https://numpy.org/devdocs/reference/generated/numpy.linspace.html))

After reading numpy's documents, I understand that `arange` has a fixed spacing while `linspace` has fixed number of samples. The fixed number of samples give evenly spaced results. The fixed spacing in `arange` may give "evenly" spaced results without end point. My suggestion would be to change the wording of description for the function from "evenly spaced values" to "fix spaced values".

Another question is whether this function is elemental. If it is, why not change "pure" to "elemental"? In numpy, 'arange' is suitable to array arguments.

zoziha:
> Another question is whether this function is elemental. If it is, why not change "pure" to "elemental"? In numpy, 'arange' is suitable to array arguments.

Thanks, `fixed spaced values` is indeed better. 

This function is not elemental. It does not seem to meet the definition of `elemental` function.
I have not used numpy, but it seems that `numpy.arange` does not state that it supports array arguments. 

Do you mean the return value is an array (not only a vector)? In addition, it is difficult for the existing `fortran` syntax to implement a multi-dimensional array generic interface. (see https://github.com/j3-fortran/fortran_proposals/issues/153)

Jim-215-Fisher:
> > Another question is whether this function is elemental. If it is, why not change "pure" to "elemental"? In numpy, 'arange' is suitable to array arguments.
> 
> Thanks, `fixed spaced values` is indeed better.
> 
> This function is not elemental. It does not seem to meet the definition of `elemental` function.
> I have not used numpy, but it seems that `numpy.arange` does not state that it supports array arguments.
> 
> Do you mean the return value is an array (not only a vector)? In addition, it is difficult for the existing `fortran` syntax to implement a multi-dimensional array generic interface. (see [j3-fortran/fortran_proposals#153](https://github.com/j3-fortran/fortran_proposals/issues/153))

You are right. numpy.arange does not support array argument. Never mind.

Jim-215-Fisher (approved):

doc/specs/stdlib_math.md
```diff
@@ -275,3 +275,75 @@ program demo_logspace_rstart_cbase
...
+This is an `intent(in)` and `optional` argument.   
+The default `by` value is `1`.
+
+#### Warning
+If `by = 0`, the `by` argument will be corrected to `1/1.0` by the internal process of the `arange` function.   
+If `by < 0`, the `by` argument will be corrected to `abs(by)` by the internal process of the `arange` function. 
+
+### Return value
+
+Return a vector of fixed spaced spaced values.
```
milancurcic, line 322:
```suggestion
Returns an array of fixed-space values.
```

doc/specs/stdlib_math.md
```diff
@@ -275,3 +275,75 @@ program demo_logspace_rstart_cbase
...
+
+Experimental
+
+### Class
+
+Pure function.
+
+### Description
+
+Create a vector of the `integer/real` type with given fixed spaced values within a given interval.
```
milancurcic, line 290:
```suggestion
Creates an array of the `integer/real` type with fixed-spaced values of given spacing, within a given interval.
```

doc/specs/stdlib_math.md
```diff
@@ -275,3 +275,75 @@ program demo_logspace_rstart_cbase
...
+
+#### Note
+
+Because of the `i` (`huge(integer :: i) = 2147483647`) index inside the `arange` function , the dimensional maximum length of array created by the `arange` function is `2147483647`.
```
milancurcic, line 294: Is there a use case to create larger arrays? I can't think of one, but if there is, we could use `int64` to allow larger arrays.

doc/specs/stdlib_math.md
```diff
@@ -275,3 +275,75 @@ program demo_logspace_rstart_cbase
...
+
+Create a vector of the `integer/real` type with given fixed spaced values within a given interval.
+
+#### Note
+
+Because of the `i` (`huge(integer :: i) = 2147483647`) index inside the `arange` function , the dimensional maximum length of array created by the `arange` function is `2147483647`. 
+
+### Syntax
+
+`result = [[stdlib_math(module):arange(interface)]](start [, end, by])`
```
milancurcic, line 298: `by` for the third argument sounds odd to me. I'm trying to think who to say this in words: Arange from start to end by...? It doesn't sound right. 

Why not call this `increment`, or just `incr` if `increment` is too long? Or `step`, like in numpy.arange?

src/stdlib_math_arange.fypp
```diff
@@ -0,0 +1,54 @@
...
+        ${t1}$, intent(in), optional :: end, by
+        ${t1}$, allocatable :: result(:)
+        
+        ${t1}$ :: start_, end_, by_
+        integer :: i
+
+        start_ = merge(start, 1.0_${k1}$, present(end))
+        end_   = merge(end,   start,      present(end))
+        by_    = sign(merge(merge(by, 1.0_${k1}$, by /= 0.0_${k1}$), &
+                 1.0_${k1}$, present(by)), end_ - start_)
```
milancurcic, line 21: If `end` or `by` are not present, one of these lines will be standard non-conforming and possibly segfault. We should use `optval` from `stdlib_optval` instead.

src/stdlib_math_arange.fypp
```diff
@@ -0,0 +1,54 @@
...
+        ${t1}$, intent(in), optional :: end, by
+        ${t1}$, allocatable :: result(:)
+        
+        ${t1}$ :: start_, end_, by_
+        ${t1}$ :: i
+
+        start_ = merge(start, 1_${k1}$, present(end))
+        end_   = merge(end,   start,    present(end))
+        by_    = sign(merge(merge(by, 1_${k1}$, by /= 0_${k1}$), &
+                 1_${k1}$, present(by) ), end_ - start_)
```
milancurcic, line 45: Ditto here, `optval` instead of `merge`.

milancurcic (changes requested):
Thank you for this PR! I think the API is mostly good, I gave some naming suggestions to use instead of `by`, and I think there is an issue with using `present()` inside `merge()`, so I suggested using `optval()` instead.

doc/specs/stdlib_math.md
zoziha: In fact, what `arange` returns is a one-dimensional `vector`, the word `array` may be ambiguous?

doc/specs/stdlib_math.md
zoziha: I think the length of `2147483647` as an array dimension is very sufficient. `int64` seems unnecessary, but I may subconsciously explain this in the document. Does it need to be removed?

doc/specs/stdlib_math.md
zoziha: OK, `step`.😏

doc/specs/stdlib_math.md
milancurcic: To me it's the other way around--I think of vector as a mathematical concept and array as a programming one. I think array is also a more Fortrannic word. But it's not a big deal, just a suggestion, pick your favorite, either works well 👍

doc/specs/stdlib_math.md
milancurcic: Okay, great, I was just wondering. I think it's okay that it stays int32. Also, the note is helpful IMO.

src/stdlib_math_arange.fypp
zoziha: I tested it (see https://github.com/fortran-lang/stdlib/issues/448#issuecomment-893110485), it is indeed what you said, I will change this kind of `merge` to `optval`.

doc/specs/stdlib_math.md
zoziha: I think it can be taken as `one-dimensional array`

milancurcic (approved):

awvwgk (approved):

milancurcic:
With 3 approvals, I'll go ahead and merge. Thank you @zoziha and reviewers!

PR #481, opened 2021-08-01
[stdlib_linalg] Update eye function.

zoziha:
- [x] Update `eye` function, add support that matrix **is not square**.
- [x] Based on storage considerations, `integer(int8)` will used as the return value type.
- [x] allocation or automatic: **allocation because of `optional` `dim2`!**(see https://github.com/fortran-lang/stdlib/pull/481#discussion_r693286644)
#### Description
That `eye` creates a non-square matrix would be a nice update.

#### Notes
More routines to do, not this PR
see #476.


Jim-215-Fisher:
I will review this one.

doc/specs/stdlib_linalg.md
```diff
@@ -101,21 +101,29 @@ end program demo_diag5
...
+`dim1`: Shall be a scalar of default type `integer`.
+This is an `intent(in)` argument. 
+
+`dim2`: Shall be a scalar of default type `integer`.
+This is an `intent(in)` and `optional` argument. 
 
 ### Return value
 
-Returns the identity matrix, i.e. a square matrix with ones on the main diagonal and zeros elsewhere. The return value is of type `integer(int8)`.
+Returns the identity matrix, i.e. a square matrix with ones on the main diagonal and zeros elsewhere. The return value is of type `integer`.
```
Jim-215-Fisher, line 126: The matrix may not be square since dim1 may differ from dim2

Jim-215-Fisher: Why the results has to be type `integer`? what about `real`? what about other kinds? I remember python has eye function for different type.

doc/specs/stdlib_linalg.md
```diff
@@ -124,7 +132,10 @@ program demo_eye1
     use stdlib_linalg, only: eye
     implicit none
     real :: a(3,3)
+    real :: b(3,4)
     A = eye(3)
+    A = eye(3,3)
+    B = eye(3,4)
```
Jim-215-Fisher, line 138: May be print out a matrix will be more explanatory here.

doc/specs/stdlib_linalg.md
ivan-pi: The original reasons for having only an integer version can be found here https://github.com/fortran-lang/stdlib/issues/169#issuecomment-611074126. Other kinds are achieved by implicit or explicit type promotion.

Some other possibilities are explored in the thread @zoziha created at Discourse: https://fortran-lang.discourse.group/t/fortran-function-return-value-polymorphism/1350

The best solution would be to submit a proposal to J3 to allow specifying an optional return kind for function results, see https://github.com/j3-fortran/fortran_proposals/issues/91.

Jim-215-Fisher: The reasoning in the comment is only valid to the `eye` function, isn't it?

zoziha: > Why the results has to be type `integer`? what about `real`? what about other kinds? I remember python has eye function for different type.

Such an `eye` function can be converted to the `real/complex` type (see https://fortran-lang.discourse.group/t/fortran-function-return-value-polymorphism/1350/11?u=zoziha), but this is indeed an experimental solution. I hope that such a proposal (see https://github.com/j3-fortran/fortran_proposals/issues/91) can be passed.

Jim-215-Fisher (approved):

src/stdlib_linalg.fypp
```diff
@@ -82,20 +82,28 @@ module stdlib_linalg
...
+    pure function eye(dim1, dim2) result(result)
+
+        integer, intent(in) :: dim1
+        integer, intent(in), optional :: dim2
+        integer, allocatable :: result(:, :)
+
+        integer :: dim2_
+        integer :: i
 
+        dim2_ = merge(dim2, dim1, present(dim2))
```
awvwgk, line 98: Please use `optval` instead of `merge` here to avoid segmentation faults with compilers that don't extend `merge` with short circuiting behaviour.

awvwgk (approved):
Thanks for sharing, @zoziha. Thanks for reviewing @Jim-215-Fisher. Looks good to me, except for one issue with `merge`.

doc/specs/stdlib_linalg.md
```diff
@@ -101,30 +101,63 @@ end program demo_diag5
...
+This is an `intent(in)` and `optional` argument. 
+
+#### Note
+
+Because of `huge(integer :: i) = 2147483647`, the dimensional maximum length of array created by the `eye` function is `2147483647`. 
 
 ### Return value
 
-Returns the identity matrix, i.e. a square matrix with ones on the main diagonal and zeros elsewhere. The return value is of type `integer(int8)`.
+Return the identity matrix, i.e. a square matrix with ones on the main diagonal and zeros elsewhere. The return value is of type `integer`.
```
nshaffer, line 130:
```suggestion
Return the identity matrix, i.e. a matrix with ones on the main diagonal and zeros elsewhere. The return value is of type `integer`.
```
Since this PR generalizes to non-square matrices.

doc/specs/stdlib_linalg.md
```diff
@@ -101,30 +101,63 @@ end program demo_diag5
...
-`n`: Shall be a scalar of default type `integer`. 
+`dim1`: Shall be a scalar of default type `integer`.
+This is an `intent(in)` argument. 
+
+`dim2`: Shall be a scalar of default type `integer`.
+This is an `intent(in)` and `optional` argument. 
+
+#### Note
+
+Because of `huge(integer :: i) = 2147483647`, the dimensional maximum length of array created by the `eye` function is `2147483647`.
```
nshaffer, line 126:
```suggestion
```
I don't think this note is correct, since the range of the default integer kind is processor-dependent (IIRC).

doc/specs/stdlib_linalg.md
```diff
@@ -101,30 +101,63 @@ end program demo_diag5
...
 ### Return value
 
-Returns the identity matrix, i.e. a square matrix with ones on the main diagonal and zeros elsewhere. The return value is of type `integer(int8)`.
+Return the identity matrix, i.e. a square matrix with ones on the main diagonal and zeros elsewhere. The return value is of type `integer`.
+
+#### Warning
+
+If the array that receives the return value of the `eye` function is of `real/complex` type, conversion from `integer` type to `real/complex` type will occur.  
+
+Just as `Fortran` is a strongly typed statically compiled language, be careful with the following statements:
```
nshaffer, line 136:
```suggestion
Since the result of `eye` is of `integer` type, one should be careful about using it in arithmetic expressions. For example:
```

src/tests/linalg/test_linalg.f90
```diff
@@ -81,6 +81,13 @@ subroutine test_eye
     integer :: i
     write(*,*) "test_eye"
 
+    call check(all(eye(3,3) == diag([(1,i=1,3)])), &
+               msg="all(eye(3,3) == diag([(1,i=1,3)])) failed.",warn=warn)
+    
+    rye = eye(3,4)
+    call check(sum(rye(:,1:3) - diag([(1.0_sp,i=1,3)])) < sptol, &
+               msg="sum(rye(:,1:3) - diag([(1.0_sp,i=1,3)])) < sptol", warn=warn)
```
nshaffer, line 89:
```suggestion
    call check(sum(abs(rye(:,1:3) - diag([(1.0_sp,i=1,3)]))) < sptol, &
               msg="sum(abs(rye(:,1:3) - diag([(1.0_sp,i=1,3)]))) < sptol", warn=warn)
```
Probably should sum the absolute values, right?

nshaffer (commented):
Made a few suggestions, but mostly seems good to me. Let me know if you have any questions about my review.

nshaffer (approved):

doc/specs/stdlib_linalg.md
```diff
@@ -101,30 +101,57 @@ end program demo_diag5
...
+`dim1`: Shall be a scalar of default type `integer`.
+This is an `intent(in)` argument. 
+
+`dim2`: Shall be a scalar of default type `integer`.
+This is an `intent(in)` and `optional` argument. 
 
 ### Return value
 
-Returns the identity matrix, i.e. a square matrix with ones on the main diagonal and zeros elsewhere. The return value is of type `integer(int8)`.
+Return the identity matrix, i.e. a matrix with ones on the main diagonal and zeros elsewhere. The return value is of type `integer`.
```
jvdp1, line 126: Why did you change the return value from `integer(int8)` to `integer`?

src/stdlib_linalg.fypp
```diff
@@ -82,20 +83,28 @@ module stdlib_linalg
...
-    end function eye
+    !> Version: experimental
+    !>
+    !> Constructs the identity matrix.
+    !> ([Specification](../page/specs/stdlib_linalg.html#eye-construct-the-identity-matrix))
+    pure function eye(dim1, dim2) result(result)
+
+        integer, intent(in) :: dim1
+        integer, intent(in), optional :: dim2
+        integer, allocatable :: result(:, :)
```
jvdp1, line 94: Could you avoid the `allocatable` status, like for other functions in `stdlib`?

jvdp1 (commented):
Thanks for this PR. I have a few suggestions/comments/questions.

src/stdlib_linalg.fypp
zoziha: (see https://github.com/fortran-lang/stdlib/pull/478#discussion_r680392916, and https://community.intel.com/t5/Intel-Fortran-Compiler/how-to-set-the-stack-size/td-p/933530)
It seems that `ifort` places the automatic array in the heap by default, which leads to a `stack overflow` when the dimension of the automatic array is large. Or is there any disadvantage of the allocatable array, need to avoid it?

doc/specs/stdlib_linalg.md
zoziha: Is there any specific difference between `integer(int8)` and `integer`? I may think `integer` is a bit more common, but there is no difference in actual use.🤣

ivan-pi: The default integer typically uses 32-bits of storage. The use of int8 was suggested by Milan, to save storage I believe. Personally I'm fine with using the default integer.

jvdp1: Yes, it was mainly to save storage. Since the result would be mainly temporary, and most likely converted, I am in favor to keep it as `integer(int8)`

src/stdlib_linalg.fypp
jvdp1: My comment was based on @certik and @milancurcic 's comments ([this comment](https://github.com/fortran-lang/stdlib/issues/113#issuecomment-576028842) and following ones).

Furthermore, the rules for stack vs heap can be different between compilers/OS, and can be also changed within a compiler through a compiler option (e.g., using Gfortran with `-Ofast` will imply `-fstack-arrays`, that will put all arrays of unknown size and temporary arrays onto stack memory, which can create issues for very large arrays).

Therefore, I am in favor to always use automatic arrays when possible.

@certik @milancurcic @awvwgk @ivan-pi Any comments on that topic?

zoziha: This may indeed be a big controversy, it seems to be: @awvwgk agrees with allocation array, @jvdp1 agrees with automatic array.
By default, `ifort`'s automatic array solution faces large-scale arrays, it is easier to report an error: stack overflow. (see https://github.com/fortran-lang/stdlib/pull/478#discussion_r680392916, https://github.com/fortran-lang/stdlib/pull/477#issuecomment-890441564, and https://community.intel.com/t5/Intel-Fortran-Compiler/how-to-set-the-stack-size/td-p/933530)

doc/specs/stdlib_linalg.md
zoziha: Agree, I will change it. (Also `zeros/ones/empty`)

src/stdlib_linalg.fypp
milancurcic: Yes, I suggest using automatic array wherever possible, and `allocatable` otherwise. In this case, it seems like you need to make it `allocatable` because the second dimension depends on the presence of the optional dummy argument. 

However, if you re-write this to be two specific procedures under one generic name, instead of using the `optional` argument, then I think you don't need `allocatable` for the result.

awvwgk: Yes, I'm not a fan of automatic arrays, since usage with `ifort` will lead to stack overflows. The option to allocate automatic arrays on the heap rather than on the stack is there, for `ifort` it happens to be buggy (heap allocation is not always freed). `gfortran` is doing a little better here, but it is an issue as well.

So far automatic array usage has resulted in a significant amount of support request from users for me. Setting the system stack with `ulimit -s unlimited` works, still this is something the user has to remember.

milancurcic: @awvwgk I don't think that's a good reason to avoid a language feature. If a feature (and a very old one) doesn't work with a compiler, let's report it. Avoiding it only helps permeate the bug with the compiler. I've been using this with ifort since 2010 and haven't seen an issue. It's possible that I didn't notice it.

The downside to the `allocatable` approach is that it's always on the heap. The user doesn't have the choice and always gets the less-performant implementation. Also, in the `optional` argument approach, it's possible (but I'm not sure) that the compiler would evaluate the branch inside the function at run-time rather than compile-time. If at run-time, that would be another performance penalty. It's possible that performance is less important for `eye`, I don't know. And the advantage to `allocatable` approach is, of course, that it's always on the heap. :)

@zoziha When resolving conversations as a GitHub feature, I suggest to leave a comment about what you decided, otherwise discussions get hidden but aren't really resolved. Maybe we should document this in the Code Review guide.

zoziha:
@milancurcic Ok, got it.
1. I changed the return value type of the `eye` function to `integer(int8)`;
2. `allocatable` is used because of the existence of `optional` `dim2` argument.

At present, the questions raised by @jvdp1 , I think they are answered and resolved.

milancurcic:
Okay, thanks. If needed we can revisit the allocatable result later. If there are no objections let's merge this tomorrow.

milancurcic (approved):

src/stdlib_linalg.fypp
jvdp1: I agree with @milancurcic: issues with some compilers/options/OS/... are not good reasons to avoid a feature of the language.
Futhermore, because it is a small function that would be most often (I guess) used in expressions (e.g., arr1 = eye(5) * scalar), I am also in favor to apply @milancurcic suggestion (i.e., to have 2 specific procedures under one generic name)

milancurcic: @jvdp1 okay if we do it in a separate PR and merge this one first? We could evaluate the performance difference too.

jvdp1:
Ok for me.

Le dim. 22 août 2021 à 17:22, Milan Curcic ***@***.***> a
écrit :

Issue #438, opened 2021-06-16
[STDLIB_STATS] need to upgrade `stdlib_stats` codes about compilation efficiency


zoziha:
## Overview: Compilation time is too long.
When compiling, I found that compiling `stdlib_stats` uses a lot of computer resources, especially RAM, which is related to the high-dimensional matrix dimensions defined in `stdlib_stats`, which greatly reduces the efficiency of `stdlib` and improves **the overall compilation time** of `stdlib`.
![image](https://user-images.githubusercontent.com/32035096/122204524-cbf12f80-ced1-11eb-9c16-edd6376ea9de.png)

It took my computer (CPU: intel i5 8250U) more  than **two hours** to compile `stdlib` completely,
![image](https://user-images.githubusercontent.com/32035096/122210472-445aef00-ced8-11eb-9ba8-874f58076935.png)
When `RANK=15`, the compiled volume of `stdlib` reached **747MB**.

I took a quick look at the source code and thought that there might be a better way to replace **the polymorphic interface with such a large number of multi-dimensional array arguments**.
(see [high-dimensional matrix dimensions](https://github.com/zoziha/stdlib/blob/38d3c0736de7872dd7b45705cc0d9f82a52790b3/src/fpm/stdlib_stats_moment.f90#L17593))
(see [RANK](https://github.com/fortran-lang/stdlib/blob/88adb9f85239806c32a762e8fa49ca102de5d607/src/common.fypp#L54))
## My understanding is: Rethink, need to be more flexible.

The length within a single dimension defined by Fortran can theoretically be infinitely expanded, but the number of dimensions needs to be manually defined by the user.
In the future, we will also build a large number of functions that use matrices. The current implementation of `stdlib_stats` is unreasonable, not adaptable and needs to be improved, (see [stdlib_stats_moment.fypp](https://github.com/fortran-lang/stdlib/blob/88adb9f85239806c32a762e8fa49ca102de5d607/src/stdlib_stats_moment.fypp#L18)).

`stdlib_stats` presets several basic dimensions to form a polymorphic interface, and sets multiple judgments (see [condition judgments](https://github.com/zoziha/stdlib/blob/38d3c0736de7872dd7b45705cc0d9f82a52790b3/src/fpm/stdlib_stats_moment.f90#L7234)) on the number of processing dimensions, resulting in a decrease in compilation speed and an increase in compilation load.

https://github.com/fortran-lang/stdlib/issues/281
[...]

PR #488, opened 2021-08-19
[stdlib_math] add `is_close` routines.

zoziha:
- [x] Add `is_close` for `real/complex` value tests and checks
- [x] The default value of `rel_tol/abs_tol`
- [x] add `all_close`. (gcc 9.0 not support `select rank`)(use `stdlib_stats` solution)
- [x] equal_nan in `is_close/all_close`(need to discuss) https://github.com/fortran-lang/stdlib/pull/488#issuecomment-932045225 
- [x] `fypp` style tests.
- [x] `sqrt(epsilon(..))`

When we are testing, `is_close` and `all_close` are very useful.

APIs are:
```fortran
!! abs(a - b) <= max(rtol*max(abs(a), abs(b)), atol)
bool = [[stdlib_math(module): is_close(interface)]] (a, b [, rel_tol=sqrt(epsilon(..)), abs_tol=0.0, equal_nan=.false.]) 
bool = [[stdlib_math(module):all_close(interface)]] (a, b [, rel_tol=sqrt(epsilon(..)), abs_tol=0.0, equal_nan=.false.]) 
```
#### Prior Arts
+ [python math.isclose](https://docs.python.org/3.9/library/math.html#math.isclose) and [PEP 485, isclose](https://www.python.org/dev/peps/pep-0485/#numpy-isclose)
+ [numpy.isclose](https://numpy.org/doc/stable/reference/generated/numpy.isclose.html) and [numpy.allclose](https://numpy.org/doc/stable/reference/generated/numpy.allclose.html)


doc/specs/stdlib_logic.md
```diff
@@ -0,0 +1,66 @@
...
+
+The tolerance values are positive, typically very small numbers. The relative difference `(rtol*abs(b))` and the absolute difference `atol` are added together to compare against the absolute difference between `a` and `b`.
+
+```fortran
+abs(a - b) <= atol + rtol*abs(b)
+```
+
+#### Syntax
+
+`res = [[stdlib_math(module):is_close(interface)]] (a, b [, rtol, atol, equal_nan])`
```
zoziha, line 23:
```suggestion
`bool = [[stdlib_logic(module):is_close(interface)]] (a, b [, atol, rtol])`
```

milancurcic:
Thanks, @zoziha. There's already PR #353 by @arjenmarkus which we reviewed a little bit, and discussed on a monthly call, but there didn't seem to be much enthusiasm from others. Perhaps we can revisit this now and get more feedback to see if people would like to see something like this in stdlib.

Overall, I think it's in scope and useful. I like having a function like is_close(), and I like tolerant operators in #353.

It'd be great if we can try and follow the [Workflow](https://github.com/fortran-lang/stdlib/blob/master/WORKFLOW.md), I think we'd more easily narrow down to a) whether a function/module is in scope; b) whether there is interest; c) what kind of API would people like to see.

arjenmarkus:
I would like to contribute to this, either in discussion or code ;). There
is a lot of misunderstanding regarding floating-point calculations and
comparisons.

Op do 19 aug. 2021 om 05:27 schreef Milan Curcic ***@***.***>:

ivan-pi:
There are two more related issues:
* #145 
* #327 (underlying issue for PR #353)

We should consider some of the other name suggestions too. Personally I am not convinced this needs a new module. Could the `stdlib_math` module be home for this?

milancurcic:
I also like the math module better for this.

bilderbuchi:
While you are following `numpy.isclose`, I will caution the following notes in the docs, which you elided above:

**Warning The default atol is not appropriate for comparing numbers that are much smaller than one (see Notes).**


> For finite values, isclose uses the following equation to test whether two floating point values are equivalent.
> absolute(a - b) <= (atol + rtol * absolute(b))
> Unlike the built-in math.isclose, the above equation is not symmetric in a and b – it assumes b is the reference value – so that isclose(a, b) might be different from isclose(b, a). Furthermore, the default value of atol is not zero, and is used to determine what small values should be considered close to zero. The default value is appropriate for expected values of order unity: if the expected values are significantly smaller than one, it can result in false positives. atol should be carefully selected for the use case at hand. A zero value for atol will result in False if either a or b is zero.



For these reasons, I personally tend to prefer the standard lib [`math.isclose()` implementation](https://docs.python.org/3.9/library/math.html#math.isclose):

> math.isclose(a, b, *, rel_tol=1e-09, abs_tol=0.0)
>
>   Return True if the values a and b are close to each other and False otherwise.
>
>    Whether or not two values are considered close is determined according to given absolute and relative tolerances.
>
>    rel_tol is the relative tolerance – it is the maximum allowed difference between a and b, relative to the larger absolute value of a or b. For example, to set a tolerance of 5%, pass rel_tol=0.05. The default tolerance is 1e-09, which assures that the two values are the same within about 9 decimal digits. rel_tol must be greater than zero.
>
>    abs_tol is the minimum absolute tolerance – useful for comparisons near zero. abs_tol must be at least zero.
>
>    If no errors occur, the result will be: abs(a-b) <= max(rel_tol * max(abs(a), abs(b)), abs_tol).
>
>    The IEEE 754 special values of NaN, inf, and -inf will be handled according to IEEE rules. Specifically, NaN is not considered close to any other value, including NaN. inf and -inf are only considered close to themselves.

zoziha:
Thank you for your suggestion @bilderbuchi ! `python.isclose` gave us a lot of inspiration, maybe without rigorous thinking, is the following implementation feasible?:

```fortran
!! abs(a - b) <= rtol*max(abs(a), abs(b)) + atol
bool = [[stdlib_math(module):is_close(interface)]] (a, b [, rtol=1.0e-5, atol=0.0]) 
```

The value of `rtol` can be negotiated, and the value of `atol`, we default to `0.0`?

bilderbuchi:
I would not have expected the errors to be added together, but as mentioned I'm not a floating-point specialist.

If there are no good reasons to do otherwise, I think there is value in following already established popular approaches, if only to make it easier for developers to be familiar with a function's behaviour (i.e. you can say in the docs "behaves like Julia and Python stdlib"). 

In this case, Julia's `Base.isapprox` linked to in #145 and Python's `math.isclose` do the same thing, i.e. 
```
norm(x-y) <= max(atol, rtol*max(norm(x), norm(y)))
``` 
It probably makes sense to just follow that, if you agree with my assesssment of `numpy.isclose`'s drawbacks (see also the [PEP485 rationale on this](https://www.python.org/dev/peps/pep-0485/#numpy-isclose)).

Indeed, atol default should be 0.0, and rtol is probably a matter of taste

zoziha:
Maybe the API still needs discussion, but I just tried to update `stdlib_math : is_close` according to @bilderbuchi 's suggestion: now the `is_close` function is consistent with the python stdlib `math.isclose`.

bilderbuchi:
@zoziha do you think it in-scope/a reasonable effort to also add an `all_close` to compare two array, leveraging the new `is_close`? I noticed that that is something that @awvwgk was interested in (https://github.com/fortran-lang/stdlib/pull/494#issuecomment-903317347).

bilderbuchi:
Also, I don't know where the stdlib lies on the spectrum between "complete features, seldomly changed" and "frequent iteration", but considering the API of `is_close`, it might be useful to add the following switch (from numpy.isclose):
> equal_nan (bool, default false): Whether to compare NaN’s as equal. If True, NaN’s in a will be considered equal to NaN’s in b in the output array.

If that is not available, checking two arrays with missing values (nan) for closeness will never succeeed because two nans are never considered close.

zoziha:
@bilderbuchi In fact, the Fortran language has limited support for arrays of uncertain rank.
I think Fortran should enhance this kind of grammar, and the relevant proposal is here: https://github.com/j3-fortran/fortran_proposals/issues/157
At present, we can implement `all_close` in the following three ways.
1. use `all(is_close(a, b))`.
2. Similar to `stdlib_stats`, this will cause the compiled binary file to be a bit large, see https://github.com/fortran-lang/stdlib/issues/438 .
3. Implement the select rank structure with limited support:
```fortran
logical function all_close(a, b, rel_tol, abs_tol) result(close)

real(rk), intent(in) :: a(..), b(..), rel_tol, abs_tol

select rank(a)
rank(1)
    select rank(b)
    rank(1)
        close = all(is_close(a, b, rel_tol, abs_tol))

    rank default
        error stop "The ranks of a and b are not equal"

    end select

rank(2)
    select rank(b)
    rank(2)
        close = all(is_close(a, b, rel_tol, abs_tol))

    rank default
        error stop "The ranks of `a` and `b` are not equal"

    end select

...

rank default
    error stop "The rank of `a` is too large to be supported"
end select

end function all_close
```

zoziha:
Because `all_close` is based on `is_close`, it is actually equivalent to `all(is_close())`, so I implemented `all_close` using the `select rank` structure.
As I mentioned in https://github.com/fortran-lang/stdlib/issues/438 , because Fortran's syntax is not flexible enough for arrays of uncertain dimensions, the `stdlib_stats` solution will cause the binary files compiled by the `stdlib_stats` package to be larger.

But it seems that `gcc 9.0` does not support the `select rank` structure!
```bash
   14 |         select rank(a)
      |        1
Error: Unclassifiable statement at (1)
/home/runner/work/stdlib/stdlib/build/src/stdlib_math_all_close.f90:22:15:
```

Romendakil:
> Because `all_close` is based on `is_close`, it is actually equivalent to `all(is_close())`, so I implemented `all_close` using the `select rank` structure. As I mentioned in #438 , because Fortran's syntax is not flexible enough for arrays of uncertain dimensions, the `stdlib_stats` solution will cause the binary files compiled by the `stdlib_stats` package to be larger.
> 
> But it seems that `gcc 9.0` does not support the `select rank` structure!
> 
> ```shell
>    14 |         select rank(a)
>       |        1
> Error: Unclassifiable statement at (1)
> /home/runner/work/stdlib/stdlib/build/src/stdlib_math_all_close.f90:22:15:
> ```

Yes, according to their feature page, https://gcc.gnu.org/wiki/Fortran2018Status, select rank has been implemented for v10.0 in October 2019.

zoziha:
Because gcc9.0 used by `stdlib` does not support `select rank`, I can only use the `stdlib_stats` solution. 
In the future, if `stdlib` is no longer compatible with gcc9.0, we can use the `select rank` solution again, I believe its experience is better.

awvwgk:
> Because gcc9.0 used by `stdlib` does not support `select rank`, I can only use the `stdlib_stats` solution.
> In the future, if `stdlib` is no longer compatible with gcc9.0, we can use the `select rank` solution again, I believe its experience is better.

So far we follow GCC's recommendation of supported releases. Once GCC 12 is available, GCC 9 will most likely get dropped and we can upgrade our requirements for stdlib.

src/tests/math/test_math_all_close.f90
```diff
@@ -0,0 +1,38 @@
...
+
+    subroutine test_math_all_close_real
+
+        real :: x(4, 4), random(4, 4)
+
+        call random_number(random)
+        x = 1.0
+
+        call check(all_close(x+1.0e-11*random, x), msg="REAL: all_close(x+1.0e-11*random, x) failed.")
+        call check(all_close(x+1.0e-5 *random, x), msg="REAL: all_close(x+1.0e-5 *random, x) failed.", warn=.true.)
```
awvwgk, line 21: Why do we always warn here?

zoziha: This may be because I thought it was more interesting when I wrote it **before**. I think I should be more rigorous. I will cancel the pseudo-random numbers here and replace them with exact numbers for comparison.
And I will consider the judgment of adding NAN. (I will set this PR as **a draft** until I improve these)
Thank you for your help :) @awvwgk

zoziha: I re-corrected the test program, this problem has been resolved now.

zoziha:
I added the `equal_nan` argument to `is_close` and `all_close` to control whether NANs need to equal, and updated the corresponding test program.

src/stdlib_math_is_close.fypp
```diff
@@ -0,0 +1,43 @@
...
+        logical :: equal_nan_
+
+        equal_nan_ = optval(equal_nan, .false.)
+        
+        if (ieee_is_nan(a) .or. ieee_is_nan(b)) then
+            close = merge(.true., .false., equal_nan_ .and. ieee_is_nan(a) .and. ieee_is_nan(b))
+        else
+            close = abs(a - b) <= max( abs(optval(rel_tol, 1.0e-9_${k1}$)*max(abs(a), abs(b))), &
+                                       abs(optval(abs_tol, 0.0_${k1}$)) )
+        end if
```
zoziha, line 26: Here is the internal core business of `**_close`

awvwgk: The relative tolerance is kind of an arbitrary choice here, could we derive it from something like `sqrt(epsilon(...))` (should be around 1.0e-9 for `0.0_dp`).

zoziha: `1.0e-9` here is the default relative tolerance value from the python standard library: https://docs.python.org/3.9/library/math.html#math.isclose
```python
math.isclose(a, b, *, rel_tol=1e-09, abs_tol=0.0)
```

Using an exact value `1.0e-9` may help users easy-to-remember.

doc/specs/stdlib_math.md
```diff
@@ -388,3 +388,146 @@ program demo_math_arange
...
+    
+    y   = -3
+    NAN = sqrt(y)
+    z   = (1.0, 1.0)
+    
+    print *, all_close(z+cmplx(1.0e-11, 1.0e-11), z)     !! T
+    print *, NAN, all_close([NAN], [NAN]), all_close([NAN], [NAN], equal_nan=.true.) 
+                                                         !! NAN, F, T
+                                                
+    call check(all_close(x, [2.0, 2.0], rel_tol=1.0e-6, abs_tol=1.0e-3), &
```
jvdp1, line 528: idem here: I don't think the `check` is needed to help the reader to understand the aim/behaviour of `all_close`.

src/stdlib_math_is_close.fypp
jvdp1: Should this value depend on the kind of `real`. E.g.
```fortran
print*,epsilon(1.)
end
```
returns 
```
  1.1920929E-07
```
with ifort 2021 on my computer.

`sqrt(epsilon(..))` sounds therefore a good suggestion (or a defined value close to this result, and provided in the spec).

jvdp1 (commented):
Thank you @zoziha . Here are some comments and suggestions.

src/stdlib_math_is_close.fypp
zoziha:
```fortran
print '(*(es0.4,2x))', epsilon(1.0), sqrt(epsilon(1.0))
print '(*(es0.4,2x))', epsilon(1.0d0), sqrt(epsilon(1.0d0))
print '(*(es0.4,2x))', epsilon(1.0_16), sqrt(epsilon(1.0_16))
end
```
Results:
```sh
1.1921E-7  3.4527E-4
2.2204E-16  1.4901E-8
1.9259E-34  1.3878E-17
```
It seems that `1.0e-9` is a compromise choice, and using `sqrt(epsilon(..))` also seems good.

zoziha:
Three changes: 
- Use `test-drive` as a test library for `**_close`;
- `1.0e-9` -> `sqrt(epsilon(..))`;
- Other updates based on recommendations.

awvwgk (approved):
Thanks, looks good to me.

src/stdlib_math_is_close.fypp
```diff
@@ -0,0 +1,47 @@
...
+    use, intrinsic :: ieee_arithmetic, only: ieee_is_nan
+    implicit none
+    
+    #:for k1 in REAL_KINDS
+    real(${k1}$), parameter :: sqrt_eps_${k1}$ = sqrt(epsilon(1.0_${k1}$))
+    #:endfor
+    
+contains
+
+    #! Determines whether the values of `a` and `b` are close.
```
jvdp1, line 14:
```suggestion
    ! Determines whether the values of `a` and `b` are close.
```

jvdp1 (approved):
I only have a small edit. I'll commit it and merge this PR.

Thank you @zoziha

src/stdlib_math_is_close.fypp
jvdp1, line 14:
```suggestion
    ! Determines whether the values of `a` and `b` are close.
```

Issue #489, opened 2021-08-19
Add `angle` function for `stdlib_math`: Return the angle of the complex argument.


zoziha:
**Description**

`angle` computes the phase angle of `complex` scalar in the interval [-π,π]. The angles in `theta` are such that `z = abs(z)*exp((0.0, theta))`.

```fortran
theta = [[stdlib_math(module):angle(interface) (z)  !! z: `complex` type
```
Inspired by `forlab/angle`(MIT license), has been implemented in this branch (~~[add_angle](https://github.com/fortran-fans/stdlib/commit/5976bb8f8fcda4350baa626176d8f4898b3e6f83) and~~ [add_arg](https://github.com/fortran-fans/stdlib/commit/3cb132619b5dc83e0f65ce8e2fc04cb833fa0def)) (PR will be made later.)

**Prior Art**

1. [forlab/angle](https://github.com/keurfonluu/Forlab/blob/62b80c031fe5f283e6696be0353c13f4ba5886de/src/lib/forlab.f90#L697~L734)
2. [matlab/angle](https://ww2.mathworks.cn/help/matlab/ref/angle.html?lang=en)
3. [numpy/angle](https://numpy.org/doc/stable/reference/generated/numpy.angle.html)

arjenmarkus:
Mathematically speaking, this is the argument of the complex number -
https://mathworld.wolfram.com/ComplexArgument.html. A more appropriate
name, in my view, despite the prior art you included, is arg.

Op do 19 aug. 2021 om 13:15 schreef zoziha ***@***.***>:

nshaffer:
Linking a related issue on math utility functions where this was mentioned: #150 

There I suggested the name "phase", but "angle" is OK too. Arjen's suggestion of "arg" works as well if we are not concerned about that name clashing with general programming usage. 

Many compilers include trigonometric functions that work with degrees instead of radians, usually named "sind", "atand", etc. These are extensions, but they on the workplan for the next standard (202x), in addition to "pi" trig functions that work with angles in radians in multiples of pi. Let's keep that in mind, so that when these are standardized, we can consider "argd" and "argpi" functions as well. That is, I suggest we not follow numpy's example with the optional "deg" argument.

I do not have the time to implement this right now, but I would be happy to review it when a PR is ready.

milancurcic:
I don't work with complex numbers often, though this seems useful. I like `phase`, `angle`, and `arg` about equally,

arjenmarkus:
[...]

PR #498, opened 2021-08-27
[stdlib_math] add `arg/argd/argpi`

zoziha:
- [x] add `arg/argd/argpi` functions
- [x] new style tests.

#### Description
(idea issue discussions see #489 )
`arg` computes the phase angle (radian version) of `complex` scalar in the interval `(-π,π]`. 
The angles in `theta` are such that `z = abs(z)*exp((0.0, theta))`.
`arg` API is:
```fortran
result = [[stdlib_math(module):arg(interface)]](z)
```
Referring to the Fortran202X, I added `argd/argpi`.
image
(see [19-203r1.txt](https://j3-fortran.org/doc/year/19/19-203r1.txt) and [19-204r1.txt](https://j3-fortran.org/doc/year/19/19-204r1.txt))

**Prior Art**

1. [forlab/angle](https://github.com/keurfonluu/Forlab/blob/62b80c031fe5f283e6696be0353c13f4ba5886de/src/lib/forlab.f90#L697~L734)
2. [matlab/angle](https://ww2.mathworks.cn/help/matlab/ref/angle.html?lang=en)
3. [numpy/angle](https://numpy.org/doc/stable/reference/generated/numpy.angle.html)


src/tests/math/test_math_arg.f90
```diff
@@ -0,0 +1,41 @@
...
+    call test_math_argpi_complex
+    print *, "All tests in `test_math_arg` passed."
+
+contains
+
+    subroutine test_math_arg_complex
+        use stdlib_math, only: arg
+        use stdlib_error, only: check
+
+        call check(arg(2.0*exp((0.0, 0.5))) - 0.5 < tol, msg="arg(2.0*exp((0.0, 0.5))) failed.")
```
gareth-nx, line 15: I haven't looked closely, but here I wonder if there should be an `abs` to prevent the test passing if the function result is substantially less than the target value? Something like  `call check(abs(arg(2.0*exp((0.0, 0.5))) - 0.5) < tol, msg="arg(2.0*exp((0.0, 0.5))) failed.")`

zoziha: Yes, this is my negligence, I will correct it next time. 🤞

src/tests/math/test_math_arg.f90
```diff
@@ -0,0 +1,41 @@
...
+    call test_math_argpi_complex
+    print *, "All tests in `test_math_arg` passed."
+
+contains
+
+    subroutine test_math_arg_complex
+        use stdlib_math, only: arg
+        use stdlib_error, only: check
+
+        call check(abs(arg(2.0*exp((0.0e0, 0.5))) - 0.5) < tol, msg="arg(2.0*exp((0.0, 0.5))) failed.")
```
gareth-nx, line 15: A general comment on the tests here for all variants of these routines. 

Even though this routine is simple, I think it would be good to make "tighter" tests, for example by confirming that (for `argpi`) `z = abs(z)*exp((0.0, theta*π))` is true to a high tolerance, likely higher than 1e-03 (perhaps some multiple of `epsilon(1.0_theprecision)` 

It would also be good to check a greater range of complex numbers (in different quadrants, and considering corner cases such as the discontinuity in the degrees function). 

I also think there should be tests for other supported real kinds, with appropriate tolerances. You could auto-generate these with fypp, see example in the tests for the statistics module.

zoziha: Thanks for your advice @gareth-nx . 
I used a smaller tolerance `epsilon (1.0_sp)` to test `arg/argd/argpi`, it seems that it is not very necessary to use fypp to test `kind=dp/qp`?

fiolj:
I don't know if there is an issue holding back this PR. I've looked into the files and I think they are fine.
The only thing that perhaps may be added is the inclusion of a use case with arrays in the tests. May be replacing in `test_math.arg` the line:

```f90
  print *, argpi(2.0*exp((0.0, PI))), argpi(2.0*exp(cmplx(0.0, -PI))) !! -0.999999940,         0.999999940
```
by something like:

```f90
  print *, argpi([2.0*exp((0.0, PI)), 2.0*exp(cmplx(0.0, -PI))]) !! -0.999999940,             0.999999940
```

jvdp1:
> I don't know if there is an issue holding back this PR.

This branch has some conflicts, but most likely a lack of reviewers too. I will try to review it in the next days.

zoziha:
Thank you for your review @fiolj , I will use `testdirve` next time to refactor these test programs and solve these conficts. 
Thank you for your review @jvdp1 , in fact, the PR content `arg/angle` is very easy to implement and review, before or after I modify the test programs, you can put forward your recommended changes.

zoziha:
```
# MacOS gfortran 9~11 failed
  Starting argd-cmplx-sp ... (16/26)
   179.999985    
       ... argd-cmplx-sp [FAILED]
  Message: test_nonzero_scalar
```
Very strange, `gfortran` on `MacOS` calculated the phase angle of single-precision `cmplx(-1, 0)` and got `179.999985`! It should be `180.0`, but other platforms and other precisions are in line with expectations.

Should I increase the tolerance value, now `tol=epsilon(..)`?

fiolj:
> ```shell
> # Only 1 error: MacOS gfortran 9~11 real(sp) argd failed
>   Starting argd-cmplx-sp ... (16/26)
>    179.999985    
>        ... argd-cmplx-sp [FAILED]
>   Message: test_nonzero_scalar
> ```
> 
> Very strange, `gfortran` on `MacOS` calculated the phase angle of single-precision `cmplx(-1, 0)`(only `argd` of `real(sp)` type on MacOS) and got `179.999985`! It should be `180.0`, but other platforms and other precisions are in line with expectations.
> 
> Should I increase the tolerance value, now `tol=epsilon(..)`? Let `tol=sqrt(epsilon(..))`?

I can't see what is the problem, and cannot test on that operating system. 
I am not sure that I understand correctly how it works, but if there is a difference (even if it were smaller than tol) and we scale it by 180/PI the difference will be larger than tol. 
This, however, does not explain the observed difference in this case.

Being very simple I would try to see what values are giving in MacOs the expressions involved.
Could somebody try a program like the following? May be we can understand what is happening

```f90
!> test_argd
program test_argd
    implicit none
  !> Single precision real numbers
  integer, parameter :: sp = selected_real_kind(6)
  real(sp), parameter :: tol = epsilon(1.0_sp)
  complex(sp) :: z = (-1.0_sp, 0.0_sp)
  real(sp) :: angle
  real(kind=sp), parameter :: PI = 4.0_sp*atan(1.0_sp)

  angle = aimag(log(z))

  print *, 'PI =', PI
  print *, 'angle =', angle
  print *, 'Atan(z) =', atan(aimag(z), real(z))
  print *, 'tol =', tol
  print *, 'angle=', angle*180/PI, 'diff=', 180._sp - angle*180/PI
  !
  print *, 'obtained error =', abs(180._sp - 179.999985_sp)
  print *, 'original error =', abs(180._sp - 179.999985_sp)*PI/180._sp
end program test_argd

```

src/stdlib_math.fypp
```diff
@@ -294,6 +299,36 @@ module stdlib_math
...
+    !> ([Specification](../page/specs/stdlib_math.html#argd))
+    interface argd
+        #:for k1 in CMPLX_KINDS
+        procedure :: argd_${k1}$
+        #:endfor
+    end interface argd
+
+    !> Version: experimental
+    !>
+    !> `argpi` computes the phase angle of IEEE circular version in the interval (-1.0,1.0].
```
jvdp1, line 324: I suggest to mention also this IEEE property in the specs.

src/tests/math/test_stdlib_math.fypp
```diff
@@ -203,6 +210,69 @@ contains
 #:endif
 
     end subroutine test_clip_rqp_bounds
+    
+    #:for k1 in CMPLX_KINDS
```
jvdp1, line 214: Do you plan to only add tests for zeros arrays?

jvdp1 (commented):
LGTM. I have only a major question regarding the tests: do you plan only to test for zero values?

src/tests/math/test_stdlib_math.fypp
jvdp1: my mistake: you also test for one non-zero value.

fiolj:
I apologize for being so late with this comment, but there is a specific reason to use logarithms to calculate the angle? It seems unnecessary to evaluate the logarithm of the real part, if we are only keeping the angle. 

If I am not missing something may be we can change the code to something like:

```diff
1 file changed, 3 insertions(+), 3 deletions(-)
src/stdlib_math.fypp | 6 +++---

modified   src/stdlib_math.fypp
@@ -348,7 +348,7 @@ contains
         ${t1}$, intent(in) :: z
         real(${k1}$) :: result
 
-        result = aimag(log(z))
+        result = atan2(aimag(z), real(z))
 
     end function arg_${k1}$
 
@@ -356,7 +356,7 @@ contains
         ${t1}$, intent(in) :: z
         real(${k1}$) :: result
 
-        result = aimag(log(z))*180.0_${k1}$/PI_${k1}$
+        result = atan2(aimag(z), real(z))*180.0_${k1}$/PI_${k1}$
 
     end function argd_${k1}$
 
@@ -364,7 +364,7 @@ contains
         ${t1}$, intent(in) :: z
         real(${k1}$) :: result
 
-        result = aimag(log(z))/PI_${k1}$
+        result = atan2(aimag(z), real(z))/PI_${k1}$
 
     end function argpi_${k1}$
     #:endfor

```

zoziha:
In fact, it is also possible to use `atan2`. I did not test the specific performance. If we use `atan2`, we need to add another judgment:
```fortran
result = merge(0.0_rk, atan2(z%im, z%re), abs(z) == 0.0_rk)
! or
result = merge(0.0_rk, atan2(z%im, z%re), z%im == 0.0_rk .and. z%re == 0.0_rk)
```
```fortran
print *, log(cmplx(0.0,0.0))  !! (-Infinity,0.0000000E+00)
```
image

zoziha:
There may indeed be an unstable place here:
It can be found in the compilation check of GFortran:
```sh
app/example.f90:8:17:

     8 | print *, log(cmplx(0.0,0.0))
       | 1
Error: Complex argument of LOG at (1) cannot be zero
```
But at runtime, `log(cmplx(0.0,0.0))` can be calculated. Perhaps this is one of the reasons for using `atan2`.

zoziha:
I tried to use the `atan2` solution, but CI of `macos` still failed. It seems that it might be a bug in `gfortran` on `macos`?

doc/specs/stdlib_math.md
```diff
@@ -388,3 +388,132 @@ program demo_math_arange
...
+`z`: Shall be a `complex` scalar/array.
+This is an `intent(in)` argument.
+
+### Return value
+
+Returns the `real` type phase angle (radian version) of the `complex` argument `z`.
+
+#### Notes
+
+Although the angle of the complex number `0` is undefined, `arg((0,0))` returns the value `0`.
```
gareth-nx, line 422: I would move this into the 'Return value' section

doc/specs/stdlib_math.md
```diff
@@ -388,3 +388,132 @@ program demo_math_arange
...
+`z`: Shall be a `complex` scalar/array.
+This is an `intent(in)` argument.
+
+### Return value
+
+Returns the `real` type phase angle (degree version) of the `complex` argument `z`.
+
+#### Notes
+
+Although the angle of the complex number `0` is undefined, `argd((0,0))` returns the value `0`.
```
gareth-nx, line 465: Also suggest moving this into the `Return value` section

doc/specs/stdlib_math.md
```diff
@@ -388,3 +388,132 @@ program demo_math_arange
...
+`z`: Shall be a `complex` scalar/array.
+This is an `intent(in)` argument.
+
+### Return value
+
+Returns the `real` type phase angle (circular version) of the `complex` argument `z`.
+
+#### Notes
+
+Although the angle of the complex number `0` is undefined, `argpi((0,0))` returns the value `0`.
```
gareth-nx, line 508: Suggest moving this into the `Return value` section

src/tests/math/test_stdlib_math.fypp
gareth-nx: I think a few more tests are needed. 

For convenience these could use vector arguments, which would cover a range of cases at once and exploit the `elemental` nature of the routines.

As one suggestion - define `theta` as an angle taking 100 evenly spaced values on the unit circle. Compute the associated complex numbers, and then check that the results of `arg*` are an evenly spaced sequence with the expected range, and accurate to the expected precision.

gareth-nx (commented):
Thanks for this -- a few changes suggested here.

src/tests/math/test_stdlib_math.fypp
gareth-nx: You might consider modifying this code, which implements a test like above (double precision only, but with obvious fypp changes). 
```fortran
program arg_routines
    use stdlib_kinds, only: sp, dp, qp
    use stdlib_math, only: arg, argd, argpi
    implicit none

    integer, parameter :: N = 1000
    real(dp), parameter :: pi = acos(-1.0_dp), tol = 1000.0_dp * epsilon(1.0_dp)
    real(dp) :: theta(N), results(N), expected_results(N)
    complex(dp) :: z(N)
    integer:: i

    theta = [ ((i-1)*1.0_dp/(N+1), i = 1, N) ] ! Fraction of one circular rotation

    ! Complex numbers with evenly spaced arguments (and varying modulus)
    z = cmplx(cos(theta*2*pi), sin(theta*2*pi), kind=dp)*(theta + tol)

    ! Argd should range from 0->180, then jump to -180->0
    ! The value 
    results = argd(z)
    expected_results = theta * 360_dp
    where(expected_results > 180.0_dp) expected_results = expected_results - 360.0_dp
    if(all(abs(results - expected_results) < tol)) then
        print*, "PASS"
    else
        print*, 'FAIL'
    end if

    ! Arg should range from 0->pi, then jump to -pi->0
    results = arg(z)
    expected_results = expected_results/180.0_dp * pi
    if(all(abs(results - expected_results) < tol)) then
        print*, "PASS"
    else
        print*, 'FAIL'
    end if

    ! Argpi should range from 0->1, then jump to -1->0
    results = argpi(z)
    expected_results = expected_results/pi
    if(all(abs(results - expected_results) < tol)) then
        print*, "PASS"
    else
        print*, 'FAIL'
    end if

end program
```

zoziha:
Three changes:
- Internal implementation: `aimag(log(..)) -> atan2(..)`
- Tolerance of unit tests: `epsilon(..) -> sqrt(epsilon(..))`
- Updated testing of the array content for more angles.

gareth-nx (approved):
Looks good to me, thanks.

fiolj:
>     * Tolerance of unit tests: `epsilon(..) -> sqrt(epsilon(..))` (Since gfortran on MacOS failed, maybe a bug)

Changes look good. Minor issue: using sqrt(epsilon) for tolerance is probably too lax, especially for double precision where typical  errors (for example in the values of the tests) are aprox 1.e-14 and tolerance is about 1.e-8. Probably something like 
tol = A*epsilon(...) with values of A=500, or A=1000, would be more adequate.
Anyway,  as it is looks good and, in my opinion, could be merged

awvwgk (approved):
Looks good to me. Do you mind adding this addition to the `CHANGELOG.md` file?

zoziha:
Fine, I will add later.

zoziha:
`CHANGELOG.md` has been updated, and an unnecessary addition of `**_close` examples in `stdlib_math.md` has been removed.

awvwgk:
Thanks a lot, will merge once the CI has passed.

Issue #482, opened 2021-08-04
More matrix property checks


ghbrown:
**Description**

I'd been tossing around this proposal for a while, but was spurred into action by @zoziha's [recent issue](https://github.com/fortran-lang/stdlib/issues/476) which proposed `is_square` and `is_symmetric` checks.

This proposal suggests built-in matrix property checks for common properties like triangularity, hermiticity, etc. These kinds of checks are not difficult to write inline with a main program, but having them as a one line call with an obvious name would make any calling code shorter and more readable, and I believe such checks are common enough to warrant inclusion in the standard library.

In addition to first two proposed by zoziha, I am proposing a few extras. Below is my full list for now, but additions are welcome:
- `is_square`
- `is_symmetric`
- `is_skew_symmetric`
- `is_triangular` (upper, lower, diagonal (upper and lower), and false none results)
- `is_hermitian`
- `is_hessenberg` (upper, lower, tridiagonal (upper and lower), and false/none results)
etc.

**Prior Art**

- Numpy *does not* appear to have an `is_symmetric` or any other of the proposed checks (after light digging)
- MatLab has an `issymmetric` that uses an option to also work as a skew symmetric test, it takes a matrix and returns a bool
- Julia has `issym` ([here](http://www.jlhub.com/julia/manual/en/function/issym) and see links for other `is*` tests), `isdiag`, `istriu`, `istril`, `ishermitian`, all of which take a matrix and return a bool

**API Discussion**
`is_square`, `is_symmetric`, `is_skew_symmetric`, and `is_hermitian` will probably be four separate tests that all take a matrix and return a bool*
*The one exception I might suggest is perhaps `is_square` having a form like `is_square(A,n)` that also sets n=dim(A). Personally I have found this extremely useful when checking inputs to my own liinear algebra routines for methods or decompositions in which only square matrices are permitted. I believe this form of is_square could be snuck into the interface.

[...]

PR #494, opened 2021-08-21
Add testing module to allow better structuring of test suites

awvwgk:
*For prior discussions on this topic see https://github.com/fortran-lang/stdlib/issues/162*

This is an attempt to improve the test suite in stdlib. It uses a vendored version of [test-drive], the testing framework I wrote for fpm, named `stdlib_test` for now. It currently has the status of a testing dependency, *i.e.* the module is available in stdlib's test suite but not part of stdlib's public API, downstream projects cannot use it.

[test-drive]: https://github.com/awvwgk/test-drive
[vegetables]: https://gitlab.com/everythingfunctional/vegetables

*Why vendoring a dependency?*

- [test-drive] can be automatically fetched and included in CMake, meson and fpm projects
- embedding in the manual Makefile turned out to be difficult, I gave up on this after several attempts
- can be devendored and replaced with a proper dependency later (if make is dropped)

I'm not a fan of vendoring dependencies, but I found no better solution given the constraints of stdlib's build systems.

*Why not [vegetables]?*

- [vegetables] can be automatically fetched and included in CMake and fpm projects
- similar issue as before, we can't use it in make
- vendoring is not an option because of its dependencies

I ported the tests for `stdlib_ascii` module by replacing the `check` function of `stdlib_error` and added the test suite for the testing frame itself as well. The former uses the finalizer to check for failed tests while the latter is propagating the error back. Both will correctly report on failure, the latter approach can additionally recover and run the remaining tests as well.

I'm opening this for feedback. Let me know what you think and whether we want to adopt this testing framework.

If we agree on this approach I will move all existing tests in the new format. In a follow-up patch I would try to bundle all tests into a single test executable and allow selection of the individual suites by the framework. This simplifies the logic for setting up the tests and it be easier to adopt fypp in the test suite without over-engineering the CMake macro / make include file to generate the tests.

<details>
<summary>Output from <code>stdlib_ascii</code> tester</summary>

```
# Testing: ascii
  Starting is_alphanum_short ... (1/38)
       ... is_alphanum_short [PASSED]
  Starting is_alphanum_long ... (2/38)
       ... is_alphanum_long [PASSED]
  Starting is_alpha_short ... (3/38)
       ... is_alpha_short [PASSED]
  Starting is_alpha_long ... (4/38)
       ... is_alpha_long [PASSED]
  Starting is_lower_short ... (5/38)
       ... is_lower_short [PASSED]
  Starting is_lower_long ... (6/38)
       ... is_lower_long [PASSED]
[...]

PR #272, opened 2020-12-20
Probability Distribution and Statistical Functions -- Uniform Distribution Module

Jim-215-Fisher:
This is part of probability distribution and statistical functions for uniform distribution. In addition to original real variable functions, integer and complex variables related functions were added. The files uploaded are:

src/stdlib_stats_distribution_uniform.fypp
src/CMakeLists.txt
src/Makefile.manual
src/tests/stats/test_distribution_uniform.f90
src/tests/CMakeLists.txt
src/tests/Makefile.manual
doc/specs/index.md
doc/specs/stdlib_stats_distribution_uniform.md

PR #499, opened 2021-08-28
[stdlib_linalg] matrix property checks

ghbrown:
As discussed in #482 and #476, this adds common matrix property checks (all returning a `logical`), including:
- `is_square(A)`
- `is_diagonal(A)`, *extended for nonsquare matrices as well*
- `is_symmetric(A)`
- `is_skew_symmetric(A)`
- `is_hermitian(A)`
- `is_triangular(A,uplo)`, input `uplo` in `{'u',U','l','L'}` to decide if checking for upper or lower, *also extended to nonsquare matrices*
- `is_hessenberg(A,uplo)`, input `uplo` in `{'u',U','l','L'}` to decide if checking for upper or lower, *also extended to nonsquare matrices*

`cmake` and `make` builds appear to go fine (with a few warnings). Testing via `cmake` also gives quite a few warnings specifically related to the relevant tests, but all pass with no problem. The tests are especially long (~2000 lines worth), but I've discussed assisting in porting them over to something friendlier (`fypp` / `test-drive`/ etc.) after this PR is finished.

Documentation has not been written yet, I wanted to check in midway and make sure everything looked acceptable. Thanks.


src/tests/linalg/test_linalg.f90
```diff
@@ -564,18 +672,2055 @@ subroutine test_outer_product_int64
...
+    should_be_true_sf = is_diagonal(A_true_sf)
+    should_be_false_sf = is_diagonal(A_false_sf)
+    should_be_true_ts = is_diagonal(A_true_ts)
+    should_be_false_ts = is_diagonal(A_false_ts)
+    true_when_working_s = (should_be_true_s .and. (.not. should_be_false_s)) !combine results
+    true_when_working_sf = (should_be_true_sf .and. (.not. should_be_false_sf))
+    true_when_working_ts = (should_be_true_ts .and. (.not. should_be_false_ts))
+    true_when_working = (true_when_working_s .and. true_when_working_sf .and. true_when_working_ts)
+    call check(true_when_working, &
+         msg="true_when_working failed.",warn=warn)
```
awvwgk, line 864: Can you please make those separate tests are at least calls separate calls to `check`?

src/stdlib_linalg.fypp
```diff
@@ -108,4 +201,197 @@ contains
...
+        ${t1}$ :: zero
+        integer :: A_shape(2), m, n, o, i, j
+        zero = 0 !zero of relevant type
+        A_shape = shape(A)
+        m = A_shape(1)
+        n = A_shape(2)
+        do j = 1, n !loop over all columns
+            o = min(j-1,m) !index of row above diagonal (or last row)
+            do i = 1, o !loop over rows above diagonal
+                if (.not. (A(i,j) .eq. zero)) then
```
ivan-pi, line 230: Is there any difference between the following two checks?
```
(A(i,j) == zero)

! vs

(A(i,j) < tiny(A))
```
I'm just wondering if there is an advantage to either of them.

src/stdlib_linalg.fypp
```diff
@@ -108,4 +201,197 @@ contains
...
+      function is_triangular_${t1[0]}$${k1}$(A,uplo) result(res)
+        ${t1}$, intent(in) :: A(:,:)
+        character, intent(in) :: uplo
+        logical :: res
+        ${t1}$ :: zero
+        integer :: A_shape(2), m, n, o, i, j
+        zero = 0 !zero of relevant type
+        A_shape = shape(A)
+        m = A_shape(1)
+        n = A_shape(2)
```
ivan-pi, line 329:
```suggestion
        integer :: m, n, o, i, j
        zero = 0 !zero of relevant type
        m = size(A,1)
        n = size(A,2)
```

I think it's fine to get rid of one variable here.

src/tests/linalg/test_linalg.f90
ghbrown: I think separate tests would be overkill, as you would have to repeat all the setup and teardown many many times.

Separate `check` calls should be fine though. The reason you want separate `check` calls is to determine which case(s) fail slightly more quickly and directly, right? When you say "those", I assume you want a `check` for every `should_be_*_*`, so 6 in this case?

src/stdlib_linalg.fypp
ghbrown: In my reading [here](https://stackoverflow.com/questions/65657679/tinyx-intrinsic-function), and a very simple test, there does appear to be some difference.

`(A(i,j) == zero)` will be true when `A(i,j)` is zero or smaller in magnitude than the smallest subnormal number (underflowing, not representable with any precision, effectively zero). The smallest subnormal is accessed by `tiny(x)*epsilon(x)`, and subnormals are available so long as IEEE754 is supported (exceedingly likely).

Meanwhile, `A(i,j) < tiny(A)` may be true even if `A(i,j)` is subnormal (hence still representable and not strictly equal to zero under IEEE754). I think it is probably not a good idea to use this version of the check, since you are effectively not taking full advantage of IEEE754 (even though your matrix may be VERY nearly diagonal, etc.). This being said, I believe other contributors have expressed less desire to know whether their matrices were exactly diagonal (for example), and  more desire to know whether the matrix is nearly diagonal (off diagonals being less than some possibly user defined `tol`. Such an implementation would be a very simple extension to the methods given here, and would let users define `tol` instead of imposing `tol=tiny(A)` somewhat arbitrarily (in the scheme of floating point) on them.

Two versions which I believe are equivalent and utilize subnormal numbers:
```fypp
(A(i,j) == zero) !using my definition of zero variable in the code
!and
(abs(A(i,j)) < tiny(A)*epsilon(A)) !obviously precompute the right side
```

My final opinion (if I understood all of this correctly) would be to keep the code as written since:
- I believe it's slightly easier to understand (and requires less knowledge of floating point) while being equivalent to a version that more clearly acknowledges IEEE754
- it does not require an `abs` call (hence it may be slightly faster)

src/stdlib_linalg.fypp
ghbrown: I agree, thanks for pointing that out. I'm not going to commit your suggestion just because I have to fix it in multiple places/functions (which I'll do).

src/tests/linalg/test_linalg.f90
ghbrown: The latest commit includes this edit, so I'm going to resolve this for now. Feel free to change the status if needed.

ghbrown:
I just realized the tests for `is_hermitian` are missing completely, I will address this.

ghbrown:
The most recent commit says `is_hamiltonian` which is clearly incorrect and meant to be `is_hermitian`, but I will not change it in the interest of not force pushing.

src/stdlib_linalg.fypp
ivan-pi: Thanks for doing the research. I agree you should keep the code as already written.

We can discuss the `is_diagonal(A,tol)` (or perhaps is `is_nearly_diagonal` or `is_diagonally_dominant`) in  #482 first.

ghbrown:
I believe with the docs added all of the major tasks on my side are complete.

I was getting an error from one of the test builds, but I believe I have now fixed that. Please let me know if this is not the case.

src/stdlib_linalg.fypp
```diff
@@ -80,8 +88,93 @@ module stdlib_linalg
...
+    !!
+    !! Checks if a matrix (rank-2 array) is skew-symmetric
+    !! ([Specification](../page/specs/stdlib_linalg.html#description_7))
+    #:for k1, t1 in RCI_KINDS_TYPES
+      module procedure is_skew_symmetric_${t1[0]}$${k1}$
+    #:endfor
+  end interface is_skew_symmetric
+
+
+  ! Check for Hermiticity
```
ivan-pi, line 140:
```suggestion
  ! Check for Hermicity
```

src/stdlib_linalg.fypp
```diff
@@ -80,8 +88,93 @@ module stdlib_linalg
...
+    !!
+    !! Checks if a matrix (rank-2 array) is triangular
+    !! ([Specification](../page/specs/stdlib_linalg.html#description_9))
+    #:for k1, t1 in RCI_KINDS_TYPES
+      module procedure is_triangular_${t1[0]}$${k1}$
+    #:endfor
+  end interface is_triangular
+  
+
+  ! Check for matrix being Hessenberg
```
ivan-pi, line 164:
```suggestion
  ! Check for Hessenbergness
```
I wonder is this way to form a quality out of a noun? I'm not a native English speaker, so feel free to reject this.

src/stdlib_linalg.fypp
```diff
@@ -108,4 +201,200 @@ contains
...
+        do j = 1, n !loop over all columns
+            o = min(j-1,m) !index of row above diagonal (or last row)
+            do i = 1, o !loop over rows above diagonal
+                if (.not. (A(i,j) .eq. zero)) then
+                  res = .false.
+                  return
+                end if
+            end do
+            do i = o+2, m !loop over rows below diagonal
+                if (.not. (A(i,j) .eq. zero)) then
```
ivan-pi, line 233:
```suggestion
                if ((A(i,j) .ne. zero)) then
```

Is their some reason you prefer the explicit negation instead of using the *not equal* (`.ne.` or `/=`) operator? This also applies in other places.

I guess a good compiler will probably rewrite this to minimize operations anyways, but it's still good to clear up what the intentions are.

src/stdlib_linalg.fypp
ghbrown: Interesting, the only version I had heard was "Hermiticity", which looks common enough in [quantum mechanics](https://aip.scitation.org/doi/10.1063/1.1418246), but is very near to "hermeticity" (being hermetically sealed).
But "Hermicity" appears to be the only one with a [wiktionary page](https://en.wiktionary.org/wiki/hermicity) and seems more common in [mathematics](https://www.researchgate.net/publication/334325996_Cramer%27s_rules_for_the_system_of_quaternion_matrix_equations_with_e_-Hermicity).
I'm a bit divided, so I decided to check the [worldwide popularity](https://trends.google.com/trends/explore?date=all&q=hermicity,hermiticity) which shows that "Hermiticity" is slightly more common.
I could still be convinced either way though.

src/stdlib_linalg.fypp
ghbrown: No good reason. I always forget which language use which negated equality with symbols (between `!=, ~=, /=`, and more), so I just reverted to something I knew would work (while forgetting about `.neq.`).

I'll make the changes, thanks for catching this.

arjenmarkus:
Hermiticity comes from the name of the French mathematician Charles
Hermite, who invented the concept. Hermeticity comes from the Greek god
Hermes, who in Renaissance times and earlier was also known as Hermes
Trismegistus and in that guise he was the god/patron "saint" of the
alchemists. So if Iwere you, I'd opt for the first spelling ;), even though
the second is more common in everyday speech.

Op ma 27 sep. 2021 om 16:36 schreef Gabriel Brown ***@***.***

src/stdlib_linalg.fypp
ghbrown: I think your suggestion is very reasonable, as would be (to a lesser extent) "Hessenbergity" and probably a few others. I hesitated to use any one of them just because I could not find any established noun versions at all, where "triangularity" and friends were somewhat popular and readily searchable.
I am fine with leaving it as is, or changing it to your suggestion if we're okay being the ones to set the standard for matrix property nouns.

ghbrown:
@arjenmarkus Thanks for the etymology!
Yes, the use of "Hermeticity" here was never in question, I just brought it up as a possible point against the use of the very similar "Hermiticity". Our decision is between "Hermiticity" and "Hermicity", which I believe both stem from Hermite.

src/stdlib_linalg.fypp
ghbrown: All changes should be made. Thanks again.

src/stdlib_linalg.fypp
```diff
@@ -108,4 +201,200 @@ contains
         end do
       end function trace_${t1[0]}$${k1}$
     #:endfor
+
+
+    #:for k1, t1 in RCI_KINDS_TYPES
+      pure function is_square_${t1[0]}$${k1}$(A) result(res)
+        ${t1}$, intent(in) :: A(:,:)
+        logical :: res
+        res = (size(A,1) .eq. size(A,2))
```
milancurcic, line 210: Modern operator and remove redundant parens

```suggestion
        res = size(A,1) == size(A,2)
```

src/stdlib_linalg.fypp
```diff
@@ -108,4 +201,200 @@ contains
...
+        logical :: res
+        ${t1}$ :: zero
+        integer :: m, n, o, i, j
+        zero = 0 !zero of relevant type
+        m = size(A,1)
+        n = size(A,2)
+        do j = 1, n !loop over all columns
+            o = min(j-1,m) !index of row above diagonal (or last row)
+            do i = 1, o !loop over rows above diagonal
+                if (A(i,j) .ne. zero) then
```
milancurcic, line 227:
```suggestion
                if (A(i,j) /= zero) then
```

src/stdlib_linalg.fypp
```diff
@@ -108,4 +201,200 @@ contains
...
+                  res = .false.
+                  return
+                end if
+            end do
+            do i = o+2, m !loop over rows below diagonal
+                if (A(i,j) .ne. zero) then
```
milancurcic, line 233:
```suggestion
                if (A(i,j) /= zero) then
```

src/stdlib_linalg.fypp
```diff
@@ -108,4 +201,200 @@ contains
...
+        logical :: res
+        integer :: n, i, j
+        if (.not. is_square(A)) then
+           res = .false.
+           return !nonsquare matrices cannot be symmetric
+        end if
+        n = size(A,1) !symmetric dimension of A
+        do j = 1, n !loop over all columns
+            do i = 1, j-1 !loop over all rows above diagonal
+                if (A(i,j) .ne. A(j,i)) then
```
milancurcic, line 256:
```suggestion
                if (A(i,j) /= A(j,i)) then
```

src/stdlib_linalg.fypp
```diff
@@ -108,4 +201,200 @@ contains
...
+        logical :: res
+        integer :: n, i, j
+        if (.not. is_square(A)) then
+           res = .false.
+           return !nonsquare matrices cannot be skew-symmetric
+        end if
+        n = size(A,1) !symmetric dimension of A
+        do j = 1, n !loop over all columns
+            do i = 1, j !loop over all rows above diagonal (and diagonal)
+                if (A(i,j) .ne. -A(j,i)) then
```
milancurcic, line 279:
```suggestion
                if (A(i,j) /= -A(j,i)) then
```

src/stdlib_linalg.fypp
```diff
@@ -108,4 +201,200 @@ contains
...
+        logical :: res
+        integer :: n, i, j
+        if (.not. is_square(A)) then
+           res = .false.
+           return !nonsquare matrices cannot be Hermitian
+        end if
+        n = size(A,1) !symmetric dimension of A
+        do j = 1, n !loop over all columns
+            do i = 1, j !loop over all rows above diagonal (and diagonal)
+                if (A(i,j) .ne. conjg(A(j,i))) then
```
milancurcic, line 313:
```suggestion
                if (A(i,j) /= conjg(A(j,i))) then
```

src/stdlib_linalg.fypp
```diff
@@ -108,4 +201,200 @@ contains
...
+      function is_triangular_${t1[0]}$${k1}$(A,uplo) result(res)
+        ${t1}$, intent(in) :: A(:,:)
+        character, intent(in) :: uplo
+        logical :: res
+        ${t1}$ :: zero
+        integer :: m, n, o, i, j
+        zero = 0 !zero of relevant type
+        m = size(A,1)
+        n = size(A,2)
+        if ((uplo .eq. 'u') .or. (uplo .eq. 'U')) then !check for upper triangularity
```
milancurcic, line 334:
```suggestion
       if (uplo == 'u' .or. uplo == 'U') then !check for upper triangularity
```

Or even

```suggestion
       if (to_lower(uplo) == 'u') then !check for upper triangularity
```

where `to_lower` is imported from `stdlib_ascii`.

src/stdlib_linalg.fypp
```diff
@@ -108,4 +201,200 @@ contains
...
+          do j = 1, n !loop over all columns
+              o = min(j-1,m) !index of row above diagonal (or last row)
+              do i = o+2, m !loop over rows below diagonal
+                  if (A(i,j) .ne. zero) then
+                    res = .false.
+                    return
+                  end if
+              end do
+          end do
+        else if ((uplo .eq. 'l') .or. (uplo .eq. 'L')) then !check for lower triangularity
```
milancurcic, line 344: As above, use modern operators, and can use `to_lower` from `stdlib_ascii`.

src/stdlib_linalg.fypp
```diff
@@ -108,4 +201,200 @@ contains
...
+      function is_hessenberg_${t1[0]}$${k1}$(A,uplo) result(res)
+        ${t1}$, intent(in) :: A(:,:)
+        character, intent(in) :: uplo
+        logical :: res
+        ${t1}$ :: zero
+        integer :: m, n, o, i, j
+        zero = 0 !zero of relevant type
+        m = size(A,1)
+        n = size(A,2)
+        if ((uplo .eq. 'u') .or. (uplo .eq. 'U')) then !check for upper Hessenberg
```
milancurcic, line 373: As above.

src/stdlib_linalg.fypp
```diff
@@ -108,4 +201,200 @@ contains
...
+          do j = 1, n !loop over all columns
+              o = min(j-2,m) !index of row two above diagonal (or last row)
+              do i = o+4, m !loop over rows two or more below main diagonal
+                  if (A(i,j) .ne. zero) then
```
milancurcic, line 377:
```suggestion
                  if (A(i,j) /= zero) then
```

src/stdlib_linalg.fypp
```diff
@@ -108,4 +201,200 @@ contains
...
+                    res = .false.
+                    return
+                  end if
+              end do
+          end do
+        else if ((uplo .eq. 'l') .or. (uplo .eq. 'L')) then !check for lower triangularity
+          do j=1,n !loop over all columns
+              o = min(j-1,m) !index of row above diagonal (or last row)
+              do i=1,o !loop over rows above diagonal
+                  if (A(i,j) .ne. zero) then
```
milancurcic, line 348:
```suggestion
                  if (A(i,j) /= zero) then
```

src/stdlib_linalg.fypp
```diff
@@ -108,4 +201,200 @@ contains
...
+          do j = 1, n !loop over all columns
+              o = min(j-2,m) !index of row two above diagonal (or last row)
+              do i = o+4, m !loop over rows two or more below main diagonal
+                  if (A(i,j) .ne. zero) then
+                    res = .false.
+                    return
+                  end if
+              end do
+          end do
+        else if ((uplo .eq. 'l') .or. (uplo .eq. 'L')) then !check for lower Hessenberg
```
milancurcic, line 383: As above.

src/stdlib_linalg.fypp
```diff
@@ -108,4 +201,200 @@ contains
...
+          do j = 1, n !loop over all columns
+              o = min(j-2,m) !index of row two above diagonal (or last row)
+              do i = 1, o !loop over rows one or more above main diagonal
+                  if (A(i,j) .ne. zero) then
```
milancurcic, line 387:
```suggestion
                  if (A(i,j) /= zero) then
```

milancurcic (approved):
Thanks! I left a number of specific comments on style, namely on using the modern relational operators instead of the old ones (still standard, but deprecated by Metcalf, Reid & Cohen).

src/stdlib_linalg.fypp
ivan-pi: Let's keep it with "hermiticity" in that case. Just for the record, that wiktionary also has a [hermiticity](https://en.wiktionary.org/wiki/hermiticity#English) page.

ghbrown:
@milancurcic Thanks for the style notes. I always preferred the alphabetic relational operators since they read easily (like Python), but did not realize they had been deprecated (too bad). I will update to conform.
As for the redundant parentheses, I prefer them as they help keep "thoughts" together which can be hard when you have expressions like  `a == b .or. c == d` instead of  `(a == b) .or. (c == d)` (which is much more transparent to me).
However, your suggestion of using `to_lower()` largely avoids the need for such parentheses, and I think it's a much more elegant approach.
Thanks for the review, I'll get on these changes.

src/stdlib_linalg.fypp
ivan-pi: Personally, I'm kind of in favor of keeping the original here as it 1) keeps the independence of modules and 2) is less expensive than the call to `to_lower`. Granted, such matrix checks are typically not a performance bottleneck.

ghbrown:
@milancurcic I've gone ahead and shifted to the modern operators with the latest commit. I'm also going to close all of the related suggestions in the chain, but feel free to reopen if anything needs to be revisited.
In light of @ivan-pi 's comment about whether or not to use `to_lower()` I have left that as is for now until we come to a decision.

src/stdlib_linalg.fypp
ivan-pi: @milancurcic, do you agree to leave the character checking as is?

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,285 @@ program demo_outer_product
...
+    implicit none
+    real :: A_true(2,2), A_false(2,2)
+    logical :: res
+    A_true = reshape([1., 3., 3., 4.], shape(A_true))
+    A_false = reshape([1., 0., 3., 4.], shape(A_false))
+    res = is_symmetric(A_true)
+    !res = .true.
+    res = is_symmetric(A_false)
+    !res = .false.
+end program demo_is_symmetric
```
ivan-pi, line 325:
```suggestion
program demo_is_symmetric
    use stdlib_linalg, only: is_symmetric
    implicit none
    real :: A(2,2), B(2,2)
    logical :: res
    A = reshape([1., 3., 3., 4.], shape(A))
    B = reshape([1., 0., 3., 4.], shape(B))
    res = is_symmetric(A) ! returns .true.
    res = is_symmetric(B) ! returns .false.
end program demo_is_symmetric
```

In light of the discussion at https://github.com/fortran-lang/stdlib/pull/539#discussion_r717399771, one more nitpick would be to revise the demonstration examples. 

@ghbrown, @milancurcic How do you feel about the clarity of my proposed patch?

milancurcic: I like it!

src/stdlib_linalg.fypp
milancurcic: Yes, not using `to_lower` is fine, though I recommend using the modern relational operators here as well (this may have been updated already)

doc/specs/stdlib_linalg.md
ghbrown: I don't have a strong opinion on this one, but since other people seem to like the `return` version better I'm happy to make the change.
I'll go ahead and edit the other tests as well.

ghbrown: All examples should be updated, thanks for the feedback.

ivan-pi (approved):

ivan-pi:
This PR has now two approvals. @awvwgk, do you have time to give it a second look?

jvdp1:
This PR has somce conflicts. @ghbrown Could you resolve them please?
This PR has also 2 approvals. @awvwgk it seems that all your comments/requests were answered. Can we then merge this PR, when the conflicts will be resolved?

awvwgk (commented):
Would it be possible to separate the tests in a separate test executable? With 2k lines this is quite a thing and the test file gets quite difficult to follow and to work with (not required, however those tests have to be redone for #494 anyway by somebody).

A more robust linking to the specification would also be much appreciated, while the rest of the linalg module is doing it in a similar way, it is not really a reason to keep this practice up.

Otherwise this looks good to me.

src/stdlib_linalg.fypp
```diff
@@ -80,8 +88,93 @@ module stdlib_linalg
     #:endfor
   end interface outer_product
 
+
+  ! Check for squareness
+  interface is_square
+    !! version: experimental
+    !!
+    !! Checks if a matrix (rank-2 array) is square
+    !! ([Specification](../page/specs/stdlib_linalg.html#description_4))
```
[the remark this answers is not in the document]
ghbrown: Yes, the tests are quite unwieldy, which is why I originally inquired about using `fypp` or similar for testing, and I believe you or somebody else kindly pointed me to #494. I'm happy to separate them out however you'd like.

I'm not sure about what you mean by "a more specific link". As you noted, I've just been following the general structure inside the `stdlib_linalg` module, but if there is another module whose style/structure you'd like me to use that has more robust links etc., please let me know.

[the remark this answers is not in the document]
jvdp1: > Can we have a more specific link here instead of the forth description? Same for all other links to the specs below.

Before merging, I usually check the links, and update them if needed (I am currently doing it for #272; I am linking to the main title of each procedure). I can have a check for this PR when I am finished with #272 to help @ghbrown

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,271 @@ program demo_outer_product
...
+
+Checks if a matrix is square
+
+### Syntax
+
+`d = [[stdlib_linalg(module):is_square(interface)]](A)`
+
+### Arguments
+
+`A`: Shall be a rank-2 array
```
jvdp1, line 226: instead of A, I would suggest to use the term `array`, which is more explicit to me.
```suggestion
`d = [[stdlib_linalg(module):is_square(interface)]](array)`

### Arguments

`array`: Shall be a rank-2 array
```

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,271 @@ program demo_outer_product
...
+
+### Return value
+
+Returns a logical value that is true if the input matrix is square, and false otherwise.
```
jvdp1, line 230:
```suggestion
Returns a `logical` scalar that is `.true.` if the input matrix is square, and `.false.` otherwise.
```

src/stdlib_linalg.fypp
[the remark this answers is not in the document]
awvwgk: I'll have to rebase #494 once this is merged. Chances are good that I have to deal with a 2000 line merge conflict for this file. Not impossible to handle, just a bit more tedious.

As for the links, there are somewhat more specific links available like https://stdlib.fortran-lang.org/page/specs/stdlib_linalg.html#eye-construct-the-identity-matrix for the `eye` function. If the synopsis is not included in the title, the link will also become much shorter.

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,271 @@ program demo_outer_product
...
+
+Checks if a matrix is diagonal
+
+### Syntax
+
+`d = [[stdlib_linalg(module):is_diagonal(interface)]](A)`
+
+### Arguments
+
+`A`: Shall be a rank-2 array
```
jvdp1, line 263:
```suggestion
`d = [[stdlib_linalg(module):is_diagonal(interface)]](array)`

### Arguments

`array`: Shall be a rank-2 array
```

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,271 @@ program demo_outer_product
...
+
+### Return value
+
+Returns a logical value that is true if the input matrix is diagonal, and false otherwise.
```
jvdp1, line 267:
```suggestion
Returns a `logical` scalar that is `.true.` if the input matrix is diagonal, and `.false.` otherwise.
```

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,271 @@ program demo_outer_product
...
+
+Checks if a matrix is symmetric
+
+### Syntax
+
+`d = [[stdlib_linalg(module):is_symmetric(interface)]](A)`
+
+### Arguments
+
+`A`: Shall be a rank-2 array
```
jvdp1, line 301:
```suggestion
`d = [[stdlib_linalg(module):is_symmetric(interface)]](array)`

### Arguments

`array`: Shall be a rank-2 array
```

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,271 @@ program demo_outer_product
...
+
+### Return value
+
+Returns a logical value that is true if the input matrix is symmetric, and false otherwise.
```
jvdp1, line 305:
```suggestion
Returns a `logical` scalar that is `.true.` if the input matrix is symmetric, and `.false.` otherwise.
```

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,271 @@ program demo_outer_product
...
+
+Checks if a matrix is skew-symmetric
+
+### Syntax
+
+`d = [[stdlib_linalg(module):is_skew_symmetric(interface)]](A)`
+
+### Arguments
+
+`A`: Shall be a rank-2 array
```
jvdp1, line 338:
```suggestion
`d = [[stdlib_linalg(module):is_skew_symmetric(interface)]](A)`

### Arguments

`A`: Shall be a rank-2 array
```
```suggestion
`d = [[stdlib_linalg(module):is_skew_symmetric(interface)]](array)`

### Arguments

`array`: Shall be a rank-2 array
```

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,271 @@ program demo_outer_product
...
+
+### Return value
+
+Returns a logical value that is true if the input matrix is skew-symmetric, and false otherwise.
```
jvdp1, line 342:
```suggestion
Returns a `logical` scalar that is `.true.` if the input matrix is skew-symmetric, and `.false.` otherwise.
```

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,271 @@ program demo_outer_product
...
+
+Checks if a matrix is Hermitian
+
+### Syntax
+
+`d = [[stdlib_linalg(module):is_hermitian(interface)]](A)`
+
+### Arguments
+
+`A`: Shall be a rank-2 array
```
jvdp1, line 375:
```suggestion
`d = [[stdlib_linalg(module):is_hermitian(interface)]](array)`

### Arguments

`array`: Shall be a rank-2 array
```

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,271 @@ program demo_outer_product
...
+
+### Return value
+
+Returns a logical value that is true if the input matrix is Hermitian, and false otherwise.
```
jvdp1, line 379:
```suggestion
Returns a `logical` scalar that is `.true.` if the input matrix is Hermitian, and `.false.` otherwise.
```

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,271 @@ program demo_outer_product
...
+
+Checks if a matrix is triangular
+
+### Syntax
+
+`d = [[stdlib_linalg(module):is_triangular(interface)]](A,uplo)`
+
+### Arguments
+
+`A`: Shall be a rank-2 array
```
jvdp1, line 412:
```suggestion
`d = [[stdlib_linalg(module):is_triangular(interface)]](array,uplo)`

### Arguments

`array`: Shall be a rank-2 array
```

src/stdlib_linalg.fypp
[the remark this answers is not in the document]
ghbrown: Gotcha, I'll start working on these. Thanks.

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,271 @@ program demo_outer_product
...
+
+`uplo`: Shall be a single character from `{'u','U','l','L'}`
+
+### Return value
+
+Returns a logical value that is true if the input matrix is the type of triangular specified by `uplo` (upper or lower), and false otherwise.
+Note that the definition of triangular used here allows nonsquare matrices to be triangular.
```
jvdp1, line 419:
```suggestion
Note that the definition of triangular used in this implementation allows nonsquare matrices to be triangular.
```

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,271 @@ program demo_outer_product
...
+
+### Arguments
+
+`A`: Shall be a rank-2 array
+
+`uplo`: Shall be a single character from `{'u','U','l','L'}`
+
+### Return value
+
+Returns a logical value that is true if the input matrix is the type of triangular specified by `uplo` (upper or lower), and false otherwise.
```
jvdp1, line 418:
```suggestion
Returns a `logical` scalar that is `.true.` if the input matrix is the type of triangular specified by `uplo` (upper or lower), and `.false.` otherwise.
```

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,271 @@ program demo_outer_product
...
+
+Checks if a matrix is Hessenberg
+
+### Syntax
+
+`d = [[stdlib_linalg(module):is_hessenberg(interface)]](A,uplo)`
+
+### Arguments
+
+`A`: Shall be a rank-2 array
```
jvdp1, line 453:
```suggestion
`d = [[stdlib_linalg(module):is_hessenberg(interface)]](array,uplo)`

### Arguments

`array`: Shall be a rank-2 array
```

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,271 @@ program demo_outer_product
...
+
+`uplo`: Shall be a single character from `{'u','U','l','L'}`
+
+### Return value
+
+Returns a logical value that is true if the input matrix is the type of Hessenberg specified by `uplo` (upper or lower), and false otherwise.
```
jvdp1, line 459:
```suggestion
Returns a `logical` scalar that is `.true.` if the input matrix is the type of Hessenberg specified by `uplo` (upper or lower), and `.false.` otherwise.
```

doc/specs/stdlib_linalg.md
```diff
@@ -206,3 +206,271 @@ program demo_outer_product
...
+Note that the definition of Hessenberg used here allows nonsquare matrices to be Hessenberg.
```
jvdp1, line 460:
```suggestion
Note that the definition of Hessenberg used in this implementation allows nonsquare matrices to be Hessenberg.
```

src/stdlib_linalg.fypp
```diff
@@ -108,4 +201,200 @@ contains
...
+    #:endfor
+
+
+    #:for k1, t1 in RCI_KINDS_TYPES
+      pure function is_diagonal_${t1[0]}$${k1}$(A) result(res)
+        ${t1}$, intent(in) :: A(:,:)
+        logical :: res
+        ${t1}$ :: zero
+        integer :: m, n, o, i, j
+        zero = 0 !zero of relevant type
```
jvdp1, line 221: Why is `zero` not a parameter? Using parameter may allow some optimizations.

```suggestion
        ${t1}$, parameter :: zero = 0 !zero of relevant type
        integer :: m, n, o, i, j
```

src/stdlib_linalg.fypp
```diff
@@ -5,13 +5,21 @@ module stdlib_linalg
...
   public :: eye
   public :: trace
   public :: outer_product
+  public :: is_square
+  public :: is_diagonal
+  public :: is_symmetric
+  public :: is_skew_symmetric
+  public :: is_hermitian
+  public :: is_triangular
+  public :: is_hessenberg
```
jvdp1, line 22: Would it be wise to move the implementations of all these functions fo submodules?

src/stdlib_linalg.fypp
```diff
@@ -108,4 +201,200 @@ contains
...
+
+
+    #:for k1, t1 in (REAL_KINDS_TYPES + INT_KINDS_TYPES)
+      pure function is_hermitian_${t1[0]}$${k1}$(A) result(res)
+        ${t1}$, intent(in) :: A(:,:)
+        logical :: res
+        if (.not. is_square(A)) then
+           res = .false.
+           return !nonsquare matrices cannot be Hermitian
+        end if
```
jvdp1, line 297: This is not needed, because it is also checked by `is_symmetric`

src/stdlib_linalg.fypp
```diff
@@ -108,4 +201,200 @@ contains
...
+
+
+    #:for k1, t1 in RCI_KINDS_TYPES
+      function is_triangular_${t1[0]}$${k1}$(A,uplo) result(res)
+        ${t1}$, intent(in) :: A(:,:)
+        character, intent(in) :: uplo
+        logical :: res
+        ${t1}$ :: zero
+        integer :: m, n, o, i, j
+        zero = 0 !zero of relevant type
```
jvdp1, line 331: Same as previous: why not using `parameter`?

src/stdlib_linalg.fypp
```diff
@@ -108,4 +201,200 @@ contains
...
+
+
+    #:for k1, t1 in RCI_KINDS_TYPES
+      function is_hessenberg_${t1[0]}$${k1}$(A,uplo) result(res)
+        ${t1}$, intent(in) :: A(:,:)
+        character, intent(in) :: uplo
+        logical :: res
+        ${t1}$ :: zero
+        integer :: m, n, o, i, j
+        zero = 0 !zero of relevant type
```
jvdp1, line 370: Same as previous re: `zero`.

jvdp1 (commented):
Here are some comments. Note: I didn't check the tests, because of @awvwgk 's comment.

src/stdlib_linalg.fypp
ghbrown: I had considered that, but it seemed like it would further pollute the already cluttered `src` directory.
As long as this is fine with people, I can move them into their own submodules.

src/stdlib_linalg.fypp
ghbrown: Thanks for catching this! Fixed.

src/stdlib_linalg.fypp
ghbrown: Thanks for the suggestion, it just didn't cross my mind. I've changed all three instances.

doc/specs/stdlib_linalg.md
ghbrown: Thanks for the suggestion, these changes have been made.

doc/specs/stdlib_linalg.md
ghbrown: I agree, this is better. Changes have been made.

doc/specs/stdlib_linalg.md
ghbrown: I'm not a huge fan of this change since I think the documentation is already sufficiently (and exceedingly) clear, and having a variable (even an example variable) named after a built in structure seems like a bad idea.
I could very well be the odd one out on this though, so if other people like the proposed change I'm happy to make it.

src/stdlib_linalg.fypp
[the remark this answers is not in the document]
ghbrown: A recent commit makes the change to more specific URLs. Please let me know if this what you intended, or if I've misunderstood something. Thanks.

src/stdlib_linalg.fypp
ivan-pi: I don't have time to find the thread, but if I recall correctly our plan was that once a module (say `linalg`) reaches a certain size, we would introduce a new folder like `linalg/` in the `src/` directory. We haven't done it yet for the `stats` routines which would be another candidate. @jvdp1, what do you think?

ghbrown: A separate folder is actually what I wanted to ask for, but didn't know the overall plan or want to complicate things.
I am very much on board with having separate folders for modules.

doc/specs/stdlib_linalg.md
jvdp1: Ok. Keep it like it is now.

src/stdlib_linalg.fypp
jvdp1: > I don't have time to find the thread, but if I recall correctly our plan was that once a module (say `linalg`) reaches a certain size, we would introduce a new folder like `linalg/` in the `src/` directory. We haven't done it yet for the `stats` routines which would be another candidate. @jvdp1, what do you think?

I don't recall this discussion/thread (but it doesn't mean I didn't read it or didn't participate to it ;) ). However, I think it would be a good idea of having separate folders for modules/a set of modules related to the same topic.

For now, I would suggest to keep it as it is, or to only create submodules. We can discuss in an issue how to create folders, and if it is appropriate (also by keeping in mind that it must be compatible with fpm).

awvwgk: There is one issue to keep in mind with FORD when using subdirectories for the Fortran files, because only the basename will be visible in the documentation, like here: https://toml-f.github.io/toml-f/lists/files.html

jvdp1:
Links mentioned in the code do not work correctly. Therefore I opened a PR in @ghbrown directory (ghbrown:#1) with correct links.
@ghbrown: could you have a look at this PR, please, and if OK, to merge it?

src/stdlib_linalg.fypp
ivan-pi: Looking back, it was a single comment from Zaak Beekman in https://github.com/fortran-lang/stdlib/issues/7#issuecomment-566848401:
> As stuff gets added if src gets too cluttered you can break
out more nesting and structure. E.g. `src/strings/ascii.f90` etc. where
other string stuff gets gets grouped there.

I agree we postpone the folder changes for a new PR, which will be linked with a newly created issue.

jvdp1 (approved):
thank you @ghbrown for this PR.

jvdp1:
@ghbrown there are still 2 conflicts: `src/Makefile.manual` and `src/stdlib_linalg.fypp`. Could you solve them, please? then I think your PR will be almost ready for merging.

ghbrown:
@jvdp1 Yep, I'm eventually going to break up the tests and put the routines in submodules as requested, so I figured I would do that first before worrying about the merge conflicts.

ghbrown:
> Are the tests supposed to be in both files? Apologies if I missed part of the discussion.
No, thank you for catching this. I have made the appropriate deletions.

jvdp1:
@ghbrown could you resolve the conflicts, please?

@awvwgk There are 3 approvals for this PR.  Are all your issues with this PR resolved? Any other comments? Or could we merge it when the conflicts are resolved?

awvwgk (approved):
Looks good to me. Once the merge conflicts are resolved, this is ready to get merged.

ghbrown:
@jvdp1 I'm happy to handle the merge conflicts, but I'm not sure if I have the proper authorization (or at least know-how) this time.
The button to resolve the conflict in browser is greyed out since it's "too complex". Meanwhile, I think I would need write access to the main stdlib project to view these merge conflicts in terminal.

The other option I can think of is to simulate the merge conflicts by having an up to date version of the main project locally, then trying to merge my local branch to that, which I think should give the same merge conflicts we're seeing here. Is this the way to go? I'd appreciate a bit of guidance on this if possible, thanks!

milancurcic:
@ghbrown FWIW here's how I've been doing it, there may be other or better ways:

1. Have a local repo cloned from your fork of stdlib, at the matrix_property_checks branch
2. Add fortran-lang/stdlib as a remote repo (`git add remote upstream https://github.com/fortran-lang/stdlib`)
3. `git fetch upstream`
4. `git merge upstream/master` into matrix_property_checks
5. Resolve conflicts manually in a text editor
6. Commit changes to matrix_property_checks and push

jvdp1:
> @ghbrown FWIW here's how I've been doing it, there may be other or better ways:
> 
> 1. Have a local repo cloned from your fork of stdlib, at the matrix_property_checks branch
> 2. Add fortran-lang/stdlib as a remote repo (`git add remote upstream https://github.com/fortran-lang/stdlib`)
> 3. `git fetch upstream`
> 4. `git merge upstream/master` into matrix_property_checks
> 5. Resolve conflicts manually in a text editor
> 6. Commit changes to matrix_property_checks and push

@ghbrown Note the conflict with `test_linalg.f90` is due to the fact that `test_linalg.f90` has bee renamed `test_linalg.fypp` and that it now supports the `testdrive` procedure. Let us know if you need help with this too.

ghbrown:
@milancurcic Thanks! I had been tracking it via the `upstream` remote, but still haven't been able to materialize any merge conflicts locally.
I think my trouble is in step 4. Are you suggesting
```
git checkout matrix_property_checks
git merge upstream/master
```
will alert me of the relevant merge conflicts?

@jvdp1 Thanks for the heads up on that. I had originally wanted to use `fypp` anyway since the `f90` version was quite large, so that's good. I'll take a look at `fortran-lang/testdrive` and find existing examples in the standard library, but if I get stuck I'll reach out.

jvdp1:
> @milancurcic Thanks! I had been tracking it via the `upstream` remote, but still haven't been able to materialize any merge conflicts locally. I think my trouble is in step 4. Are you suggesting
> 
> ```
> git checkout matrix_property_checks
> git merge upstream/master
> ```
> 
> will alert me of the relevant merge conflicts?

Yes, it will. I just tested it and the two conflicts mentioned on this page are mentioned too.

> 
> @jvdp1 Thanks for the heads up on that. I had originally wanted to use `fypp` anyway since the `f90` version was quite large, so that's good. I'll take a look at `fortran-lang/testdrive` and find existing examples in the standard library, but if I get stuck I'll reach out.

The current `test_linalg.fypp` already contains `testdrive` procedures. You "just" need to transfer  yours to support it.

jvdp1:
@ghbrown did you get a chance to look at this PR? Don't hesitate to ping me if you  need help.

ghbrown:
@jvdp1 I took a quick look about a week ago when it was bumped and have a general idea of what needs to be done and slightly less of an idea of how to do it.

The real constraint for me is time, of which I'll have almost none until December 20. If it can wait that long, I'm more than happy to take care of it, as I think it would be good to see what's new with testdrive, etc. I suspect I'll be able to figure things out just fine when I actually have time to work on it, but if not I will reach out. Apologies for the delay.

jvdp1:
@ghbrown Thank you for answer. No worries. There is no rush!

ghbrown:
A bit later than planned, but with the latest commits things are hopefully much closer to completion.
Merge conflicts have been addressed and I have integrated the tests for the features using `testdrive` while taking advantage of `fypp` as much as possible.

A few outstanding concerns:
- The (ubuntu-latest, 10, make) test is failing on the manual makefiles. I assume this is a problem with my branch, but I haven't been able to extract much from the details of the failing check.
- Although it builds and tests and checks mostly okay, *I'm not convinced the tests I implemented are actually running properly (or at all)*, since even when I attempt to make them fail intentionally, I still get 100% passing. My guess is that I've done something wrong with `testdrive`, but I'm not sure.

Any suggestions on these concerns would be greatly appreciated.

awvwgk:
The manual makefile build probably fails because of a missing module dependency (those have to be added by hand, because make doesn't know how to compile Fortran without our help). Missing dependencies might or might not show up as build failure, we try to run make with as many threads as possible in the CI to ensure it almost always fails when a dependency is not specified.

Regarding test-drive, you can always run the test executable from the build directory directly to inspect the printout manually. Calling `check(error, .false.)` twice should make the test fail for sure in CTest (second call escalates the error in final procedure).

I can also have a look on those issues, if you like.

ghbrown:
@awvwgk Thanks for the suggestions. The issue with tests not failing when intended was a mistake in my `fypp`, and the manual makefile issue was a remnant of the merge conflict fixes, so both of my concerns should be addressed and fixed, with all tests passing.

A lot has changed since this PR originally got its approvals, so if further review is necessary it's no problem.

awvwgk (approved):

milancurcic:
Thanks @ghbrown and reviewers, let's merge tomorrow if there are no further objections.

jvdp1 (approved):
Before merging, could you remove the files related to the hash procedures, please? They should not appear in this PR.
Otherwise, I have no other comments. Thank you @ghbrown

ghbrown:
@jvdp1 Thanks for catching that, fixed.

Should those get added to a `.gitignore` at some point?

jvdp1:
> @jvdp1 Thanks for catching that, fixed.
> 
> Should those get added to a `.gitignore` at some point?


yes, I think it should be. But not in this PR. Could you open a new PR with those added to `.gitignore`, please?

jvdp1:
Thank you @ghbrown for this PR, and also all reviewers. I will merge it.

ghbrown:
>  yes, I think it should be. But not in this PR. Could you open a new PR with those added to .gitignore, please?

@jvdp1 Sure, no problem.

Issue #492, opened 2021-08-20
Explain that setting -DMAXRANK=4 reduces compilation time


Beliavsky:
Here is a minor suggestion for the Readme.md. It says

You can limit the maximum rank by setting -DMAXRANK=<num> in the FYPPFLAGS environment variable:

`make -f Makefile.manual FYPPFLAGS=-DMAXRANK=4`

It may not be obvious to a new user why one would do this, but this drastically reduces the compilation time compared to 

`make -f Makefile.manual`

So I suggest adding a sentence after the first command above: "Doing this substantially reduces compilation time."

gareth-nx:
Hi @Beliavsky. 

In case you'd consider using this issue to propose a simple pull request with a fix -- there is a tutorial on how to do that [here](https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request). 

My git skills are not yet sufficient to make proper contributions, but I followed that tutorial a few months ago to make a simple documentation fix, and found it relatively straightforward.

milancurcic:
Fixed by #501.

PR #501, opened 2021-08-28
Minor updates to README.md

gareth-nx:
This pull request implements the suggestion in [issue 492](https://github.com/fortran-lang/stdlib/issues/492) to explain that setting the maximum-rank reduces the compilation time with a makefile. 

It also makes clear which default PREFIX directory is used when the user doesn't specify -DCMAKE_INSTALL_PREFIX. When I first used stdlib I didn't know where this was, and found it confusing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ You can pass additional options to CMake to customize the build.
 Important options are
 
 - `-G Ninja` to use the Ninja backend instead of the default Make backend. Other build backends are available with a similar syntax.
-- `-DCMAKE_INSTALL_PREFIX` is used to provide the install location for the library.
+- `-DCMAKE_INSTALL_PREFIX` is used to provide the install location for the library. If not provided, this defaults to `stdlib_install` in the `stdlib` directory.
 - `-DCMAKE_MAXIMUM_RANK` the maximum array rank procedures should be generated for.
   The default value is chosen as 4.
   The maximum is 15 for Fortran 2003 compliant compilers, otherwise 7 for compilers not supporting Fortran 2003 completely yet.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -125,7 +125,7 @@ You can pass additional options to CMake to customize the build.
 Important options are
 
 - `-G Ninja` to use the Ninja backend instead of the default Make backend. Other build backends are available with a similar syntax.
-- `-DCMAKE_INSTALL_PREFIX` is used to provide the install location for the library.
+- `-DCMAKE_INSTALL_PREFIX` is used to provide the install location for the library. If not provided, this defaults to `stdlib_install` in the `stdlib` directory.
```
awvwgk, line 128: In fact the default is platform dependent, *e.g.* `/usr/local` on Unix and `c:/Program Files/${PROJECT_NAME}` on Windows. See https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html

awvwgk (changes requested):

README.md
gareth-nx: Oops, thank you

awvwgk (approved):

awvwgk:
@gareth-nx you might want to check your local git settings and set your [commit email address](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-github-user-account/managing-email-preferences/setting-your-commit-email-address). There is a no-reply address associated to your GitHub account which is usually save to use without leaking your actual email address. We can also squash your PR on merge and GitHub will automatically fix up the committer data.

milancurcic (approved):
Looks good, thank you!

Issue #181, opened 2020-05-06
Documentation style for stdlib


vansnyder:
The stdlib should be packaged in modules, each one a coherent set of related procedures (and types and constants). These should ultimately all be described in a consistent style, as (or at least as if) an optional part of the standard. Procedures should be described in the same format as in subclause 16.9 of part 1 of the standard. Constants and opaque types can be described as in subclause 16.10. Types that have public components and bindings could be described by a type definition that shows only type parameters, and the public components and bindings. Type-bound procedures should be described as in subclause 16.9. I prepared my proposal for special functions in this format.

jvdp1:
Thank you @vansnyder for your comments.

> Procedures should be described in the same format as in subclause 16.9 of part 1 of the standard.

Good point. The current format ([see example here for stdlib_experimental_stats](https://github.com/fortran-lang/stdlib/blob/master/src/stdlib_experimental_stats.md) ) should probably slightly modified if we want to follow exactly the same format as in the standard.

milancurcic:
I think this is a good plan for the specification docs. It would also ease the transition to the standard if and when it's proposed. Then we could evolve a specification doc -> J3 paper.

We'll also need a separate set of user friendly docs but we're not there yet.

jacobwilliams:
I must admit I am not a huge fan of the format required by the standard. The priority should be the user-facing documentation, which needs to be dynamic, web based, should have syntax highlighting, links, etc. Nobody is going to be printing this stuff out in book form, so we shouldn't be limited to that format.

milancurcic:
I don't like the standard format either, but both are important. One of the original motivations behind stdlib is to implement and evaluate in practice functions before proposing them to J3. Documenting the spec in the standard format would save us from a lot of work later. Another question is whether there is much value to have a built-in function if there is one in a stable and portable stdlib (I don't think so).

I agree that user-docs are higher priority. But we haven't done any work toward that, except some docstrings inside the code itself, which I don't think is the best solution either.

Jacob, can you help with the user-docs?

jvdp1:
I agree that the standard format is too minimalist and that the user-docs is more important.
But it would be nice if we could extract what is needed for the standard directly from `stdlib` user-docs, and I think that the current format (that contains e.g. more verbose examples) is quite close to allow it.

zbeekman:
[...]

PR #503, opened 2021-08-31
refactor documentation regarding consistency

JHenneberg:
Could someone link the file to where the documentation standard is desribed?

I was reading through some documents in `doc/specs` and realised differenced between those documents.
Mostly the layout looks like following:

```markdown
---
title: <MODULE_NAME>
---

# The `<MODULE_NAME>` module

[TOC]

## Introduction

## <OPTIONAL_SECTIONS>

## Procedures and methods provided

### `<PROCEDURE_NAME>` - <SHORT_DESCRIPTION>

#### Status

#### Description

#### Syntax

#### Class

#### Argument

#### Result Value

#### Example
```

`stdlib_ascii.md` and a lot of other modules are following this convention, but some do not. Sometimes the order of the subsections describing the procedures are different or some subsections are missing. In other documents the hierarchy level is changed. Do you feel there is a need to address this issues? Should this be continued to be discussed in #181 ?

### DIFF
--- a/doc/specs/stdlib_string_type.md
+++ b/doc/specs/stdlib_string_type.md
@@ -1130,7 +1130,7 @@ end program demo
 
 #### Description
 
-Returns a new string_type instance which holds the lowercase version of the 
+Returns a new string_type instance which holds the lowercase version of the
 character sequence hold by the input string.
 
 #### Syntax
@@ -1176,7 +1176,7 @@ end program demo
 
 #### Description
 
-Returns a new string_type instance which holds the uppercase version of the 
+Returns a new string_type instance which holds the uppercase version of the
 character sequence hold by the input string.
 
 #### Syntax
@@ -1222,12 +1222,12 @@ end program demo
 
 #### Description
 
-Returns a new string_type instance which holds the titlecase version 
-of the character sequence hold by the input string.  
-Title case: First character of every word in the sentence is converted to 
-uppercase and the rest of the characters are converted to lowercase.  
-A word is a contiguous sequence of character(s) which consists of alphabetical 
-character(s) and numeral(s) only and doesn't exclude any alphabetical character 
+Returns a new string_type instance which holds the titlecase version
+of the character sequence hold by the input string.
+Title case: First character of every word in the sentence is converted to
+uppercase and the rest of the characters are converted to lowercase.
+A word is a contiguous sequence of character(s) which consists of alphabetical
+character(s) and numeral(s) only and doesn't exclude any alphabetical character
 or numeral present next to either of its 2 ends.
 
 #### Syntax
@@ -1272,10 +1272,10 @@ end program demo_to_title
 
 #### Description
 
-Returns a new string_type instance which holds the sentencecase 
-version of the character sequence hold by the input string.  
-Sentencecase version: The first alphabetical character of the input character sequence 
-is transformed to uppercase unless it follows a numeral and the rest of the 
+Returns a new string_type instance which holds the sentencecase
+version of the character sequence hold by the input string.
+Sentencecase version: The first alphabetical character of the input character sequence
+is transformed to uppercase unless it follows a numeral and the rest of the
 characters in the sequence are transformed to lowercase.
 
 #### Syntax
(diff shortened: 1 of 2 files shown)


milancurcic:
Thanks, @JHenneberg and good question. I don't think we have a template for this, and I think we should. @jvdp1 @ivan-pi what did you use as reference when creating the first spec docs in stdlib?

jvdp1:
Hi. I used the Fortran standard doc

Le mar. 31 août 2021 à 18:40, Milan Curcic ***@***.***> a
écrit :

milancurcic (approved):
The changes look good to me, thanks @JHenneberg.

awvwgk (approved):

milancurcic:
@JHenneberg I see WIP in the title--would you like to work more on this PR or should we merge it?

JHenneberg:
I wanted to work more on it, but now that there is a discussion going on about the spec_template.md it can be merged, because depending what changes might come it could be double the work.

Aman-Godara:
How about we add 2 small sections in documentation with the name `troubleshooting` and `caution`?
We know not everyone has the patience to go through the whole documentation only to look for a particular function they are looking for. If we add a small troubleshooting section and caution section (they are not needed for all functions though but for some) as per the use case it will make documentation way easier to reference to.

For example, if a person is using `find` function of `stdlib_strings` and got stuck, well then probably it is because the output of the function is 0 (suggesting that the input substring was NOT FOUND) which can be added in `caution` section.

or if a person is using `slice` function and got weird outputs then probably because of the person hasn't specified `stride` argument which then automatically got deduced by the function leading to smart behaviour (which the user doesn't want and wasn't aware of) so this can be added in `troubleshooting` section for `slice` function.

Also, sometimes Fortan throws bad errors, so the documentation can provided some possible ways in which a user might have added errors when using a function.
For example, if you give integer indexes to a stringlist, it gives 'binding not found error' which doesn't say that use `fidx` and `bidx` instead of `integer index`.

Aman-Godara:
Also, removing those trailing double spaces which looked useless make .md file (markdown) look different:

image

taken from https://stdlib.fortran-lang.org/page/specs/stdlib_string_type.html#to_title-function.

in .md files (markdown):
`Double spaces + Enter` means new line. `Double Enter` also means new line.
`space + Enter` DOESN'T mean new line and means a single space only.
`Enter` means a single space
`more than one space` means a single space only


Previously with double spaces this used to look like this:

image

To try out one can use https://www.markdowntutorial.com or some other website but the point is that markdown files look different when viewed on GitHub from what they do when we are editing them on some editor like VS Code.

Those single spaces before Enter, I never knew, turned out to be redundant.

doc/specs/stdlib_string_type.md
```diff
@@ -1350,7 +1350,7 @@ program demo
   use stdlib_string_type
   implicit none
   type(string_type) :: string, reverse_string
-
```
Aman-Godara, line 1353: These `double spaces` are redundant because the two `Enters` did the job here of creating a new line

JHenneberg:
> Also, removing those trailing double spaces which looked useless make .md file (markdown) look different:
> 
> image
> 
> taken from https://stdlib.fortran-lang.org/page/specs/stdlib_string_type.html#to_title-function.
> 
> in .md files (markdown):
> `Double spaces + Enter` means new line. `Double Enter` also means new line.
> `space + Enter` DOESN'T mean new line and means a single space only.
> `Enter` means a single space
> `more than one space` means a single space only
> 
> Previously with double spaces this used to look like this:
> 
> image
> 
> taken from [GitHub](https://github.com/Aman-Godara/stdlib/blob/revival_string_list/doc/specs/stdlib_string_type.md#to_title-function) and NOT from https://stdlib.fortran-lang.org
> 
> To try out one can use https://www.markdowntutorial.com or some other website but the point is that markdown files look different when viewed on GitHub from what they do when we are editing them on some editor like VS Code.
> 
> Those single spaces before Enter, I never knew, turned out to be redundant.

Ok Interesting. Did not know about this 'double space' feature, but I guess its better because more intuitive to do this similar to LaTeX using 'double Enter' for new line.

PR #506, opened 2021-09-01
Portable savetxt and loadtxt

### LINKED ISSUE #505: savetxt and loadtxt are not portable
While porting stdlib_io tests to the new testing framework (PR in progress in #494), it occurred to me that the `loadtxt` tests were not testing values, but merely loading the data and printing them to stdout. When I re-wrote the tests to compare the values, though the tests passed for GFortran, they failed for ifort (see https://github.com/awvwgk/fortran-stdlib/pull/87). 

We should be able to round-trip the data without loss of information, i.e.

```
call random_number(a)
call savetxt('data.txt', a)
call loadtxt('data.txt', b)
print *, all(a == b) ! should print T
```

Currently it seems that doesn't generally work because both `loadtxt` and `savetxt` use list-directed input and output, which is not generally portable. We should use specific formats with field widths wide enough so that we can both round-trip the data and so that the files are in the same format between different compilers.

milancurcic:
This PR implements specific formats to use in `savetxt` and `loadtxt`, and improves the tests to compare exact values and not just approximates.

It depends on tests in #494, so this should be rebased to master after #494 is merged. Or is there a better strategy for this, @awvwgk? The tests in https://github.com/awvwgk/fortran-stdlib/pull/87 need this fix to pass. Should this fix go there instead?

Fixes #505.


awvwgk (approved):
Thanks for sharing. This looks like a good solution.

Note: Please *do not merge* this PR into stdlib master after review. This is part of https://github.com/fortran-lang/stdlib/pull/494 and will be included there.

milancurcic:
Sounds good. I'll close this PR and open against your branch in #494. The latest 2 commits expand the round-trip tests for tiny and huge numbers. These tests revealed a few more issues--one leading to a small fix in `number_of_rows()`, and another requiring broader width and exponent fields.

milancurcic:
Closing in favor of https://github.com/awvwgk/fortran-stdlib/pull/90.

certik:
I am struggling to review this, I see 51 changed files... That's too much.

Can we do one clean PR to port to a new test framework system (but do no other changes), and then send clean PRs with changes unrelated to the testsuite change?

awvwgk:
@certik I added a comment for the recommended review process in https://github.com/fortran-lang/stdlib/pull/494#issuecomment-913110156.

PR #508, opened 2021-09-01
github-ci: add fpm support

zoziha:
- [x] add `fpm-deployment` ci;
- [x] create a `stdlib-fpm` branch. (github-ci will do this job, see https://github.com/fortran-lang/stdlib/pull/508#issuecomment-913443862)
- [x] add `./ci/fpm.toml` and update `README.md`

#### Description
We have this urgent need for fortran-stdlib. We hope that `stdlib` and `fpm` can interact, and `stdlib-fpm` will be placed in the `fortran-lang/stdlib` repository (branch).
The CI scheme here comes from @LKedward (https://github.com/LKedward/stdlib-fpm). I have made some modifications to adapt to specific technical requirements. I am not very familiar with CI. I need to admit that the CI scheme this time has shortcomings, but I believe we can iterate and update it.

#### User efficiency
In order to improve the efficiency of users of `stdlib`, we can set the `stdlib-fpm` branch for `stdlib`, allowing users to add the following reference in their projects:
```bash
[dependencies]
stdlib = { git = "https://github.com/fortran-lang/stdlib", branch = "stdlib-fpm" }
```


zoziha:
I checked again, the entire CI process seems to be no problem, except that the `fpm test (release)` task takes a long time.
If you have any questions, you can give me feedback and I will quickly correct it.
In addition, once we think there is no problem with this CI, that we can pass it. We need to create a branch, `stdlib-fpm`, to support fpm, see [fortran-fans : stdlib-fpm](https://github.com/fortran-fans/stdlib/tree/stdlib-fpm) for **my** stdlib-fpm branch.

awvwgk:
Depending on the commit action it will create the branch all by itself, the one used for deploying fpm's documentation will create the branch automatically for example (https://github.com/marketplace/actions/deploy-to-github-pages).

awvwgk:
Also, shouldn't the deployment only happen on push actions for the default branch rather than on PRs as well?

For running the action on a PR we don't have to checkout the stdlib-fpm branch at all but just deploy the converted project to a new directory and test it. The checkout is only required for updating the branch on push and this shouldn't be required when running on PRs.

zoziha:
> Depending on the commit action it will create the branch all by itself, the one used for deploying fpm's documentation will create the branch automatically for example (https://github.com/marketplace/actions/deploy-to-github-pages).

I think stdlib-fpm needs the existence of the `fpm.toml` file. I think there are two ways that I can think of:
1. Write `fpm.toml` in `stdlib-master`. When CI is running, copy the `fpm.toml` file in `stdlib-master` to `stdlib-fpm`. In this case, `fpm.toml` needs to be maintained in **`stdlib-master`**, although I think `fpm.toml` may not need to be updated frequently.
2. Manually create a `stdlib-fpm` branch, and manually submit `fpm.toml` to `stdlib-fpm`. When CI is running, it is also possible. `fpm.toml` needs to be maintained in **`stdlib-fpm`** in this case.

Thank you for your suggestion @awvwgk , I think both methods are fine.
I prefer the first option. If everyone agrees, I will add a file: `fpm.toml` to `stdlib-master`(`./ci/fpm.toml`).

> Also, shouldn't the deployment only happen on push actions for the default branch rather than on PRs as well?

```yml
# When a PR occurs, the fpm package of `stdlib` will be generated.
on: [push, pull_request]
```

Yes, both PR and push will activate `fpm-deployment.yml`, and I will remove the incorrect comment.

zoziha:
I added `fpm.toml` and improved `fpm-deployment` CI. I have passed the test on my branch (see https://github.com/fortran-fans/stdlib/runs/3524567814). Once this PR passes, `fpm-deployment` CI should automatically create a `stdlib-fpm` branch for `fortran-lang/stdlib` (see https://github.com/fortran-lang/stdlib/pull/508#issuecomment-913443862). (Now it is failing, because github-action does not currently have permission)

.github/workflows/fpm-deployment.yml
```diff
@@ -0,0 +1,50 @@
...
+    - name: Run fpm test ⚙
+      run: |
+        cp -r stdlib-fpm stdlib-fpm-test
+        cd stdlib-fpm-test
+        fpm test
+        fpm test --profile release
+
+    # Update and deploy the f90 files generated by github-ci to the `stdlib-fpm` branch.
+    - name: Deploy 🚀
+      uses: JamesIves/github-pages-deploy-action@4.1.5
```
LKedward, line 47: In the [`stdlib-fpm`](https://github.com/LKedward/stdlib-fpm) workflow, I use the following conditional to prevent deployment for PRs. (This should fix the CI error for this PR)

```suggestion
      uses: JamesIves/github-pages-deploy-action@4.1.5
      if: github.event_name != 'pull_request'
```

LKedward:
Great stuff @zoziha - thanks for implementing this!

The CI is failing currently, because it is trying to deploy the stdlib-fpm package but we don't want to deploy for PRs, only on pushes to master - see my suggestion above.

zoziha:
@awvwgk @LKedward Thank you very much for your answers, now I understand what you mean.
I resubmitted a `git push`, hope there is no problem.

LKedward (approved):

.github/workflows/fpm-deployment.yml
```diff
@@ -0,0 +1,51 @@
...
+
+    - name: Generate stdlib-fpm package 🔧
+      run: |
+        sh ./ci/fpm-deployment.sh
+
+    - name: Install GFortran
+      run: |
+        sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-${{ env.GCC_V }} 100 \
+        --slave /usr/bin/gfortran gfortran /usr/bin/gfortran-${{ env.GCC_V }} \
+        --slave /usr/bingcov gcov /usr/bin/gcov-${{ env.GCC_V }}
```
awvwgk, line 31:
```suggestion
        --slave /usr/bin/gcov gcov /usr/bin/gcov-${{ env.GCC_V }}
```

ci/fpm-deployment.sh
```diff
@@ -0,0 +1,31 @@
...
+
+DESTDIR="stdlib-fpm"
+
+FYFLAGS="-DMAXRANK=4"
+
+mkdir -p $DESTDIR/src
+mkdir -p $DESTDIR/test
+
+# Preprocess stdlib sources
+ls src/*.fypp | cut -f1 -d. | xargs -i{} fypp {}.fypp {}.f90 $FYFLAGS
```
awvwgk, line 14: Don't do this, artifacts of this script must never exist in the source directory they were generated from as they possibly interfere with future runs. Also, `-i` is deprecated in `xargs`. Instead, directly write all artifacts in the destination directory.

I'll send a PR against your branch shortly with an updated version of the script.

awvwgk (dismissed):

ci/fpm-deployment.sh
awvwgk: See https://github.com/fortran-fans/stdlib/pull/1

LKedward:
Thanks for reviewing and updating Sebastian @awvwgk. It would be good to get another reviewer on this to move it forward.

awvwgk (approved):

LKedward (approved):
Latest updates to `fpm-deployment.sh` look good to me :+1:

zoziha:
So cool 😎😎, we can use `stdlib` in `fpm` project in the future, and `stdlib` will interact with `fpm` from now on! 🎉

README.md
```diff
@@ -176,7 +177,21 @@ You can limit the maximum rank by setting ``-DMAXRANK=<num>`` in the ``FYPPFLAGS
 make -f Makefile.manual FYPPFLAGS=-DMAXRANK=4
 ```
 
+### Build with [fortran-lang/fpm](https://github.com/fortran-lang/fpm)
 
+Fortran Package Manager (fpm) is a great package manager and build system for Fortran.
```
jvdp1, line 182:
```suggestion
Fortran Package Manager (fpm) is a package manager and build system for Fortran.   
```
I don't like words like "great". I suggest to remove it.

README.md
```diff
@@ -176,7 +177,21 @@ You can limit the maximum rank by setting ``-DMAXRANK=<num>`` in the ``FYPPFLAGS
 make -f Makefile.manual FYPPFLAGS=-DMAXRANK=4
 ```
 
+### Build with [fortran-lang/fpm](https://github.com/fortran-lang/fpm)
 
+Fortran Package Manager (fpm) is a great package manager and build system for Fortran.   
+You can build using provided `fpm.toml`:
```
jvdp1, line 183:
```suggestion
You can build `stdlib` using provided `fpm.toml`:
```

README.md
```diff
@@ -176,7 +177,21 @@ You can limit the maximum rank by setting ``-DMAXRANK=<num>`` in the ``FYPPFLAGS
...
+
+```sh
+git checkout stdlib-fpm
+fpm build --profile release
+```
+
+To use `stdlib` within your `fpm` project, add the following to your `fpm.toml` file:
```
jvdp1, line 190:
```suggestion
To use `stdlib` within your `fpm` project, add the following lines to your `fpm.toml` file:
```

jvdp1 (approved):
Great and highly needed addition. It will benefit to both stdlib and fpm IMO. Thank you @zoziha .
I approve this PR, pending minor revisions.

zoziha:
I modified the corresponding content of `README.md`. @jvdp1 😉

urbanjost:
Seems like stdlib should go onto the fpm registry but I do not see it there (yet?)

jvdp1:
With 3 approvals, I will merge this PR. Thank you @zoziha

PR #514, opened 2021-09-06
Delete stringlist & move subroutine

Aman-Godara:
Delete function to `stdlib_stringlist`

Tasks:

- [x] Add delete function (`impure`) to stringlist
- [ ] Add test cases for the same
- [ ] Documentation


Aman-Godara:
`delete` function vs `remove` function?

`delete` functions returns the deleted element.
`remove` function does same job as to `delete` function but doesn't return the deleted element.

Any suggestions on better naming convention? Should I rather call `delete` as `pop` and think of a better name for `remove`?

epagone:
This is highly subjective but FWIW, I would prefer `pop` for the functions that return the deleted element, simply because I imagine that if I "pop" something out I still have it with me.

For the function that does **not** return the deleted element I am fine with both `remove` or `delete`, with a slight preference for `delete`.

Did you have any time to see what are the choices of other languages for similar functions?

Aman-Godara:
Python uses `pop` (returns deleted item) and `remove` (no return)
Java doesn't have any function equivalent of `pop` (or atleast I couldn't find one), it has `remove` function (no return).

`pop` function is used whenever deleted item is returned, otherwise `remove` and `delete` functions both are used by programming languages. Though, I could find only one occurrence where `delete` was used to remove elements.
Also, some languages uses `pop()` (with no arguments) as a way to allow a user to make a `list` behave like a `stack` or a `queue`.

arjenmarkus:
How about get_and_remove or get_remove? That makes the function completely
clear, albeit via a more verbose name.

Op di 7 sep. 2021 om 15:29 schreef Aman Godara ***@***.***>:

awvwgk:
I think `pop` is a reasonable name to a procedure to remove and return the removed element. For only deleting elements, `delete`, `remove` or `drop` might be suitable. The names `delete` and `destroy` could be associated with a deconstructor of the entire list.

epagone:
Based on my experience with past discussions, we are trying to provide a familiar interface to the user of `stdlib` following well-established conventions from other languages (like python, for example). @arjenmarkus suggests more "Fortranic" names (that I like too) but they might be inconsistent with other parts of `stdlib`.

Based on the other suggestions, my preference is
- `pop`: removes and returns the removed element
-  `drop`: removes an element without returning it
- `destroy`: de-constructs the list and frees the memory

I do not have strong opinions on this, though.

Aman-Godara:
I agree with `destroy` (please note that we have a function named `clear` in `stdlib_stringlist` which resets the list and NOT destroys it). Checkout this example from **python** to understand the difference:

image

Aman-Godara:
I added a basic range feature to `pop` and `drop` which will remove all strings at indexes in the interval [`first`, `last`].
Out of bounds indexes will be ignored. There is no concept of `stride` currently for `pop` and `drop`. We can discuss the behaviour of non-basic range feature for `pop` and `drop`.

I think atleast a basic range feature should be added to `get` function as well, it will also improve the way `pop_engine` function is written.

epagone:
Thank you.

I personally don't like that out of bounds indexes are ignored: I'd like to be notified (at least with a warning, if not by an error) in these cases. Based on my experience, out of bounds indexes are the consequence of upstream bugs that would be difficult to identify, if ignored. However, I believe that a choice in this sense might have been already made during the initial implementation of `stringlist` for other methods.

epagone:
I am under the impression that adding the `get` function (although, as I said, it's a good idea) is out of scope for this PR. Am I right?

src/stdlib_stringlist_type.f90
```diff
@@ -710,26 +736,182 @@ end subroutine insert_before_stringarray_int_impl
...
+
+    !> Version: experimental
+    !>
+    !> Removes strings present at indexes in interval ['first', 'last']
+    !> Stores captured popped strings in array 'capture_popped'
+    !> No return
+    pure subroutine pop_drop_engine( list, first, last, capture_popped )
+        class(stringlist_type), intent(inout)                   :: list
+        type(stringlist_index_type), intent(in)                 :: first, last
+        type(string_type), allocatable, intent(out), optional   :: capture_popped(:)
```
epagone, line 815:
```suggestion
        type(string_type), allocatable, intent(out), optional   :: popped_string(:)
```
Wouldn't be more clear to rename this `popped_string`?

src/stdlib_stringlist_type.f90
```diff
@@ -710,26 +736,182 @@ end subroutine insert_before_stringarray_int_impl
...
+    end subroutine pop_drop_engine
+
+    !> Version: experimental
+    !>
+    !> Removes the string present at stringlist_index 'idx' in stringlist 'list'
+    !> Returns the removed string
+    function pop_idx_impl( list, idx )
+        class(stringlist_type), intent(inout)           :: list
+        type(stringlist_index_type), intent(in)         :: idx
+        type(string_type)                               :: pop_idx_impl
```
epagone, line 865: What `_impl` means here (and in other places)? It is a bit obscure to me.

Aman-Godara: `impl` means implementation.

Aman-Godara:
> I am under the impression that adding the `get` function (although, as I said, it's a good idea) is out of scope for this PR. Am I right?

Yeah, that's right. Let me rename PR.

src/stdlib_stringlist_type.f90
epagone: I believe it's redundant and can be cause of potential confusion. I would suggest to remove the suffix.

Aman-Godara: I will have to somehow differentiate ti from `pop_idx`.

epagone: I think the short form of the type bound procedure is not necessary in this case: I would only use `pop_idx` or `pop_index`. The same goes with all the other TBPs with the `_impl` suffix.

Aman-Godara: I didn't understand it @epagone, can you please elaborate on this? May be you can **insert a suggestion** and then I will take it forward.

epagone: Ok, sure :+1:

src/stdlib_stringlist_type.f90
```diff
@@ -84,8 +84,20 @@ module stdlib_stringlist_type
...
-        generic, public   :: get                    => get_string_idx
+        procedure         :: get_idx                        =>  get_idx_impl
+        procedure         :: get_range_idx                  =>  get_range_idx_impl
+        generic, public   :: get                            =>  get_idx,                        &
+                                                                get_range_idx
+        
+        procedure         :: pop_idx                        =>  pop_idx_impl
+        procedure         :: pop_range_idx                  =>  pop_range_idx_impl
+        generic, public   :: pop                            =>  pop_idx,                        &
+                                                                pop_range_idx
```
epagone, line 95:
```suggestion
        procedure         :: pop_index               ! or `pop_idx`, if you wish
        procedure         :: pop_range_index         ! or `pop_range_idx`, if you wish
        generic, public   :: pop                =>  pop_index, pop_range_index
```

In the executable statements you can use `pop_index` or `pop_range_index`.

The same can be done for the other TBPs with the `_impl` suffix.

The rationale of the (optional) short version of TBPs is to help you create shorter, more convenient names, not to create additional naming problems :wink: .

Please let me know if it's still unclear.

Aman-Godara: Thanks, got it. Wish I had known this before 😅.

Aman-Godara:
This PR is blocked by: #552

Issue #504, opened 2021-08-31
Create a spec document template


milancurcic:
Currently we don't have a Markdown template for spec documents. We should have one. Benefits include:

* Makes it easier for contributors to write spec docs.
* Ensures consistency if followed.

This could be in a file SPEC_TEMPLATE.md or similar, placed in the root directory, and linked to from WORKFLOW.md and CONTRIBUTING.md. 

Inspired by @JHenneberg in #503.

JHenneberg:
Short summary of the more or less status quo:
* General structure of the spec file
  * For now mostly like this:
```markdown
---
title: <MODULE_NAME>
---

# The `<MODULE_NAME>` module

[TOC]

## Introduction

## <OPTIONAL_SECTIONS>

## Procedures and methods provided

### `<PROCEDURE_NAME>` - <SHORT_DESCRIPTION>
```
* How should procedures be documented:
  * For now mostly according to the Fortran Standard [ISO 1539-1 2018](https://j3-fortran.org/doc/year/18/18-007r1.pdf) was followed and extended by the subsection `Status` and `Syntax` but missing the section `Result Characteristics`:
```markdown
### `<PROCEDURE_NAME>` - <SHORT_DESCRIPTION>

#### Status

[<STABLE>, <EXPERIMENTAL>]

#### Description

#### Syntax

#### Class

[<IMPURE>, <PURE>, <ELEMENTAL>,]

#### Arguments

#### Result Value

#### Example
```

milancurcic:
Should Class be:

```
#### Class

[<IMPURE>, <PURE>, <ELEMENTAL>,] [<FUNCTION> <SUBROUTINE>]
```

?

JHenneberg:
@milancurcic I think that would be usefull, even though there is the section `#### Syntax` it would be redundant.

The `Syntax` section is not needed in the Fortran Std 2018 files because the complete function/procedure is written in the headline.

Fortran Std or [GCC](https://gcc.gnu.org/onlinedocs/gfortran/ICHAR.html) Style:

-----------------------------------------------------------------------------------------------------------

### `ICHAR(C, [, KIND])`

#### Description

Returns the code for the character in the first character position of `C` in the system’s native character set.

#### Class

Elemental function.

#### Arguments

* `C` : shall be of type character and of length one. Its value shall be that of a character capable of representation in the processor.
* `KIND` (optional): shall be a scalar integer constant expression.

#### Result Characteristics

[...]

PR #517, opened 2021-09-07
adding SPEC_TEMPLATE.md #504

JHenneberg:


CONTRIBUTING.md
```diff
@@ -31,6 +31,17 @@ You are welcome to propose changes to the style guide by
 [opening an issue](https://github.com/fortran-lang/stdlib/issues/new/choose) or
 [starting a discussion](https://github.com/fortran-lang/stdlib/discussions/new).
 
+## Documentation
```
Aman-Godara, line 34: A new point needs to be added to the 6 bullet points above.

awvwgk:
Do you think it would be better to maintain such template in the wiki? The wiki might already be world-editable, if this is not the case, I can change the permissions there.

awvwgk:
An alternative would be to add the spec template to the pages directory such that it will be visible together with the specs at stdlib.fortran-lang.org.

JHenneberg:
> Do you think it would be better to maintain such template in the wiki? The wiki might already be world-editable, if this is not the case, I can change the permissions there.

Personally I am not a fan of the wiki pages, because when you check out the repository you do not have access to them. Also to keep consistency with the other documents `WORKFLOW.md` etc I think it is better to have the `SPEC_TEMPLATE.md`. 

This is just the PR resulting from the discussion in #504 , so we should continue the discussion there.

milancurcic:
FWIW I also don't like the WIki pages. Ideally when creating a new spec document, I'd like to just copy the spec template md file from the repo and fill it out.

PR #519, opened 2021-09-10
Add the parse of `access` in `mode` of the `open` function, to close #465.

### LINKED ISSUE #465: [Feedback] The inconsistency of `stdlib` internal modules about `access` keyword of opening files: stdlib_io_open, stdlib_logger
**Describe the issue**

I tried to use the `stdlib_logger` module, and used the `open` function in `stdlib_io`, and found that there was an adaptability problem. The root cause was that `stdlib_logger` required to open the log file as `access=sequential`, while the `stdlib_io:open` function defaulted to open the file as `access=stream` , I was wondering if this is a problem and how to solve this problem?
```fortran
program test_io_logger
    use stdlib_logger, only: global_logger
    use stdlib_io, only: open
    implicit none
    integer :: log_unit

    log_unit = open("log.txt","w+")
    call global_logger%add_log_unit(log_unit)
        !! ERROR STOP unit in stdlib_logger % set_log_unit is not "sequential".

end program test_io_logger
```
**Links**

+ https://github.com/fortran-lang/stdlib/blob/d0f13b08e4778961cb6d9c93886a65764bc5762d/src/stdlib_io.fypp#L266
+ https://github.com/fortran-lang/stdlib/blob/d0f13b08e4778961cb6d9c93886a65764bc5762d/src/stdlib_logger.f90#L384-L385

zoziha:
```fortran
select case (mode_(4:4))
    case('u'); access_ = 'sequential'  !! u: universal
    case('d'); access_ = 'direct'
    case('s'); access_ = 'stream'
end select
```
- [x] add `sequential` and `direct` `access` support in the `open` function.

#### Description

To fix #465 , I think `open` as a more basic function, there are three types of access (`sequential, direct, stream`) are supported in the fortran syntax, so keeping `stream` as the default and adding support for `sequential and direct` is currently acceptable.

### DIFF
--- a/doc/specs/stdlib_io.md
+++ b/doc/specs/stdlib_io.md
@@ -49,7 +49,7 @@ Experimental
 
 ### Description
 
-Returns the unit number of a file opened to read, to write, or to read and write. The file might be a text file or a binary file. All files are opened using a streamed access.
+Returns the unit number of a file opened to read, to write, or to read and write. The file might be a text file or a binary file. All files are opened using a streamed access by default.
 
 ### Syntax
 
@@ -71,9 +71,12 @@ Returns the unit number of a file opened to read, to write, or to read and write
 | `'+'` | open for updating (reading and writing) |
 | `'b'` | binary mode |
 | `'t'` | text mode (default) |
+| `'u'` | sequential access |
+| `'d'` | direct access |
+| `'s'` | stream access (default) |
 
 
-The default `mode` is `'rt'` (i.e. open for reading a text file). The `mode` may include one of the four different methods for opening a file (i.e., `'r'`, `'w'`, `'x'`, and `'a'`). These four methods can be associated with the character `'+'` to open the file for updating. In addition, it can be specified if the file should be handled as a binary file (`'b'`) or a text file (`'t'`).
+The default `mode` is `'rts'` (i.e. open for reading a text file). The `mode` may include one of the four different methods for opening a file (i.e., `'r'`, `'w'`, `'x'`, and `'a'`). These four methods can be associated with the character `'+'` to open the file for updating. In addition, it can be specified if the file should be handled as a binary file (`'b'`) or a text file (`'t'`).
 
 `iostat` (optional): Shall be a scalar of type `integer` that receives the error status of `open`, if provided. If no error exists, `iostat` is zero.
 
--- a/src/stdlib_io.fypp
+++ b/src/stdlib_io.fypp
@@ -211,7 +211,7 @@ contains
     character(*), intent(in), optional :: mode
     integer, intent(out), optional :: iostat
 
-    character(3) :: mode_
+    character(4) :: mode_
     character(:),allocatable :: action_, position_, status_, access_, form_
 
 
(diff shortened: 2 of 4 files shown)


milancurcic:
Thanks @zoziha. Though potentially useful (other than for fixing the logger), I'm not convinced that these are necessary.

 A downside is that it complicates the API--we're now adding the 4th (Fortran-specific) letter to the syntax. It's a development and maintenance cost for us, and a learning cost for the user.

The upside is that it'd now work with the stdlib logger, and users who want to use `sequential` or `direct` access with `stdlib_io` `open`, now can. 

I wonder if we should deprecate `sequential` and `direct` access from the `stdlib_io` POV. Is there anything that we can do with `sequential` or `direct` access that we can't do with `stream` access? If yes, the solution to #465 would be to update the logger to use stream access.

certik:
My feeling is similar to @milancurcic in that the beauty of the `open` function is that it behaves similarly to Python or C and it simplifies the usage. The letters / API that it allows are cross platform, there is nothing Fortran or C or Python specific in it and most people quickly understand the usage. This PR adds some Fortran specific usage, and to be honest I myself would not be able to explain the difference between "u", "d" and "s", without studying. It seems it's not worth doing --- if a specific application demands this level of control, then using the intrinsic open statement seems like the way to go.

zoziha:
Okay, I think what you said is correct and reasonable. It seems that we should try to modify the sequential requirements in stdlib_logger to stream. (see #465 )

jvdp1:
FWIW I agree with @milancurcic and @certik regarding the additional letters in `open()`.

I think we discussed with @wclodius2 the differences between `sequential` vs `stream` for `stdlib_logger` when he developed it. But I can't find back this discussion. So I may be wrong. I highly suggest to discuss this "issue" with @wclodius2 before opening a PR.

wclodius2:
I am on travel again. Travel and catching up on activities delayed by travel, makes it hard to code and document the code. Right now my coding is focussed on hash functions and hash tables. I had thought I was caught up with the coding for hashes, but [SMHasher](https://github.com/rurban/smhasher) added some more tests (particularly for bad seeds), that showed problems with some of the hash codes I had adapted, and added some hash codes that were better than what I was using, so I have been redoing my hash codes. When I get these into `stdlib` I will  be willing to go back and redo some  earlier codes.

jvdp1:
> I am on travel again. Travel and catching up on activities delayed by travel, makes it hard to code and document the code. Right now my coding is focussed on hash functions and hash tables. I had thought I was caught up with the coding for hashes, but [SMHasher](https://github.com/rurban/smhasher) added some more tests (particularly for bad seeds), that showed problems with some of the hash codes I had adapted, and added some hash codes that were better than what I was using, so I have been redoing my hash codes. When I get these into `stdlib` I will be willing to go back and redo some earlier codes.

@wclodius2 Nice that you are working on the hash codes. If you need some help, don't hesitate to ask for it.
Re: `stdlib_logger`, we will first see what will result from the future discussion ;)

zoziha:
Thank you for your comments. It seems that this PR is unnecessary, and I will cancel this PR.

PR #520, opened 2021-09-11
[stdlib_io] add `disp`(display variable values formatted).

zoziha:
- [x] add `disp`.

Note: Because `disp` is a very important IO routine, I hope it will **be discussed and approved more widely** before it is adopted.

#### Description
**Print** your data to a output `unit` with a `header`. ([more detail](https://github.com/fortran-lang/stdlib/blob/d4cce043d079bca01b4c7629696e73f315e3207e/doc/specs/stdlib_io.md#disp---display-the-value-of-the-vairable))
```fortran
call [[stdlib_io(module):disp(interface)]]  &
( [x , header, unit=output_unit, brief=.false., format="g0.4", width=80, sep="  " ] )
```
Example:
```fortran
use stdlib_strings, only: to_string
use stdlib_io,      only: disp, open

call disp("It is a note.")
call disp(x, "I am a scalar:")

i = open("array.txt")
call disp(A(1:50, 2:4), "Let me see the array:", unit=i, brief=.true.)
do j = 1, 3
    call disp(B(2:10,:,j), "Array B, index: "//to_string(j), unit=i)
end do
close(i)
```

#### Prior Art
+ [keurfonluu/Forlab.disp](https://github.com/keurfonluu/Forlab/blob/62b80c031fe5f283e6696be0353c13f4ba5886de/src/lib/forlab.f90#L2712)
+ [matlab.disp](https://ww2.mathworks.cn/help/matlab/ref/disp.html?lang=en)
+ [armadillo/brief_print](http://arma.sourceforge.net/docs.html#brief_print) (good suggestion from @St-Maxwell )
image
+ [ctsa/mdisplay](https://github.com/zoziha/fortsa/blob/256001363bd250dc81c5a20090049052033a9fe8/src/ctsa/matrix.c#L368)

#### Notes
`disp` is the first routine I submitted to `stdlib`, because the original PR discussion became **very long and not suitable for review** (see #445).
And I resubmitted this PR now.


doc/specs/stdlib_io.md
```diff
@@ -131,3 +131,130 @@ program demo_savetxt
...
+call disp( A(i, j, 2, :, 1:10) [, header, unit, brief] )    !! `i, j, ...` can be determined by `do` loop.
+```
+
+For `complex` type, the output format is `*(A25, 1X)`; 
+For    other types, the output format is `*(A12, 1X)`.
+
+To prevent users from accidentally passing large-length arrays to `disp`, causing unnecessary io blockage:
+1. If the `brief` argument is not specified, `disp` will print **the brief array content with a length of `10*50` by default**.
+2. Specify `brief=.true.`, `disp` will print **the brief array content with a length of `5*5`**;
+3. Specify `brief=.false.`, `disp` will print **`all` the contents of the array**.
```
jvdp1, line 161: So if `brief` is not provided (i.e. the default), then it is neither `.true.` nor `.false.`, right?
This sounds a bit counter-intuitive to me.

doc/specs/stdlib_io.md
```diff
@@ -131,3 +131,130 @@ program demo_savetxt
...
+
+### Syntax
+
+`call [[stdlib_io(module):disp(interface)]]([x, header, unit, brief])`  
+
+### Arguments
+
+`x`: Shall be a `logical/integer/real/complex/string_type` scalar or `logical/integer/real/complex` and rank-1/rank-2 array.
```
jvdp1, line 169: Why not array of characters or of string_type?

doc/specs/stdlib_io.md
```diff
@@ -131,3 +131,130 @@ program demo_savetxt
...
+
+```fortran
+call disp( A(i, j, 2, :, 1:10) [, header, unit, brief] )    !! `i, j, ...` can be determined by `do` loop.
+```
+
+For `complex` type, the output format is `*(A25, 1X)`; 
+For    other types, the output format is `*(A12, 1X)`.
+
+To prevent users from accidentally passing large-length arrays to `disp`, causing unnecessary io blockage:
+1. If the `brief` argument is not specified, `disp` will print **the brief array content with a length of `10*50` by default**.
```
jvdp1, line 159: I am not in favour of this behaviour. I would expect that when brief is not provided, then the defualt is false.

jvdp1 (commented):
Thank you for this PR. Here are some comments on the specs.

doc/specs/stdlib_io.md
zoziha: The **character length** of array of `string_type` is often **uncertain**, and it is difficult to print them uniformly.
And we already have `stringlist`, its internal elements are **private**, and the outside is inaccessible.
(see [stdlib_stringlist_type.f90#L53~L54](https://github.com/fortran-lang/stdlib/blob/0d1cc1c171e670a004f8bd472ab4181434f2b734/src/stdlib_stringlist_type.f90#L53~L54))

doc/specs/stdlib_io.md
zoziha: This approach is reasonable and achievable. `10*50` is also sufficient for general needs. My concern mainly stems from the fact that Fortran is a compiled language and users may pass large arrays to `disp`.
Of course, this can be unnecessary, because we can also pass a large array to `print` in normal usage:
```fortran
print *, x !! x(1:100000, 1:100000)
```

zoziha: This approach is reasonable and achievable. `10*50` is also sufficient for general needs. My concern mainly stems from the fact that Fortran is a compiled language and users may pass large arrays to `disp`.
Of course, this can be unnecessary, because we can also pass a large array to `print` in normal usage:
```fortran
print *, x !! x(1:100000, 1:100000)
```

doc/specs/stdlib_io.md
zoziha: `brief` is `.false.` now.

doc/specs/stdlib_io.md
zoziha: Support for `string_type` type is added: `string_type` scalar, rank-1/rank-2 array.
Only `character(*)` **scalar** is supported, `string_type` array should be more easy to use than `character(*)` array.
:)

doc/specs/stdlib_io.md
zoziha: Same as comment here, https://github.com/fortran-lang/stdlib/pull/520/#discussion_r720709803
> `brief` is `.false.` by default now.

zoziha:
I redesigned and updated the internal implementation of `disp`, it supports more control arguments, and now also supports the corresponding array of `string_type` type.
Therefore, poor internal design may be introduced. Please suggest amendments to make the disp more formal. 😊
### Questions
- [ ] width=80 by default.
Considering that `disp` has the need to output to the screen, `80` is a commonly used size.
- [ ] Whether the characters are left-justified or right-justified?
It seems that the left alignment needs to be neater.
- [ ] Is the parameter `sep` required? How to determine the name `sep`?
sep(separator).
- [ ] The compilation volume of the `stdlib_io_disp.o` submodule under `gfortran` has reached `800KB`. 
I have tried my best to improve it. If there is a way to improve the compilation volume, please let me know and I will try it.

Thank you :)

src/stdlib_io_disp.fypp
```diff
@@ -0,0 +1,310 @@
...
+        #{if r1 == 2 and k1 != "string_type"}#integer :: j#{endif}#
+        #{if k1 != "string_type"}#type(string_type), allocatable :: x_str${ranksuffix(r1)}$#{endif}#
+        #{if r1 != 0}#type(string_type) :: array_info#{endif}#
+        
+        #! State default values
+        unit_   = optval(unit, output_unit)
+        brief_  = optval(brief, .false.)
+        format_ = optval(format, "g0.4")
+        
+        #! width have to be greater than or equal 80 by default
```
zoziha, line 45:
```suggestion
        #! width have to be greater than or equal to 80 by default
```

doc/specs/stdlib_io.md
```diff
@@ -131,3 +131,126 @@ program demo_savetxt
...
+### Syntax
+
+`call [[stdlib_io(module):disp(interface)]]( [x, header, unit, brief, format, width, sep] )`  
+
+### Arguments
+
+- `x`: Shall be a `logical/integer/real/complex/character(len=*)/string_type` scalar or `logical/integer/real/complex/string_type` and rank-1/rank-2 array.
+This argument is `intent(in)` and `optional`.
+
+- `header`: Shall be a `character(len=*)` scalar.
```
jvdp1, line 159:
```suggestion
- `header`: Shall be a `character` scalar. 
```

doc/specs/stdlib_io.md
```diff
@@ -131,3 +131,126 @@ program demo_savetxt
...
+### Arguments
+
+- `x`: Shall be a `logical/integer/real/complex/character(len=*)/string_type` scalar or `logical/integer/real/complex/string_type` and rank-1/rank-2 array.
+This argument is `intent(in)` and `optional`.
+
+- `header`: Shall be a `character(len=*)` scalar. 
+This argument is `intent(in)` and `optional`.
+
+- `unit`: Shall be an `integer` scalar, linked to an IO stream.
+This argument is `intent(in)` and `optional`.<br>
```
jvdp1, line 163:
```suggestion
This argument is `intent(in)` and `optional`.
```

doc/specs/stdlib_io.md
```diff
@@ -131,3 +131,126 @@ program demo_savetxt
...
+The default value is `output_unit` from `iso_fortran_env` module.
+
+- `brief`: Shall be a `logical` scalar, controls an abridged version of the `x` array to be outputted.
+This argument is `intent(in)` and `optional`.<br>
```
jvdp1, line 167:
```suggestion
This argument is `intent(in)` and `optional`.
```

doc/specs/stdlib_io.md
```diff
@@ -131,3 +131,126 @@ program demo_savetxt
...
+The default value is `.false.`
+
+- `format`: Shall be a `character(len=*)` scalar.
```
jvdp1, line 170:
```suggestion
- `format`: Shall be a `character` scalar.
```

doc/specs/stdlib_io.md
```diff
@@ -131,3 +131,126 @@ program demo_savetxt
...
+This argument is `intent(in)` and `optional`.<br>
```
jvdp1, line 171:
```suggestion
This argument is `intent(in)` and `optional`.
```

doc/specs/stdlib_io.md
```diff
@@ -131,3 +131,126 @@ program demo_savetxt
...
+The default value is `g0.4`.
+
+- `width`: Shall be an `integer` scalar, controls the outputted maximum width (`>=80`).
```
jvdp1, line 174: This sentence is unclear

doc/specs/stdlib_io.md
```diff
@@ -131,3 +131,126 @@ program demo_savetxt
...
+This argument is `intent(in)` and `optional`.<br>
+The default value is `80`.
+
+- `sep`: Shall be a `character(len=*)` scalar, separator.
```
jvdp1, line 178:
```suggestion
- `sep`: Shall be a `character(len=*)` scalar, and represents the separator used for delimiting the columns.
```
Is this the aim of `sep`?

doc/specs/stdlib_io.md
```diff
@@ -131,3 +131,126 @@ program demo_savetxt
...
+This argument is `intent(in)` and `optional`.<br>
+The default value is "&ensp;&ensp;", two spaces.
```
jvdp1, line 180: What do you mean by `two spaces`?

doc/specs/stdlib_io.md
```diff
@@ -131,3 +131,126 @@ program demo_savetxt
...
+
+### Output
+
+The result is to print `header` and `x` on the screen (or another output `unit/file`) in this order.<br>
```
jvdp1, line 184:
```suggestion
The result is to print `header` and `x` on the screen (or another output `unit/file`) in this order.
```

doc/specs/stdlib_io.md
```diff
@@ -131,3 +131,126 @@ program demo_savetxt
...
+If `disp` is not passed any arguments, a blank line will be printed.
+
+### Example
+
+```fortran
```
jvdp1, line 189: I suggest to reduce the example. One or two `disp` should be enough for the reader to understand its aim.

src/stdlib_io.fypp
```diff
@@ -28,6 +29,36 @@ module stdlib_io
     FMT_COMPLEX_SP = '(*(es15.8e2,1x,es15.8e2))', &
     FMT_COMPLEX_DP = '(*(es24.16e3,1x,es24.16e3))', &
     FMT_COMPLEX_QP = '(*(es44.35e4,1x,es44.35e4))'
+    
+  !> version: experimental
+  !>
+  !> Display a scalar, vector or matrix formatted.
+  !> ([Specification](../page/specs/stdlib_io.html#display-the-value-of-the-variable))
```
jvdp1, line 36:
```suggestion
  !> ([Specification](../page/specs/stdlib_io.html#disp-display-the-value-of-the-variable))
```

src/stdlib_io_disp.fypp
```diff
@@ -0,0 +1,308 @@
...
+        width_  = optval(width, 80)
+        width_  = merge(width_, 80, width_ > 80)
+        
+        sep_    = optval(sep, "  ")
```
jvdp1, line 47: Why "two spaces" instead of "one space"?

jvdp1 (commented):
Thank you for this proposition. Here are some comments.

Regarding the tests, I guess that they should be moved to support the `test-drive` features, introduced recently in stdlib. However, since this PR is open since a while, I would understand that you would prefer to not do it.

PR #521, opened 2021-09-11
Ensure module output directory is generated in configure stage

awvwgk:
- dependent projects using the stdlib's build interface will resolve the module output directory before it is created and error without this fix


ivan-pi:
Is this something I can test, by forking [stdlib-cmake-example](https://github.com/fortran-lang/stdlib-cmake-example) and trying to build a project with several modules?

awvwgk:
I'm already fixing this behaviour in the CMake example, you would have to remove my fix from there first to check:

https://github.com/fortran-lang/stdlib-cmake-example/blob/61e5fb0a1df900d8ebfe216b0f60f1a5eeefcdf9/config/cmake/Findfortran_stdlib.cmake#L149-L151

awvwgk:
Won't happen with a single dependency, so this can't be tested with the CMake example.

I usually hit issues with non-existent include directories when working with five to six interdependent CMake projects and multiple levels of dependencies, looks something like this:

```
CMake Error in CMakeLists.txt:
  Imported target "toml-f::toml-f" includes non-existent path

    "/home/awvwgk/projects/src/git/mctc-tools/tblite/__build/_deps/toml-f-build/include"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.
```

ivan-pi (approved):
I'm not well versed in CMake, but the change is small and I trust you have tested this so +1.

As a side note: do you think it makes sense to add a link in your comment, e.g. "see PR #521", or is this unnecessary because it can always be queried using git?

milancurcic (approved):
Thanks!

jvdp1 (approved):
Sounds good to me. Thank you.

PR #528, opened 2021-09-17
Reduce redundancy caused by optional arguments

### LINKED ISSUE #524: Rewrite parts of stdlib where optval could be used
While exploring the `stdlib` I discovered the `optval` function and shortly after a routine which could have made use of it but it doesn't. Should we search for such optional arguments and replace the `if`-`else`-blocks with the `optval`-function?
If there are no other plans which would interfere with this, I think this should be done because it shortens the code which makes the code easier to understand.

If I find time to build the `stdlib` I could do the job and open a PR. Please tell me if I have to keep something in mind, otherwise I will just try it.

Carltoffel:
With this PR I tried to reduce redundancy, primary by using the `optval` function.

Some discoveries I made while working on this:
- In `stdlib_io` the `open` statement is used with an optional `iostat` argument. Since `open` is an intrinsic function and not a routine, the optional argument cannot be directly passed, which leads to the `open` statement being defined twice. If it doesn't cause too much overhead I suggest to declare an extra variable which will always be passed to the `iostat` argument of the `open` statement. If the optional `iostat` argument of the `open` function is present, the result will then be written to the optional argument.
- The `simps_weights` function (`stdlib_quadrature_simps.fypp`) contains an nested `if` statement which could be rewritten as `select case ( optval(even, 0) ) ...`. I'm not sure if this would improve the code, but it would reduce the `if`-nesting by one `if`.
- `stdlib_stat_moment` has some functions which could be rewritten using the `optval` function, but as I mentioned in the commit message this is inefficient. Instead I introduced a new variable which de-duplicates the formular.
- OT: I saw a `min(max(...))` construct which could be replaced with `clip`... I guess there are more of these, so maybe we should search and replace them where possible just like I did with `present()` and `optval()` in this PR.

closes #524

### DIFF
--- a/src/stdlib_stats_moment_all.fypp
+++ b/src/stdlib_stats_moment_all.fypp
@@ -22,6 +22,7 @@ contains
         ${t1}$ :: res
 
         real(${k1}$) :: n
+        ${t1}$ :: center_
 
         if (.not.optval(mask, .true.)) then
           res = ieee_value(1._${k1}$, ieee_quiet_nan)
@@ -31,10 +32,11 @@ contains
         n = real(size(x, kind = int64), ${k1}$)
 
         if (present(center)) then
-         res = sum((x - center)**order) / n
+          center_ = center
         else
-         res = sum((x - mean(x))**order) / n
+          center_ = mean(x)
         end if
+        res = sum((x - center_)**order) / n
 
       end function ${RName}$
     #:endfor
@@ -52,6 +54,7 @@ contains
         real(dp) :: res
 
         real(dp) :: n
+        real(dp) :: center_
 
         if (.not.optval(mask, .true.)) then
           res = ieee_value(1._dp, ieee_quiet_nan)
@@ -61,10 +64,11 @@ contains
         n = real(size(x, kind = int64), dp)
 
         if (present(center)) then
-         res = sum((real(x, dp) - center)**order) / n
+          center_ = center
         else
-         res = sum((real(x, dp) - mean(x))**order) / n
+          center_ = mean(x)
         end if
+        res = sum((real(x, dp) - center_)**order) / n
 
       end function ${RName}$
     #:endfor
@@ -82,14 +86,16 @@ contains
         ${t1}$ :: res
 
         real(${k1}$) :: n
+        ${t1}$ :: center_
 
         n = real(count(mask, kind = int64), ${k1}$)
 
         if (present(center)) then
-         res = sum((x - center)**order, mask) / n
+          center_ = center
         else
-         res = sum((x - mean(x, mask))**order, mask) / n
+          center_ = mean(x, mask)
         end if
+        res = sum((x - center_)**order, mask) / n
 
       end function ${RName}$
     #:endfor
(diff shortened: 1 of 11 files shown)


src/stdlib_stats_moment_all.fypp
```diff
@@ -61,10 +64,11 @@ contains
         n = real(size(x, kind = int64), dp)
 
         if (present(center)) then
-         res = sum((real(x, dp) - center)**order) / n
+          center_ = center
         else
-         res = sum((real(x, dp) - mean(x))**order) / n
+          center_ = mean(x)
         end if
```
milancurcic, line 70: Could this not also be replaced by `optval`?

```suggestion
        center_ = optval(center, mean(x))
```

and likewise in two other instances below?

milancurcic (approved):
Thanks @Carltoffel, this all looks good to me.

src/stdlib_stats_moment_all.fypp
```diff
@@ -31,10 +32,11 @@ contains
         n = real(size(x, kind = int64), ${k1}$)
 
         if (present(center)) then
-         res = sum((x - center)**order) / n
+          center_ = center
         else
-         res = sum((x - mean(x))**order) / n
+          center_ = mean(x)
         end if
+        res = sum((x - center_)**order) / n
```
jvdp1, line 39: Why not (probably due to inefficiency?):
```fortran
res = sum(( x - optval(center, mean(x))**order) / n
```

src/stdlib_stats_moment_all.fypp
jvdp1: Is it ok for efficiency? I mean: would `mean(x)` not be evaluated, even if not needed?

jvdp1 (commented):
Thank you  for these improvements. Regarding `stdlib_stats_moment`, I wonder if using `optval()` will not affect its efficiency?

src/stdlib_stats_moment_all.fypp
milancurcic: I thought the same, and I worried that maybe some compiler may place an `optval()` call in every iteration of the loop, that wouldn't be good. But I don't know. Evaluating the scalar `center_` first seems safer.

milancurcic: Sorry, I read it again and now understand what you mean. Ignore my previous comment. I don't think my question/suggestion was good anymore, because `mean(x)` would be evaluated if refactored to `optval()`, so let's not do it.

gareth-nx:
I'm cautious about this pull request, because I don't understand the performance implications of generally replacing `present` with a function call. See a few comments [here](https://github.com/fortran-lang/stdlib/issues/524) -- just wanting to check whether we understand the implications.

src/stdlib_stats_moment_all.fypp
Carltoffel: Indeed. I tested it with a large array and gfortran was slower every time, even when enabling `-Ofast -flto`. `optval` doesn't seem to be efficient here.

jvdp1: thank yoju for testing. Futhermore, using `center_` __might__ add an additional temporary array, especially for the case when `center` is present. Therefore, for the different modules of `stdlib_stats_moment`, I suggest to revert the changes and to keep the original version.

Carltoffel: I don't understand `fypp`... is there a way for `center` to be an array? If it is only scalar the compiler should be able to optimise it.

jvdp1: In this case, `center` is indeed a scalar. In `stdlib_stats_moment.fypp`, there is a similar case, however `center` is an array.

I agree that, since it is a scalar in this case, the potential cost is not high, and the compiler should be able to optimise it.

jvdp1 (approved):
Based on all discussions in this PR, and in #524, I approve this PR as it is currently. Thank you @Carltoffel .

milancurcic:
Let's merge this tomorrow (Sep 21) if there are no objections.

Carltoffel:
Thank you all for the respectful discussion about the performance. I've learned a lot!

PR #530, opened 2021-09-19
Make it clearer how the user can control compiler optimization

gareth-nx:
These edits to the README attempt to offer simple guidance on compiling stdlib with optimization flags.

In this discussion [here](https://github.com/fortran-lang/stdlib/issues/524) it became clear that not everyone understands how to apply compiler optimizations to stdlib, although the discussion shows that is quite important for performance. 

I think some simple pointers would help users who do not have much experience with build tools (like me!).

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ cmake -B build
 You can pass additional options to CMake to customize the build.
 Important options are
 
-- `-G Ninja` to use the Ninja backend instead of the default Make backend. This makes it easy to specify compiler flags for the build. Other build backends are available with a similar syntax.
+- `-G Ninja` to use the Ninja backend instead of the default Make backend. Other build backends are available with a similar syntax.
 - `-DCMAKE_INSTALL_PREFIX` is used to provide the install location for the library. If not provided the defaults will depend on your operating system, [see here](https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html). 
 - `-DCMAKE_MAXIMUM_RANK` the maximum array rank procedures should be generated for.
   The default value is chosen as 4.
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ Important options are
 For example, to configure a build using the Ninja backend while specifying compiler flags `FFLAGS`, generating procedures up to rank 7, and installing to your home directory, use
 
 ```sh
-export FFLAGS="-O3 -flto"
+export FFLAGS="-O3"
 cmake -B build -G Ninja -DCMAKE_MAXIMUM_RANK:String=7 -DCMAKE_INSTALL_PREFIX=$HOME/.local
 ```
 
--- a/README.md
+++ b/README.md
@@ -182,7 +182,10 @@ You can limit the maximum rank by setting ``-DMAXRANK=<num>`` in the ``FYPPFLAGS
 make -f Makefile.manual FYPPFLAGS=-DMAXRANK=4
 ```
 
-You can edit the compiler and its optimization flags by specifying the `FC` and `FFLAGS` variables in Makefile.manual. The default `FFLAGS` does not include any optimization, and you may want to add `-O3 -flto` to its definition. 
+You can also specify the compiler and compiler-flags by setting the ``FC`` and ``FFLAGS`` environmental variables. Among other things, this facilitates use of compiler optimizations that are not specified in the Makefile.manual defaults.
+```sh
+make -f Makefile.manual FYPPFLAGS=-DMAXRANK=4 FC=gfortran FFLAGS="-O3 -flto"
+```
 
 ### Build with [fortran-lang/fpm](https://github.com/fortran-lang/fpm)
 
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ cmake -B build
 You can pass additional options to CMake to customize the build.
 Important options are
 
-- `-G Ninja` to use the Ninja backend instead of the default Make backend. Other build backends are available with a similar syntax.
+- `-G Ninja` to use the Ninja backend instead of the default Make backend. This makes it easy to specify compiler flags for the build. Other build backends are available with a similar syntax.
 - `-DCMAKE_INSTALL_PREFIX` is used to provide the install location for the library. If not provided the defaults will depend on your operating system, [see here](https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html). 
 - `-DCMAKE_MAXIMUM_RANK` the maximum array rank procedures should be generated for.
   The default value is chosen as 4.
@@ -134,9 +134,10 @@ Important options are
   Compiling with maximum rank 15 can be resource intensive and requires at least 16 GB of memory to allow parallel compilation or 4 GB memory for sequential compilation.
 - `-DBUILD_SHARED_LIBS` set to `on` in case you want link your application dynamically against the standard library (default: `off`).
 
-For example, to configure a build using the Ninja backend and generating procedures up to rank 7, which is installed to your home directory use
+For example, to configure a build using the Ninja backend while specifying compiler flags `FFLAGS`, generating procedures up to rank 7, and installing to your home directory, use
 
 ```sh
+export FFLAGS="-O3 -flto -fPIC"
 cmake -B build -G Ninja -DCMAKE_MAXIMUM_RANK:String=7 -DCMAKE_INSTALL_PREFIX=$HOME/.local
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -125,7 +125,7 @@ cmake -B build
 You can pass additional options to CMake to customize the build.
 Important options are
 
-- `-G Ninja` to use the Ninja backend instead of the default Make backend. Other build backends are available with a similar syntax.
+- `-G Ninja` to use the Ninja backend instead of the default Make backend. This makes it easy to specify compiler flags for the build. Other build backends are available with a similar syntax.
```
awvwgk, line 128: This is incorrect, specifying compiler flags is independent of the build backend used.

README.md
```diff
@@ -134,9 +134,10 @@ Important options are
   Compiling with maximum rank 15 can be resource intensive and requires at least 16 GB of memory to allow parallel compilation or 4 GB memory for sequential compilation.
 - `-DBUILD_SHARED_LIBS` set to `on` in case you want link your application dynamically against the standard library (default: `off`).
 
-For example, to configure a build using the Ninja backend and generating procedures up to rank 7, which is installed to your home directory use
+For example, to configure a build using the Ninja backend while specifying compiler flags `FFLAGS`, generating procedures up to rank 7, and installing to your home directory, use
 
 ```sh
+export FFLAGS="-O3 -flto -fPIC"
```
awvwgk, line 140: What is reason to request position independent code via the Fortran flags? This should be done via the [`CMAKE_POSITION_INDEPENDENT_CODE`](https://cmake.org/cmake/help/latest/variable/CMAKE_POSITION_INDEPENDENT_CODE.html#variable:CMAKE_POSITION_INDEPENDENT_CODE) variable instead if needed. Also for the manual makefile build this flag is absent.

README.md
```diff
@@ -177,6 +178,8 @@ You can limit the maximum rank by setting ``-DMAXRANK=<num>`` in the ``FYPPFLAGS
 make -f Makefile.manual FYPPFLAGS=-DMAXRANK=4
 ```
 
+You can edit the compiler and its optimization flags by specifying the `FC` and `FFLAGS` variables in Makefile.manual. The default `FFLAGS` does not include any optimization, and you may want to add `-O3 -flto` to its definition.
```
awvwgk, line 181: Editing makefiles shouldn't be encouraged here, instead can we modify the makefile to accept the environment variables as default? Something along the lines of:

```make
FC ?= gfortran
FFLAGS ?= ...
```

The manual makefile would be used with

```
make -f Makefile.manual FFLAGS="-O3 -flto"
```

awvwgk (changes requested):

README.md
gareth-nx: Thanks, I'll fix this. 

At some point, I tried recompiling stdlib while specifying FFLAGS. However it didn't work until I used Ninja. 

But perhaps my problem was that I didn't delete the "build" directory (and perhaps also my CMAKE_INSTALL_PREFIX directory). 

Is it necessary to delete some build directories in that kind of situation? If yes I would like to add some statement about that, within this pull request.

awvwgk: The important file in your build directory is `CMakeCache.txt`. It contains the state of the complete build directory. By deleting this file you force a reconfiguration, you could also, if you know what you are doing, modify variables in this cache directly. Deleting the complete build directory will always be the safest bet.

README.md
gareth-nx: Thanks for the comment. I think `-fPIC` should be removed, and will implement that. 

The reason I added it here is because it was part of the performance solution discussed [here](https://github.com/fortran-lang/stdlib/issues/524), that @jvdp1 figured out. I wasn't aware of other means to specify it. 

However I just checked now, and find `-fPIC` it isn't needed to get good performance in that case -- just `-flto`. So best to avoid the extra complexity here.

awvwgk: Position independent code is usually used to link shared object, for linking executables from static libraries it is not required. Building with `-DBUILD_SHARED_LIBS=true` to get shared objects will automatically enable position independent code. Enabling it for static libraries is usually only required if you plan to manually link it into a shared object later.

ivan-pi: I think the `-fPIC` option is important if you want to use stdlib as a shared library. But it should be discussed in a different section (where also @awvwgk's answer for CMake can be added).

README.md
gareth-nx: Agree that's a nice way of doing it. 

The only potential issue I can think of: Suppose a user accidently has `FC` or `FFLAGS` already defined, and are not aware of that-- then they might get confusing results. 

But if the README encourages users to define those two variables, then I doubt that will be a problem.

awvwgk: I'm hearing this argument a lot, how do you set `FC` or `FFLAGS` by accident? Usually those variables are set by environment modules which load compilers, where else?

README.md
gareth-nx: Thanks -- I added a statement about deleting the build directory.

README.md
gareth-nx: A situation would be where a user has a script that they know can be used to setup the environment for some software -- but they don't know/understand the contents of the script.

For example the [clawpack software installation instructions](https://www.clawpack.org/installing.html) encourage users to define: 

```
export CLAW=$HOME/clawpack_src/clawpack-v5.8.0
export FC=gfortran
```

I have a script "setup_clawpack.sh" in my home directory that does this. If using clawpack, I would simply source the script and then do stuff, but not understand how my environment was modified. And similarly for other software.

Obviously users with enough experience will appreciate the risks here, but I suspect many users do not.

Romendakil: I am bit confused. I thought at the end the main use case will be that stdlib will exist precompiled on a system and just gets linked by the user. This brings me to the question (different issue probably) whether the interfaces of stdlib are bind(C), even if they are intended to be used only in Fortran. But they are many not so unlikely situations where a user would want to use a gfortran-compiled stdlib to be used in a project compiled with the Intel compiler or vice versa.

gareth-nx: > But they are many not so unlikely situations where a user would want to use a gfortran-compiled stdlib to be used in a project compiled with the Intel compiler or vice versa.

My understanding is that this isn't possible - which is a general issue with Fortran and the module file format (not specifically a limitation of stdlib).

awvwgk: @Romendakil So far building stdlib on demand using fpm or the [CMake template](https://github.com/fortran-lang/stdlib-cmake-example) seems the most straight-forward way to use stdlib without having to deal with Fortran ABI issues.

In the future stdlib might be shipped together with compilers and installed in a location specific to the respective compiler.

Romendakil: No, that's not true. Of course, you cannot use compiled modules from a different compiler, but we are regularly using in our code third-party libraries that are only available for gfortran and include them as shared libraries in our code compiled with ifort or nagfor.

gareth-nx: @Romendakil Interesting -- so you can do this by simply having the Fortran interfaces be `bind(C)`?

Romendakil: @gareth-nx yes. For some you don't need this, it works even across compilers. But even with `bind(C)` I think you are not totally guaranteed to get always what you expect, as the accompanying C compiler of the two Fortran compilers could be different, and e.g. have different bit sizes of their integer or float data types. However, in general it works. The usual situation is that most of my colleagues in theoretical physics do development only with gfortran, and they have some sloppiness which is tolerated by gfortran. So I cannot compile their code with Nagfor. I do development for our own code mostly with Nagfor, but as the third-party code is bind(C) I can nevertheless link it and use it correctly.

gareth-nx: I have pushed a commit that edits Makefile.manual as suggested by @awvwgk . Further, in the README there is also explanation of how to specify FC and FFLAGS when building with make.

Recall the purpose is to make the process of compiling with optimization clearer for users, especially novices, while keeping the document brief.

awvwgk (approved):

README.md
```diff
@@ -134,9 +134,10 @@ Important options are
   Compiling with maximum rank 15 can be resource intensive and requires at least 16 GB of memory to allow parallel compilation or 4 GB memory for sequential compilation.
 - `-DBUILD_SHARED_LIBS` set to `on` in case you want link your application dynamically against the standard library (default: `off`).
 
-For example, to configure a build using the Ninja backend and generating procedures up to rank 7, which is installed to your home directory use
+For example, to configure a build using the Ninja backend while specifying compiler flags `FFLAGS`, generating procedures up to rank 7, and installing to your home directory, use
 
 ```sh
+export FFLAGS="-O3 -flto"
```
jvdp1, line 140:
```suggestion
export FFLAGS="-O3"
```
I propose to remove "-lto" such that the proposed `FFLAGS` is (almost?) independant of the compiler

README.md
```diff
@@ -162,6 +163,10 @@ cmake --install build
 
 Now you have a working version of stdlib you can use for your project.
 
+If at some point you wish to recompile stdlib with different options, you might
```
jvdp1, line 166:
```suggestion
If at some point you wish to recompile `stdlib` with different options, you might
```

jvdp1 (approved):
Thank you for these additions. I approve pending minor comments.

README.md
gareth-nx: Thanks, agree that is a good trade-off.

README.md
gareth-nx: Thanks, done

milancurcic (approved):
Thank you, @gareth-nx!

Issue #387, opened 2021-04-11
CMake compiler options for build configurations


awvwgk:
The CMake build files currently set options using the `add_compile_options` function, which adds the respective argument independently from the actual build configuration (we call it profile in fpm).

For context: The CMake build configuration is found in `CMAKE_BUILD_TYPE` and is empty by default (`NoConfig`), other common profiles are `Debug`, `RelWithDebInfo`, `Release`, and `MinSizeRel`, but one can basically use any name. The global compiler options are taken from `CMAKE_Fortran_FLAGS` and the build configuration specific ones from `CMAKE_Fortran_FLAGS_${CMAKE_BUILD_TYPE}` (with the value of `CMAKE_BUILD_TYPE` as uppercase...).

As I understood @milancurcic the current options set with `add_compile_options` are the debug / development profile, but they are currently added for all build configurations. Instead we should try to set those options to the `CMAKE_Fortran_FLAGS_DEBUG` variable.

milancurcic:
Yes, what I meant in that comment was that we could do this, not that we currently do it.

PR #536, opened 2021-09-20
Fix conversion warnings in logspace

Carltoffel:
#### What do you think about these changes?

**pro**
fixes 12 warnings (the remaining 12 warnings should be easy to fix, too), which are ~ 60 lines to `stderr`

**contra**
complexity of code (readability, maintainability, etc.)

*Maybe there are better options to fix the warnings, please let me know.*

### DIFF
--- a/src/stdlib_math_logspace.fypp
+++ b/src/stdlib_math_logspace.fypp
@@ -45,21 +45,33 @@ contains
     module procedure ${RName}$
       ${t1}$ :: exponents(max(n, 0))
       exponents = linspace(start, end, n)
-      res = base ** exponents
+      #: if t1[0] == 'r'
+        res = base ** exponents
+      #: else
+        res = cmplx(base, kind=${k1}$) ** cmplx(exponents, kind=${k1}$)
+      #: endif
     end procedure
 
     #:set RName = rname("logspace", 1, t1, k1, "n_cbase")
     module procedure ${RName}$
       ${t1}$ :: exponents(max(n, 0))
       exponents = linspace(start, end, n)
-      res = base ** exponents
+      #: if t1[0] == 'r'
+        res = real(base ** cmplx(exponents, kind=${k1}$))
+      #: else
+        res = base ** cmplx(exponents, kind=${k1}$)
+      #: endif
     end procedure
 
     #:set RName = rname("logspace", 1, t1, k1, "n_ibase")
     module procedure ${RName}$
       ${t1}$ :: exponents(max(n, 0))
       exponents = linspace(start, end, n)
-      res = base ** exponents
+      #: if t1[0] == 'r'
+        res = base ** exponents
+      #: else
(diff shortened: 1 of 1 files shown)


gareth-nx:
Obviously this is a subjective judgement. My feeling is that, if the original code is legal, then the extra code complexity isn't worth it.

It would be ideal if were compiler options to turn off this kind of warning. But I don't know of any.

Carltoffel:
> It would be ideal if were compiler options to turn off this kind of warning. But I don't know of any.

Usually it is the other way around: You can turn these warnings **on** with compiler options. Build systems tend to activate every warning available, because these warnings are useful (in general).

gareth-nx:
Yes, I think for this reason gfortran supports turning warnings off, like (`-Wno-implicit`) as well as on (`-Wimplicit`). So you can use `-Wall` and just silence a few of them.

awvwgk:
> It would be ideal if were compiler options to turn off this kind of warning. But I don't know of any.

We are getting a lot of false-positive and spurious warning from our current CMake setup. See https://github.com/fortran-lang/stdlib/issues/387 for further details.

Carltoffel:
> We are getting a lot of false-positive and spurious warning from our current CMake setup. See #387 for further details.

Are those conversion warnings I fixed really false-positive? There are implicit conversions happening, so I think these are true-positive warnings. The question is, should we:
- Make the conversions explicit (verbose & complex code)
- Hide all conversion warnings (might hide useful warnings)
- Leave it as is (very verbose output, might also hide useful warnings)

awvwgk:
We are using `-Wconversion-extra` which basically flags every conversion even if it is safe to use.

https://github.com/fortran-lang/stdlib/blob/8bfcdf9fdf8d3d897eee76512ba5083ecab29f96/CMakeLists.txt#L27

Carltoffel:
Okay, so do we want to leave the code as is and simply ignore the warnings?

awvwgk:
My suggestion would be to check our compiler warning options first and trim those down to the useful ones. Once we get rid of the 90% false positives, we fix the remaining warnings.

gareth-nx:
I wonder if removing `-Wconversion-extra` is a good trade-off here?

Carltoffel:
I did a quick test if we can leave out the `kind=` arguments, because this would make the code a little more readable. Unfortunately this doesn't solves the `Conversion from COMPLEX(4) to COMPLEX(16)` warnings. But maybe this will be an useful option with a different set of compiler flags.

jvdp1:
Most of them are due to specific compiler flags. I would first remove the useless ones, and then check the remaining warnings.

epagone:
> My suggestion would be to check our compiler warning options first and trim those down to the useful ones. Once we get rid of the 90% false positives, we fix the remaining warnings.

I agree with this. The next step would be to open a bug report with the compiler(s) that should not report false positives obfuscating the useful warnings. I don't think that we should bend stdlib to the needs of the compiler(s).

Issue #292, opened 2021-01-07
Is stdlib versioned?


awvwgk:
While writing package files for stdlib (see #291) I encountered a quite simple issue what puzzled me: What is the version number of stdlib?

It turns out that some package files simply don't work if there is no version number specified, *i.e.* pkg-config will not use any pc-file without a Version entry. Not specifying any version is therefore not possible if I want to use stdlib in a non-CMake project. I could *just define* a dummy version number for my purpose, but this calls for trouble.

Maybe this issue has been discussed already somewhere, but I didn't found any hint on this here.

milancurcic:
Currently, no. It should be. You can't think of stdlib still being in perpetual 0.1.0, but that version number is not specified anywhere.

There have been discussions scattered across this repo, monthly call, and Discourse:

* #216 
* [July 2020 discussion on stdlib structure](https://youtu.be/_ojLReFvjbs?t=702)
* [Stabilization of Rust language and stdlib features](https://fortran-lang.discourse.group/t/stabilization-of-rust-language-and-stdlib-features/212?u=milancurcic)

TL'DR:

* We started with the idea of having 2 separate namespaces/APIs, experimental and stable. Everything goes into experimental (equivalent to 0.x.x), and when parts of stdlib make it into the Standard, they become "stable" (equivalent to 1.x.x). Separate namespaces/APIs means that experimental functions are in separate modules from stable ones. So there could be a `stdlib_experimental_io` module, and a `stdlib_stable_io` module. Functions can exist only in one or the other. But of course, this wasn't too meaningful because we don't have anything stable. So every module has the word "experimental" in it.
* This structure was challenged by @wclodius2 in #216. I mostly agreed and we discussed it on a call and had a follow-up discussion on Discourse.
* We tentatively agreed to unify "experimental" and (future) "stable" into the same namespace, and start semantic versioning. We didn't start it in practice.
* We also agreed to keep track of what's experimental or stable on a per procedure basis, by noting in the source code docstrings and spec docs, similar to Rust.

So, let's start versioning and making releases, like we do for fpm.

awvwgk:
Thanks for the comprehensive summary. Really looking forward to see the first release of stdlib.

awvwgk:
With #291 merged there is now a dummy version in place.

jvdp1:
> So, let's start versioning and making releases, like we do for fpm.

I agree with that. A couple of questions:
* Should we start with, let say, the version 0.1.0?
[...]

PR #537, opened 2021-09-25
Add a changelog for the current features of stdlib

awvwgk:
This is a first step in the initial release of stdlib (see #292). To get proper release notes we have to make a list of features at this occasion we can also start to create a changelog. I might missed a few relevant PRs when going through one and a half year worth of PRs. Please feel free to amend the changelog if I missed an important PR.

This changelog currently ignores the work on the build systems and the workflow automations. Not sure if we want to list a few important changes there or would just go with *support for CMake, manual Makefiles and fpm available*.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,93 @@
+# Unreleased
+
+- new module `stdlib_ascii`
+  [#32](https://github.com/fortran-lang/stdlib/pull/32)
+- new module `stdlib_bitsets`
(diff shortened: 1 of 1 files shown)


jvdp1 (approved):
Thanks. LGTM. Most important PRs are there IMO.

milancurcic (approved):
I agree, this is quite comprehensive. Thank you for collecting it all.

PR #539, opened 2021-09-26
Add function gcd

### LINKED ISSUE #515: GCD (Greatest Common Divisior)
As suggested in #365, providing `gcd` function out of the box will be a good addition.

**Description**

Have a look at these comments to know more:
https://github.com/fortran-lang/stdlib/issues/365#issuecomment-907683782
https://github.com/j3-fortran/fortran_proposals/issues/221#issuecomment-907638974

**Prior Art**

Many languages provide `gcd` out of the box.

Sideboard:
Closes #515

1. Add elemental integer function gcd to calculate the greatest common divisor of two numbers
  a. Handle negative input as positive
2. Add tests (special, normal, swapped, negative, integer kinds)

### DIFF
--- a/doc/specs/stdlib_math.md
+++ b/doc/specs/stdlib_math.md
@@ -91,6 +91,51 @@ program demo_clip_real
 end program demo_clip_real
 ```
 
+### `gcd` function
+
+#### Description
+
+Returns the greatest common divisor of two integers.
+
+#### Syntax
+
+`res = [[stdlib_math(module):gcd(interface)]] (a, b)`
+
+#### Status
+
+Experimental
+
+#### Class
+
+Elemental function.
+
+#### Argument(s)
+
+`a`: One integer with `intent(in)` to get the divisor for.
+`b`: Another integer with `intent(in)` to get the divisor for.
+
+Note: All arguments must be integers of the same `kind`.
+
+#### Output value or Result value
+
+Returns an integer of the same `kind` as that of the arguments.
+
+#### Examples
+
+##### Example 1:
+
+```fortran
+program demo_gcd
+  use stdlib_math, only: gcd
+  implicit none
+  integer :: a, b, c
+  
+  a = 48
+  b = 18
+  c = gcd(a, b) ! returns 6
+end program demo_gcd
+```
+
 ### `linspace` - Create a linearly spaced rank one array
 
 #### Description
--- a/src/stdlib_math.fypp
+++ b/src/stdlib_math.fypp
@@ -8,7 +8,7 @@ module stdlib_math
 
     implicit none
     private
-    public :: clip, linspace, logspace
+    public :: clip, gcd, linspace, logspace
     public :: EULERS_NUMBER_SP, EULERS_NUMBER_DP, EULERS_NUMBER_QP
     public :: DEFAULT_LINSPACE_LENGTH, DEFAULT_LOGSPACE_BASE, DEFAULT_LOGSPACE_LENGTH
     public :: arange
@@ -27,6 +27,16 @@ module stdlib_math
         module procedure clip_${k1}$
     #:endfor
     end interface clip
+
+    !> Returns the greatest common divisor of two integers
+    !> ([Specification](../page/specs/stdlib_math.html#gcd))
+    !>
+    !> Version: experimental
+    interface gcd
+    #:for k1, t1 in INT_KINDS_TYPES
+        module procedure gcd_${k1}$
+    #:endfor
+    end interface gcd
 
     interface linspace
     !! Version: Experimental
@@ -290,6 +300,27 @@ contains
 
         res = max(min(x, xmax), xmin)
     end function clip_${k1}$
+
+    #:endfor
+
+    #:for k1, t1 in INT_KINDS_TYPES
+    !> Returns the greatest common divisor of two integers of kind ${k1}$
+    !> using the Euclidean algorithm.
+    elemental function gcd_${k1}$(a, b) result(res)
+        ${t1}$, intent(in) :: a
+        ${t1}$, intent(in) :: b
+        ${t1}$ :: res
+
+        ${t1}$ :: rem, tmp
+
+        rem = min(abs(a), abs(b))
+        res = max(abs(a), abs(b))
+        do while (rem /= 0_${k1}$)
+          tmp = rem
+          rem = mod(res, rem)
+          res = tmp
+        end do
+    end function gcd_${k1}$
 
     #:endfor
 end module stdlib_math
--- a/src/tests/math/test_stdlib_math.f90
+++ b/src/tests/math/test_stdlib_math.f90
@@ -1,7 +1,7 @@
 ! SPDX-Identifier: MIT
 
 program test_stdlib_math
-    use stdlib_math, only: clip
+    use stdlib_math, only: clip, gcd
     use stdlib_error, only: check
     use stdlib_kinds, only: int8, int16, int32, int64, sp, dp, qp
     implicit none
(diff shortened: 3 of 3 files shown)


Sideboard:
I appreciate tips about tests. More, less, special cases …

milancurcic:
Thanks @Sideboard, the implementation looks good to me. Tests seem fine as well. 

We should add the FORD docstrings to the function, and the spec as well. You can look at other functions for examples. Let us know if you need help with that.

General question for everybody, do we have a guide on when to put a function implementation in its own submodule and when to leave it in the module? I see that we have `clip` in the module but the others have their own submodules.

Sideboard:
Ok, I can add them. I looked at clip for reference. Thought it was in the main file because of its size. The subfile question extends to the corresponding tests.

Sideboard:
The spec only says "integer" and "another integer" for now.

doc/specs/stdlib_math.md
```diff
@@ -91,6 +91,53 @@ program demo_clip_real
...
+
+```fortran
+program demo_gcd
+  use stdlib_math, only: gcd
+  use stdlib_kinds, only: int32
+  implicit none
+  integer(int32) :: a, b, c
+  
+  a = 48_int32
+  b = 18_int32
```
milancurcic, line 135:
```suggestion
  a = 48
  b = 18
```

Kind suffix not necessary, compiler will cast it automatically.

Sideboard: I took clip as a reference. Probably not the best choice.

Sideboard: I would rather strip all the kinds for the example for simplicity. Or is it recommended to use explicit kinds in the examples? Still use a and b as input variables or simplify to use literals in the call?

milancurcic: I suggest keeping `a` and `b` as declared variables. I would also just use default `integer` without `int32`. Makes the example simpler, and I can't think of a situation where it's problematic (i.e. default `integer` is `int32` on all platforms that I know of).

Sideboard: Ok, I thought there was some underlying theme in the fact that clip uses int32 explicitly and didn't want to break them right next to it.

aslozada:
> I appreciate tips about tests. More, less, special cases …

A test with nested functions, maybe.
`gcd(a,gcd(b,gcd(c,gcd(e,gcd(f,...gcd(y,x))))))`

Sideboard:
> A test with nested functions, maybe.
> `gcd(a,gcd(b,gcd(c,gcd(e,gcd(f,...gcd(y,z))))))`

Does this really test anything new? If `gcd(y, z)` correctly returns `yz` what does `gcd(x, yz)` tell us?

Is there interest in a version of gcd that works with rank-1 arrays for more than two arguments?

Carltoffel:
> Is there interest in a version of gcd that works with a rank-1 array for more than two arguments?

What do you mean? Could you give an example, please?

Sideboard:
> Edit: Something like gcd([20, 30, 40]) -> 10?

Yes, that's what I mean.

ivan-pi:
> General question for everybody, do we have a guide on when to put a function implementation in its own submodule and when to leave it in the module? I see that we have `clip` in the module but the others have their own submodules.

On the long term I think the public modules should contain only the interfaces, and all of the implementation should be placed in sub-modules. This would facilitate compiler vendors providing their own optimized implementations in the (distant?) future. It would also make it easier to swap between debug and release binaries by re-linking. I think the current reasoning was simply when launching a new module with a small function, it was just easier to prepare a single module.

A second reason in favor of separating interface and implementation *might* be to use a stdlib library compiled by another Fortran compiler. In this case you would only need to compile the interface (see https://github.com/fortran-lang/stdlib/pull/530#discussion_r711945701 and the discussion that follows). As an example, Intel MKL ships libraries that can be used with both Intel and gfortran compilers.

@milancurcic, should I make a new issue to discuss this? We can then edit the developer docs in a new PR.

milancurcic:
Yes @ivan-pi, thanks, that would be good.

Sideboard:
Should anything be changed for this PR?

src/stdlib_math.fypp
```diff
@@ -292,4 +298,27 @@ contains
     end function clip_${k1}$
 
     #:endfor
+
+    #:for k1, t1 in INT_KINDS_TYPES
+    !> Determine the greatest common divisor of two integers
+    !> ([Specification](../page/specs/stdlib_math.html#gcd))
+    !>
+    !> Version: experimental
```
Aman-Godara, line 306: This part needs to be moved up at `gcd` interface. You can write any comments here specific to this particular procedure like:
```suggestion
    !> Returns greatest common divisor of two integers of kind ${k1}$
```

doc/specs/stdlib_math.md
```diff
@@ -91,6 +91,52 @@ program demo_clip_real
...
+Experimental
+
+#### Class
+
+Elemental function.
+
+#### Argument(s)
+
+`a`: An integer.
+`b`: Another integer.
```
Aman-Godara, line 115: This part needs to be improved a bit.
You can mention `intent` here. If possible try to make it similar to other procedures mentioned in this doc to be consistent.

doc/specs/stdlib_math.md
```diff
@@ -91,6 +91,52 @@ program demo_clip_real
...
+
+#### Examples
+
+##### Example 1:
+
+```fortran
+program demo_gcd
+  use stdlib_math, only: gcd
+  implicit none
+  integer :: a, b, c
```
Aman-Godara, line 131: Should `kind` be mentioned explicitly here to emphasize the point made in line 117?

doc/specs/stdlib_math.md
```diff
@@ -91,6 +91,52 @@ program demo_clip_real
...
+  
+  a = 48
+  b = 18
+  c = gcd(a, b)
+  ! c == 6
```
Aman-Godara, line 136: I am trying to think like a beginner for whom it is the first programming language.

if `==` is accepted as a .true. statement then it is fine, I personally would have preferred `<--` (or any other random text) to indicate that the value of c is 6 because `==` operator, I think, is little ambiguous as it can return `.false.` also. Your call, shouldn't be a problem either way.

Aman-Godara (approved):
Looks great to me! I have left few comments, some of them are just open ended question for everyone. Thank you @Sideboard.

doc/specs/stdlib_math.md
milancurcic: We could also write it something like this:

```
c = gcd(a, b) ! returns 6
```

src/stdlib_math.fypp
Sideboard: I switched back and forth between "returns" and "determines", it's not uniform among the existing procedures as far as I remember.

Is there some guide about what goes where? I did not know if the doc scripts would recognize annotations to the interface declarations.

doc/specs/stdlib_math.md
Sideboard: I am not sure what to write for the a and b. Don't want to repeat the description of the function.

I thought intent would be read from the code itself.

doc/specs/stdlib_math.md
Sideboard: There was explicit kinds here initially but I was asked to remove them since int32 is the default kind.

doc/specs/stdlib_math.md
Sideboard: I find arrows confusing. I like the return version.

src/stdlib_math.fypp
14NGiestas: With "returns" it get's clear that this procedure is expected to return a value.

src/stdlib_math.fypp
```diff
@@ -292,4 +298,27 @@ contains
...
+        ${t1}$ :: rem, tmp
+
+        rem = min(abs(a), abs(b))
+        res = max(abs(a), abs(b))
+        do while (rem /= 0_${k1}$)
+          tmp = rem
+          rem = mod(res, rem)
+          res = tmp
+        end do
+    end function gcd_${k1}$
```
14NGiestas, line 321: If anyone ever wonder about performance in GCD there is a interesting benchmark
[here (in C)](https://hbfs.wordpress.com/2013/12/10/the-speed-of-gcd/). In conclusion it seems that this implementation is clear and fast enough.
PS: There is also the [gmplib's documentation](https://gmplib.org/manual/Greatest-Common-Divisor-Algorithms) that explores other GCD algorithms.

I believe this patch is good to go

14NGiestas (approved):
There is only a minor (non-)issue about the wording in the docs but I believe this is good to go.

ivan-pi (approved):
Thanks @14NGiestas for the interesting link. There was a [nice response](https://lemire.me/blog/2013/12/26/fastest-way-to-compute-the-greatest-common-divisor/) to that blog post from Prof Daniel Lemire (@lemire). 

Pending the small issues in the documentation, I approve merging this.

Sideboard:
Is there a consensus on where the annotation should be, the interface or the implementation?

Sideboard:
Both descriptive "Converts …" and imperative "Return …" can be found among the function annotations.

14NGiestas (approved):

Aman-Godara (approved):

Aman-Godara:
@milancurcic this still needs one more approval to be merged.

src/tests/math/test_stdlib_math.f90
```diff
@@ -95,4 +95,22 @@ program test_stdlib_math
...
+
+    ! gcd function
+    ! testing format: check(gcd(a, b) == correct answer)
+    call check(gcd(0_int8, 0_int8) == 0_int8, 'gcd failed for gcd(0, 0)', warn=.true.)
+    call check(gcd(2_int8, 0_int8) == 2_int8, 'gcd failed for gcd(2, 0)', warn=.true.)
+    call check(gcd(3_int8, 3_int8) == 3_int8, 'gcd failed for gcd(3, 3)', warn=.true.)
+    call check(gcd(9_int8, 6_int8) == 3_int8, 'gcd failed for gcd(9, 6)', warn=.true.)
+    call check(gcd(6_int8, 9_int8) == 3_int8, 'gcd failed for gcd(6, 9)', warn=.true.)
+    call check(gcd(-9_int8, 6_int8) == 3_int8, 'gcd failed for gcd(-9, 6)', warn=.true.)
+    call check(gcd(9_int8, -6_int8) == 3_int8, 'gcd failed for gcd(9, -6)', warn=.true.)
```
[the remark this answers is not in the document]
Sideboard: I added (0, -2). Now I see that it's not exactly what was asked for but probably close enough.

Sideboard:
I thought the approvals were invalidated when I pushed to the branch but a grey check mark means that the reviewer does not have write-access necessary for the merge. Should I request a review from someone with those rights? Who?

milancurcic:
Thanks all, let's merge it.

PR #543, opened 2021-09-29
Fix string concat

Carltoffel:
Found a bug in `string_type` concat where `lhs` and `rhs` were the wrong way around. Added proper tests and fixed it.

### DIFF
--- a/src/tests/string/test_string_operator.f90
+++ b/src/tests/string/test_string_operator.f90
@@ -103,9 +103,11 @@ contains
 
         a = "a"
         b = "b"
-        call check("a" //  b  == "ab")
-        call check( a  // "b" == "ab")
-        call check( a  //  b  == "ab")
+        call check( "a" //  b  == "ab" )
+        call check( a  // "b" == "ab" )
+        call check( a  //  b  == "ab" )
+        call check( a  //  ""  == "a" )
+        call check( ""  //  b  == "b" )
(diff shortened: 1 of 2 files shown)


ivan-pi (approved):
Thanks for catching this.

awvwgk (approved):
Nice catch, thanks.

src/tests/string/test_string_operator.f90
```diff
@@ -99,11 +99,13 @@ subroutine test_ne
...
+        type(string_type) :: a, b
 
-        string = "Hello, "
-        string = string // "World!"
-        call check(len(string) == 13)
+        a = "a"
+        b = "b"
+        call check("a" //  b  == "ab")
+        call check( a  // "b" == "ab")
+        call check( a  //  b  == "ab")
```
Aman-Godara, line 108:
```suggestion
        call check( "a" //  b  == "ab" )
        call check( a  // "b" == "ab" )
        call check( a  //  b  == "ab" )
        call check( a  //  ""  == "a" )
        call check( ""  //  b  == "b" )
```

Added test cases for empty character scalars (`""`).

Carltoffel:
Nice addition! I wonder if the `char // char` test is necessary, since an intrinsic operator is tested here.

milancurcic:
Thanks!

PR #552, opened 2021-10-10
fixed bug in stringlist

aman-godara:
If the inserted `slist` has length 0, then `insert_at` can give **seg-fault error** in this code:

```fortran
do i = work_idxn + post_length - pre_length, post_length
         list%stringarray(inew) = slist%stringarray(i)
         inew = inew + 1
end do
```

Resolved by: adding an **if-else** statement to check for length of `slist`

### DIFF
--- a/src/stdlib_stringlist_type.f90
+++ b/src/stdlib_stringlist_type.f90
@@ -643,7 +643,7 @@ contains
 
         integer                                         :: i
         integer                                         :: work_idxn, inew
-        integer                                         :: pre_length, post_length
+        integer                                         :: pre_length, post_length, temp
 
         pre_length  = slist%len()
         if (pre_length > 0) then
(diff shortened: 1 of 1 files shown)


jvdp1:
What is the status of this PR? How can we help to move forward?

Aman-Godara:
@jvdp1 Thanks for contacting. PR is stuck here, I tried many different things. I have mentioned the details of the problem [here](https://github.com/fortran-lang/stdlib/pull/552#issuecomment-945098369).

I am planning to make these test cases readable for others (I should have thought to do that earlier).
But I don't understand why all test cases passes on my GCC compiler but not here on GitHub. This is also preventing me from diving deep into the issue because I have to make a commit everytime and push it to GitHub to see if it works or not. That's why you will see those "temp commits" in the logs.

jvdp1:
Strange...
I also tested [this commit bb30a55a53b3728232727a3afdce168900b0be66](https://github.com/fortran-lang/stdlib/pull/552/commits/bb30a55a53b3728232727a3afdce168900b0be66) using Fedora with GCC 11.2.0, and all tests passed. 
I'll try to look further in the next days...

Aman-Godara:
Any luck? @jvdp1

jvdp1:
> Any luck? @jvdp1

No :( And then the holidays arrived. It is also difficult to run tests as I don't have a MacOS computer. Anyone with a MaCOS to test this issue?

PR #556, opened 2021-11-01
fix some FORD links

jvdp1:
I was reading and testing the module `stdlib_stringlist_type.f90` when I found a few issues related to FORD links. These should fix some of them


awvwgk (approved):

gareth-nx (approved):
Thanks for that.

awvwgk:
With two approvals I'm going ahead and merge.

PR #558, opened 2021-11-04
Pin specific fpm version

### LINKED ISSUE #557: Problems with FPM and updating a PR
I have submitted a PR for hash functions and pushed my initial code a couple of weeks ago with no problems. I then found some problems with a module and pushed the correction. This led to a message with the header "[wclodius2/stdlib] Run failed: fpm-deployment - hash_functions (7be16f3)". The error report is for 
Install fpm latest release

with the detailed message

Run fortran-lang/setup-fpm@v3
10
use-haskell: false
11
fpm-version: latest
12
fpm-repository: https://github.com/fortran-lang/fpm
13
This platform is linux
14
Fetching fpm from https://github.com/fortran-lang/fpm/releases/download/current/fpm-current-linux-x86_64
15
Error: Error while trying to fetch fpm - please check that a version exists at the above release url.
16
undefined
17
Error: The "path" argument must be of type string. Received type undefined

It looks like FPM has become required since the initial push. I can see three issues in the error report:
1. It is unable to find the specified version of fpm at the specified path.
2. The specified version is for Linux while I do development in MacOSX.
3. The path argument is not specified to be a string (although it seems to show a path in the report)

What can I do to fix it on my end? Is there something that needs to be fixed in the use of fpm by the repository?

awvwgk:
Closes #557 
Reported in https://github.com/fortran-lang/setup-fpm/issues/6

### DIFF
--- a/.github/workflows/fpm-deployment.yml
+++ b/.github/workflows/fpm-deployment.yml
@@ -33,7 +33,7 @@ jobs:
     - name: Install fpm latest release
       uses: fortran-lang/setup-fpm@v3
       with:
-        github-token: ${{ secrets.GITHUB_TOKEN }}
+        fpm-version: 'v0.4.0'
 
     - name: Run fpm test ⚙
       run: |


zoziha (approved):
Looks good. ❤

LKedward (approved):
Cheers

PR #562, opened 2021-11-12
Minor update `pure/elemental` in `string_type` module

zoziha:
- [x] `pure elemental -> elemental`: ✔
- interface `string_type` procedure
  - `new_string`
  - `new_string_from_integer`
  - `new_string_from_logical`
- [x] ~~`pure -> elemental`: ❓(Update: ❌)~~
- ~~interface `char` procedure~~
  - ~~`char_string`~~
  - ~~`char_string_range`~~
- ~~private function `maybe`~~
- [x] add `elemental`: ❓(Update: ✔)
- interface `move` procedure
  - `move_string_string`
- [x] add `pure`: ✔
- interface `move` procedure
  - `move_char_string`
  - `move_char_string`
  - `move_char_char`

#### Description
Here I have updated the elemental attributes of the routine. There are two areas that are doubtful ❓, but I think they can be `elemental`.

#### Another proposal

There is another issue. 
`string_type` interface relies on the `to_string` routine, it can convert `character`, `integer`, and `logical` variables into strings and **be elemental** because `type(string_type)` can be non-allocatable, but it does not support the `format` argument. 
While `to_string` supports `integer`, `real`, `complex` and `logical` types, support `format` argument, but not elemental.

`string_type` seems to have a **elemental advantage** to construct strings. It seems that we can take advantage of this and let `string_type` add support for `real` and `complex` types, and support the `format` argument?

### DIFF
--- a/doc/specs/stdlib_string_type.md
+++ b/doc/specs/stdlib_string_type.md
@@ -528,7 +528,7 @@ Return the character sequence represented by the string.
 
 #### Class
 
-Elemental function.
+Pure function.
 
 #### Argument
 
@@ -618,7 +618,7 @@ Return a substring from the character sequence of the string.
 
 #### Class
 
-Elemental function.
+Pure function.
 
 #### Argument
 
--- a/doc/specs/stdlib_string_type.md
+++ b/doc/specs/stdlib_string_type.md
@@ -528,7 +528,7 @@ Return the character sequence represented by the string.
 
 #### Class
 
-Pure function.
+Elemental function.
 
 #### Argument
 
@@ -618,7 +618,7 @@ Return a substring from the character sequence of the string.
 
 #### Class
 
-Pure function.
+Elemental function.
 
 #### Argument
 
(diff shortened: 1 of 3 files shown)


milancurcic:
> ￼ pure -> elemental: ￼
> interface char procedure
> char_string
> char_string_range
> private function maybe
> ￼ add elemental: ￼
> interface move procedure
> move_string_string

I also think both of these are safe to be `elemental`. In the first case we go from `character` to `string_type`, so an input `character` array will need to be fixed length no matter what. In the second case it's all `string_type` to `string_type` so I don't see an issue.

Others are cosmetic and straightforward, so I think this PR can go forward.

milancurcic (approved):
Thank you!

awvwgk (changes requested):
There are some limitations in `char_string` and `char_string_range` which lead me to use `pure` rather than `elemental`. Namely, there is an issue with `elemental` which can result in an uncatchable runtime error. While useful, the subtle consequences of `elemental` usage might lead to hard to debug runtime errors for some use cases.

zoziha:
Thanks @awvwgk , I understand what you mean now.
`char_string, char_string_range, maybe`, I kept their `pure` attributes, and add test code for `elemental` `move` of `type(string_type)`.

awvwgk (approved):
Thanks, looks good to me.

PR #579, opened 2021-11-25
Add module for handling version information of stdlib

### LINKED ISSUE #576: stdlib version information function
### Motivation

It would be good to have a function `stdlib_version()`

```
function stdlib_version() result(r)
    character(len=:), allocatable :: r
    r = "v0.1.0"
end function
```

This information is useful for external build tools, and for bug reporting.

### Prior Art

_No response_

### Additional Information

_No response_

awvwgk:
Closes #576


doc/specs/stdlib_version.md
```diff
@@ -0,0 +1,64 @@
...
+---
+
+# The `stdlib_version` module
+
+[TOC]
+
+## Introduction
+
+The `stdlib_version` module contains the version of the standard library.
+The version information an be used as compile time constant or retrieved from a getter function at runtime.
```
ivan-pi, line 12:
```suggestion
The version information can be used as compile time constant or retrieved from a getter function at runtime.
```

ci/fpm.toml
```diff
@@ -1,5 +1,5 @@
 name = "stdlib"
-version = "0.1.0"
+version = "VERSION"
```
ivan-pi, line 2: Is this to be understood as a reference to the version file? Wouldn't this complicate the version retrieval in the pretty output from fpm @LKedward is working on in https://github.com/fortran-lang/fpm/pull/622?

API-doc-FORD-file.md
```diff
@@ -9,6 +9,9 @@ media_dir: doc/media
 fpp_extensions: fypp
 preprocess: true
 macro: MAXRANK=3
+       PROJECT_VERSION_MAJOR=0
+       PROJECT_VERSION_MINOR=0
+       PROJECT_VERSION_PATCH=0
```
ivan-pi, line 14:
```suggestion
       PROJECT_VERSION_MAJOR=0
       PROJECT_VERSION_MINOR=1
       PROJECT_VERSION_PATCH=0
```

Makefile.manual
```diff
@@ -4,6 +4,11 @@ FC ?= gfortran
 FFLAGS ?= -Wall -Wextra -Wimplicit-interface -fPIC -g -fcheck=all
 FYPPFLAGS ?=
 
+FYPPFLAGS += \
+             -DPROJECT_VERSION_MAJOR=$(shell cut -d. -f1 VERSION) \
+             -DPROJECT_VERSION_MINOR=$(shell cut -d. -f2 VERSION) \
+             -DPROJECT_VERSION_PATCH=$(shell cut -d. -f3 VERSION)
+
```
ivan-pi, line 11: Is the `shell cut` expected to run on Windows machines too?

ci/fpm.toml
awvwgk: No, the fpm model will see no difference between a version number specified in the manifest and a version number read from a file, since the deserializer will normalize this information. Using a version file will ensure we keep the numbers in sync everywhere.

API-doc-FORD-file.md
ivan-pi: Or is this overridden during the fypp processing?

awvwgk: This is supposed to be a stub, similar like the `MAXRANK=3` argument above, which would would not allow to actually compile stdlib.

doc/specs/stdlib_version.md
```diff
@@ -0,0 +1,64 @@
...
+In case the standard library is dynamically linked, the version number retrieved from the getter might mismatch the compile time constants provided from the version built against.
+Therefore, it is recommended to retrieve the version information always at runtime.
+
+
+## Constants provided by `stdlib_version`
+
+### `stdlib_version_string`
+
+String contant representing the version number.
```
ivan-pi, line 21:
```suggestion
String constant representing the version number.
```

Makefile.manual
awvwgk: Can the manual makefile run on Windows? I always thought this is a Unix-only thing.

ivan-pi (approved):
The CMake syntax to read a list is just terrible.

Pending the small issues, LGTM. Thanks.

Makefile.manual
ivan-pi: The Quickstart Fortran installer comes with Make included, but I haven't tested it. I remember your advice was ninja + CMake is the better approach for Windows. Personally I'm fine leaving as it is until things break.

ci/fpm.toml
ivan-pi: Okay, I've found it's mentioned here: https://github.com/fortran-lang/fpm/blob/main/manifest-reference.md#project-version

Makefile.manual
awvwgk: Windows is particular in this case, the `shell` command might not work at all if no POSIX shell is available on the platform, but it might pickup the git-bash which is also included in the installer and could contain other small Unix command line utilities.

I can have a look into parsing files in pure make, which should be possible, but the result might look even more convoluted than the CMake version string splitting ;)

brocolis: `shell cut` fails for me on Windows cmd.exe and Msys2, even with `cut.exe` in PATH

```
fypp -DMAXRANK=4 stdlib_optval.fypp stdlib_optval.f90
common.fypp:4: error: exception occured when setting variable(s) 'PROJECT_VERSION' to '"{}.{}.{}".format(PROJECT_VERSION_MAJOR, PROJECT_VERSION_MINOR, PROJECT_VERSION_PATCH)' [FyppFatalError]
error: name 'PROJECT_VERSION_MAJOR' is not defined [NameError]
gnumake[1]: *** [Makefile.manual:74: stdlib_optval.f90] Error 1
```

doc/specs/stdlib_version.md
```diff
@@ -0,0 +1,64 @@
...
+#### Example
+
+```fortran
+program demo_version
+    use stdlib_version, only : get_stdlib_version
+    implicit none
+    character(len=:), allocatable :: version
+    call get_stdlib_version(string=version)
+    print '(a)', version
+end program demo_to_lower
```
brocolis, line 63:
```suggestion
end program demo_version
```

brocolis (commented):
@awvwgk many thanks for this PR. I tested it on Windows using CMake+Ninja.
Regarding the issue with $(shell cut ..), I'm trying to find a solution but no luck yet.

brocolis:
OK I found a combination that works for me on Windows

``` 
set FC=gfortran
set FYPPFLAGS=-DMAXRANK=4
gnumake -f Makefile.manual
```

Thanks @awvwgk

brocolis (approved):
Thanks @awvwgk

brocolis:
I just noticed `ubuntu-latest` CI is failing at `make -f Makefile.manual FYPPFLAGS="-DMAXRANK=4" -j`. It is the same issue. If we remove `FYPPFLAGS="-DMAXRANK=4"` from the command line and set it as environment variable, it will work, I think. That's how I did on my Windows.

Makefile.manual
awvwgk: Found a solution using the make commands `file`, `subst` and `word`.

API-doc-FORD-file.md
ivan-pi: Don't these macros get applied when FORD runs fypp to preprocess the sources for the documentation? At least that's what I understand from here: https://github.com/Fortran-FOSS-Programmers/ford/wiki/Project-File-Options

Makefile.manual
```diff
@@ -4,6 +4,12 @@ FC ?= gfortran
 FFLAGS ?= -Wall -Wextra -Wimplicit-interface -fPIC -g -fcheck=all
 FYPPFLAGS ?=
 
+VERSION := $(subst ., ,$(file <VERSION))
```
ivan-pi, line 7:
```suggestion
VERSION := $(subst ., ,$(file < VERSION))
```
This is an input redirection right?

awvwgk: There is no space used in the make docs: https://www.gnu.org/software/make/manual/make.html#File-Function

API-doc-FORD-file.md
awvwgk: It does show up in the final docs for the compile time constant:

image

Otherwise the value set won't affect the output of the documentation. I would prefer to leave this as a stub, so we don't have to worry about it becoming outdated when we release a new version.

doc/specs/stdlib_version.md
```diff
@@ -0,0 +1,64 @@
...
+#### Class
+
+Pure subroutine.
+
+#### Argument
+
+`major`: shall be an intrinsic integer type. It is an optional, `intent(out)` argument.
+`minor`: shall be an intrinsic integer type. It is an optional, `intent(out)` argument.
+`patch`: shall be an intrinsic integer type. It is an optional, `intent(out)` argument.
+`string`: shall be a deferred length character type. It is an optional, `intent(out)` argument.
```
ivan-pi, line 52: Just wondering if a fixed-length string can also be used to collect the output?

ivan-pi (approved):

Makefile.manual
ivan-pi: There seems to be a space here in the syntax definition `$(file op filename[,text])`, where op is one of >, <, <<. This is a nit-pick.

doc/specs/stdlib_version.md
awvwgk: The dummy argument is currently `character(len=:), allocatable, intent(out), optional`, since we are using a subroutine rather than a function. The rationale is that we want a stable API *and* ABI for this procedure, we could also go for multiple functions for getting the individual values, to make use of it in a more functional way.

API-doc-FORD-file.md
ivan-pi: Comments are not allowed in the FORD meta-data section. Is there somewhere we could document these values are stubs, and don't reflect the actual library?

brocolis (approved):
@awvwgk looks good.

ivan-pi:
@jvdp1, @milancurcic, can one of you give this a third approval before merging?

doc/specs/stdlib_version.md
```diff
@@ -0,0 +1,64 @@
...
+---
+
+# The `stdlib_version` module
+
+[TOC]
+
+## Introduction
+
+The `stdlib_version` module contains the version of the standard library.
+The version information can be used as compile time constant or retrieved from a getter function at runtime.
```
jvdp1, line 12: Not sure about this, just a feeling:
```suggestion
The version information can be used as a compile time constant or retrieved from a getter function at runtime.
```

doc/specs/stdlib_version.md
```diff
@@ -0,0 +1,64 @@
...
+In case the standard library is dynamically linked, the version number retrieved from the getter might mismatch the compile time constants provided from the version built against.
+Therefore, it is recommended to retrieve the version information always at runtime.
```
jvdp1, line 14: Why not only allow the correct approach to do it?

jvdp1 (approved):
LGTM. Just a couple of minor comments.
Thank you!

doc/specs/stdlib_version.md
awvwgk: For the compact version I can see an application to verify the compatibility of API and/or ABI at runtime by a construct like:
```f90
use stdlib_version, only : stdlib_version_compact, get_stdlib_version
implicit none
integer :: major
call get_stdlib_version(major=major)
if (major > stdlib_version_compact / 10000) then
  error stop "Incompatible version of stdlib used"
end if
end
```

Similarly, `stdlib_version_string` and `get_stdlib_version(string=string)` can be used to provide a printout to inform about the stdlib versions used at build time and run time.

Not sure whether this is a realistic application of this information.

jvdp1: Ok. Thank you for the explanation. You can merge this PR from my point-of-view.

milancurcic (approved):
Very nice addition, thank you.

Would it be useful for this module to also provide compile-time constants of compiler version and options used to build it?

```
module m
  use iso_fortran_env, only: compiler_version, compiler_options
  character(*), parameter :: stdlib_compiler_version = compiler_version()
  character(*), parameter :: stdlib_compiler_options = compiler_options()
end module m

program p
  use m
  print *, stdlib_compiler_version
  print *, stdlib_compiler_options
end
```

It works nicely with GFortran, but ifort 2021.3.0 doesn't seem to allow these intrinsic in constant expressions:

```
$ ifort test.f90 
test.f90(3): error #6263: This intrinsic function is invalid in constant expressions.   [COMPILER_VERSION]
  character(*), parameter :: stdlib_compiler_version = compiler_version()
-------------------------------------------------------^
test.f90(4): error #6263: This intrinsic function is invalid in constant expressions.   [COMPILER_OPTIONS]
  character(*), parameter :: stdlib_compiler_options = compiler_options()
-------------------------------------------------------^
test.f90(8): error #7013: This module file was not generated by any release of this compiler.   [M]
  use m
------^
compilation aborted for test.f90 (code 1)
```

which I think is incorrect. Anyhow, just an idea, something we could tackle in another PR.

awvwgk:
Good idea, I opened a new issue for this one.

Issue #229, opened 2020-09-08
ANSI Colors support


14NGiestas:
## Colors Module
In order to support colors and styling (bold, underline and etc) I propose a small utility to handle colored output, namely `stdlib_colors`. The use cases can vary, but the main application would be the richer logger and errors which helps a lot to visually identify what kind of message is right away.

![image](https://user-images.githubusercontent.com/14851487/92472268-9df6f600-f1af-11ea-9be9-d9e1ef57379d.png)

[Figure 1 - coloredlogs python library showing a use case](https://pypi.org/project/coloredlogs/)

## Few remarks from the community in #193 
> @mobius-eng commented on 20 May
> (...)
> Julia (and Python and R) are meant to be used interactively. Thus, colors play an important role with the output to the terminal. By contrast, Fortran programs are meant to be run in the "batch mode" with log going to the file.

> @aradi commented on 24 May
> As for colors: I am fine, as long as they are optional. I think most people typically run Fortran programs in batch systems with redirected output. Seeing a lot of color control sequences when opening the output files is rather disturbing...

> @jvdp1 commented on 23 May
> (...)
> Colors are not needed for me.

> @wclodius2
> A problem with using colors is that `output_unit` can be directed to a file where the "color codes” will be distracting.

## My hack and test
- Well, gnu utils have support to colored output for eons and when it gets redirected it handles correctly, how is that possible? So I did some research in order to figure out how they are made and found [this](https://unix.stackexchange.com/questions/314324/redirect-output-of-a-program-to-a-file-fails/314369#314369).
- So basically we test if there is a teletypewriter device (A interactive terminal) connected with the `isatty` function (from `<unistd.h>`), if `isatty` returns true then we write the ANSI colors, otherwise no colors in the output.
- A lot of compiler vendors support this function as a extension, however is not a standard so we can't rely on it. That being said I think it should be included in the `stdlib_system` module as a new function binding the C one (and alternatives to other platforms).
- In order to show that it works I made a small change in a program to output colors from the fortran-wiki and its [here](https://gist.github.com/14NGiestas/6cb7522441dd81bd221776f70189d5f5)

## Some other related projects that could inspire the API
- [FACE](https://github.com/szaghi/FACE) by @szaghi
- [Foul](http://foul.sourceforge.net/), as pointed by @arjenmarkus]

certik:
I think something along the `FACE` and `Foul` libraries would be good for stdlib. I like using colors a lot in a terminal, it helps.

[...]

PR #580, opened 2021-11-27
Add terminal and color escape sequences

awvwgk:
- implement low-level API for color support
- style, foreground and background color enumerators
- ~~true color (24-bit) types~~
- generation of escape strings via to_string function

Questions:
- how to test this?
- aim for compile time conversion?

Related
- https://github.com/fortran-lang/stdlib/issues/229
- https://github.com/awvwgk/fortty/blob/main/src/escape.f90
- https://github.com/urbanjost/M_attr/blob/main/src/M_attr.f90
- https://github.com/jupyter-xeus/cpp-terminal/blob/master/cpp-terminal/base.hpp


14NGiestas:
Looking good.
Testing if they produce the expected code escapes should be enough since this is a low-level API,
users should be aware that no checks are done to ensure that the terminal will support colors.

src/tests/terminal/test_colors.f90
```diff
@@ -0,0 +1,106 @@
+! SPDX-Identifier: MIT
+
+module test_colors
+    use stdlib_colors
```
certik, line 4: I think we should be using explicit imports, that is, list all the things to import here.

certik:
I am personally not a fan using derived types. The ideal for this would be to use enumerations (@fortranfan would probably approve:), in fact that's what we ended up using in C++ that you linked above.

We also have to initialize the terminal (especially on Windows).

I don't know if there is a way to simplify this somehow. If not, then the present approach is fine.

certik:
One question: this module talks about "colors", but these are specifically terminal colors, and in fact only "ansi escape sequences", but it seems all terminals now converges to support those, including Windows.

certik:
And since we require to setup a terminal (on Windows) even to just print colors, we might as well use this machinery for other terminal things, such as the goal of the `cpp-terminal` library. Then we can write good terminal applications. 

I think this is within scope of stdlib.

ivan-pi:
I am not a user of terminal colors (yet), however upon reviewing the module I was wondering if abbreviating the components as `r, g, b` would be just as clear but more concise. This might be just a personal preference, but I think it gives a nice clean look when initializing a color using the structure constructor:

```fortran

type(fg_color24) :: my_color

my_color = fg_color24(r=34, g=12, b=10)   ! vs
my_color = fg_color24(red=34, green=12, blue=10)
```

ivan-pi:
I am not a user of terminal colors (yet), however upon reviewing the module I was wondering if abbreviating the components as `r, g, b` would be just as clear but more concise. This might be just a personal preference, but I think it gives a nice clean look when initializing a color using the structure constructor:

```fortran

type(fg_color24) :: my_color

my_color = fg_color24(r=34, g=12, b=10)   ! vs
my_color = fg_color24(red=34, green=12, blue=10)
```

src/stdlib_colors.f90
```diff
@@ -0,0 +1,182 @@
...
+    !> Identifier for the default background color
+    type(bg_color_enum), parameter :: bg_color_default = bg_color_enum(9)
+
+    !> Offset for background color
+    integer, parameter :: bg_offset = 40
+    !> Prefix for background true color
+    character(len=*), parameter :: bg_color24_prefix = "48;2;"
+
+    !> Escape sequence for terminal style and color
+    character(len=*), parameter :: esc = char(27) // "["
```
ivan-pi, line 136:
```suggestion
    character(len=*), parameter :: esc = achar(27) // "["
```

The character in the ascii sequence is supposed to be more portable.

awvwgk:
Thanks for the feedback. Building something like `cpp-terminal` in stdlib for Fortran would be awesome, eventually we could just use `stdlib_terminal` for this purpose. The color escape codes would be the first step (maybe as `stdlib_terminal_colors`) to define our low-level API for handling escape sequences.

I usually start building this kind of projects outside of stdlib and than notice that simple things like `to_string` for integer output are something I'm already missing. Usually, I'm just copying the snippet, but I figured it would be better to start doing things properly and just contribute such things directly to stdlib.

ivan-pi:
> I am personally not a fan using derived types. The ideal for this would be to use enumerations (@FortranFan would probably approve:), in fact that's what we ended up using in C++ that you linked above.

I don't think the fact Fortran doesn't support proper enums should be a show stopper. Here is a small quote from the Lua community: 

> Our motto in the design of Lua has always been "mechanisms instead of policies." By policy, we mean a methodical way of using existing mechanisms to build a new abstraction. Encapsulation in the C language provides a good example of a policy. The ISO C specification offers no mechanism for modules or interfaces. Nevertheless, C programmers leverage existing mechanisms (such as file inclusion and external declarations) to achieve those abstractions.

Derived types offer a nice and type-safe way to organize code, even if they might lead to reduced performance in some cases. 
I don't think this usage case is so performance-critical, that derived types would represent an issue. Moreover, when the derived types are default-constructed and used as parameters, an optimizing compiler can just copy the value and insert it in the right place of the client code (at least the NAG compiler is known to do this).

For ANSI Escape sequences specifically, the other two options I see available are 1) using a character string, like the [M_attr](https://github.com/urbanjost/M_attr/blob/608b9d3b40be9ff2590c23d2089781fd4da76344/src/M_attr.f90#L237) module does, or 2) use an array of 3 integers. (In principle you could encode all 24 bits in a single integer, but I think it makes the code less clear.) 

Would it be feasible to overload the `//` operator for the foreground and background color types?

awvwgk:
> Would it be feasible to overload the `//` operator for the foreground and background color types?

I actually thought about overloading `operator(+)` to create combinations from `style`, `fg_color` and `bg_color`. This will create a bit of combinatorics blowup if not designed properly, therefore I just wen with the simple variant first to get the discussion going.

ivan-pi:
Here's the example @certik posted in #229:
```cpp
 std::string text = "Some text with "
            + color(fg::red) + color(bg::green) + "red on green"
            + color(bg::reset) + color(fg::reset) + " and some "
            + color(style::bold) + "bold text" + color(style::reset) + ".";
        std::cout << text << std::endl;
```

The same can be achieved in Fortran. I guess using `operator(+)` between the colors and styles makes more sense, however I'd still go for `//` to attach the escape sequence to a string. I'd also leave it unsymmetric, i.e. the escape sequence is always the left operand:

```
type(fg_color) :: green, red

print *, green // "Sentence in green"   ! valid
print *, "Red sentence" // red          ! invalid 
```

doc/specs/index.md
```diff
@@ -27,7 +27,8 @@ This is and index/directory of the specifications (specs) for each new module/fe
  - [stats_distributions_uniform](./stdlib_stats_distribution_uniform.html) - Uniform Probability Distribution
  - [string\_type](./stdlib_string_type.html) - Basic string support
  - [strings](./stdlib_strings.html) - String handling and manipulation routines
- - [stringlist_type](./stdlib_stringlist_type.html) - 1-Dimensional list of strings
+ - [stringlist_type](./stdlib_stringlist_type.html) - 1-Dimensional list of
```
14NGiestas, line 30:
```suggestion
 - [stringlist_type](./stdlib_stringlist_type.html) - 1-Dimensional list of strings
```
I believe you didn't mean to remove "strings" in this line

certik:
> I don't think the fact Fortran doesn't support proper enums should be a show stopper.

I agree. All I am advocating for is to consider and search for the simplest solution that involves the least amount of "layers" (such as derived types or even more OO approach). Btw, I am not convinced the C++ I approach I took is the best either. 

I guess the only approach I can think of is to have functions like `color_fg_bright_3bit` and it would accept an integer and we define integer constants like `color_red`, but it's easy to pass the wrong integer in. It's less safe than the current approach with derived types I think.

That's why I CCed @fortranfan, because this use case is far from unique, I think this is actually quite common.

urbanjost:
Since Fortran now supports (some more than others) everything from procedural to functional to OOP programming there are a lot of options and all of us have different approaches (some of mine and others discussed in M_esc and to a lesser extent M_attr) but if going with OOP and user defined types I might want a string that can have attributes (bg color, fg color, blink, underline, ...) that I can set that I would otherwise use like the STRING type in stdlib but that on output I could _optionally_ print with all the attributes applied; albeit if putting out a lot of short strings with similar attributes it would take some work to not produce redundant output when all the strings have a common attribute.

  I think some example programs solving some common use cases would help.

Looks good given the approach taken, but unless you want to expand this to allow building panels and./or ASCII graphics it seems a little overkill; but since it works and something is needed and it has the attribute of being expandable  to include positioning and clearing I am good with it. 

Having been using an ncurses interface and more recently M_attr I am having a hard time getting through the "not what I am used to" roadblock while trying it, but everything I have tried so far has worked.

 I personally need something I can easily read and write from an external file and turn off so I need more of an abstraction of the attributes I can easily embed in text files; but I think a lot of people just want an easy way to color some text and this works for that so I am good with this.

FortranFan:
@certik wrote Nov. 28, 2021 11:52 AM EDT:

> I guess the only approach I can think of is to have functions like `color_fg_bright_3bit` and it would accept an integer and we define integer constants like `color_red`, but it's easy to pass the wrong integer in. It's less safe than the current approach with derived types I think.
> 
> That's why I CCed @FortranFan, because this use case is far from unique, I think this is actually quite common.

Thanks @certik.

I agree the use case here is hardly unique: scientific and technical computing is replete with the need to work with **named constants** with particular **scopes** in **type-safe** manner even in compute-intensive sections of codebases.

The need here to work with color codes for terminal escape sequences is but one instance that some might consider is outside of hard-code number crunching but it should not trivialized.

Nonetheless **proper ENUMs** are a bridge way too far for Fortran, the solution in the next revision Fortran 202X is far from optimal.

Thus moving ahead with this Fortran stdlib item in sync with all your consensus is as good as it can get.

jvdp1:
@awvwgk is this ready for review?

awvwgk:
I think we have agreed that terminal escape sequences are generally in-scope for stdlib. The question now is how make best use of them in an actual application and I don't really have a definite answer, this touches points like:

1. how to represent a color (string, enum, integer, derived type, ...)
2. applying color to strings (function, binary/unary operator, ...)
3. ways to enable/disable color support at runtime (atty detection, ...)

I haven't touched point 3 here in this PR, but I think this is one of the most important points to cover. However, I don't think there is a best answer.

The most practical solution I was able to come up with is [here](https://github.com/awvwgk/pretty-diagnostics/blob/071792f468a014c5f054f6afa091c06f42ac81a2/src/diagnostic/color.f90#L116-L190), using a derived type to hold all escape sequences and initialize them with empty string if color support is disabled, while deferring the decision of color usage to the user of the library. But I doubt this is the best strategy, since it requires explicit initialization.

14NGiestas:
Point 3 is not a blocking issue for this PR to go IMO, it produces all the code escapes needed to color a text. We can just let the user of the module deal with it and provide helpers later on (E.g: a portable `isatty` function).
Some other options that do not involve a higher-level abstraction: 
- creating a private global variable (module-wise) and a function to turn on / off the colors.
- [using an environment variable](https://no-color.org/) (I don't know if that is portable, however it worth doing it)

jvdp1:
@awvwgk Thank you for the answers. I can't answer your questions. However, I would suggest to finish this PR as is, and ask users some feedbacks regarding these 3 questions. With fpm, people could test it quite easily IMO.

milancurcic (approved):
Thank you for this addition. I reviewed the files and played with it locally. It works as expected and I think it can be merged as is, and we can add any higher-level functionality in later PRs.

I agree with @certik that we should look for the simplest and lowest-level approach, though I think the derived types here are appropriate. Alternatively, the color derived types could also be expressed as `integer :: fg_color24(3)` arrays, but to me the current implementation is more intuitive.

What's not so intuitive to me is that we have two separate types `fg_color24` and `bg_color24` (and corresponding "enum" types) that are otherwise exactly the same. As I understand it, this is designed like this so that a generic `to_string` can figure out whether to emit a background or foreground styling. As a user, I would prefer for colors to be just colors (e.g. type `color24`), but use different functions (`to_string_fg`, `to_string_bg`, etc.) to apply different kind of stylings. In other words, in my mind foregroundness and backgroundness are not properties of a color, but a property of a styling operation. I hope that makes sense. But I also understand this is a matter of style, so I'm not opposed to the current approach.

One question that came up for me is if the color components of the derived types should be `integer(int8)` (as that would make the types truly 24-bit :)). Memory footprint would be four times smaller (are there any use cases where the memory use would be important, i.e. large arrays of color values? I can't think of it. In general graphics sure, but in terminal? I don't know. Maybe somebody will start writing rich-color roguelikes in Fortran). The default constructor would be uglier (e.g. `fg_color24(0_int8, 127_int8, 255_int8)`), but that could be easily fixed by providing a custom type constructor function. Another advantage could be type casting: Currently if you pass values out-of-range (e.g. < 0 or > 255) the style simply won't work but the program would otherwise be correct. But if you pass out-of-range values to `int8`, a compiler could emit helpful warnings if enabled.

milancurcic:
@awvwgk Do you think there is sufficient interest and feedback to wrap up and merge this PR? If yes, I'd be happy to help resolve the merge conflicts.

awvwgk:
I just revisited the API for color escape sequences for a different project and came up with a more robust scheme in https://github.com/tblite/tblite/pull/37. The implementation in question can be found at [`terminal.f90`](https://github.com/tblite/tblite/blob/d432deeb0159699207386b71e2aaa835fe09ddfe/src/tblite/context/terminal.f90), if anyone is interested, feel free to use the linked file under MIT license for stdlib.

14NGiestas (approved):
Now colors are just colors.
We lost the true color (24-bit) types (which I'm fine with).
The only thing a little off is that the derived type is called `ansi_color` but it represents not only a color but an entire `ansi_style` or `ansi_escape`.

Excellent usage example in that asciinema.

src/stdlib_terminal_colors.f90
```diff
@@ -0,0 +1,138 @@
...
+    public :: bg_color_black, bg_color_red, bg_color_green, bg_color_yellow, bg_color_blue, &
+        & bg_color_magenta, bg_color_cyan, bg_color_white, bg_color_default
+
+    public :: to_string, operator(+), operator(//)
+
+
+
+    !> Container for terminal escape code
+    type :: ansi_color
+        private
```
14NGiestas, line 23: There is one issue with intel (and it seems it follows the standard 2008), you can either change this to `public`
or change the lines where the error occurred to initialize each component individually

```
/Users/runner/work/stdlib/stdlib/src/stdlib_terminal_colors_operator.f90(19): error #6053: Structure constructor may not have components with the PRIVATE attribute   [STYLE]
            style=merge(rval%style, lval%style, rval%style >= 0), &
------------^
/Users/runner/work/stdlib/stdlib/src/stdlib_terminal_colors_operator.f90(20): error #6053: Structure constructor may not have components with the PRIVATE attribute   [FG]
            fg=merge(rval%fg, lval%fg, rval%fg >= 0), &
------------^
/Users/runner/work/stdlib/stdlib/src/stdlib_terminal_colors_operator.f90(21): error #6053: Structure constructor may not have components with the PRIVATE attribute   [BG]
            bg=merge(rval%bg, lval%bg, rval%bg >= 0))
```

awvwgk:
```suggestion
```

awvwgk:
I removed the true color support because I didn't use it for my applications so far. Regarding the naming of the type, I'm open for suggestions, the currently chosen in this PR might not be the best choice.

awvwgk:
Fpm project is up at https://github.com/awvwgk/fortty

milancurcic:
It seems like we have a few approvals and no objects, so let's give it a few more days and merge on Monday, May 23, if there are no objections.

Thanks Sebastian and all reviewers!

certik:
As discussed in #658, it seems we all agree to go with the route of "merge much quicker to experimental, and reserve the right to completely change or even remove later". So I am fine to merge this as is and go from there. And for other PRs, let's do this in a matter of days or weeks, not months.

(And my objections above were not meant to derail the merging of this. In light of the above, it now seems to me it is much more important to merge quickly and iterate, rather than wait.)

awvwgk:
> The only thing a little off is that the derived type is called `ansi_color` but it represents not only a color but an entire `ansi_style` or `ansi_escape`.

Just rebased and renamed the derived type to `ansi_code` to better reflect that it can contain a style while keeping the name short (and avoiding British / American English variants).

ivan-pi:
`ansi_code` seems very reasonable given that "ANSI code" redirects to [ANSI Escape Codes](https://en.wikipedia.org/wiki/ANSI_escape_code) on Wikipedia.

doc/specs/stdlib_terminal_color.md
```diff
@@ -0,0 +1,304 @@
...
+
+Support terminal escape sequences to produce styled and colored terminal output.
+
+
+## Derived types provided
+
+
+### ``ansi_code`` type
+
+The ``ansi_code`` type represent an ANSI escape sequence with a style, forground
```
ivan-pi, line 20:
```suggestion
The ``ansi_code`` type represent an ANSI escape sequence with a style, foreground
```

doc/specs/stdlib_terminal_color.md
```diff
@@ -0,0 +1,304 @@
...
+```
+
+
+### ``operator(//)``
+
+Concatenate an escape code with a string and turn it into an actual escape sequence
+
+#### Syntax
+
+`code = lval + rval`
```
ivan-pi, line 274:
```suggestion
`colored_str = code // str`
```

ivan-pi: I read only later that left and right values could be either codes (in which case // is same as +) or strings. Best give a hint the return type is a string.

doc/specs/stdlib_terminal_color.md
```diff
@@ -0,0 +1,304 @@
...
+
+Pure function.
+
+#### Argument
+
+``lval``: Style, foreground or background code of ``ansi_code`` type or a character string,
+          this argument is ``intent(in)``.
+``rval``: Style, foreground or background code of ``ansi_code`` type or a character string,
+          this argument is ``intent(in)``.
+
```
ivan-pi, line 286: The way this is currently written one might wrongly interpret that codes can be "added" using the the // operator. Only after I checked the interface I realized this was not the case.

I've got no good suggestions how to amend this.

src/stdlib_terminal_colors_to_string.f90
```diff
@@ -0,0 +1,40 @@
...
+            if (code%style > 0 .and. code%style < 10) str = str // ";" // chars(code%style)
+            if (code%fg >= 0 .and. code%fg < 10) str = str // ";3" // chars(code%fg)
+            if (code%bg >= 0 .and. code%bg < 10) str = str // ";4" // chars(code%bg)
+            str = str // "m"
+        else
+            str = ""
+        end if
+    end function to_string_ansi_code
+
+    !> Check whether the code describes any color or is just a stub
```
ivan-pi, line 30:
```suggestion
    !> Check whether the code describes any color / style or is just a stub
```

src/stdlib_terminal_colors_to_string.f90
```diff
@@ -0,0 +1,40 @@
...
+
+    !> Transform a color code into an actual ANSI escape sequence
+    pure module function to_string_ansi_code(code) result(str)
+        !> Color code to be used
+        type(ansi_code), intent(in) :: code
+        !> ANSI escape sequence representing the color code
+        character(len=:), allocatable :: str
+
+        if (anycolor(code)) then
+            str = achar(27) // "[0"  ! Always reset the style
```
ivan-pi, line 20: Maybe introduce a named parameter for `achar(27)` just to make it clear it's the ascii escape code:
```
character(len=1), parameter :: esc = achar(27)
```

src/tests/terminal/test_colors.f90
```diff
@@ -0,0 +1,82 @@
...
+            ]
+    end subroutine collect_colors
+
+    subroutine test_fg_color(error)
+        !> Error handling
+        type(error_type), allocatable, intent(out) :: error
+        character(len=:), allocatable :: str
+
+        str = to_string(fg_color_red)
+        call check(error, ichar(str(1:1)), 27)
```
ivan-pi, line 28:
```suggestion
        call check(error, iachar(str(1:1)), 27)
```

src/tests/terminal/test_colors.f90
```diff
@@ -0,0 +1,82 @@
...
+        call check(error, str(2:), "[0;31m")
+    end subroutine test_fg_color
+
+    subroutine test_bg_color(error)
+        !> Error handling
+        type(error_type), allocatable, intent(out) :: error
+        character(len=:), allocatable :: str
+
+        str = to_string(bg_color_yellow)
+        call check(error, ichar(str(1:1)), 27)
```
ivan-pi, line 39:
```suggestion
        call check(error, iachar(str(1:1)), 27)
```

src/tests/terminal/test_colors.f90
```diff
@@ -0,0 +1,82 @@
...
+        call check(error, str(2:), "[0;43m")
+    end subroutine test_bg_color
+
+    subroutine test_style(error)
+        !> Error handling
+        type(error_type), allocatable, intent(out) :: error
+        character(len=:), allocatable :: str
+
+        str = to_string(style_bold)
+        call check(error, ichar(str(1:1)), 27)
```
ivan-pi, line 50:
```suggestion
        call check(error, iachar(str(1:1)), 27)
```

doc/specs/index.md
```diff
@@ -33,6 +33,8 @@ This is and index/directory of the specifications (specs) for each new module/fe
  - [string\_type](./stdlib_string_type.html) - Basic string support
  - [stringlist_type](./stdlib_stringlist_type.html) - 1-Dimensional list of strings
  - [strings](./stdlib_strings.html) - String handling and manipulation routines
+ - [stringlist_type](./stdlib_stringlist_type.html) - 1-Dimensional list of strings
```
14NGiestas, line 36:
```suggestion
```
It's a duplicate of line 34

ivan-pi (commented):
A few comments:
1) I'm not 100% convinced by the naming of the module and constants. The derived type name `ansi_code` is good, but I wonder if `stdlib_ansi` or `stdlib_ansi_codes` would be better? For the colors, I feel like the "color" word are just extra characters. There is the actual color name in the variable name anyway, and it could be just stated in the documentation that parameters prepended with `fg_` and `bg_` are colors. A second consideration I had was to prepend `ansi_` to all named parameters for extra clarity, but I'm not fully decided if I like it or not.
2) I find the right-overwriting rule of the addition operator a bit nonintuitive the way it reads in the documentation. I suppose the correct usage pattern is to add up to one code from each group (fg, bg, and style), so the over-writing direction doesn't actually matter, unless you are "recycling" code variables. The graphics language [Asymptote](https://asymptote.sourceforge.io/) offers something similar for combining different colored pens:
> Pens may be added together with the nonassociative binary operator +. This will add the colors of the two pens. All other non-default attributes of the rightmost pen will override those of the leftmost pen. Thus, one can obtain a yellow dashed pen by saying dashed+red+green or red+green+dashed or red+dashed+green. 
3) It would be interesting to look at some micro-optimizations for the ANSI code to string translation step in the future. Granted, colors are typically not used in performance critical sections. I'm just being pedantic.
4) We can offer a functional API in addition to the operator one. This would reset the sequence by default. There could also be a second function with a logical flag for resetting.
  ```
  pure function colorize(code, str)
    type(ansi_code), intent(in) :: code
    character(len=:), allocatable :: colorize
    colorize = code // str // style_reset
  end function
```
5) Should string type be supported to? Can also be deferred to a new PR.
6) [`NO_COLOR`](https://no-color.org/) could be easily supported in the future by changing the behavior of `to_string_ansi_code`.
7) Perhaps I missed the discussion, but what was the reason for making the structure constructor private? Is it to force users to initialize it explicitly after the declaration section? This ties back to comment 2).

Overall, I think it's a great how this new "low-level" interface turned out. I say low-level even if derived types and operator overloading is involved. After-all this is the "standard" library, and compiler vendors can choose to implement this in a more primitive fashion if they really want to. In that case `ansi_code` should probably be a sequence type to prevent access of the internals through inheritance.

File: src/stdlib_terminal_colors_to_string.f90 (hunk 9 above)
Comment (awvwgk) [reply src/stdlib_terminal_colors_to_string.f90:20]:
```suggestion
            str = esc // "[0"  ! Always reset the style
```

jvdp1:
@awvwgk If you are satisfied with this PR, I suggest that you merge it, such that users can test it, e.g., through `fpm`.

awvwgk:
I'm already using the module in a project and will soon also adopt it in TOML Fortran, the design works nicely so far for my usage.

Now that we have the tree-shaking in fpm I could adopt it TOML Fortran via stdlib, but I'm still hesitant to limit myself in the compiler choice (I have to support GCC 5 and NAG 7 which are both not supported with stdlib but currenly work with TOML Fortran).

milancurcic:
Thanks Sebastian for the contribution and all reviewers. I'll merge now.

PR #584, opened 2021-12-01
Replace the call to sort by select in stdlib_stats_median

jvdp1:
Based on the performance tests reported in stdlib_selection.md, I replaced all calls to `sort` by `select` in `stdlib_stats_median.fypp`.

### DIFF
--- a/src/stdlib_stats_median.fypp
+++ b/src/stdlib_stats_median.fypp
@@ -44,7 +44,7 @@ contains
         call select(x_tmp, c, val)
 
         if (mod(n, 2_int64) == 0) then
-          call select(x_tmp, c+1, val1, left = c)
+          val1 = minval(x_tmp(c+1:n))  !instead of call select(x_tmp, c+1, val1, left = c)
           #:if t1[0] == 'r'
             res = (val + val1) / 2._${o1}$
           #:else
@@ -111,7 +111,7 @@ contains
                 call select(x_tmp, c, val)
 
                 if (mod(n, 2) == 0) then
-                    call select(x_tmp, c+1, val1, left = c)
+                    val1 = minval(x_tmp(c+1:n))
                     res${reduce_subvector('j', rank, fi)}$ = &
                     #:if t1[0] == 'r'
                         (val + val1) / 2._${o1}$
@@ -171,7 +171,7 @@ contains
         call select(x_tmp, c, val)
 
         if (mod(n, 2_int64) == 0) then
-          call select(x_tmp, c+1, val1, left = c)
+          val1 = minval(x_tmp(c+1:n))
           #:if t1[0] == 'r'
             res = (val + val1) / 2._${o1}$
           #:else
@@ -242,7 +242,7 @@ contains
                 call select(x_tmp, c, val)
 
                 if (mod(n, 2_int64) == 0) then
-                    call select(x_tmp, c+1, val1, left = c)
+                    val1 = minval(x_tmp(c+1:n))
                     res${reduce_subvector('j', rank, fi)}$ = &
                     #:if t1[0] == 'r'
                         (val + val1) / 2._${o1}$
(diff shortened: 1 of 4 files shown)


src/stdlib_stats_median.fypp
```diff
@@ -43,16 +41,18 @@ contains
 
         x_tmp = reshape(x, [n])
 
-        call sort(x_tmp)
+        call select(x_tmp, c, val)
 
         if (mod(n, 2_int64) == 0) then
+          call select(x_tmp, c+1, val1, left = c)
```
jvdp1, line 47: I am wondering if 
```fortran
val1 = minval(x_tmp(c+1:n))
```
would be more efficient than
```fortran
call select(x_tmp, c+1, val1, left = c)
```
, since `all(x_tmp(c+1:n)  > x_tmp(c))`?

gareth-nx: Yes I anticipate `minval` should be faster will only require one scan of the data.

src/stdlib_stats_median.fypp
```diff
@@ -107,17 +108,18 @@ contains
                   end if
                 #:endif
 
-                call sort(x_tmp)
+                call select(x_tmp, c, val)
 
                 if (mod(n, 2) == 0) then
+                    call select(x_tmp, c+1, val1, left = c)
```
gareth-nx, line 114: As above, `minval` is likely faster (below I won't repeat)

src/tests/stats/test_median.fypp
```diff
@@ -117,12 +117,6 @@ contains
         call check(error, median(d2odd_${k1}$),  1._dp&
                     , 'median(d2odd_${k1}$): uncorrect answer'&
                     , thr = dptol)
-        if (allocated(error)) return
```
gareth-nx, line 120: Here, did you remove one-too-many `if(allocated(error))` ?

gareth-nx (commented):
Thanks very much!

gareth-nx:
I just realised that it will also be necessary to update the specs.

At the moment the specs indicated that after a call to `median`, the input array is sorted. Whereas now it will only be partially sorted.

src/tests/stats/test_median.fypp
jvdp1: It is the last test, so it will return anyway. I can add it back, if it clarifies things.

src/stdlib_stats_median.fypp
jvdp1: Done. I replaced the second `select` by `minval`

src/stdlib_stats_median.fypp
jvdp1: done

jvdp1:
> I just realised that it will also be necessary to update the specs.
> 
> At the moment the specs indicated that after a call to `median`, the input array is sorted. Whereas now it will only be partially sorted.

Indeed, I forgot it. It is now done. Note that the input array is an `intent(in)` and it is therefore not modified. A copy of the array is always done.

leonfoks:
The side effect of having a modified input array is actually beneficial to the speed of KdTrees once those are implemented or any other recursive partitioning that uses the median.
Can an optional argument be added like “in-place” (like a lot of numpy methods) so a user can specifically request in place memory manipulation?

src/tests/stats/test_median.fypp
gareth-nx: Oh makes sense -- I don't mind, as you like.

gareth-nx:
> The side effect of having a modified input array is actually beneficial to the speed of KdTrees once those are implemented or any other recursive partitioning that uses the median. Can an optional argument be added like “in-place” (like a lot of numpy methods) so a user can specifically request in place memory manipulation?

@leonfoks In this case, one could just use `select` or `arg_select` directly -- and the in-place partial sort will happen. 

However that is not exactly the same as computing the median (if there are an even number of elements). Does the KdTree application need this variant of the median, or does the result of `select` or `arg_select` alone suffice?

leonfoks:
Good point @gareth-nx ! I forgot that median was wrapping the select methods...  Ignore my last comment!

gareth-nx (approved):
LGTM, thanks

milancurcic (approved):
Looks good, thank you!

PR #585, opened 2021-12-01
stdlib_selection: correction of typos and addition of some checks

jvdp1:
* I corrected some typos in `stdlib_selection.md`
* I added some checks to ensure that `k` in `select(x, k, kth, letft, right)` is within the interval `[left: right]`.

ping @gareth-nx

### DIFF
--- a/src/stdlib_selection.fypp
+++ b/src/stdlib_selection.fypp
@@ -88,7 +88,7 @@ contains
           if(l > r .or. l < 1_ip .or. r > size(a, kind=ip) &
               .or. k < l .or. k > r                        & !i.e. if k is not in the interval [l; r]  
               ) then
-              error stop "select must have 1 <= k <= size(a), and 1 <= left <= right <= size(a)";
+              error stop "select must have  1 <= left <= k <= right <= size(a)";
           end if
 
           searchk: do
@@ -205,7 +205,7 @@ contains
           if(l > r .or. l < 1_ip .or. r > size(a, kind=ip) &
               .or. k < l .or. k > r                        & !i.e. if k is not in the interval [l; r]  
               ) then
-              error stop "arg_select must have 1 <= k <= size(a), and 1 <= left <= right <= size(a)";
+              error stop "arg_select must have 1 <= left <= k <= right <= size(a)";
           end if
 
           searchk: do
--- a/src/stdlib_selection.fypp
+++ b/src/stdlib_selection.fypp
@@ -85,9 +85,9 @@ contains
           r = size(a, kind=ip)
           if(present(right)) r = right
 
-          if(k < 1_ip .or. k > size(a, kind=ip) .or. l > r .or. l < 1_ip .or. &
-              k < l .or. k > r .or.      &             !i.e. if k is not in the interval [l; r]  
-              r > size(a, kind=ip)) then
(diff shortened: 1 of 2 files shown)


src/stdlib_selection.fypp
```diff
@@ -86,6 +86,7 @@ contains
           if(present(right)) r = right
 
           if(k < 1_ip .or. k > size(a, kind=ip) .or. l > r .or. l < 1_ip .or. &
+              k < l .or. k > r .or.      &             !i.e. if k is not in the interval [l; r]
```
gareth-nx, line 89: Good catch.

gareth-nx (approved):
Thanks for that, looks good.

src/stdlib_selection.fypp
gareth-nx: Actually with these changes, I think we can also remove the criteria `k < 1_ip .or. k > size(a, kind=ip)`

awvwgk (approved):
Looks good to me

src/stdlib_selection.fypp
jvdp1: good point. Indeed, they can be removed.

jvdp1: Done

gareth-nx (approved):
Thanks!

ivan-pi:
The error message now doesn't mention the condition that k be in the interval [l, r]. Maybe breaking into two if blocks would make it easier on the eye. But I don't have strong feelings about this.

gareth-nx:
I just noticed another minor issue in `stdlib_selection.md`.  At line 26, there should be a new line before `* select`. 

At the moment you can see we are missing a newline in the [html view](https://stdlib.fortran-lang.org/page/specs/stdlib_selection.html), just under the "Overview of the module" section. Interestingly this formatting issue doesn't occur with the github markdown view.

If you can add a newline as part of this pull request, great. Otherwise I can make a pull request later.

jvdp1:
> The error message now doesn't mention the condition that k be in the interval [l, r]. Maybe breaking into two if blocks would make it easier on the eye. But I don't have strong feelings about this.

Thank you @ivan-pi . I modified the lines as follows:
```
error stop "select must have 1 <= left <= k <= right <= size(a)";
```

jvdp1:
> I just noticed another minor issue in `stdlib_selection.md`. At line 26, there should be a new line before `* select`.
> 
> At the moment you can see we are missing a newline in the [html view](https://stdlib.fortran-lang.org/page/specs/stdlib_selection.html), just under the "Overview of the module" section. Interestingly this formatting issue doesn't occur with the github markdown view.
> 
> If you can add a newline as part of this pull request, great. Otherwise I can make a pull request later.

Done. Thank you @gareth-nx

gareth-nx:
@jvdp1 In case you make further changes here, could you please also add a suitable ford comment at the top of the stdlib_selection.fypp, such as:
```
!! Quickly find the k-th smallest value of an array, or the index of the k-th smallest value.
```
This statement will then appear on the stdlib module html page: https://stdlib.fortran-lang.org/lists/modules.html

Thanks.

gareth-nx (approved):
Thanks

jvdp1:
Mainly fixing typos and links for FORD. Thank you to the reviewers.

PR #588, opened 2021-12-03
Error on no tests in CTest

### LINKED ISSUE #587: Possible issue: "No tests were found!!!" in CI Build (ubuntu-latest, 9, cmake)
I just noticed that no tests are found fo the CI "Build (ubuntu-latest, *, cmake"). See, e.g. [this action](https://github.com/fortran-lang/stdlib/runs/4398555214?check_suite_focus=true).
Therefore, sometimes, the actions are ok, while they should failed it the tests were properly run.

Did I miss somet discussions? Is this on purpose?

awvwgk:
We are currently pinning CMake to version 3.16 in our CI, which doesn't allow to raise an error in case CTest doesn't find any tests. Unless there is a particular reason to test with CMake 3.16, we can use this PR to use the preinstalled CMake 3.22 version.

Closes #587

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -43,10 +43,6 @@ jobs:
       with:
         python-version: 3.x
 
-    - name: Install CMake Linux
-      if: contains(matrix.os, 'ubuntu')
-      run: ci/install_cmake.sh
-
     - name: Install fypp
       run: pip install --upgrade fypp
 
@@ -85,7 +81,12 @@ jobs:
 
     - name: test
       if: ${{ contains(matrix.build, 'cmake') }}
-      run: ctest --test-dir ${{ env.BUILD_DIR }} --parallel --output-on-failure
+      run: >-
+        ctest
+        --test-dir ${{ env.BUILD_DIR }}
+        --parallel
+        --output-on-failure
+        --no-tests=error
 
     - name: Install project
       if: ${{ contains(matrix.build, 'cmake') }}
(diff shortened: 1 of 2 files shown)


jvdp1 (approved):
LGTM.
I believe @certik wrote the `install_cmake.sh`, but I can't remember the reason. @certik, what was the aim of this (if it is actually you who did this)?

awvwgk:
I thought it was used to test the oldest supported CMake version, but the minimum required CMake version is 3.14. Maybe the original Ubuntu image 16.04 or 18.04 did not have a sufficiently new CMake version preinstalled?

jvdp1:
> I thought it was used to test the oldest supported CMake version, but the minimum required CMake version is 3.14. Maybe the original Ubuntu image 16.04 or 18.04 did not have a sufficiently new CMake version preinstalled?

Yes, it could be that.

LKedward (commented):
I assume the CMake version was pinned to ensure that there were no breaking changes for the minimum required version, but I can't be sure. The changes to the CI all look fine to me. What is the motivation here? i.e. what situation would cause CMake to not find any tests?

jvdp1:
>  i.e. what situation would cause CMake to not find any tests?

I have no idea why CMake didn't find the tests. I observed it for several actions, and then seemed to work fine again...

LKedward (approved):

jvdp1:
Two approvals... I'll merge this PR. Thank you @awvwgk

PR #593, opened 2021-12-11
Probability Distribution and Statistical Functions -- Uniform Distribution Module

Jim-215-Fisher:
This is a minor update version to correct precision issue. Based on the conversation [here](https://github.com/fortran-lang/stdlib/pull/273#discussion_r762466332), we should provide all precision for  pdf and cdf functions. Originally, uniform distribution provides only single precision for pdf and cdf.


jvdp1 (approved):
Great. Thank you.

awvwgk (approved):

gareth-nx (changes requested):
Thanks very much for this! 

I noticed some minor issues:
* In the documentation for `pdf_uniform`, there is no explanation of what the `loc` parameter does. I think you just need to say something like: `f(x)` is zero for `x<loc` or `x > (loc+scale)`, and otherwise .... (continue to current definition). 
* In the documentation for `cdf_uniform`, you should also mention that it is zero for `x < loc`, and one for `x > (loc+scale)`. 

I also have a question about `MESSENE_NUMBER` which is used to normalise the random integers to produce uniform random variables. Currently it is always double precision. Does this reflect that the random number algorithm is fundamentally double precision (like in the case of the normal distribution module). If yes, could you please add a statement in the documentation to that effect? Or, should we have the precision of `MESSENE_NUMBER` varying with the precision of the inputs?

Jim-215-Fisher:
> Thanks very much for this!
> 
> I noticed some minor issues:
> 
>     * In the documentation for `pdf_uniform`, there is no explanation of what the `loc` parameter does. I think you just need to say something like: `f(x)` is zero for `x<loc` or `x > (loc+scale)`, and otherwise .... (continue to current definition).
> 
>     * In the documentation for `cdf_uniform`, you should also mention that it is zero for `x < loc`, and one for `x > (loc+scale)`.
> 
> 
> I also have a question about `MESSENE_NUMBER` which is used to normalise the random integers to produce uniform random variables. Currently it is always double precision. Does this reflect that the random number algorithm is fundamentally double precision (like in the case of the normal distribution module). If yes, could you please add a statement in the documentation to that effect? Or, should we have the precision of `MESSENE_NUMBER` varying with the precision of the inputs?

Alright, I will add description for `loc` in pdf_uniform and cdf_uniform. In terms of `MESSENE_NUMBER`,  yes the algorithm is fundamentally limited to double precision. The paper referred had only studied the double precision case. I will add a note in the specs.

gareth-nx (approved):
LGTM, thankyou.

Issue #592, opened 2021-12-08
`optval` issue in `stdlib_random` module


zoziha:
### Description

During a development, I found an unexpected situation: swapping `optval(s, si)` in the `stdlib_random` for `if (present(s)) then` form, and unit tests of `macos` and `ubuntu` systems failed.

- The first commit: https://github.com/zoziha/stdlib/commit/92a19496099434a9de0abe590a86e41bc09962e5
- The next commit: https://github.com/zoziha/stdlib/commit/d1a37ebed5e7ffd8724f50687f4a0914d9212600

![image](https://user-images.githubusercontent.com/32035096/145241965-98082dab-18ec-48d0-ad03-fc7807f7455e.png)
(see https://github.com/zoziha/stdlib/commits/update_random)

**Possible Problem**
https://github.com/fortran-lang/stdlib/blob/100f76098c834df29a800b3dcb1a0a59f295ae12/src/stdlib_random.fypp#L123
Maybe:

```fortran
res = res + int01 !?
```

cc @Jim-215-Fisher 

### Expected Behaviour

CI failed: https://github.com/zoziha/stdlib/actions/runs/1554566528

Expected behaviour: `optval` == `if (present(..)) then ..`

### Version of stdlib

100f76098c834df29a800b3dcb1a0a59f295ae12

### Platform and Architecture

MacOS, GFortran 9 10 11; Ubuntu, GFortran 10

zoziha:
The reason I modified the `stdlib_random` was because I planned to generate a stdlib with only `real64` and `int32`, so that I could use lightweight `stdlib` (see [dp-stdlib](https://github.com/zoziha/dp-stdlib)) within fpm projects.

gareth-nx:
@zoziha For what it's worth I've previously run into compiler problems at this part of the code (when using inter-procedural optimization). Although it would be good to hear from @Jim-215-Fisher, to my knowledge the line ` si = res + int01` is not a coding error -- the intention is indeed to update the random number generator state.

gareth-nx:
To give a bit more context - if I compile stdlib with: 
```
make -B -f Makefile.manual ADD_FYPPFLAGS=-DMAXRANK=4 FC=gfortran FFLAGS="-O3 -flto"
```
then I get a segfault on the test program in `stdlib/src/tests/stats/test_random`, which I believe is also related to this part of the code (it manifests in repeated calls to `splitmix64`). 

But if I remove the link-time-optimization then it works OK, i.e. using
```
[...]

PR #594, opened 2021-12-12
Minor update to makefile installation instructions

gareth-nx:
This removes the use of the option `-flto` in the instructions to compile with a Makefile.

The reason for doing that is that currently, `-flto` can cause some tests to break - see [discussion here](https://github.com/fortran-lang/stdlib/issues/592#issuecomment-991863294).

Note -- this pull request initially had a failing github test. But that issue must be unrelated to this commit, because the changes here only affect README.md


awvwgk (approved):

jvdp1 (approved):
OK for me. But do we know what the problem is?
If we know it, it could be interesting to add a warning in the specs.

gareth-nx:
> OK for me. But do we know what the problem is? If we know it, it could be interesting to add a warning in the specs.

Unfortunately I don't know what the problem is.

jvdp1:
Ok. I'll merge it. Thank you.

PR #597, opened 2021-12-12
Add getline to read whole line from formatted unit

### LINKED ISSUE #595: Procedure to read whole line to deferred length character
### Motivation

Just wrote a tutorial for using stdlib with fpm and noticed that I can't read a whole line into a deferred length character.

See: https://awvwgk.github.io/fpm-docs/en/tutorial/dependencies.html

### Prior Art

For the tutorial I used the following snippet

```f90
  !> Read a whole line from a formatted unit into a deferred length character variable
  subroutine getline(unit, line, iostat, iomsg)
    !> Formatted IO unit
    integer, intent(in) :: unit
    !> Line to read
    character(len=:), allocatable, intent(out) :: line
    !> Status of operation
    integer, intent(out) :: iostat
    !> Error message
    character(len=:), allocatable, optional :: iomsg

    integer, parameter :: bufsize = 512
    character(len=bufsize) :: buffer, msg
    integer :: chunk, stat

    allocate(character(len=0) :: line)
    do
      read(unit, '(a)', advance='no', iostat=stat, iomsg=msg, size=chunk) buffer
      if (stat > 0) exit
      line = line // buffer(:chunk)
      if (stat < 0) then
        if (is_iostat_eor(stat)) stat = 0
        exit
      end if
    end do

    if (stat /= 0 .and. present(iomsg)) iomsg = trim(msg)
    iostat = stat
  end subroutine getline
```

### Additional Information

Due to [gcc#100875](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100875) and https://github.com/fortran-lang/stdlib/issues/354 I couldn't use `string_type` for the tutorial.

awvwgk:
<details><summary>Alternative implementation by John @urbanjost in M_io</summary>

```f90
!!##NAME
!!    getline(3f) - [M_io] read a line from specified LUN into allocatable string up to line length limit
!!    (LICENSE:PD)
!!
!!##SYNTAX
!!   function getline(line,lun) result(ier)
!!
!!    character(len=:),allocatable,intent(out) :: line
!!    integer,intent(in),optional              :: lun
!!    integer,intent(out)                      :: ier
!!
!!##DESCRIPTION
!!    Read a line of any length up to programming environment maximum
!!    line length. Requires Fortran 2003+.
!!
!!    It is primarily expected to be used when reading input which will
!!    then be parsed.
!!
!!    The input file must have a PAD attribute of YES for the function
!!    to work properly, which is typically true.
!!
!!    The simple use of a loop that repeatedly re-allocates a character
!!    variable in addition to reading the input file one buffer at a
!!    time could (depending on the programming environment used) be
!!    inefficient, as it could reallocate and allocate memory used for
!!    the output string with each buffer read.
!!
!!##OPTIONS
!!    LINE   line read
!!    LUN    optional LUN (Fortran logical I/O unit) number. Defaults
!!           to stdin.
!!##RETURNS
!!    IER    zero unless an error occurred. If not zero, LINE returns the
!!           I/O error message.
!!
!!##EXAMPLE
!!
!!   Sample program:
!!
!!    program demo_getline
!!    use,intrinsic :: iso_fortran_env, only : stdin=>input_unit
!!    use M_io, only : getline
!!    implicit none
!!    character(len=:),allocatable :: line
!!       open(unit=stdin,pad='yes')
!!       INFINITE: do while (getline(line)==0)
!!          write(*,'(a)')'['//line//']'
!!       enddo INFINITE
!!    end program demo_getline
!!
!!##AUTHOR
!!    John S. Urban
!!
!!##LICENSE
!!    Public Domain
function getline(line,lun) result(ier)
implicit none

! ident_11="@(#)M_io::getline(3f): read a line from specified LUN into allocatable string up to line length limit"

character(len=:),allocatable,intent(out) :: line
integer,intent(in),optional              :: lun
integer                                  :: ier
character(len=4096)                      :: message

integer,parameter                        :: buflen=1024
character(len=:),allocatable             :: line_local
character(len=buflen)                    :: buffer
integer                                  :: isize
integer                                  :: lun_local

   line_local=''
   ier=0
   if(present(lun))then
      lun_local=lun
   else
      lun_local=stdin
   endif
   open(lun_local,pad='yes')

   INFINITE: do                                                      ! read characters from line and append to result
      read(lun_local,iostat=ier,fmt='(a)',advance='no',size=isize,iomsg=message) buffer ! read next buffer (might use stream I/O for files
                                                                     ! other than stdin so system line limit is not limiting
      if(isize.gt.0)line_local=line_local//buffer(:isize)            ! append what was read to result
      if(is_iostat_eor(ier))then                                     ! if hit EOR reading is complete unless backslash ends the line
         ier=0                                                       ! hitting end of record is not an error for this routine
         exit INFINITE                                               ! end of reading line
     elseif(ier.ne.0)then                                            ! end of file or error
        line=trim(message)
        exit INFINITE
     endif
   enddo INFINITE
   line=line_local                                                   ! trim line
end function getline
```
</details>
Closes #595

### DIFF
--- a/doc/specs/stdlib_io.md
+++ b/doc/specs/stdlib_io.md
@@ -223,3 +223,51 @@ program demo_savenpy
     call save_npy('example.npy', x)
 end program demo_savenpy
 ```
+
+## `getline`
+
+### Status
+
+Experimental
+
+### Description
+
+Read a whole line from a formatted unit into a string variable
+
+### Syntax
+
+`call [[stdlib_io(module):getline(interface)]](unit, line[, iostat][, iomsg])`
+
+### Arguments
+
+`unit`: Formatted input unit.
+        This argument is `intent(in)`.
+
+`line`: Deferred length character or `string_type` variable.
+        This argument is `intent(out)`.
+
+`iostat`: Default integer, contains status of reading from unit, zero in case of success.
+          It is an optional argument, in case not present the program will halt for non-zero status.
+          This argument is `intent(out)`.
+
+`iomsg`: Deferred length character value, contains error message in case `iostat` is non-zero.
+         It is an optional argument, error message will be dropped if not present.
+         This argument is `intent(out)`.
+
+### Example
+
+```fortran
+program demo_getline
+    use, intrinsic :: iso_fortran_env, only : input_unit, output_unit
+    use stdlib_io, only: getline
+    implicit none
+    character(len=:), allocatable :: line
+    integer :: stat
+
+    call getline(input_unit, line, stat)
+    do while(stat == 0)
+      write(output_unit, '(a)') line
+      call getline(input_unit, line, stat)
+    end do
+end program demo_getline
+```
--- a/doc/specs/stdlib_io.md
+++ b/doc/specs/stdlib_io.md
@@ -236,12 +236,14 @@ Read a whole line from a formatted unit into a string variable
 
 ### Syntax
 
-`call [[stdlib_io(module):getline(interface)]](unit, line[, iostat][, iomsg])`
+`call [[stdlib_io(module):getline(interface)]] (unit, line[, iostat][, iomsg])`
+`call [[stdlib_io(module):getline(interface)]] (line[, iostat][, iomsg])`
 
 ### Arguments
 
 `unit`: Formatted input unit.
         This argument is `intent(in)`.
+        If `unit` is not specified standard input is used.
 
 `line`: Deferred length character or `string_type` variable.
         This argument is `intent(out)`.
--- a/src/stdlib_io.fypp
+++ b/src/stdlib_io.fypp
@@ -357,7 +357,7 @@ contains
     character(len=bufsize) :: buffer, msg
     integer :: chunk, stat
 
-    allocate(character(len=0) :: line)
+    line = ""
     do
       read(unit, '(a)', advance='no', iostat=stat, iomsg=msg, size=chunk) buffer
       if (stat > 0) exit
--- a/src/stdlib_io.fypp
+++ b/src/stdlib_io.fypp
@@ -11,10 +11,11 @@ module stdlib_io
   use stdlib_error, only: error_stop
   use stdlib_optval, only: optval
   use stdlib_ascii, only: is_blank
+  use stdlib_string_type, only : string_type
   implicit none
   private
   ! Public API
-  public :: loadtxt, savetxt, open
+  public :: loadtxt, savetxt, open, getline
 
   ! Private API that is exposed so that we can test it in tests
   public :: parse_mode
@@ -30,6 +31,14 @@ module stdlib_io
     FMT_COMPLEX_DP = '(*(es24.16e3,1x,es24.16e3))', &
     FMT_COMPLEX_XDP = '(*(es26.18e3,1x,es26.18e3))', &
     FMT_COMPLEX_QP = '(*(es44.35e4,1x,es44.35e4))'
+
+  !> Version: experimental
+  !>
+  !> Read a whole line from a formatted unit into a string variable
+  interface getline
+    module procedure :: getline_char
+    module procedure :: getline_string
+  end interface getline
 
   interface loadtxt
     !! version: experimental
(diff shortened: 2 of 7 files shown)


src/stdlib_io.fypp
```diff
@@ -331,4 +340,59 @@ contains
...
+    !> Status of operation
+    integer, intent(out), optional :: iostat
+    !> Error message
+    character(len=:), allocatable, optional :: iomsg
+
+    integer, parameter :: bufsize = 512
+    character(len=bufsize) :: buffer, msg
+    integer :: chunk, stat
+
+    allocate(character(len=0) :: line)
```
milancurcic, line 360: Is this still not widely supported?

```suggestion
    line = ''
```

src/stdlib_io.fypp
```diff
@@ -331,4 +340,59 @@ contains
...
+    !> Formatted IO unit
+    integer, intent(in) :: unit
+    !> Line to read
+    character(len=:), allocatable, intent(out) :: line
+    !> Status of operation
+    integer, intent(out), optional :: iostat
+    !> Error message
+    character(len=:), allocatable, optional :: iomsg
+
+    integer, parameter :: bufsize = 512
```
milancurcic, line 356: 512 seems reasonable. Is there any rule of thumb on the best value for this? I assume not, as it depends on what data the users will read. If it later shows that it's important, we could make it user settable by making `bufsize` an optional argument and `buffer` an allocatable.

doc/specs/stdlib_io.md
```diff
@@ -223,3 +223,51 @@ program demo_savenpy
     call save_npy('example.npy', x)
 end program demo_savenpy
 ```
+
+## `getline`
```
milancurcic, line 227: What do you think about `readline` for the name? It seems closer to the description and what the subroutine does. "get" is appropriate, but less specific than "read". `readline` is also how Python and Julia call it.

milancurcic (approved):
Looks great, thanks! I left a few comments and questions.

doc/specs/stdlib_io.md
awvwgk: I'm associating `readline` with the GNU readline library: https://en.wikipedia.org/wiki/GNU_Readline

milancurcic: Do you think there's a risk that some users may confuse a Fortran subroutine `readline` with GNU readline?

GNU readline (and Python, Julia) aside, do you like `getline` better than `readline`?

I'd like to get some more feedback on this, let's discuss it on the monthly call.

awvwgk: There is `std::getline` in C++ (https://www.cplusplus.com/reference/string/string/getline) and the C `getline` procedure (https://man7.org/linux/man-pages/man3/getline.3.html).

urbanjost: There are several C interfaces to readline, and readline is well-known enough I have avoided using that name, also I make a habit of prefixing anything that is a C interface with system_* so it it clearer to me it is introducing a C dependency;  but I find I like having a low-level and a high-level read of a line  [(see getline and read_line])( https://urbanjost.github.io/M_io/man3.html) so wondering if  a  line_get, and line_read, and line_edit, ... family might emerge?  Might just be personal taste by I have had similiar functionality for a long time and I like a "read a line as-is" sometimes, and a "read a line and get rid of likely DOS line terminators and allow for line continuation with backslash" version, and a mini-editor (using  https://urbanjost/github/M_history of course :>) and when I am sure it will be available an interface to the C readline (which possibly introduces GNU issues but it the most supported one remaining that I know of) 

Since users can rename these (but not if they become true intrinsics!) maybe std_getline, std_readline, would be a convention to use?

milancurcic: As agreed on the call, let's go forward with `getline`, as is.

src/stdlib_io.fypp
awvwgk: Not sure, best is something that captures most usual line lengths at once. The actual buffer length should be an implementation detail and not concern the user or clutter the API in my opinion.

milancurcic:
@urbanjost Does this look good to merge to you?

urbanjost:
If IOSTAT is optional on the call and the user leaves it off, how to you tell you hit end-of-file? 

The only other thing I see is a matter of taste. If LINE comes before UNIT then UNIT could be optional also and default to stdin. I think that is preferable to encouraging users to put in a 5; although if they put in INPUT_UNIT it is not bad, but requires a USE statement and using a long variable name. 

 There really should be a note somewhere the input file requires PAD mode to be on or it will confuse someone sooner or later. 

Could have sworn something similar was in stdlib allready but apparently not; I find a procedure like this very handy, especially for prototyping. 

I have a version that ignored trailing ctrl_J and ctrl_M characters because of a problem with MSWindows and Linux, especially when reading files from the other platform but it looks like that has settled down, as I tried it with three compilers and did not see anything wrong at least with formatted sequential files. I think it would still be a problem with stream I/O but this would seem to (almost?) always be used with that.

awvwgk:
> If IOSTAT is optional on the call and the user leaves it off, how do you tell you hit end-of-file?

This should work in analogy with a usual `read` going past an end of file without having `iostat` specified:
```
At line 46 of file /home/awvwgk/projects/src/git/stdlib/src/tests/io/test_getline.f90
Fortran runtime error: End of file
```

> The only other thing I see is a matter of taste. If LINE comes before UNIT then UNIT could be optional also and default to stdin. I think that is preferable to encouraging users to put in a 5; although if they put in INPUT_UNIT it is not bad, but requires a USE statement and using a long variable name.

I added another version of `getline` which doesn't require to pass the `unit` as first argument in analogy to `read(unit, *)`/`write(unit, *)` and `read *`/`print *`.

> There really should be a note somewhere the input file requires PAD mode to be on or it will confuse someone sooner or later.

I'm checking the `pad` case now. Note that `open(unit=unit, pad="yes")` for an unopened unit will open it, therefore I'm now checking first whether the unit is opened using `inquire(unit=unit, opened=opened)`. However, `-1` must not be passed to inquire, so another thing to catch.

jvdp1 (approved):
LGTM.
My only concern is the lenght of the buffer. 512 characters would be often too small for some of my cases.

awvwgk:
I created a test with different bufferlengths for reading 3600 lines, with the longest lines having 120000 chars, with short lines having 1200 chars. Timings in seconds.

buffersize | long lines | mixed lines | short lines
--- | ---
512 | 44.8 | 15.1 | 8.3e-2
1024 | 24.0 | 7.6 | 7.9e-2
2048 | 12.5 | 4.7 | 7.3e-2
4096 | 8.9 | 2.6 | 8.2e-2 
8192 | 4.8 | 1.9 | 7.9e-2

Larger buffersizes indeed give a sizable speedup, as the time consuming step is the read of each chunk from the unit. Also, it is good to see that small buffer.

src/stdlib_io.fypp
awvwgk:
```suggestion
    integer, parameter :: bufsize = 4096
```

PR #600, opened 2021-12-15
Fix iomsg allocation in save_npy

ivan-pi:
This is currently just a draft for a minor issue remaining in #581.

I've added a test to check that iomsg is deallocated if present, although this should technically be guaranteed by a standard-conforming compiler when the `intent(out)` attribute is present. 

To make sure the message gets allocated I need a way to trigger `stat /= 0`. Is their any known way of doing this for a new file?

On more thought, I'm afraid that both `stat` and `iomsg` only communicate the result of the command `close()`, and not if opening or writing were unsuccessful. I'm not sure whether the standard stipulates these commands be connected (in other words whether the same I/O error will trigger the same iostat flag in any of the three commands).

### DIFF
--- a/src/stdlib_io_npy_save.fypp
+++ b/src/stdlib_io_npy_save.fypp
@@ -128,7 +128,9 @@ contains
         end if
 
         if (present(iomsg)) then
-            iomsg = "Failed to write array to file '"//filename//"'"
+            if (stat /= 0) then
+                iomsg = "Failed to write array to file '"//filename//"'"
+            end if
         end if
     end subroutine save_npy_${t1[0]}$${k1}$_${rank}$
   #:endfor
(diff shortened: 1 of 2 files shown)


src/stdlib_io_npy_save.fypp
```diff
@@ -128,7 +128,9 @@ contains
         end if
 
         if (present(iomsg)) then
-            iomsg = "Failed to write array to file '"//filename//"'"
+            if (stat /= 0) then
```
awvwgk, line 131: Why not check for `present(iomsg) .and. stat /= 0`?

awvwgk (approved):
Good catch, thanks for fixing it

src/stdlib_io_npy_save.fypp
ivan-pi: I did that initially, but then decided to go with nested if's to avoid the `.and.`.

jvdp1 (approved):
Thank you

milancurcic (approved):
Thank you!

PR #603, opened 2021-12-19
Implement trueloc/falseloc

### LINKED ISSUE #568: true_pos
In a Fortran Discourse thread [Indexing arrays by an array of logicals](https://fortran-lang.discourse.group/t/indexing-arrays-by-an-array-of-logicals/2264) I mentioned a `true_pos` function I had written to simulate logical indexing, and @ivan-pi suggested that it be considered for stdlib.  To deal with the case of a `lbound /= 1` mentioned by mecej4, I have added an optional `lower_bound` argument. If `tf(:)` has the same size as `x(:)`, one can write `x(true_pos(tf))` to refer to elements of x where `tf` is `.true.` One can also write `pack(x,tf)`, but that cannot appear on the LHS of an assignment. I have used `true_pos` hundreds of times in my codes -- maybe it would be generally useful. Since Fortran has `minloc`, `maxloc`, and `findloc`, maybe `true_pos` should be renamed `trueloc`.

```
pure function true_pos(tf, lower_bound) result(ipos)
! return in ipos the positions of the true elements in tf(:)
logical, intent(in)           :: tf(:)
integer, intent(in), optional :: lower_bound 
integer                       :: ipos(count(tf))
integer                       :: i,j
j = 0
do i=1,size(tf)
   if (tf(i)) then
      j = j + 1
      ipos(j) = i
   end if
end do
if (present(lower_bound)) ipos = ipos + lower_bound - 1
end function true_pos
```

awvwgk:
Closes #568

### DIFF
--- a/doc/specs/stdlib_array.md
+++ b/doc/specs/stdlib_array.md
@@ -33,6 +33,10 @@ Turn a logical mask into an index array by selecting all true values.
 
 `lbound`: Lower bound of the array to index. This argument is `optional` and `intent(in)`.
 
+#### Return value
+
+Returns an array of default integer size, with a maximum length of `size(array)` elements.
+
 #### Examples
 
 ```fortran
@@ -66,6 +70,10 @@ Turn a logical mask into an index array by selecting all false values.
 `array`: List of default logical arrays. This argument is `intent(in)`.
 
 `lbound`: Lower bound of the array to index. This argument is `optional` and `intent(in)`.
+
+#### Return value
+
+Returns an array of default integer size, with a maximum length of `size(array)` elements.
 
 #### Examples
 
--- a/doc/specs/stdlib_array.md
+++ b/doc/specs/stdlib_array.md
@@ -22,6 +22,8 @@ Experimental
 #### Description
 
 Turn a logical mask into an index array by selecting all true values.
+Provides similar functionality like the built-in `where` or the intrinsic procedures `merge` and `pack` when working with logical mask.
+The built-in / intrinsics are usually preferable to `trueloc`, unless the access to the index array is required.
 
 #### Syntax
 
@@ -64,6 +66,8 @@ Experimental
 #### Description
 
 Turn a logical mask into an index array by selecting all false values.
+Provides similar functionality like the built-in `where` or the intrinsic procedures `merge` and `pack` when working with logical mask.
+The built-in / intrinsics are usually preferable to `falseloc`, unless the access to the index array is required.
 
 #### Syntax
 
--- a/doc/specs/stdlib_array.md
+++ b/doc/specs/stdlib_array.md
@@ -25,7 +25,11 @@ Turn a logical mask into an index array by selecting all true values.
 
 #### Syntax
 
-`call [[trueloc(function)]] (array[, lbound])`
+`loc = [[trueloc(function)]] (array[, lbound])`
+
+#### Class
+
+Pure function.
 
 #### Arguments
 
@@ -40,14 +44,14 @@ Returns an array of default integer size, with a maximum length of `size(array)`
 #### Examples
 
 ```fortran
-program demo
+program demo_trueloc
   use stdlib_array, only : trueloc
   implicit none
   real, allocatable :: array(:)
   allocate(array(500))
   call random_number(array)
   array(trueloc(array > 0.5)) = 0.0
-end program demo
+end program demo_trueloc
 ```
 
 
@@ -63,7 +67,11 @@ Turn a logical mask into an index array by selecting all false values.
 
 #### Syntax
 
-`call [[falseloc(function)]] (array[, lbound])`
+`loc = [[falseloc(function)]] (array[, lbound])`
+
+#### Class
+
+Pure function.
 
 #### Arguments
 
@@ -78,12 +86,12 @@ Returns an array of default integer size, with a maximum length of `size(array)`
 #### Examples
 
 ```fortran
-program demo
+program demo_falseloc
   use stdlib_array, only : falseloc
   implicit none
   real, allocatable :: array(:)
   allocate(array(-200:200))
   call random_number(array)
   array(falseloc(array < 0.5), lbound(array)) = 0.0
-end program demo
+end program demo_falseloc
 ```
--- a/doc/specs/stdlib_array.md
+++ b/doc/specs/stdlib_array.md
@@ -1,0 +1,81 @@
+---
+title: array
+---
+
+# The `stdlib_array` module
+
+[TOC]
+
+## Introduction
+
+Module for index manipulation and array handling tasks.
+
+## Procedures and methods provided
+
+
+### `trueloc`
+
+#### Status
+
+Experimental
+
+#### Description
+
+Turn a logical mask into an index array by selecting all true values.
+
+#### Syntax
+
+`call [[trueloc(function)]] (array[, lbound])`
+
+#### Arguments
+
+`array`: List of default logical arrays. This argument is `intent(in)`.
+
+`lbound`: Lower bound of the array to index. This argument is `optional` and `intent(in)`.
+
+#### Examples
+
+```fortran
+program demo
+  use stdlib_array, only : trueloc
+  implicit none
+  real, allocatable :: array(:)
+  allocate(array(500))
+  call random_number(array)
+  array(trueloc(array > 0.5)) = 0.0
+end program demo
+```
+
+
+### `falseloc`
+
+#### Status
+
+Experimental
+
+#### Description
+
+Turn a logical mask into an index array by selecting all false values.
+
+#### Syntax
+
+`call [[falseloc(function)]] (array[, lbound])`
+
+#### Arguments
+
+`array`: List of default logical arrays. This argument is `intent(in)`.
+
+`lbound`: Lower bound of the array to index. This argument is `optional` and `intent(in)`.
+
+#### Examples
+
+```fortran
+program demo
+  use stdlib_array, only : falseloc
+  implicit none
+  real, allocatable :: array(:)
+  allocate(array(-200:200))
+  call random_number(array)
+  array(falseloc(array < 0.5), lbound(array)) = 0.0
+end program demo
+```
--- a/src/stdlib_array.f90
+++ b/src/stdlib_array.f90
@@ -18,7 +18,7 @@ contains
     !> Locations of true elements
     integer :: loc(count(array))
 
-    loc = logicalloc(array, .true., lbound)
+    call logicalloc(loc, array, .true., lbound)
   end function trueloc
 
   !> Return the positions of the false elements in array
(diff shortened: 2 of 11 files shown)


src/stdlib_array.f90
```diff
@@ -0,0 +1,60 @@
...
+contains
+
+  !> Return the positions of the true elements in array
+  pure function trueloc(array, lbound) result(loc)
+    !> Mask of logicals
+    logical, intent(in) :: array(:)
+    !> Lower bound of array to index
+    integer, intent(in), optional :: lbound
+    !> Locations of true elements
+    integer :: loc(count(array))
```
gareth-nx, line 19: I wonder how it will perform if `size(array) == 0`? I suspect it will still work (and return a logical array of length 0), but it would be good to check this and either include such a test in the test-suite, or else document that it won't work.

gareth-nx: I won't repeat this comment below, but obviously it applies to all the routines

doc/specs/stdlib_array.md
```diff
@@ -0,0 +1,81 @@
...
+
+#### Syntax
+
+`call [[trueloc(function)]] (array[, lbound])`
+
+#### Arguments
+
+`array`: List of default logical arrays. This argument is `intent(in)`.
+
+`lbound`: Lower bound of the array to index. This argument is `optional` and `intent(in)`.
```
gareth-nx, line 34: Somewhere we should document the `return value`.

doc/specs/stdlib_array.md
```diff
@@ -0,0 +1,81 @@
...
+
+#### Syntax
+
+`call [[falseloc(function)]] (array[, lbound])`
+
+#### Arguments
+
+`array`: List of default logical arrays. This argument is `intent(in)`.
+
+`lbound`: Lower bound of the array to index. This argument is `optional` and `intent(in)`.
```
gareth-nx, line 68: As above, suggest to document the return value

src/tests/array/test_logicalloc.f90
```diff
@@ -0,0 +1,154 @@
...
+      ]
+  end subroutine collect_logicalloc
+
+  subroutine test_trueloc_where(error)
+    !> Error handling
+    type(error_type), allocatable, intent(out) :: error
+
+    integer :: ndim
+    real, allocatable :: avec(:), bvec(:), cvec(:)
+
```
gareth-nx, line 33: Good for the tests to include some corner cases, such as `size(array) == 0`, and also `array(:) = .true.` and `array(:) = .false.`, for both `trueloc` and `falseloc` (unless they do not support that, in which case just document it in the specs).

gareth-nx (commented):
Thanks -- a few simple suggestions here, which should be easy to address.

gareth-nx (approved):
Thanks

milancurcic:
I have a concern about the implementation.

Since `trueloc` and `falseloc` call `logicalloc` and they're all functions, there will be a double copy of the result array on each call of `trueloc` and `falseloc`. It's possible that some of the poor performance shown [here](https://github.com/fortran-lang/stdlib/issues/568#issuecomment-997374204) can be explained by the double copy. So maybe there is a performance penalty in doing it this way.

Why not implement the functionality of `logicalloc` in both `trueloc` and `falseloc` (and duplicate a few lines of code) and avoid the extra function call and array copy?

And, is `falseloc` really needed? You could call `trueloc(.not. condition)` to get that. I understand it's a matter of style and has no right or wrong answer.

FWIW, my preference would be, in this order:

1. Have only `trueloc` and avoid calling `logicalloc`
2. Have `trueloc` and `falseloc` and avoid calling `logicalloc`
3. Current implementation.

awvwgk:
Implementation in 7b5ffac 
```
[Timing] trueloc: 0.0218s where: 0.0043s ratio:  5.0
[Timing] trueloc: 0.0218s merge: 0.0044s ratio:  4.9
[Timing] falseloc: 0.0223s where: 0.0051s ratio:  4.4
[Timing] falseloc: 0.0233s merge: 0.0053s ratio:  4.4
```

Using subroutine instead of function
```
[Timing] trueloc: 0.0181s where: 0.0043s ratio:  4.2
[Timing] trueloc: 0.0168s merge: 0.0044s ratio:  3.8
[Timing] falseloc: 0.0174s where: 0.0050s ratio:  3.5
[Timing] falseloc: 0.0178s merge: 0.0052s ratio:  3.4
```

Duplicated inline implementation
```
[Timing] trueloc: 0.0184s where: 0.0042s ratio:  4.4
[Timing] trueloc: 0.0168s merge: 0.0046s ratio:  3.7
[Timing] falseloc: 0.0177s where: 0.0049s ratio:  3.6
[Timing] falseloc: 0.0177s merge: 0.0053s ratio:  3.3
```

jvdp1:
> Implementation in [7b5ffac](https://github.com/fortran-lang/stdlib/commit/7b5fface9f0556d18248fef61d1e307a9247e0f8)
> 
> ```
> [Timing] trueloc: 0.0218s where: 0.0043s ratio:  5.0
> [Timing] trueloc: 0.0218s merge: 0.0044s ratio:  4.9
> [Timing] falseloc: 0.0223s where: 0.0051s ratio:  4.4
> [Timing] falseloc: 0.0233s merge: 0.0053s ratio:  4.4
> ```
> 
> Using subroutine instead of function
> 
> ```
> [Timing] trueloc: 0.0181s where: 0.0043s ratio:  4.2
> [Timing] trueloc: 0.0168s merge: 0.0044s ratio:  3.8
> [Timing] falseloc: 0.0174s where: 0.0050s ratio:  3.5
> [Timing] falseloc: 0.0178s merge: 0.0052s ratio:  3.4
> ```
> 
> Duplicated inline implementation
> 
> ```
> [Timing] trueloc: 0.0184s where: 0.0042s ratio:  4.4
> [Timing] trueloc: 0.0168s merge: 0.0046s ratio:  3.7
> [Timing] falseloc: 0.0177s where: 0.0049s ratio:  3.6
> [Timing] falseloc: 0.0177s merge: 0.0053s ratio:  3.3
> ```

Nice comparison. However, is this comparison fair for `trueloc`, since the aim is a bit different than `where` and `merge`? Furthermore, `trueloc` must initialize an `integer array, which is not the case of where and merge, if I understand it well.

awvwgk:
> Nice comparison. However, is this comparison fair for `trueloc`, since the aim is a bit different than `where` and `merge`? Furthermore, `trueloc` must initialize an `integer array, which is not the case of where and merge, if I understand it well.

Indeed, `merge` and `where` are both language intrinsic constructs and give the compiler many opportunities for optimization, due to the `elemental` character merge is somewhat slower than `where`. For `trueloc` we have to construct an index array (function call, stack allocation, loop over array and increment counter), which can only be slower, with link time optimization the compiler might be able to optimize `trueloc` away again and regain the performance loss.

I would consider `trueloc` and `falseloc` a convenience feature, for actual compute intense task `where` is preferable (also offers worksharing via OMP).

awvwgk:
Hmm, now I want to have timers in test-drive and stdlib.

milancurcic (approved):

doc/specs/stdlib_array.md
```diff
@@ -0,0 +1,89 @@
...
+
+#### Return value
+
+Returns an array of default integer size, with a maximum length of `size(array)` elements.
+
+#### Examples
+
+```fortran
+program demo
```
jvdp1, line 43:
```suggestion
program demo_trueloc
```

doc/specs/stdlib_array.md
```diff
@@ -0,0 +1,89 @@
...
+  use stdlib_array, only : trueloc
+  implicit none
+  real, allocatable :: array(:)
+  allocate(array(500))
+  call random_number(array)
+  array(trueloc(array > 0.5)) = 0.0
+end program demo
```
jvdp1, line 50:
```suggestion
end program demo_trueloc
```

doc/specs/stdlib_array.md
```diff
@@ -0,0 +1,89 @@
...
+
+Experimental
+
+#### Description
+
+Turn a logical mask into an index array by selecting all true values.
+
+#### Syntax
+
+`call [[trueloc(function)]] (array[, lbound])`
```
jvdp1, line 28: Is it a function of a subroutine?
In the demo, it is a function.

doc/specs/stdlib_array.md
```diff
@@ -0,0 +1,89 @@
...
+
+Experimental
+
+#### Description
+
+Turn a logical mask into an index array by selecting all false values.
+
+#### Syntax
+
+`call [[falseloc(function)]] (array[, lbound])`
```
jvdp1, line 66: Function or subroutine?

doc/specs/stdlib_array.md
```diff
@@ -0,0 +1,89 @@
...
+`lbound`: Lower bound of the array to index. This argument is `optional` and `intent(in)`.
+
+#### Return value
+
+Returns an array of default integer size, with a maximum length of `size(array)` elements.
+
+#### Examples
+
+```fortran
+program demo
```
jvdp1, line 81:
```suggestion
program demo_falseloc
```

doc/specs/stdlib_array.md
```diff
@@ -0,0 +1,89 @@
...
+  use stdlib_array, only : falseloc
+  implicit none
+  real, allocatable :: array(:)
+  allocate(array(-200:200))
+  call random_number(array)
+  array(falseloc(array < 0.5), lbound(array)) = 0.0
+end program demo
```
jvdp1, line 88:
```suggestion
end program demo_falseloc
```

doc/specs/stdlib_array.md
```diff
@@ -0,0 +1,89 @@
...
+## Procedures and methods provided
+
+
+### `trueloc`
+
+#### Status
+
+Experimental
+
+#### Description
```
jvdp1, line 22: Mention somewhere that it is a pure function.

src/stdlib_array.f90
```diff
@@ -0,0 +1,60 @@
...
+  implicit none
+  private
+
+  public :: trueloc, falseloc
+
+contains
+
+  !> Return the positions of the true elements in array
+  pure function trueloc(array, lbound) result(loc)
+    !> Mask of logicals
```
jvdp1, line 14:
```suggestion
    !! Version: experimental
    !!
    !! [Specification](LINK_FOR_FORD)
    !> Mask of logicals
```

jvdp1 (changes requested):
Thahnk you.
Here are some minor comments

doc/specs/stdlib_array.md
awvwgk:
```suggestion
`loc = [[falseloc(function)]] (array[, lbound])`
```

doc/specs/stdlib_array.md
awvwgk:
```suggestion
`loc = [[trueloc(function)]] (array[, lbound])`
```

gareth-nx:
> I would consider `trueloc` and `falseloc` a convenience feature, for actual compute intense task `where` is preferable (also offers worksharing via OMP).

I suggest that something like this is stated in the specs to help guide users.

gareth-nx:
Looks like this is good to be merged? @jvdp1, any further comments?

jvdp1 (approved):
LGTM. Thank you @awvwgk

PR #604, opened 2021-12-19
Add get_argument, get_variable and set_variable

awvwgk:
- cleanup of `stdlib_system` module + documentation
- new procedure `get_argument` to retrieve command line arguments in deferred length characters and `string_type` variables
- new procedure `get_variable` to retrieve environment variables in deferred length characters and `string_type` variables
- new procedure `set_variable` to write environment variables

### DIFF
--- a/doc/specs/stdlib_system.md
+++ b/doc/specs/stdlib_system.md
@@ -1,0 +1,169 @@
+---
+title: system
+---
+
+# The `stdlib_system` module
+
+[TOC]
+
+## Introduction
+
+Interface to system functions.
+
+## Procedures and methods provided
+
+
+### `sleep`
+
+#### Status
+
+Experimental
+
+#### Description
+
+Wait for a given period of time.
+
+#### Syntax
+
+`call [[sleep(subroutine)]] (time)`
+
+
+#### Arguments
+
+`time`: Time to wait in millisecs.
+
+
+#### Examples
+
+```fortran
+program demo
+    use stdlib_system, only: sleep
+    implicit none
+    call sleep(150)
+end program demo
+```
+
+
+### `get_argument`
+
+#### Status
+
+Experimental
+
+#### Description
+
+Obtain the nth command line argument
+
+#### Syntax
+
+`call [[get_argument(interface)]] (idx, val[, stat])`
+
+#### Arguments
+
+`idx`: Index of the command line argument to retrieve.
+       This is an `intent(in)` argument of default integer kind.
+
+`val`: Value of command line argument on return.
+       Can be either a deferred length character or a [[string_type]].
+       This is an `intent(out)` argument.
+
+`stat`: Default integer, contains status of operation, zero in case of success.
+        It is an optional argument, in case not present the program will halt for non-zero status.
+        This argument is `intent(out)`.
+
+
+#### Examples
+
+```fortran
+program demo
+    use stdlib_system, only: get_argument
+    implicit none
+    character(len=:), allocatable :: prog
+    call get_argument(0, prog)
+    print '(a)', prog
+end program demo
+```
+
+
+
+### `get_variable`
+
+#### Status
+
+Experimental
+
+#### Description
+
+Obtain the value of an environment variable
+
+#### Syntax
+
+`call [[get_variable(interface)]] (var, val[, stat])`
+
+#### Arguments
+
+`val`: Name of the environment variable to read.
+       Can be either a default character or a [[string_type]].
+       This is an `intent(in)` argument.
+
+`val`: Value of the environment variable on return.
+       Can be either a deferred length character or a [[string_type]].
+       This is an `intent(out)` argument.
+
+`stat`: Default integer, contains status of operation, zero in case of success.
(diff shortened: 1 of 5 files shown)


awvwgk:
Implementing this in Fortran seems not possible due to a long-standing bug (= feature) in GCC (see [gcc#42954](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=42954), [gcc#55007](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=55007)). This also means the macro for sleep never worked. Only way forward seems to implement it in C directly.

urbanjost:
I found I almost always wanted a default for an environment variable and that I often had an alternate variable to look at so I think having a default would be nice;  and because a functional approach is so easily nestable maybe consider a function instead of a subroutine, which also eliminates some issues with whether the result is allocatable or fixed length. If it is a function with a default things like in [system_getenv](https://urbanjost.github.io/general-purpose-fortran/system_getenv.3m_system.html)
are easier:
```fortran
   program demo_system_getenv
   use M_system, only : ge=>system_getenv
   implicit none 
   character(len=:),allocatable :: TMPDIR,USER
   
      ! look first for USER then LOGNAME then USERNAME
      USER=ge('USER', ge('LOGNAME', ge('USERNAME', 'UNKNOWN')))

      TMPDIR= ge('TMPDIR', ge('TMP', ge('TEMPDIR', ge('TEMP', '/tmp'))))
      write(*,*)'favorite scratch area is ',TMPDIR 

   end program demo_system_getenv
```

awvwgk:
Using a function sounds like a better idea, how do you handle the return status in this case? Also, do you have a good idea how to set variables?

urbanjost:
I actually just return a null string on error. I like functions to be pure, but you could return an optional parameter, but that is something I prefer to use a subroutine for; so the alternative would be a separate function that just returned an error code; sort of like having to check the allocation of a variable before allocating it, or testing for association.  In practice I have found I rarely care about any error status; and if I did there is always the intrinsic itself for that case.  In the M_system module there are procedures for setting environment variables but they only work on a POSIX platform; but it includes the routines for reading the environment table as well, which is handy for things like "show me all environment variable names starting with "FPM_" and for dumping the environment so you can restore it in a different process. I know there are equivalents on MS_WIndows but I very rarely use MSWindows except with WSL or CygWin so I have never made the routines work with MSWindows. I wonder if instead of always coming up against that if we should start a POSIX library for MSWIndows?  It would not be trivial, but is obviously possible, as CygWin does it.

urbanjost:
Just about (maybe all) Fortran compilers seem to have common posix-like routines but I got tired of all the conditionals needed to use them when writing for multiple compilers; which resulted in M_system once the C interface was standardized; but I used to have wrappers for each compiler that normalized their versions of getenv, fstat, getcwd, ...; which was workable for the platforms I used. A horrible shame that got dropped from the standard; the PXF interface would have been a huge boon to Fortran but the only one I remember fully implementing it was Cray, which only had to do it for their own platform. Since we had a cray as a primary platform we converted a lot of codes to the PXF interface expecting it to become standard and were severely disappointed when everyone did not support it.

doc/specs/stdlib_system.md
```diff
@@ -0,0 +1,169 @@
...
+#### Examples
+
+```fortran
+program demo
+    use stdlib_system, only: get_argument
+    implicit none
+    character(len=:), allocatable :: prog
+    call get_argument(0, prog)
+    print '(a)', prog
+end program demo
```
jvdp1, line 84:
```suggestion
program demo_get_argument
    use stdlib_system, only: get_argument
    implicit none
    character(len=:), allocatable :: prog
    call get_argument(0, prog)
    print '(a)', prog
end program demo_get_argument
```

doc/specs/stdlib_system.md
```diff
@@ -0,0 +1,169 @@
...
+`val`: Value of the environment variable on return.
+       Can be either a deferred length character or a [[string_type]].
+       This is an `intent(out)` argument.
+
+`stat`: Default integer, contains status of operation, zero in case of success.
+        It is an optional argument, in case not present the program will halt for non-zero status.
+        This argument is `intent(out)`.
+
+
+#### Examples
```
jvdp1, line 118:
```suggestion
#### Example
```

doc/specs/stdlib_system.md
```diff
@@ -0,0 +1,169 @@
...
+
+```fortran
+program demo
+    use stdlib_system, only: set_variable
+    implicit none
+    call set_variable("OMP_NUM_THREADS", "1")
+end program demo
```
jvdp1, line 168:
```suggestion
```fortran
program demo_set_variable
    use stdlib_system, only: set_variable
    implicit none
    call set_variable("OMP_NUM_THREADS", "1")
end program demo_set_variable
```

jvdp1 (commented):
Thank you. This sounds good to me.
Question: how to add tests for such procedures?

awvwgk:
Since the implementation of `set_variable` is still broken and I don't find a good way to fix it, I'll drop this patch for now.

PR #605, opened 2021-12-19
[stdlib_math] Add function `diff`

zoziha:
- [x] Add `diff`: Differences and approximate derivatives

#### Description

Computes differences between adjacent elements of an array.

**Note**: Only `diff` for `rank-1/rank-2` arrays at this stage. (If there is a higher `rank` demand in the future, we can consider extending it.)

API: 
For a rank-1 array
```fortran
y = [[stdlib_math(module):diff(interface)]](x [, n=1, prepend=<no value>, append=<no value>])
```
and for a rank-2 array
```fortran
y = [[stdlib_math(module):diff(interface)]](x [, n=1, dim=1, prepend=<no value>, append=<no value>])
```

- `X`: `real/integer` types.
- `n`: Difference order.
- ...

#### Note

Originally `diff` should be putted in `stdlib_linalg`, but considering that `diff` can support `date` drived types in the future, now it is putted in `stdlib_math`.

#### Prior Art
- [keurfonluu/forlab](https://github.com/keurfonluu/Forlab/blob/master/src/lib/forlab.f90#L2673) (Fortran)
- [`diff` in Matlab](https://ww2.mathworks.cn/help/matlab/ref/diff.html?lang=en)
- [`diff` in numpy](https://numpy.org/doc/stable/reference/generated/numpy.diff.html)

### DIFF
--- a/doc/specs/stdlib_math.md
+++ b/doc/specs/stdlib_math.md
@@ -658,11 +658,11 @@ Computes differences between adjacent elements of an array.
 
 #### Syntax
 
-For rank-1 array
+For a rank-1 array
 ```fortran
 y = [[stdlib_math(module):diff(interface)]](x [, n, prepend, append])
 ```
-and for rank-2 array
+and for a rank-2 array
 ```fortran
 y = [[stdlib_math(module):diff(interface)]](x [, n, dim, prepend, append])
 ```
@@ -677,35 +677,36 @@ Pure function.
 
 #### Arguments
 
-Note: The `x`, `prepend` and `append` arguments must have the same `type`, `kind` and `rank`.
-
-`x`: Shall be a `real/integer` and `rank-1/rank-2` array.
+`x`: The array to take a difference of.
+Shall be a `real/integer` and `rank-1/rank-2` array.
 This argument is `intent(in)`.
 
-`n`: Shall be an `integer` scalar.
-This argument is `intent(in)` and `optional`, which is `1` by default.
-It represents to calculate the n-th order difference.
+`n`: How many times to iteratively calculate the difference.
+Shall be an `integer` scalar.
+This argument is `intent(in)` and `optional`, and has value of `1` by default.
 
-`dim`: Shall be an `integer` scalar.
-This argument is `intent(in)` and `optional`, which is `1` by default.
-It gives the dimension of the input array along which the difference is calculated, between `1` and `rank(x)`.
+`dim`: The dimension of the input array along which to calculate the difference.
+Its value must be between `1` and `rank(x)`.
+Shall be an `integer` scalar.
+This argument is `intent(in)` and `optional` and has a value of `1` by default.
 
-`prepend`: Shall be a `real/integer` and `rank-1/rank-2` array, which is no value by default.
-This argument is `intent(in)` and `optional`.
-
-`append`: Shall be a `real/integer` and `rank-1/rank-2` array, which is no value by default.
-This argument is `intent(in)` and `optional`.
+`prepend`, `append`: Arrays to prepend or append to a along axis prior to performing the difference.
+The dimension and shape must match a except along axis.
+Shall be a `real/integer` and `rank-1/rank-2` array.
+This argument is `intent(in)` and `optional`, which is no value by default.
 
 Note: 
-- If the value of `n` is less than or equal to `0`, the return value of `y` is `x`. (Not recommended)
-- If the value of `dim` is not equal to `1` or `2`,
-`1` will be used by the internal process of `diff`. (Not recommended)
+- The `x`, `prepend` and `append` arguments must have the same `type`, `kind` and `rank`.
+- If the value of `n` is less than or equal to `0` (which is not recommended), the return value of `diff` is `x`.
+- If the value of `dim` is not equal to `1` or `2` (which is not recommended),
+`1` will be used by the internal process of `diff`.
+
 
 #### Result value
 
-Note: That `y` generally has one fewer element than `x`.
-
-Returns a `real/integer` and `rank-1/rank-2` array.
+Returns the finite difference of the input array.
+Shall be a `real/integer` and `rank-1/rank-2` array.
+When both `prepend` and `append` are not present, the result `y` has one fewer element than `x` alongside the dimension `dim`.
 
 #### Example
 
@@ -720,16 +721,16 @@ program demo_diff
     integer :: A(3, 3) = reshape([1, 7, 17, 3, 11, 19, 5, 13, 23], [3, 3])
     integer :: Y(3, 2)
     
-    print *, diff(i)        !! [0, 1, 1, 2, 3, 5]
-    print *, diff(x, 2)     !! [5.0, 5.0, 5.0, 5.0]
+    print *, diff(i)        ! [0, 1, 1, 2, 3, 5]
+    print *, diff(x, 2)     ! [5.0, 5.0, 5.0, 5.0]
     
     Y = diff(A, n=1, dim=2)
-    print *, Y(1, :)        !! [2, 2]
-    print *, Y(2, :)        !! [4, 2]
-    print *, Y(3, :)        !! [2, 4]
+    print *, Y(1, :)        ! [2, 2]
+    print *, Y(2, :)        ! [4, 2]
+    print *, Y(3, :)        ! [2, 4]
 
-    print *, diff(i, prepend=[0]) !! [1, 0, 1, 1, 2, 3, 5]
-    print *, diff(i, append=[21]) !! [0, 1, 1, 2, 3, 5, 8]
+    print *, diff(i, prepend=[0]) ! [1, 0, 1, 1, 2, 3, 5]
+    print *, diff(i, append=[21]) ! [0, 1, 1, 2, 3, 5, 8]
     
 end program demo_diff
 ```
--- a/doc/specs/stdlib_math.md
+++ b/doc/specs/stdlib_math.md
@@ -677,7 +677,7 @@ Pure function.
 
 #### Arguments
 
-Note: The `x`, `prepend` and `append` arguments must have same `type`, `kind` and `rank`.
+Note: The `x`, `prepend` and `append` arguments must have the same `type`, `kind` and `rank`.
 
 `x`: Shall be a `real/integer` and `rank-1/rank-2` array.
 This argument is `intent(in)`.
@@ -688,7 +688,7 @@ It represents to calculate the n-th order difference.
 
 `dim`: Shall be an `integer` scalar.
 This argument is `intent(in)` and `optional`, which is `1` by default.
-It represents to calculate the difference along which dimension.
+It gives the dimension of the input array along which the difference is calculated, between `1` and `rank(x)`.
 
 `prepend`: Shall be a `real/integer` and `rank-1/rank-2` array, which is no value by default.
 This argument is `intent(in)` and `optional`.
@@ -728,5 +728,8 @@ program demo_diff
     print *, Y(2, :)        !! [4, 2]
     print *, Y(3, :)        !! [2, 4]
 
+    print *, diff(i, prepend=[0]) !! [1, 0, 1, 1, 2, 3, 5]
+    print *, diff(i, append=[21]) !! [0, 1, 1, 2, 3, 5, 8]
+    
 end program demo_diff
 ```
--- a/doc/specs/stdlib_math.md
+++ b/doc/specs/stdlib_math.md
@@ -654,15 +654,17 @@ end program demo_math_all_close
 
 #### Description
 
-Computes differences between adjacent elements of the input array.
+Computes differences between adjacent elements of an array.
 
 #### Syntax
 
+For rank-1 array
 ```fortran
-!> For rank-1 array
-Y = [[stdlib_math(module):diff(interface)]](X [, n])
-!> and for rank-2 array
-Y = [[stdlib_math(module):diff(interface)]](X [, n, dim])
+y = [[stdlib_math(module):diff(interface)]](x [, n, prepend, append])
+```
+and for rank-2 array
+```fortran
+y = [[stdlib_math(module):diff(interface)]](x [, n, dim, prepend, append])
 ```
 
 #### Status
@@ -675,10 +677,9 @@ Pure function.
 
 #### Arguments
 
-Note: If the value of `dim` is not equal to `1` or `2`,
-`1` will be used by the internal process of `diff`. (Not recommended)
+Note: The `x`, `prepend` and `append` arguments must have same `type`, `kind` and `rank`.
 
-`X`: Shall be a `real/integer` and `rank-1/rank-2` array.
+`x`: Shall be a `real/integer` and `rank-1/rank-2` array.
 This argument is `intent(in)`.
 
 `n`: Shall be an `integer` scalar.
@@ -689,9 +690,20 @@ It represents to calculate the n-th order difference.
 This argument is `intent(in)` and `optional`, which is `1` by default.
 It represents to calculate the difference along which dimension.
 
+`prepend`: Shall be a `real/integer` and `rank-1/rank-2` array, which is no value by default.
+This argument is `intent(in)` and `optional`.
+
+`append`: Shall be a `real/integer` and `rank-1/rank-2` array, which is no value by default.
+This argument is `intent(in)` and `optional`.
+
+Note: 
+- If the value of `n` is less than or equal to `0`, the return value of `y` is `x`. (Not recommended)
+- If the value of `dim` is not equal to `1` or `2`,
+`1` will be used by the internal process of `diff`. (Not recommended)
+
 #### Result value
 
-Note: That `Y` has one fewer element than `X`.
+Note: That `y` generally has one fewer element than `x`.
 
 Returns a `real/integer` and `rank-1/rank-2` array.
 
--- a/doc/specs/stdlib_math.md
+++ b/doc/specs/stdlib_math.md
@@ -528,3 +528,72 @@ program demo_math_all_close
     
 end program demo_math_all_close
 ```
+
+### `diff`
+
+#### Description
+
+Computes differences between adjacent elements of the input array.
+
+#### Syntax
+
+```fortran
+!> For rank-1 array
+Y = [[stdlib_math(module):diff(interface)]](X [, n])
+!> and for rank-2 array
+Y = [[stdlib_math(module):diff(interface)]](X [, n, dim])
+```
+
+#### Status
+
+Experimental.
+
+#### Class
+
+Pure function.
+
+#### Arguments
+
+Note: If the value of `dim` is not equal to `1` or `2`,
+`1` will be used by the internal process of `diff`. (Not recommended)
+
+`X`: Shall be a `real/integer` and `rank-1/rank-2` array.
+This argument is `intent(in)`.
+
+`n`: Shall be an `integer` scalar.
+This argument is `intent(in)` and `optional`, which is `1` by default.
+It represents to calculate the n-th order difference.
+
+`dim`: Shall be an `integer` scalar.
+This argument is `intent(in)` and `optional`, which is `1` by default.
+It represents to calculate the difference along which dimension.
+
+#### Result value
+
+Note: That `Y` has one fewer element than `X`.
+
+Returns a `real/integer` and `rank-1/rank-2` array.
+
+#### Example
+
+```fortran
+program demo_diff
+
+    use stdlib_math, only: diff
+    implicit none
+    
+    integer :: i(7) = [1, 1, 2, 3, 5, 8, 13]
+    real    :: x(6) = [0, 5, 15, 30, 50, 75]
+    integer :: A(3, 3) = reshape([1, 7, 17, 3, 11, 19, 5, 13, 23], [3, 3])
+    integer :: Y(3, 2)
+    
+    print *, diff(i)        !! [0, 1, 1, 2, 3, 5]
+    print *, diff(x, 2)     !! [5.0, 5.0, 5.0, 5.0]
+    
+    Y = diff(A, n=1, dim=2)
+    print *, Y(1, :)        !! [2, 2]
+    print *, Y(2, :)        !! [4, 2]
+    print *, Y(3, :)        !! [2, 4]
+
+end program demo_diff
+```
--- a/src/stdlib_math.fypp
+++ b/src/stdlib_math.fypp
@@ -15,6 +15,7 @@ module stdlib_math
 #:endif
     public :: DEFAULT_LINSPACE_LENGTH, DEFAULT_LOGSPACE_BASE, DEFAULT_LOGSPACE_LENGTH
     public :: arange, is_close, all_close
+    public :: diff
 
     integer, parameter :: DEFAULT_LINSPACE_LENGTH = 100
     integer, parameter :: DEFAULT_LOGSPACE_LENGTH = 50
@@ -326,6 +327,26 @@ module stdlib_math
         #:endfor
         #:endfor
     end interface all_close
+    
+    !> Version: experimental
+    !>
+    !> Computes differences between adjacent elements of the input array.
+    !> ([Specification](../page/specs/stdlib_math.html#diff))
+    interface diff
+        #:set RI_KINDS_TYPES = REAL_KINDS_TYPES + INT_KINDS_TYPES
+        #:for k1, t1 in RI_KINDS_TYPES
+        pure module function diff_1_${k1}$(x, n) result(Y)
+            ${t1}$, intent(in) :: x(:)
+            integer, intent(in), optional :: n
+            ${t1}$, allocatable :: Y(:)
+        end function diff_1_${k1}$
+        pure module function diff_2_${k1}$(X, n, dim) result(Y)
+            ${t1}$, intent(in) :: X(:, :)
+            integer, intent(in), optional :: n, dim
+            ${t1}$, allocatable :: Y(:,:)
+        end function diff_2_${k1}$
+        #:endfor
+    end interface diff
 
 contains
 
--- a/src/stdlib_math.fypp
+++ b/src/stdlib_math.fypp
@@ -362,20 +362,22 @@ module stdlib_math
     
     !> Version: experimental
     !>
-    !> Computes differences between adjacent elements of the input array.
+    !> Computes differences between adjacent elements of an array.
     !> ([Specification](../page/specs/stdlib_math.html#diff))
     interface diff
         #:set RI_KINDS_TYPES = REAL_KINDS_TYPES + INT_KINDS_TYPES
         #:for k1, t1 in RI_KINDS_TYPES
-        pure module function diff_1_${k1}$(x, n) result(Y)
+        pure module function diff_1_${k1}$(x, n, prepend, append) result(y)
             ${t1}$, intent(in) :: x(:)
             integer, intent(in), optional :: n
-            ${t1}$, allocatable :: Y(:)
+            ${t1}$, intent(in), optional :: prepend(:), append(:)
+            ${t1}$, allocatable :: y(:)
         end function diff_1_${k1}$
-        pure module function diff_2_${k1}$(X, n, dim) result(Y)
-            ${t1}$, intent(in) :: X(:, :)
+        pure module function diff_2_${k1}$(X, n, dim, prepend, append) result(y)
+            ${t1}$, intent(in) :: x(:, :)
             integer, intent(in), optional :: n, dim
-            ${t1}$, allocatable :: Y(:,:)
+            ${t1}$, intent(in), optional :: prepend(:, :), append(:, :)
+            ${t1}$, allocatable :: y(:, :)
         end function diff_2_${k1}$
         #:endfor
     end interface diff
--- a/src/stdlib_math_diff.fypp
+++ b/src/stdlib_math_diff.fypp
@@ -1,4 +1,4 @@
-#! Originally written in 2016 by Keurfon Luu (keurfonluu@outlook.com)
+#! Inspired by original code (MIT license) written in 2016 by Keurfon Luu (keurfonluu@outlook.com)
 #! https://github.com/keurfonluu/Forlab/blob/master/src/lib/forlab.f90#L2673
 
 #:include "common.fypp"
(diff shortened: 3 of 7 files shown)


src/stdlib_math_diff.fypp
```diff
@@ -0,0 +1,71 @@
...
+
+    implicit none
+
+contains
+
+    #! `diff` computes differences of arrays of the ${t1}$ type.
+    
+    #:for k1, t1 in RI_KINDS_TYPES
+    pure module function diff_1_${k1}$(X, n) result(Y)
+        ${t1}$, intent(in) :: X(:)
```
gareth-nx, line 16: Not sure if this interface has been discussed anywhere so far? A few potential issues that come to mind are:
* This routine allocates its output on every call. I can imagine in some situations that could be inefficient. Users might wish to make a more efficient `diff` call by providing the output data. For that case a subroutine interface may be preferable.
* When computing differences in practice, there are different approaches that may be desired to treat the end-points. The current routines reduce the size of the array. But one might also wish to preserve the array size and do some kind of extrapolation at the missing entry. There can be a number of cases here (e.g. forward differences, backward differences, centred differences, with different implications for end-points).

Currently the routine won't support such things. While that might prove to be fine, I'd like to see some discussion to ensure we've chosen the right interface interface.

jvdp1:
What is the state-of-the-art of such a function (e.g., how is it designed in other languages?)?

src/stdlib_math_diff.fypp
```diff
@@ -12,27 +12,47 @@ contains
...
+            return
+        end if
 
         Y = X
         do i = 1, n_
             Y = Y(2:) - Y(:size(Y) - 1)
         end do
+        
+        if (size_prepend > 0) Y = [prepend(:), Y]
+        if (size_append > 0) Y = [Y, append(:)]
```
gareth-nx, line 42: I think that in this implementation, the meaning of `prepend` and `append` is different to in numpy (?). 

In numpy the idea is that we compute the standard `diff` of the array defined by `[prepend, X, append]`. So the output will have a size equal to `(-1 + size(prepend) + size(X) + size(append))`. 

Here is an illustration in numpy:
```python
import numpy
# Here is the standard diff, where the output_size is (input_size - 1)
numpy.diff(numpy.array([-2, -1, 0, 2, 3, 6, 10, 20]))
# array([ 1,  1,  2,  1,  3,  4, 10])

# Here we do the same thing, but the first 2 elements are set with prepend, and 
# the last 2 with append. It gives the same answer as above.
numpy.diff(numpy.array([0, 2, 3, 6]), prepend = [-2, -1], append=[10, 20])
# array([ 1,  1,  2,  1,  3,  4, 10])
```

zoziha: Sorry, I misunderstood it, now it's corrected.

src/stdlib_math_diff.fypp
```diff
@@ -0,0 +1,102 @@
...
+        if (size_prepend > 0) then
+            Y = [prepend, X]
+        else
+            Y = X
+        end if
+        if (size_append > 0) Y = [Y, append]
+        
+        do i = 1, n_
+            Y = Y(2:) - Y(:size(Y) - 1)
+        end do
```
gareth-nx, line 43: I think there's another difference with numpy here. In the latter, the output size is `(size(x) + size(prepend) + size(append) - n)`, for example:
```python
In [1]: import numpy

In [2]: numpy.diff([1, 4, 7, 20])
Out[2]: array([ 3,  3, 13])

In [3]: numpy.diff([1, 4, 7, 20], n=2)
Out[3]: array([ 0, 10])

# Cases using prepend and append

In [4]: numpy.diff([1, 4, 7, 20], n=2, prepend=[1,1])
Out[4]: array([ 0,  3,  0, 10])

In [5]: numpy.diff([1, 4, 7, 20], n=2, prepend=[1,1], append=[50, 50])
Out[5]: array([  0,   3,   0,  10,  17, -30])

```

zoziha: The feature of the allocatable array is used here, and the length of the array is updated according to the result each time the loop is looped.
```fortran
integer :: x(8) = [1, 1, 1, 4, 7, 20, 50, 50]
integer, allocatable :: y(:)

y = x
do i = 1, 2
    Y = Y(2:) - Y(:size(Y) - 1)
end do

print "(1x, *(i0, 2x))", Y  !! [ 0,  3,  0,  10,  17,  -30 ]

end
```

gareth-nx: Oh my mistake, interesting!

gareth-nx: It would be good to understand whether all the reallocations lead to performance hits when `n` is large. An alternative approach would be to use `Y(2:) = Y(2:) - Y(:size(Y)-1)` in the loop, and then after the loop, have a line `Y = Y(n_+1:)`

doc/specs/stdlib_math.md
```diff
@@ -649,3 +649,72 @@ program demo_math_all_close
     
 end program demo_math_all_close
 ```
+
+### `diff`
+
+#### Description
+
+Computes differences between adjacent elements of the input array.
```
jvdp1, line 657:
```suggestion
Computes differences between adjacent elements of an array.
```

doc/specs/stdlib_math.md
```diff
@@ -649,3 +649,72 @@ program demo_math_all_close
...
+
+#### Syntax
+
+```fortran
+!> For rank-1 array
+Y = [[stdlib_math(module):diff(interface)]](X [, n])
+!> and for rank-2 array
+Y = [[stdlib_math(module):diff(interface)]](X [, n, dim])
+```
```
jvdp1, line 666:
````suggestion
!> For rank-1 array
```fortran
Y = [[stdlib_math(module):diff(interface)]](X [, n])
```
and for a rank-2 array
```fortran
Y = [[stdlib_math(module):diff(interface)]](X [, n, dim])
```
````

File: doc/specs/stdlib_math.md [hunk 6]
```diff
@@ -649,3 +649,72 @@ program demo_math_all_close
...
+Experimental.
+
+#### Class
+
+Pure function.
+
+#### Arguments
+
+Note: If the value of `dim` is not equal to `1` or `2`,
+`1` will be used by the internal process of `diff`. (Not recommended)
```
jvdp1, line 679: I suggest to mention that after defining all arguments. Because, at this stage, the reader doesn't know what `dim` is (and for what it is used

doc/specs/stdlib_math.md
```diff
@@ -649,3 +649,72 @@ program demo_math_all_close
...
+
+`dim`: Shall be an `integer` scalar.
+This argument is `intent(in)` and `optional`, which is `1` by default.
+It represents to calculate the difference along which dimension.
+
+#### Result value
+
+Note: That `Y` has one fewer element than `X`.
+
+Returns a `real/integer` and `rank-1/rank-2` array.
```
jvdp1, line 696:
```suggestion
Returns a `real/integer` and `rank-1/rank-2` array. The size of `Y` is the size of `X` minus 1.
```

src/stdlib_math.fypp
```diff
@@ -359,6 +359,28 @@ module stdlib_math
         #:endfor
         #:endfor
     end interface all_close
+    
+    !> Version: experimental
+    !>
+    !> Computes differences between adjacent elements of the input array.
```
jvdp1, line 365:
```suggestion
    !> Computes differences between adjacent elements of an array.
```

src/stdlib_math.fypp
```diff
@@ -359,6 +359,28 @@ module stdlib_math
...
+            ${t1}$, allocatable :: Y(:)
+        end function diff_1_${k1}$
+        pure module function diff_2_${k1}$(X, n, dim, prepend, append) result(Y)
+            ${t1}$, intent(in) :: X(:, :)
+            integer, intent(in), optional :: n, dim
+            ${t1}$, intent(in), optional :: prepend(:, :), append(:, :)
+            ${t1}$, allocatable :: Y(:, :)
+        end function diff_2_${k1}$
+        #:endfor
+    end interface diff
```
jvdp1, line 383: Should this be extended to all dimensions supported by Fortran (up to 7 or to 15)?

src/stdlib_math_diff.fypp
```diff
@@ -0,0 +1,102 @@
+#! Originally written in 2016 by Keurfon Luu (keurfonluu@outlook.com)
+#! https://github.com/keurfonluu/Forlab/blob/master/src/lib/forlab.f90#L2673
```
jvdp1, line 2: what is the license? Please mention it (maybe also in the specs)?

src/stdlib_math_diff.fypp
```diff
@@ -0,0 +1,102 @@
...
+
+#:include "common.fypp"
+#:set RI_KINDS_TYPES = REAL_KINDS_TYPES + INT_KINDS_TYPES
+submodule (stdlib_math) stdlib_math_diff
+
+    implicit none
+
+contains
+
+    #! `diff` computes differences of arrays of the ${t1}$ type.
```
jvdp1, line 12:
```suggestion
    #! `diff` computes differences of adjacent elements of an array of the ${t1}$ type.
```

src/stdlib_math_diff.fypp
```diff
@@ -0,0 +1,102 @@
...
+    
+    #:for k1, t1 in RI_KINDS_TYPES
+    pure module function diff_1_${k1}$(X, n, prepend, append) result(Y)
+        ${t1}$, intent(in) :: X(:)
```
jvdp1, line 16: Could you use lowercases, please (as mentioned in the "style guide")?

src/stdlib_math_diff.fypp
jvdp1, line 43: Why is there an iterative process? the `i` is not involved in the loop body.

src/stdlib_math_diff.fypp
```diff
@@ -0,0 +1,102 @@
...
+            allocate(Y(0))
+            return
+        end if
+
+        if (size_prepend > 0) then
+            Y = [prepend, X]
+        else
+            Y = X
+        end if
+        if (size_append > 0) Y = [Y, append]
```
jvdp1, line 39: Is it not possible to avoid this reallocation?
E.g.
1) allocate `Y` with the full size
2) copy all elements in their right place

src/stdlib_math_diff.fypp
```diff
@@ -0,0 +1,102 @@
...
+        
+        if (dim_ == 1) then
+            allocate(tmp(size(X, 1)+size_prepend+size_append, size(X, 2)))
+            if (size_prepend > 0) tmp(:size_prepend, :) = prepend
+            tmp(size_prepend+1:size(X, 1)+size_prepend, :) = X
+            if (size_append > 0) tmp(size(X, 1)+size_prepend+1:, :) = append
+            Y = tmp
+            do i = 1, n_
+                Y = Y(2:, :) - Y(:size(Y, 1) - 1, :)
+            end do
```
jvdp1, line 85: `i` is not involved in the loop body.

jvdp1 (commented):
Thank you for this PR. Here are some suggestions and comments.

src/stdlib_math.fypp
zoziha: Fortran does not seem to be as easy to create an array interface for any `rank` like dynamic languages python and matlab. From a cost point of view, it seems more appropriate to only support `rank-1/rank-2`.

If there is a higher demand for `rank` in the future, maybe we can expand it?

gareth-nx: The `stdlib_stats` routines use some fancy fypp to enable treating any rank and `dim`. See around [this part of the code](https://github.com/fortran-lang/stdlib/blob/03b1695fdd514c5ad9dc5e1292b6eda178d28ab4/src/stdlib_stats_median.fypp#L93)

src/stdlib_math_diff.fypp
```diff
@@ -0,0 +1,117 @@
...
+            return
+        end if
+
+        allocate(work(size_x + size_prepend + size_append))
+        if (size_prepend > 0) work(:size_prepend) = prepend
+        work(size_prepend+1:size_prepend+size_x) = x
+        if (size_append > 0) work(size_prepend+size_x+1:) = append
+        
+        do i = 1, n_
+            work(2:) = work(2:) - work(1:size(work)-1)
```
jvdp1, line 47: I don't understand why there is a loop, since `i` is not involve in the expression 
```fortran
            work(2:) = work(2:) - work(1:size(work)-1)
```

As written now, it will repear `n_` the same opereation. What do I miss?

gareth-nx: @jvdp1 I agree it is just repeating the difference operation `n_` times. But that is the intention -- and I can't see how this would be done more cleanly without a loop?

jvdp1: Ok, I undertand now the meaning of `n_`. Thank you. Fine as is.

zoziha: When I submitted `diff` for the **first** time, I took advantage of the allocatable array feature. This is correct, but I am worried that the repeated allocation of arrays is not easy to understand, and maybe **not efficient**. (see https://github.com/fortran-lang/stdlib/pull/605#discussion_r776642724)

So the **second** time I adopted @gareth-nx  scheme and set up a `work` array. In fact, in the case of a large array `x` inputted in `diff`, an extra large array `work` was created to **increase memory usage** (about space complexity).

In addition, there are some unnecessary calculations ([see code](https://github.com/fortran-lang/stdlib/blob/8cc2d66e89512d04593689c79bea9432345d3ddf/src/stdlib_math_diff.fypp#L47)), which I updated in a recent commit.
At the same time, I use `move_alloc` to try to reduce the space complexity of the second scheme ([see code](https://github.com/fortran-lang/stdlib/blob/8cc2d66e89512d04593689c79bea9432345d3ddf/src/stdlib_math_diff.fypp#L50)), and I did a test ([see gist](https://gist.github.com/zoziha/e5986c4a9b752dbe0d95da90b6a5997e)):

```text
 Test case with N=    10000000 ; reps=          10 ; size_append_prepend=           1 ; order=           1
 >>> Time ratio (t_malloc_diff / t_diff) :   1.15813529    
 Test case with N=       10000 ; reps=       10000 ; size_append_prepend=           1 ; order=           1
 >>> Time ratio (t_malloc_diff / t_diff) :   1.50465608    
 Test case with N=         100 ; reps=     1000000 ; size_append_prepend=           1 ; order=           1
 >>> Time ratio (t_malloc_diff / t_diff) :   1.51453304    
 
 Test case with N=    10000000 ; reps=          10 ; size_append_prepend=           0 ; order=           1
 >>> Time ratio (t_malloc_diff / t_diff) :   1.21079636    
 Test case with N=       10000 ; reps=       10000 ; size_append_prepend=           0 ; order=           1
 >>> Time ratio (t_malloc_diff / t_diff) :   1.12502217    
 Test case with N=         100 ; reps=     1000000 ; size_append_prepend=           0 ; order=           1
 >>> Time ratio (t_malloc_diff / t_diff) :   1.47107768    
 
 Test case with N=       10000 ; reps=       10000 ; size_append_prepend=           1 ; order=          10
 >>> Time ratio (t_malloc_diff / t_diff) :   1.05520868    
 Test case with N=       10000 ; reps=       10000 ; size_append_prepend=           1 ; order=         100
 >>> Time ratio (t_malloc_diff / t_diff) :  0.971785545    
 
 Test case with N=       10000 ; reps=       10000 ; size_append_prepend=           0 ; order=          10
 >>> Time ratio (t_malloc_diff / t_diff) :  0.970007539    
 Test case with N=       10000 ; reps=       10000 ; size_append_prepend=           0 ; order=         100
 >>> Time ratio (t_malloc_diff / t_diff) :  0.933800876    
```
Using `move_alloc`, in the demonstration of `reps=1,000,000` on the same array, it obviously took more time. This is normal, it will lower the memory usage, but increase the calculation time.

zoziha:
I updated some details slightly, such as using `move_alloc` to reduce the memory usage of large arrays, but the time is slightly longer. 
I think it is possible, which improves the stability of `diff`, because for small arrays, the operation is inherently faster; and for large arrays, if users are pursuing **the ultimate performance**, it is better for users to implement the `diff` code by themselves, that the performance may be purer. Will the `diff` of the standard library provide a reference for them?
```fortran
do concurrent (i = 1:n_)
    work(:, 1:size(work,2)-i) = work(:, 2:size(work,2)-i+1) - work(:, 1:size(work, 2)-i)
end do
y = work(1:size(work) - n_)
```

zoziha:
I updated some details slightly, such as using `move_alloc` to reduce the memory usage of large arrays, but the time is slightly longer. 
I think it is possible, which improves the stability of `diff`, because for small arrays, the operation is inherently faster; and for large arrays, if users are pursuing **the ultimate performance**, it is better for users to implement the `diff` code by themselves, that the performance may be purer. Will the `diff` of the standard library provide a reference for them?
```fortran
do concurrent (i = 1:n_)
    work(:, 1:size(work,2)-i) = work(:, 2:size(work,2)-i+1) - work(:, 1:size(work, 2)-i)
end do
y = work(1:size(work) - n_)
```

src/stdlib_math_diff.fypp
```diff
@@ -0,0 +1,120 @@
...
+            allocate(y(0))
+            return
+        end if
+
+        allocate(work(size_x + size_prepend + size_append))
+        if (size_prepend > 0) work(:size_prepend) = prepend
+        work(size_prepend+1:size_prepend+size_x) = x
+        if (size_append > 0) work(size_prepend+size_x+1:) = append
+        
+        do concurrent (i = 1:n_)
```
gareth-nx, line 46: Suggest not making this a `do concurrent`, but rather stick to a regular `do`. The code will not be correct if the loop is executed in parallel. 

Actually the code is correct, because the Fortran standard only requires that `do concurrent` can be executed serially in any order. However this seems to have been a design flaw, and I think most users will assume the parallel interpretation. 

So I suggest using regular `do`.

src/stdlib_math.fypp
zoziha: Yes, using fypp can achieve higher-level array rank support, but I don’t know whether we need higher-level rank support at this stage. Maybe we can be conservative?

src/stdlib_math_diff.fypp
zoziha: Suddenly my brain is out of order, I will correct it. -_-

src/stdlib_math_diff.fypp
```diff
@@ -38,16 +37,24 @@ contains
             return
         end if
 
-        allocate(work(size_x + size_prepend + size_append))
+        #! Use a quick exit for the common case, to avoid memory allocation.
```
gareth-nx, line 40: Does the `#` need to be removed?

src/stdlib_math_diff.fypp
```diff
@@ -87,19 +93,32 @@ contains
             return
         end if
         
+        #! Use a quick exit for the common case, to avoid memory allocation.
```
gareth-nx, line 96: I think the `#` needs to be removed?

src/stdlib_math_diff.fypp
```diff
@@ -0,0 +1,137 @@
...
+        end if
+
+        block
+        ${t1}$ :: work(size_x + size_prepend + size_append)
+        if (size_prepend > 0) work(:size_prepend) = prepend
+        work(size_prepend+1:size_prepend+size_x) = x
+        if (size_append > 0) work(size_prepend+size_x+1:) = append
+        
+        do i = 1, n_
+            work(1:size(work)-i) = work(2:size(work)-i+1) - work(1:size(work)-i)
```
jvdp1, line 53: To avoid all these `size(work)`, you could define an integer `size_work = size_x + size_prepend + size_append` after defining its different component.
The compiler should remove all these calls to `size(work)`, but I think it will clarify the code (at least for me).

doc/specs/stdlib_math.md
```diff
@@ -649,3 +649,84 @@ program demo_math_all_close
...
+`x`: Shall be a `real/integer` and `rank-1/rank-2` array.
+This argument is `intent(in)`.
+
+`n`: Shall be an `integer` scalar.
+This argument is `intent(in)` and `optional`, which is `1` by default.
+It represents to calculate the n-th order difference.
+
+`dim`: Shall be an `integer` scalar.
+This argument is `intent(in)` and `optional`, which is `1` by default.
+It represents to calculate the difference along which dimension.
```
jvdp1, line 691: This sentence sounds strange to me. What do you mean?

jvdp1 (commented):
thank you. Here are a couple of minor comments.

src/stdlib_math_diff.fypp
gareth-nx: Actually it is legal (thanks to fypp). But I suggest removing anyway.

src/stdlib_math_diff.fypp
gareth-nx: On reflection it's legal fypp -- but I still suggest it is removed.

src/stdlib_math_diff.fypp
```diff
@@ -0,0 +1,137 @@
+#! Inspired by original code (MIT license) written in 2016 by Keurfon Luu (keurfonluu@outlook.com)
+#! https://github.com/keurfonluu/Forlab/blob/master/src/lib/forlab.f90#L2673
```
jvdp1, line 2: Following @gareth-nx comment: 
```suggestion
! Inspired by original code (MIT license) written in 2016 by Keurfon Luu (keurfonluu@outlook.com)
! https://github.com/keurfonluu/Forlab/blob/master/src/lib/forlab.f90#L2673
```

src/stdlib_math_diff.fypp
```diff
@@ -0,0 +1,137 @@
...
+
+#:include "common.fypp"
+#:set RI_KINDS_TYPES = REAL_KINDS_TYPES + INT_KINDS_TYPES
+submodule (stdlib_math) stdlib_math_diff
+
+    implicit none
+
+contains
+
+    #! `diff` computes differences of adjacent elements of an array of the ${t1}$ type.
```
jvdp1, line 12:
```suggestion
    ! `diff` computes differences of adjacent elements of an array of the ${t1}$ type.
```

src/stdlib_math_diff.fypp
zoziha: `#!` is [the comment header of `fypp`](https://fypp.readthedocs.io/en/stable/fypp.html#comment-directive), which can be replaced by `!`.

zoziha: `#!` is [the comment header of `fypp`](https://fypp.readthedocs.io/en/stable/fypp.html#comment-directive), which can be replaced by `!`.

src/stdlib_math_diff.fypp
zoziha: Good suggestion, although I don't know what the compiler will do to optimize `size(work)`, but use `size_work` to make it clear that it looks good, I have updated this code.

src/stdlib_math_diff.fypp
awvwgk: That is not a permanent link and is likely to break in case of upstream changes.

zoziha:
- `#!` is [the comment header of `fypp`](https://fypp.readthedocs.io/en/stable/fypp.html#comment-directive), which now is replaced by `!`.
- `size(work, ..)` -> `size_work` scalar. (see https://github.com/fortran-lang/stdlib/pull/605#discussion_r777189157)

src/stdlib_math_diff.fypp
jvdp1: This one is probably better:
```
https://github.com/keurfonluu/Forlab/blob/62b80c031fe5f283e6696be0353c13f4ba5886de/src/lib/forlab.f90#L2673
```
Would a link to the main project not sufficient?

doc/specs/stdlib_math.md
zoziha: [numpy.diff](https://numpy.org/doc/stable/reference/generated/numpy.diff.html):
> `axis`: `int`, `optional`
The axis along which the difference is taken, default is the last axis.

[matlab.diff](https://ww2.mathworks.cn/help/matlab/ref/diff.html?lang=en):
> `Y = diff(X,n,dim)` is the nth difference calculated along the dimension specified by `dim`. The dim input is a positive integer scalar.

src/stdlib_math_diff.fypp
awvwgk: Maybe `https://github.com/keurfonluu/Forlab` is sufficient.

doc/specs/stdlib_math.md
gareth-nx: Perhaps change the last sentence to: `It represents the dimension along which the difference is calculated.`

doc/specs/stdlib_math.md
```diff
@@ -649,3 +649,87 @@ program demo_math_all_close
...
+
+### `diff`
+
+#### Description
+
+Computes differences between adjacent elements of an array.
+
+#### Syntax
+
+For rank-1 array
```
jvdp1, line 661:
```suggestion
For a rank-1 array
```

doc/specs/stdlib_math.md
```diff
@@ -649,3 +649,87 @@ program demo_math_all_close
...
+```fortran
+y = [[stdlib_math(module):diff(interface)]](x [, n, prepend, append])
+```
+and for rank-2 array
```
jvdp1, line 665:
```suggestion
and for a rank-2 array
```

doc/specs/stdlib_math.md
```diff
@@ -649,3 +649,87 @@ program demo_math_all_close
...
+
+Experimental.
+
+#### Class
+
+Pure function.
+
+#### Arguments
+
+Note: The `x`, `prepend` and `append` arguments must have the same `type`, `kind` and `rank`.
```
jvdp1, line 680: I would suggest to move this note at the end of this section (with the other notes).

doc/specs/stdlib_math.md
```diff
@@ -649,3 +649,87 @@ program demo_math_all_close
...
+Note: 
+- If the value of `n` is less than or equal to `0`, the return value of `y` is `x`. (Not recommended)
+- If the value of `dim` is not equal to `1` or `2`,
+`1` will be used by the internal process of `diff`. (Not recommended)
+
+#### Result value
+
+Note: That `y` generally has one fewer element than `x`.
+
+Returns a `real/integer` and `rank-1/rank-2` array.
```
jvdp1, line 708:
```suggestion
Returns a `real/integer` and `rank-1/rank-2` array. When both `prepend` and `append ` are not present, the result `y` has one fewer element than `x` alongside the dimension `dim`.
```

jvdp1 (approved):
LGTM. Pending a few comments, it could be merged IMO, Thank you @zoziha for this PR.

zoziha:
Thank you @jvdp1 for your patient review, I have made changes accordingly :)

jvdp1:
Thank you @zoziha. This PR is fine for me. If @gareth-nx or @awvwgk could have a final look, it would be nice ;)

doc/specs/stdlib_math.md
```diff
@@ -649,3 +649,86 @@ program demo_math_all_close
...
+
+Experimental.
+
+#### Class
+
+Pure function.
+
+#### Arguments
+
+`x`: Shall be a `real/integer` and `rank-1/rank-2` array.
```
milancurcic, line 680:
```suggestion
`x`: The array to take a difference of.
Shall be a `real/integer` and `rank-1/rank-2` array.
```

doc/specs/stdlib_math.md
```diff
@@ -649,3 +649,86 @@ program demo_math_all_close
...
+This argument is `intent(in)`.
+
+`n`: Shall be an `integer` scalar.
+This argument is `intent(in)` and `optional`, which is `1` by default.
+It represents to calculate the n-th order difference.
```
milancurcic, line 685:
```suggestion
`n`: The order of finite differencing to use.
Shall be an `integer` scalar.
This argument is `intent(in)` and `optional` and has value `1` by default.
```

doc/specs/stdlib_math.md
```diff
@@ -649,3 +649,86 @@ program demo_math_all_close
...
+
+`dim`: Shall be an `integer` scalar.
+This argument is `intent(in)` and `optional`, which is `1` by default.
+It gives the dimension of the input array along which the difference is calculated, between `1` and `rank(x)`.
```
milancurcic, line 689:
```suggestion
`dim`: The dimension of the input array along which to calculate the difference.
Its value must be between `1` and `rank(x)`.
Shall be an `integer` scalar.
This argument is `intent(in)` and `optional` and has a value of `1` by default.
```

doc/specs/stdlib_math.md
```diff
@@ -649,3 +649,86 @@ program demo_math_all_close
...
+
+`prepend`: Shall be a `real/integer` and `rank-1/rank-2` array, which is no value by default.
+This argument is `intent(in)` and `optional`.
+
+`append`: Shall be a `real/integer` and `rank-1/rank-2` array, which is no value by default.
+This argument is `intent(in)` and `optional`.
```
milancurcic, line 695: Can you add 1-sentence description of these arguments (what do they mean)?

doc/specs/stdlib_math.md
```diff
@@ -649,3 +649,86 @@ program demo_math_all_close
...
+
+Note: 
+- The `x`, `prepend` and `append` arguments must have the same `type`, `kind` and `rank`.
+- If the value of `n` is less than or equal to `0`, the return value of `y` is `x`. (Not recommended)
```
milancurcic, line 699: Do you mean `diff` instead of `y`?

doc/specs/stdlib_math.md
```diff
@@ -649,3 +649,86 @@ program demo_math_all_close
...
+- If the value of `dim` is not equal to `1` or `2`,
+`1` will be used by the internal process of `diff`. (Not recommended)
```
milancurcic, line 701: This only matters in the case of a rank-2 array, right?

If this is not recommended, why is `dim=1` the default value? I think the default value should be the recommended one.

doc/specs/stdlib_math.md
```diff
@@ -649,3 +649,86 @@ program demo_math_all_close
...
+
+
+#### Result value
+
+Returns a `real/integer` and `rank-1/rank-2` array.
+When both `prepend` and `append` are not present, the result `y` has one fewer element than `x` alongside the dimension `dim`.
```
milancurcic, line 707: It would be useful to here also state the meaning of the resulting value (the finite difference of the input array).

doc/specs/stdlib_math.md
```diff
@@ -649,3 +649,86 @@ program demo_math_all_close
...
+    print *, diff(i)        !! [0, 1, 1, 2, 3, 5]
+    print *, diff(x, 2)     !! [5.0, 5.0, 5.0, 5.0]
+    
+    Y = diff(A, n=1, dim=2)
+    print *, Y(1, :)        !! [2, 2]
+    print *, Y(2, :)        !! [4, 2]
+    print *, Y(3, :)        !! [2, 4]
+
+    print *, diff(i, prepend=[0]) !! [1, 0, 1, 1, 2, 3, 5]
+    print *, diff(i, append=[21]) !! [0, 1, 1, 2, 3, 5, 8]
```
milancurcic, line 731: Are the comments on these lines used by FORD in any way? If not, let's just use a single "!" for comments in examples. Otherwise, newcomers to the language may be confused why are two !! needed here.

milancurcic (approved):
Looks good, thank you! I left several suggestions in the spec doc.

doc/specs/stdlib_math.md
zoziha:
```suggestion
- If the value of `dim` is not equal to `1` or `2` (which is not recommended),
`1` will be used by the internal process of `diff`.
```
How about changing it to this?

doc/specs/stdlib_math.md
```diff
@@ -649,3 +649,88 @@ program demo_math_all_close
...
+
+#### Arguments
+
+`x`: The array to take a difference of.
+Shall be a `real/integer` and `rank-1/rank-2` array.
+This argument is `intent(in)`.
+
+`n`: The order of finite differencing to use.
+Shall be an `integer` scalar.
+This argument is `intent(in)` and `optional`, which is `1` by default.
```
milancurcic, line 686: I apologize, now I realize that this wording could be confused for the accuracy order of the finite difference scheme. How about:

```suggestion
`n`: How many times to iteratively calculate the difference.
Shall be an `integer` scalar.
This argument is `intent(in)` and `optional`, and has value of `1` by default.
```

zoziha: Thanks, I feel good, now updated.

milancurcic:
@gareth-nx @awvwgk are you OK with this PR going forward?

awvwgk (approved):
Looks good to me.

PR #609, opened 2021-12-22
Consistent spec titles

### LINKED ISSUE #608: Spec titles are not consistent in the rendered docs
Spec titles in the TOC in the left-side bar are not consistent. Most are module names and some are descriptive. There are a few similar inconsistencies in the list of modules in the body of the page.

The names in the bar are controlled by the `title` attribute in each spec md document.
specs_toc

milancurcic:
This PR:

* Renames several spec title attributes to make them consistent with others (module names)
* A few similar edits to the index.md of the Spec top-level page, for consistency.

Closes #608.

### DIFF
--- a/doc/specs/stdlib_hash_procedures.md
+++ b/doc/specs/stdlib_hash_procedures.md
@@ -1,5 +1,5 @@
 ---
-title: Hash procedures
+title: hash
 ---
 
 # The `stdlib_hash_32bit` and `stdlib_hash_64bit` modules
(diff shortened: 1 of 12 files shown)


awvwgk (approved):

doc/specs/stdlib_hash_procedures.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Hash procedures
+title: hash
```
jvdp1, line 2: Actually I would prefer reading `Hash procedures` over `hash`, which is more informative to me.
E..g., `Numerical integration` can be easier understood than `quadrature`.

jvdp1 (approved):
I approve this PR because it makes all titles using the same rule. However, I would prefer more explicit titles.

doc/specs/stdlib_hash_procedures.md
milancurcic: Why don't we choose whether to list descriptive names or module names and use it consistently? I assumed we were listing module names. Can we merge this as is, and I'll open a new issue to discuss whether to switch this table to descriptive names?

jvdp1: Yes, merge it as is. At least it is consistent now. Thank you.

PR #613, opened 2021-12-31
Ignore hash testing binaries and logs

ghbrown:
During a previous PR (#499) I accidentally committed changes which included the addition of some `.bin` and `.txt` files generated in the hash function testing process. @jvdp1 asked that I remove said files and create a new PR that `.gitignore`s them.

I have ignored the `.bin` files by adding them to the top level `.gitignore`, and added `log` to the `.txt` file names so they get caught by an existing entry (`*log*.txt`) in the `.gitignore`. Alternatively, I could add `test` or `log` to the `.bin` files as well so they are caught by existing ignore entries, if this is preferable.

Related question: 
The current `.gitignore` also ends up leaving a lot of untracked files and executables from the manual make process. I think this is largely fine, since we use `fypp` which makes it hard to ignore all `.f90` files (for example), but am I correct in saying that there are no make rules to remove these in-tree artifacts (I've tried `make -f Makefile.manual clean`, but it does not seem to work)?


awvwgk (approved):

jvdp1 (approved):
Thank you. I will merge it.

PR #617, opened 2022-01-12
Made format constants public

### LINKED ISSUE #570: Provide precision-preserving edit descriptors
### Motivation

In [#505](https://github.com/fortran-lang/stdlib/issues/505#issuecomment-971396472) @epagone proposed:

> Considering that these formats are more useful than only for savetxt and loadtxt, would it make sense to make them available on their own in stdlib_io and simply use them? I am thinking, for example, of any sort of I/O where round-trip accuracy is required (e.g. writing to a CSV, JSON, TOML, YAML, etc...) to override a default behaviour that might be less strict.

### Prior Art

Currently the formats used for `savetxt` and `loadtxt` are private to the `stdlib_io` and not usable from the outside:

https://github.com/fortran-lang/stdlib/blob/57cfaf011ee11001b35bbdde43841a5a1f8cf443/src/stdlib_io.fypp#L22-L32

They work with scalars and arrays alike:

```
print FMT_REAL_DP, 1.23d0
print FMT_REAL_DP, [1.23d0, 4.56d0]
```

outputs

```
 1.2300000000000000E+000
 1.2300000000000000E+000  4.5599999999999996E+000
```

So it seems to me that the solution could be to make the existing constants public. @epagone @certik do you foresee any scenario where it wouldn't work?

### Additional Information

@epagone would you like to submit a PR for this?

lewisfish:
Hey,
This resolves #570.
Pull request makes the format constants from stdlib_io public.

Let me know if I need to do anything else as this is my first pull request, so not 100% sure of the protocol.

Lewis

### DIFF
--- a/src/stdlib_io.fypp
+++ b/src/stdlib_io.fypp
@@ -33,6 +33,9 @@ module stdlib_io
     FMT_COMPLEX_XDP = '(*(es26.18e3,1x,es26.18e3))', &
     FMT_COMPLEX_QP = '(*(es44.35e4,1x,es44.35e4))'
 
+  public :: FMT_INT, FMT_REAL_SP, FMT_REAL_DP, FMT_REAL_XDP, FMT_REAL_QP
+  public :: FMT_COMPLEX_SP, FMT_COMPLEX_DP, FMT_COMPLEX_XDP, FMT_COMPLEX_QP
+
   !> Version: experimental
   !>
   !> Read a whole line from a formatted unit into a string variable


awvwgk:
Thanks for starting this. Have a look at https://github.com/fortran-lang/stdlib/blob/HEAD/doc/specs/stdlib_io.md to add a description of the constants which should become part of the public API.

awvwgk:
Another point for discussion is how the format specifier should be exported. Do we want to be able to use it in a context like this:

```f90
use stdlib_io, only : FMT_REAL_SP, FMT_COMPLEX_SP
implicit none

print '('//FMT_REAL_SP//',1x,'//FMT_COMPLEX_SP//')', 1.0, (2.0, 3.0)
end
```

I wonder whether the `*` repeat will not mess this up.

Further, what about non-advancing output, some format specifies have a trailing skip (`1x` for int and real), other don't (complex). Should we include an explicit break for such case, *i.e.* `(es24.16e3,:,1x)`?

lewisfish:
> Thanks for starting this. Have a look at https://github.com/fortran-lang/stdlib/blob/HEAD/doc/specs/stdlib_io.md to add a description of the constants which should become part of the public API.

Added now, let me know if you have any comments on the format of this as wasn't too sure what to copy format-wise from the other examples in the docs.

> I wonder whether the `*(...)` repeat will mess this up.
> 
> Further, what about non-advancing output, some format specifies have a trailing skip (`1x` for int and real), other don't (complex). Should we include an explicit break for such case, _i.e._ `(es24.16e3,:,1x)`?

Is there any reason the complex doesn't have the trailing skip? It would make more sense if all the formats had the same output

milancurcic:
Good points, I think we should export constants without the `*(...)` and the `1x`s, and let the user adapt them for arrays if needed.

> Is there any reason the complex doesn't have the trailing skip? It would make more sense if all the formats had the same output

I think I implemented this and I don't remember why it is like this. It's possible that I did it by mistake and the tests were insufficient to make it an issue.

lewisfish:
> Good points, I think we should export constants without the `*(...)` and the `1x`s, and let the user adapt them for arrays if needed.
> 
> > Is there any reason the complex doesn't have the trailing skip? It would make more sense if all the formats had the same output
> 
> I think I implemented this and I don't remember why it is like this. It's possible that I did it by mistake and the tests were insufficient to make it an issue.

Great, I'm happy to implement this. 
save_txt uses these constants, the easiest way I see to implement this would be to remove the * and 1x from the format constants then concat them in the required places in save_txt. Does that make sense?

milancurcic:
> save_txt uses these constants, the easiest way I see to implement this would be to remove the * and 1x from the format constants then concat them in the required places in save_txt. Does that make sense?

That seems fine to me. :+1:

lewisfish:
Okay, updated the code and docs. Let me know if anything needs changed or adjusted.

doc/specs/stdlib_io.md
```diff
@@ -273,3 +273,44 @@ program demo_getline
...
+    r64 = 100.0_real64
+    c32 = cmplx(100.0_real32, kind=real32)
+    c64 = cmplx(100.0_real64, kind=real64)
+
+    print "2("//FMT_INT//",1x)", i32, i64
+    print FMT_REAL_SP, r32
+    print FMT_REAL_DP, r64
+    print FMT_COMPLEX_SP, c32
+    print FMT_COMPLEX_DP, c64
+
```
milancurcic, line 315:
```suggestion

end program demo_fmt_constants
```

doc/specs/stdlib_io.md
```diff
@@ -273,3 +273,44 @@ program demo_getline
...
+```
```
milancurcic, line 316: I suggest also showing the output of the example program, considering that these constants are all about output formatting.

milancurcic (approved):
Thank you!

awvwgk (approved):

Could you add a note about this change in https://github.com/fortran-lang/stdlib/blob/HEAD/CHANGELOG.md as well?

lewisfish:
> Thanks for sharing. Looks good to me.
> 
> Could you add a note about this change in https://github.com/fortran-lang/stdlib/blob/HEAD/CHANGELOG.md as well?

Done. Thanks for all your and @milancurcic help in getting this done!

doc/specs/stdlib_io.md
```diff
@@ -273,3 +273,45 @@ program demo_getline
...
+### Description
+
+Formatting constants for printing out integer, floating point, and complex numbers at their full precision.
+Provides formats for all kinds as defined in the `stdlib_kinds` module.
+
+### Example
+
+```fortran
+program demo_fmt_constants
+    use, intrinsic :: iso_fortran_env, only : int32, int64, real32, real64
```
jvdp1, line 292: As the desctiption mentions "all kinds as defined in the `stdlib_kinds` module, I suggest to use the kinds defined in stdlib_kinds, instead of those defined in `iso_fortran_env`. Otherwise, it may look strange.

jvdp1 (commented):
Thank you.
Small questions: is there no tests?
Also, I think that these will not be documented in Ford docs. Could you add some comments for each format (or at least one general comment) with a link to the specs, please?

doc/specs/stdlib_io.md
lewisfish: Good catch! Fixed.

lewisfish:
> Thank you. Small questions: is there no tests? Also, I think that these will not be documented in Ford docs. Could you add some comments for each format (or at least one general comment) with a link to the specs, please?

With regards to tests. There are no explicit tests. However, there are some implicit ones within the save_txt tests since save_txt uses the format specifiers. Is that sufficient?

Added some comments for each specifier. Let me know if I've done the documentation wrong as I've not used FORD before.

src/stdlib_io.fypp
```diff
@@ -21,17 +21,32 @@ module stdlib_io
   ! Private API that is exposed so that we can test it in tests
   public :: parse_mode
 
-  ! Format strings with edit descriptors for each type and kind
+  !> Version: experimental
+  !>
+  !> Format strings with edit descriptors for each type and kind
+  !> The specification of this module is available [here](../page/specs/stdlib_io.html).
```
jvdp1, line 27:
```suggestion
  !> ([Specification](../page/specs/stdlib_io.html))
```

jvdp1 (approved):
Thank you @lewisfish . It looks good to me!

milancurcic:
Thank you @lewisfish and reviewers!

Issue #619, opened 2022-01-13
gauss_legendre with interval


mjben:
### Description

Unless I missed something, there should not be x(1) = interval(1) and x(size(x)) = interval(2) in gauss_legendre subroutines, this would only be relevant for gauss_legendre_lobatto.

### Expected Behaviour

gauss_legendre(x, w, interval=[-1._dp, 1._dp]  should give same result as without interval optional.

### Version of stdlib

0.1.0

### Platform and Architecture

Linux

ivan-pi:
@hsnyder, can you comment on this issue?

The classic Gauss-Legendre quadrature indeed doesn't have nodes at the interval endpoints.

hsnyder:
mjben is correct, this is an error. Feel free to fix it if either of you want to, otherwise I will can do so over the weekend.

jvdp1:
> mjben is correct, this is an error. Feel free to fix it if either of you want to, otherwise I will can do so over the weekend.

Has this bug been fixed?

hsnyder:
I've opened a pull request to fix this issue. Sorry for the delay!

PR #622, opened 2022-01-24
fix gauss quadrature

freevryheid:
when providing intervals [-1, 1] to guass quad function I was getting a different result. I was able to fix by commenting the two lines in each of the quad functions as shown. Sorry, my editor strips trailing spaces, hence the other diffs.

```
real(dp), dimension(2), parameter :: interval = [-1_dp,1_dp]
call gauss_legendre(x,w,interval)
```


freevryheid:
Addressed by https://github.com/fortran-lang/stdlib/issues/619

PR #629, opened 2022-03-02
option to disable testing by setting BUILD_TESTING to OFF

stigh:
Removed enable_testing(), since this seems to be enabled when the subdirectory "tests" is added.
Used the intrinsic CTests name BUILD_TESTING, ref: https://cmake.org/cmake/help/latest/module/CTest.html.

From my understanding, maybe the ideal way of tackling this issue would be to handle BUILD_TESTING as an intrinsic setting in test-drive? Then we might not need the
`option(BUILD_TESTING "Define if tests should be enabled" ON)`
in CMakeLists.txt

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,7 @@ list(GET VERSION_LIST 1 PROJECT_VERSION_MINOR)
 list(GET VERSION_LIST 2 PROJECT_VERSION_PATCH)
 unset(VERSION_LIST)
 
-enable_testing()
+option(BUILD_TESTING "Define if tests should be enabled" ON)
 
 # Follow GNU conventions for installation directories
 include(GNUInstallDirs)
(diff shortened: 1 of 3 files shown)


src/CMakeLists.txt
```diff
@@ -114,7 +114,9 @@ else()
   target_sources(${PROJECT_NAME} PRIVATE f08estop.f90)
 endif()
 
-add_subdirectory(tests)
+if(BUILD_TESTING)
+  add_subdirectory(tests)
```
awvwgk, line 118:
```suggestion
  enable_testing()
  add_subdirectory(tests)
```

CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ list(GET VERSION_LIST 1 PROJECT_VERSION_MINOR)
 list(GET VERSION_LIST 2 PROJECT_VERSION_PATCH)
 unset(VERSION_LIST)
 
-enable_testing()
+option(BUILD_TESTING "Define if tests should be enabled" ON)
```
awvwgk, line 15:
```suggestion
include(CTest)
```

awvwgk (approved):
Thanks for sharing. Looks like a useful addition.

jvdp1 (approved):
LGTM. Thank you.

Issue #224, opened 2020-07-23
Handling and propagating errors in stdlib


aradi:
Before the file system layer (see #201, #220) in stdlib can be implemented, it would be important to reach some agreement on how to signalize errors within stdlib. I am not talking about logging, as this is an add on we can discuss later. I want to concentrate here on how errors can be passed between routines in stdlib.

Fortran's current approach to report status in I/O commands is to have some optional integer argument, returning the status of the operation. If the argument has not been specified, the code stops in case of errors. I'd propose to generalize this principle, but using derived types to achieve higher robustness and flexibility:

* The status should become a derived type (``type(status)``), which methods allowing to set and query it.
* The status-type has a finalizer. If it goes out of scope with an unhandled error in it (without having been handled by the caller), it calls ``error stop``.
* The derived has a special field which contains a ``class(error)`` item. It can carry arbitrary types derived from a base class, so that arbitrary complex error information can be passed around. This way, the error signaling mechanism is extensible, we could even think to add an error-hierarchy as we have in Python.

I have made a [toy project to demonstrate the principle](https://github.com/aradi/fx-error).  Please, have a look at it and let me know what you think.

A few more notes:
* This error passing should be applied to cases, where error reporting via special values in not feasible (e.g. ``change_dir()``).
* We could of course implement in the error-class to information about how it was propagate (backtrace).

This issue is related to several other error discussons, e.g. #219 , #193, #95. My suggestion is, again, to first reach agreement on the low-level of the error reporting, and we can add extended functionality, such as logging, on top later.

certik:
Would the idea still be that if the `status` argument is not provided, that it will error out right away? I think this is the key principle.

I think your approach can work.

The only possible downside that I can see is that you can notice that built-in Fortran functions never return a derived type, but only simple arguments such as integers or arrays. The reason for that is that it does not impose any particular derived type on the user. In here, we would be imposing the type `type(status)`. So that's the main downside compared to returning just an integer.

Regarding speed, the derived type can possibly be a bit slower than an integer. So maybe not a good idea for small functions in hot loops. But for IO functions, the possible overhead is probably small.

Just like integer, `type(status)` is thread safe, does not have any global state.

And the Pros are: extensible, can convey more rich error information such as a string.

And it is still relatively simple.

----

If we chose to go with returning just an integer, then we can document the various different errors as different integers, and then provide a function that prints a nice error message based on the integer number.

[...]

PR #652, opened 2022-04-19
Feature: loadtxt skiprows and max_rows

MuellerSeb:
Hey there,

I needed to skip rows with loadtxt, so I added this feature similar to the numpy routine. In addition, I also added `max_rows`.

Added optional arguments:
- `skiprows`: skipping rows at the beginning of the file (default 0)
- `max_rows`: maximum for read lines (default -1 for all lines)

Argument names are taken from `numpy.loadtxt`. All negative values for `max_rows` are interpreted as "all lines".

### DIFF
--- a/doc/specs/stdlib_io.md
+++ b/doc/specs/stdlib_io.md
@@ -17,13 +17,17 @@ Loads a rank-2 `array` from a text file.
 
 ### Syntax
 
-`call [[stdlib_io(module):loadtxt(interface)]](filename, array)`
+`call [[stdlib_io(module):loadtxt(interface)]](filename, array [, skiprows] [, max_rows])`
 
 ### Arguments
 
 `filename`: Shall be  a character expression containing the file name from which to load the rank-2 `array`.
 
 `array`: Shall be an allocatable rank-2 array of type `real`, `complex` or `integer`.
+
+`skiprows` (optional): Skip the first `skiprows` lines. If skipping more rows than present, a 0-sized array will be returned. The default is 0.
+
+`max_rows` (optional): Read `max_rows` lines of content after `skiprows` lines. The default is -1, to read all the lines.
 
 ### Return value
 
--- a/doc/specs/stdlib_io.md
+++ b/doc/specs/stdlib_io.md
@@ -27,7 +27,7 @@ Loads a rank-2 `array` from a text file.
 
 `skiprows` (optional): Skip the first `skiprows` lines. If skipping more rows than present, a 0-sized array will be returned. The default is 0.
 
-`max_rows` (optional): Read `max_rows` lines of content after `skiprows` lines. The default is -1, to read all the lines.
+`max_rows` (optional): Read `max_rows` lines of content after `skiprows` lines. A negative value results in reading all lines. A value of zero results in no lines to be read. The default value is -1.
 
 ### Return value
 
--- a/src/stdlib_io.fypp
+++ b/src/stdlib_io.fypp
@@ -81,7 +81,7 @@ module stdlib_io
 contains
 
   #:for k1, t1 in KINDS_TYPES
-    subroutine  loadtxt_${t1[0]}$${k1}$(filename, d, skiprows)
+    subroutine  loadtxt_${t1[0]}$${k1}$(filename, d, skiprows, max_rows)
       !! version: experimental
       !!
       !! Loads a 2D array from a text file.
@@ -95,6 +95,8 @@ contains
       ${t1}$, allocatable, intent(out) :: d(:,:)
       !! Skip the first `skiprows` lines, default: 0.
       integer, intent(in), optional :: skiprows
+      !! Read max_rows lines of content after skiprows lines. The default is -1, to read all the lines.
+      integer, intent(in), optional :: max_rows
       !!
       !! Example
       !! -------
@@ -113,10 +115,12 @@ contains
       !!     ...
       !!
       integer :: s
-      integer :: nrow, ncol, i, skiprows_
+      integer :: nrow, ncol, i, skiprows_, max_rows_
 
       skiprows_ = optval(skiprows, 0)
       if ( skiprows_ < 0 ) call error_stop("loadtxt: skiprows needs to be non-negative.")
+
+      max_rows_ = optval(max_rows, -1)
 
       s = open(filename)
 
@@ -128,14 +132,16 @@ contains
 
       ! determine number or rows
       nrow = number_of_rows(s) - skiprows_
-
       if ( nrow < 0 ) call error_stop("loadtxt: skipping more rows than present.")
+      if ( max_rows_ < 0 .or. max_rows_ > nrow) max_rows_ = nrow
+
+      allocate(d(max_rows_, ncol))
+
       do i = 1, skiprows_
         read(s, *)
       end do
 
-      allocate(d(nrow, ncol))
-      do i = 1, nrow
+      do i = 1, max_rows_
         #:if 'real' in t1
           read(s, "(*"//FMT_REAL_${k1}$(1:len(FMT_REAL_${k1}$)-1)//",1x))") d(i, :)
         #:elif 'complex' in t1
--- a/src/stdlib_io.fypp
+++ b/src/stdlib_io.fypp
@@ -81,7 +81,7 @@ module stdlib_io
 contains
 
   #:for k1, t1 in KINDS_TYPES
-    subroutine  loadtxt_${t1[0]}$${k1}$(filename, d)
+    subroutine  loadtxt_${t1[0]}$${k1}$(filename, d, skiprows)
       !! version: experimental
       !!
       !! Loads a 2D array from a text file.
@@ -93,6 +93,8 @@ contains
       character(len=*), intent(in) :: filename
       !! The array 'd' will be automatically allocated with the correct dimensions
       ${t1}$, allocatable, intent(out) :: d(:,:)
+      !! Skip the first `skiprows` lines, default: 0.
+      integer, intent(in), optional :: skiprows
       !!
       !! Example
       !! -------
@@ -111,7 +113,9 @@ contains
       !!     ...
       !!
       integer :: s
-      integer :: nrow, ncol, i
+      integer :: nrow, ncol, i, skiprows_
+
+      skiprows_ = optval(skiprows, 0)
 
       s = open(filename)
 
@@ -122,7 +126,12 @@ contains
       #:endif
 
       ! determine number or rows
-      nrow = number_of_rows(s)
+      nrow = number_of_rows(s) - skiprows_
+
+      if ( nrow < 0 ) call error_stop("loadtxt: skipping more rows than present.")
+      do i = 1, skiprows_
+        read(s, *)
+      end do
 
       allocate(d(nrow, ncol))
       do i = 1, nrow
--- a/src/stdlib_io.fypp
+++ b/src/stdlib_io.fypp
@@ -118,7 +118,7 @@ contains
       integer :: nrow, ncol, i, skiprows_, max_rows_
 
       skiprows_ = optval(skiprows, 0)
-      if ( skiprows_ < 0 ) call error_stop("loadtxt: skiprows needs to be non-negative.")
+      if ( skiprows_ < 0 ) call error_stop("loadtxt: skiprows must be non-negative.")
 
       max_rows_ = optval(max_rows, -1)
 
--- a/src/stdlib_io.fypp
+++ b/src/stdlib_io.fypp
@@ -125,16 +125,17 @@ contains
 
       s = open(filename)
 
-      ! determine number of columns
-      ncol = number_of_columns(s)
-      #:if 'complex' in t1
-      ncol = ncol / 2
-      #:endif
-
       ! determine number or rows
       nrow = number_of_rows(s)
       skiprows_ = min(skiprows_, nrow)
-      if ( max_rows_ < 0 .or. max_rows_ > (nrow - skiprows_)) max_rows_ = nrow - skiprows_
+      if ( max_rows_ < 0 .or. max_rows_ > (nrow - skiprows_) ) max_rows_ = nrow - skiprows_
+
+      ! determine number of columns
+      ncol = 0
+      if ( skiprows_ < nrow ) ncol = number_of_columns(s, skiprows=skiprows_)
+      #:if 'complex' in t1
+      ncol = ncol / 2
+      #:endif
 
       allocate(d(max_rows_, ncol))
 
@@ -196,17 +197,25 @@ contains
   #:endfor
 
 
-  integer function number_of_columns(s)
+  integer function number_of_columns(s, skiprows)
     !! version: experimental
     !!
     !! determine number of columns
     integer,intent(in) :: s
+    integer, intent(in), optional :: skiprows
 
-    integer :: ios
+    integer :: ios, skiprows_
     character :: c
     logical :: lastblank
 
+    skiprows_ = optval(skiprows, 0)
+
     rewind(s)
+
+    do i = 1, skiprows_
+      read(s, *)
+    end do
+
     number_of_columns = 0
     lastblank = .true.
     do
--- a/src/stdlib_io.fypp
+++ b/src/stdlib_io.fypp
@@ -117,9 +117,7 @@ contains
       integer :: s
       integer :: nrow, ncol, i, skiprows_, max_rows_
 
-      skiprows_ = optval(skiprows, 0)
-      if ( skiprows_ < 0 ) call error_stop("loadtxt: skiprows must be non-negative.")
-
+      skiprows_ = max(optval(skiprows, 0), 0)
       max_rows_ = optval(max_rows, -1)
 
       s = open(filename)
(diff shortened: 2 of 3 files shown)


src/stdlib_io.fypp
```diff
@@ -122,7 +127,12 @@ contains
       #:endif
 
       ! determine number or rows
-      nrow = number_of_rows(s)
+      nrow = number_of_rows(s) - skiprows_
+
+      if ( nrow < 0 ) call error_stop("loadtxt: skipping more rows than present.")
```
ivan-pi, line 132: Should this really be a fatal error?

src/stdlib_io.fypp
```diff
@@ -111,7 +113,10 @@ contains
       !!     ...
       !!
       integer :: s
-      integer :: nrow, ncol, i
+      integer :: nrow, ncol, i, skiprows_
+
+      skiprows_ = optval(skiprows, 0)
+      if ( skiprows_ < 0 ) call error_stop("loadtxt: skiprows needs to be non-negative.")
```
ivan-pi, line 119:
```suggestion
      if ( skiprows_ < 0 ) call error_stop("loadtxt: skiprows must be non-negative.")
```

ivan-pi (changes requested):
Looks good to me. 👍🏻 I'd invite other reviewers to voice their opinions if the skipping more rows than present should be an error or not.

src/stdlib_io.fypp
MuellerSeb: We could also cut of `nrow` at 0, which would result in an empty array.

MuellerSeb:
@ivan-pi thanks for the review. I just added `max_rows` since it would make sense to directly add it. Hope this is ok.

src/stdlib_io.fypp
MuellerSeb: This would be the `numpy` behavior. So I would also be in favor of allowing negative values for `skiprows`.

MuellerSeb:
Question from my side: Do we want to mimic the numpy behavior and signature?

src/stdlib_io.fypp
```diff
@@ -111,7 +115,12 @@ contains
       !!     ...
       !!
       integer :: s
-      integer :: nrow, ncol, i
+      integer :: nrow, ncol, i, skiprows_, max_rows_
+
+      skiprows_ = optval(skiprows, 0)
+      if ( skiprows_ < 0 ) call error_stop("loadtxt: skiprows must be non-negative.")
```
MuellerSeb, line 121: Here we could also use:
```fortran
skiprows_ = max(skiprows_, 0)
```
to mimic numpy behavior.

MuellerSeb: Done.

src/stdlib_io.fypp
MuellerSeb: gfortran also crashes when trying to read after EOF. So having a meaning full error message is preferable IMO.

jvdp1: Why not using `iostat` and `iomsg` instead of this check, and returning an appropriate message or an zero-size array when appropriate?

MuellerSeb: Unfortunately, I am not familiar with `iostat` and `iomsg`. This would only affect the following loop, right?
```fortran
do i = 1, skiprows_
  read(s, *)
end do
```

src/stdlib_io.fypp
```diff
@@ -122,10 +129,17 @@ contains
...
-      nrow = number_of_rows(s)
+      nrow = number_of_rows(s) - skiprows_
+      if ( nrow < 0 ) call error_stop("loadtxt: skipping more rows than present.")
+      if ( max_rows_ < 0 .or. max_rows_ > nrow) max_rows_ = nrow
+
+      allocate(d(max_rows_, ncol))
+
+      do i = 1, skiprows_
+        read(s, *)
+      end do
```
jvdp1, line 140:
```suggestion
      do i = 1, skiprows_
        read(s, *, iostat = io)
        if (io .ne. 0) then
         allocate(d(0, 0))
         return
        end if
      end do
```
Or alternatively,
```suggestion
      do i = 1, skiprows_
        read(s, *, iostat = io, iomsg = msg)
        if (io .ne. 0) then
         error stop msg
        end if
      end do
```
@MuellerSeb Here are some possibilities. All are fine for me. The main question IMO is that if  the program should `error stop` or return a zero-size array when the file is "empty".

MuellerSeb: Then I am still in favor of my suggested implementation or alternatively we could cut off `skip_rows_` at the number of present lines.

MuellerSeb: Maybe some more thoughts: Catching this with `iostat`, is a bit strange, since we know beforehand what is going to happen, since the number of rows in the file was calculated. Putting an `error stop msg` there, could be a bit misleading, since we know exactly what the actual problem is: skipping more lines than present.
The other option to silently creating a 0-size array is the same result as cutting of `skiprows_` at the number of remaining rows. Then a 0-size array is also created with the current code.
If we vote against the error message I implemented, I would vote for cutting off `skip_rows_`.

src/stdlib_io.fypp
```diff
@@ -122,10 +129,17 @@ contains
       #:endif
 
       ! determine number or rows
-      nrow = number_of_rows(s)
+      nrow = number_of_rows(s) - skiprows_
+      if ( nrow < 0 ) call error_stop("loadtxt: skipping more rows than present.")
+      if ( max_rows_ < 0 .or. max_rows_ > nrow) max_rows_ = nrow
```
MuellerSeb, line 134:
```suggestion
      nrow = number_of_rows(s)
      skiprows_ = min(skiprows_, nrow)
      if ( max_rows_ < 0 .or. max_rows_ > (nrow - skiprows_)) max_rows_ = nrow - skiprows_
```

14NGiestas:
Just a comment about error handling: there is still an open issue about it #224 (it's worth reading)

Since this function is dealing with the system (reading a file) this may fail (as both noted) and the best approach so far IMHO it is how `stdlib_bitsets` by wclodius2 module deals with errors: have an optional `status` and let the user decides if an error should terminate the application or not. 

Indeed imagine a [user building a GUI application](https://github.com/fortran-lang/stdlib/issues/224#issuecomment-664027001) where a fatal error would close the entire program (and loss of progress). Of course, we could add some boilerplate to `inquire` about the file's existence first (on the end-user side) but it defeats the purpose of using/having a helper function such as `loadtxt` in the first place.

I hope it helps.

MuellerSeb:
@14NGiestas then the suggestion with cutting of `skiprows_` should be fine for this pull request, since it doesn't introduce any new error behavior. If everybody is fine with it, I would commit my suggestion for this.

14NGiestas:
> @14NGiestas then the suggestion with cutting of `skiprows_` should be fine for this pull request, since it doesn't introduce any new error behavior. If everybody is fine with it, I would commit my suggestion for this.

Yep, at first glance your suggestion is ok (just make sure it is mentioned in the docs and examples)
If I recall correctly returning a zero-sized array is a behavior used in `stdlib_math` `linspace` and `logspace`.

src/stdlib_io.fypp
```diff
@@ -93,6 +93,10 @@ contains
       character(len=*), intent(in) :: filename
       !! The array 'd' will be automatically allocated with the correct dimensions
       ${t1}$, allocatable, intent(out) :: d(:,:)
+      !! Skip the first `skiprows` lines, default: 0.
```
MuellerSeb, line 96:
```suggestion
      !! Skip the first `skiprows` lines. If skipping more rows than present, a 0-sized array will be returned. The default is 0.
```

14NGiestas:
I believe the specs should also be updated at [specs/stdlib_io](https://github.com/MuellerSeb/stdlib/blob/feature_loadtxt_skiprows/doc/specs/stdlib_io.md)

src/tests/io/test_loadtxt.f90
```diff
@@ -19,6 +19,7 @@ subroutine collect_loadtxt(testsuite)
             new_unittest("loadtxt_sp_huge", test_loadtxt_sp_huge), &
             new_unittest("loadtxt_sp_tiny", test_loadtxt_sp_tiny), &
             new_unittest("loadtxt_dp", test_loadtxt_dp), &
+            new_unittest("test_loadtxt_dp_max_skip", test_loadtxt_dp_max_skip), &
```
14NGiestas, line 22:
```suggestion
            new_unittest("loadtxt_dp_max_skip", test_loadtxt_dp_max_skip), &
```
:) Uniformity

MuellerSeb:
> I believe the specs should also be updated at [specs/stdlib_io](https://github.com/MuellerSeb/stdlib/blob/feature_loadtxt_skiprows/doc/specs/stdlib_io.md)

Oh. Why are there two files for the same docs?!

14NGiestas:
@MuellerSeb the stdlib project has this little quirk where it mediates, in a manner, future standardized Fortran features.

> The [specification is a document](https://stdlib.fortran-lang.org/page/specs/index.html) that describes the API and the functionality, so that anyone can use it to create an implementation from scratch without looking at stdlib. The stdlib library then provides the reference implementation. 

It will end up being the "same"  but it serves a different purpose. This PR is a kinda undefined behavior AFAIK (since it updates a spec) and maybe we should [discuss](https://github.com/fortran-lang/stdlib/discussions) more. I'm reviewing in a good faith that is useful to have such options available but the [workflow](https://stdlib.fortran-lang.org/page/contributing/Workflow.html) doesn't seem to be ready for this specific situation.

src/stdlib_io.fypp
```diff
@@ -111,7 +115,10 @@ contains
       !!     ...
       !!
       integer :: s
-      integer :: nrow, ncol, i
+      integer :: nrow, ncol, i, skiprows_, max_rows_
+
+      skiprows_ = max(optval(skiprows, 0), 0)
+      max_rows_ = optval(max_rows, -1)
```
jvdp1, line 121: Should it be some checks to avoid `max_rows=0`? Or what is the expected behaviour in such a case?

doc/specs/stdlib_io.md
```diff
@@ -17,14 +17,18 @@ Loads a rank-2 `array` from a text file.
...
 
 ### Arguments
 
 `filename`: Shall be  a character expression containing the file name from which to load the rank-2 `array`.
 
 `array`: Shall be an allocatable rank-2 array of type `real`, `complex` or `integer`.
 
+`skiprows` (optional): Skip the first `skiprows` lines. If skipping more rows than present, a 0-sized array will be returned. The default is 0.
+
+`max_rows` (optional): Read `max_rows` lines of content after `skiprows` lines. The default is -1, to read all the lines.
```
jvdp1, line 30: What about with a negative number (other than 0), or a 0 value provided for max_rows?

jvdp1:
> It will end up being the "same" but it serves a different purpose. This PR is a kinda undefined behavior AFAIK (since it updates a spec) and maybe we should [discuss](https://github.com/fortran-lang/stdlib/discussions) more. I'm reviewing in a good faith that is useful to have such options available but the [workflow](https://stdlib.fortran-lang.org/page/contributing/Workflow.html) doesn't seem to be ready for this specific situation.

I would say that it is labelled as `experimental`. So changing the API is ok at this stage, especially that it is also backward-compatible.

jvdp1 (commented):
Changes look good to me. Returning a zero-array is fine IMO. Thank you.

src/stdlib_io.fypp
MuellerSeb: With `max_rows=0` you will get a 0-sized array. And I think that makes totally sense.

doc/specs/stdlib_io.md
MuellerSeb: Every negative number will be interpreted as: read all lines.

MuellerSeb:
@jvdp1 another option for negative values for `max_rows` could be to to interpret them as wrap-around values. Meaning, `-1` will be the full file, `-2` will leave out the last line and so on. We would just need to find a solution for to big negative numbers. It should again be cut-off at 0 read lines.

doc/specs/stdlib_io.md
jvdp1: This should be mentioned in the specs (instead of only the value `-1`).

jvdp1:
> @jvdp1 another option for negative values for `max_rows` could be to to interpret them as wrap-around values. Meaning, `-1` will be the full file, `-2` will leave out the last line and so on. We would just need to find a solution for to big negative numbers. It should again be cut-off at 0 read lines.

In this case, I would suggest to define a `parameter` `integer`, e.g., `read_all`, which would be available to the user, and then to add a procedure for checking the provided value, e.g.
```fortran
select case(max_rows_)
 case( 0 )
  ....
 case( : -1 )
   .... = real_all
 case default
  ....
end select
```

A `parameter` integer would provide the flexibility for the API (as the user doesn't need to rely on a specific number)

doc/specs/stdlib_io.md
milancurcic, line 30:
```suggestion
`max_rows` (optional): Read `max_rows` lines of content after `skiprows` lines. A negative value results in reading all lines. A value of zero results in no lines to be read. The default value is -1.
```

src/stdlib_io.fypp
```diff
@@ -93,6 +93,10 @@ contains
       character(len=*), intent(in) :: filename
       !! The array 'd' will be automatically allocated with the correct dimensions
       ${t1}$, allocatable, intent(out) :: d(:,:)
+      !! Skip the first `skiprows` lines. If skipping more rows than present, a 0-sized array will be returned. The default is 0.
+      integer, intent(in), optional :: skiprows
+      !! Read max_rows lines of content after skiprows lines. The default is -1, to read all the lines.
```
milancurcic, line 98:
```suggestion
      !! Read `max_rows` lines of content after `skiprows` lines.
      !! A negative value results in reading all lines.
      !! A value of zero results in no lines to be read.
      !! The default value is -1.
```

milancurcic (approved):
Thanks @MuellerSeb and reviewers. I made suggestions to the spec and docstrings. Anything else needed here before we merge?

MuellerSeb:
Hey @milancurcic, thanks for your cleanup. There is still the comment from @jvdp1 about the `-1` value for `max_rows` and the discussion, if other negative values should skip lines at the end.

Beside that I just noticed, that if you use `skiprows` to skip comments in the beginning for example, the following line will still try to get the number of columns from the first line in the file:
```fortran
nrow = number_of_rows(s)
```
I think that is bad.

milancurcic:
> Beside that I just noticed, that if you use skiprows to skip comments in the beginning for example, the following line will still try to get the number of columns from the first line in the file:
> ```
> nrow = number_of_rows(s)
> ```
> I think that is bad.

Do you mean number of _rows_ from the first line in the file? And if yes, I don't understand why is that not what we want. I'm not familiar with the npy format, maybe this is something specific to that.

MuellerSeb:
> > Beside that I just noticed, that if you use skiprows to skip comments in the beginning for example, the following line will still try to get the number of columns from the first line in the file:
> > ```
> > nrow = number_of_rows(s)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think that is bad.
> 
> Do you mean number of _rows_ from the first line in the file? And if yes, I don't understand why is that not what we want. I'm not familiar with the npy format, maybe this is something specific to that.

Sorry I meant columns:
```fortran
      ! determine number of columns
      ncol = number_of_columns(s)
```
Just copied the wrong line.

If you have a file like this:
```
comment
0.1 0,2
0.3 0.4
```
And read it with
```fortran
call loadtxt("log.txt", data, skiprows=1)
```
It will still think, that there is only one column, since the first line has only one (`"comment"`).

milancurcic:
I see, that makes total sense. I agree that won't work. What do you think about this approach:

* Add an optional argument `row` to `number_of_columns` which defaults to 1 (current behavior). `row` indicates the line number at which we'll determine the number of columns.
* In `loadtxt`, if `skiprows` is present, move the call to `number_of_columns` to go after the skip-rows logic, and call `number_of_columns(s, skiprows_+1)`.

MuellerSeb:
> I see, that makes total sense. I agree that won't work. What do you think about this approach:
> 
>     * Add an optional argument `row` to `number_of_columns` which defaults to 1 (current behavior). `row` indicates the line number at which we'll determine the number of columns.
> 
>     * In `loadtxt`, if `skiprows` is present, move the call to `number_of_columns` to go after the skip-rows logic, and call `number_of_columns(s, skiprows_+1)`.

Thought of almost the same. We could also just add a `skiprows` to `number_of_columns` to pass `skiprows_` there directly (would also be more consistent). I would move the call of `number_of_columns` also to have the correct value for `skiprows_`. I would just add these lines from `loadtxt` to `number_of_columns`:
```fortran
      do i = 1, skiprows_
        read(s, *)
      end do
```

milancurcic:
@MuellerSeb nice, I like that even better.

MuellerSeb:
If everyone is fine with the mechanism for `max_rows` set with negative numbers, I think this is ready for merging.

jvdp1 (approved):
LGTM. Nice catch and solution for  `number_of_columns`!

milancurcic:
Thanks all, will merge.

PR #653, opened 2022-04-19
Avoid interfering of internal make variables

### LINKED ISSUE #648: Wrong Compiler Selected When Building With Make
This line
https://github.com/fortran-lang/stdlib/blob/9f1aa24016b652d25cd0493ac9c0111eb3de197b/Makefile.manual#L3
should set `FC` to gfortran, if no other compiler is defined. But on some (don't know on how many) systems `FC` is preset to `f77`. Since I don't have `f77` (and I doubt it would be able to compile stdlib) the build process fails.
Is there anything we can do about this issue? Maybe check the value of `FC` if it is set to such a default value and print a error/warning message?
The current warning is a little bit misleading:
`f77: No such file or directory`

Testet on:
Arch Linux, GNU Make 4.3 built for x86_64-pc-linux-gnu
Fedora 35, GNU Make 4.3 built for x86_64-redhat-linux-gnu

awvwgk:
Closes #648

### DIFF
--- a/Makefile.manual
+++ b/Makefile.manual
@@ -1,5 +1,7 @@
 # Fortran stdlib Makefile
 
+MAKEFLAGS += --no-builtin-rules --no-builtin-variables
+
 FC ?= gfortran
 # -fno-range-check needed for hash functions for gfortran-9
 FFLAGS ?= -Wall -Wextra -Wimplicit-interface -fPIC -g -fcheck=all -fno-range-check


Carltoffel:
I don't know what happens here, but this is what I get:
```
fortran-stdlib$ make -f Makefile.manual                                                                                                                                        
make -f Makefile.manual --directory=src
make[1]: Entering directory '/home/burkert1/programs/fortran-stdlib/src'
Wall -Wextra -Wimplicit-interface -fPIC -g -fcheck=all -fno-range-check -c stdlib_kinds.f90
make[1]: Wall: No such file or directory
make[1]: [Makefile.manual:89: stdlib_kinds.o] Error 127 (ignored)
Wall -Wextra -Wimplicit-interface -fPIC -g -fcheck=all -fno-range-check -c stdlib_optval.f90
make[1]: Wall: No such file or directory
...
```
Looks like the `?=` operator is somehow broken on my system.

awvwgk:
Recursive make was never really my strength, dunno I would have to check the manual to figure out how you can avoid the make defaults and still pass optional values down the recursive invocations.

Carltoffel:
> Somehow the - in front of the -Wall compiler option got stripped.

It's not the compiler which cannot find the file "Wall", but `make` tries to execute `Wall` as command. `FC` is empty, therefore `Wall` is the first "command" `$(FC) $(FFLAGS)...`.
TBH this doesn't look like a `fpm` issue. This surely is a make issue. If I do `@echo ">$(FC)<"` in the `all` target, it prints `><`. If I change `?=` to `=` to set `FC` it works! Somehow `FC` is not set but cannot be overwritten. For `FFLAGS` the `?=` works as expected...

Romendakil:
> It's not the compiler which cannot find the file "Wall", but `make` tries to execute `Wall` as command. `FC` is empty, therefore `Wall` is the first "command" `$(FC) $(FFLAGS)...`. TBH this doesn't look like a `fpm` issue. This surely is a make issue. If I do `@echo ">$(FC)<"` in the `all` target, it prints `><`. If I change `?=` to `=` to set `FC` it works! Somehow `FC` is not set but cannot be overwritten. For `FFLAGS` the `?=` works as expected...

Oops, sorry, you are right. My remark was too hasty. It seems strange. ?= is a GNU extension, it is not POSIX. It might not even work on macOS.

awvwgk:
Well, I guess that won't work than.

PR #655, opened 2022-04-24
fixed 32-bit integer overflow in stdlib_io_npy

### LINKED ISSUE #647: 32bit integer overflow in stdlib_io_npy_save.fypp and load.fypp
When compiling stdlib using NAG Fortran compiler 7.0 for Windows on Windows 10, it terminates compiling with the error "Integer (KIND=3) overflow" (KIND=3 corresponds to int32 in NAG Fortran). 
Errors are detected in `to_bytes_i4` in stdlib_io_npy_save.fypp and in `get_descriptor`  in stdlib_io_npy_load.fypp .

In the procedures, the integer literal `2**32` is used in expressions as below:

https://github.com/fortran-lang/stdlib/blob/8eaa77857bcdfc56188cabdaa2e49c00c206285a/src/stdlib_io_npy_save.fypp#L63-L66

https://github.com/fortran-lang/stdlib/blob/8eaa77857bcdfc56188cabdaa2e49c00c206285a/src/stdlib_io_npy_load.fypp#L135-L143

I guess that `2**24` is probably correct instead of `2**32`.

I have understood that `to_bytes_i4` converts an integer to a byte string by converting a 32 bit decimal integer to a base-256 number and then to four ASCII characters.
In this case, the power of base number (=256) at each position are (`256**0`, `256**1`, `256**2`, `256**3`), corresponding to (`2**0`, `2**8`, `2**16`, `2**24`).

This consideration does not cover negative integers, but if my understanding is correct, the concerning expressions can be modified as follows:

```Fortran
        str = achar(mod(val, 256**1)) // &
            & achar(mod(val, 256**2) / 256**1) // &
            & achar(mod(val, 256**3) / 256**2) // &
            & achar(val / 256**3)
````

 and 

```Fortran
        if (major > 1) then
            header_len = ichar(buf(1)) &
                &      + ichar(buf(2)) * 256**1 &
                &      + ichar(buf(3)) * 256**2 &
                &      + ichar(buf(4)) * 256**3
        else
            header_len = ichar(buf(1))  &
                &      + ichar(buf(2)) * 256**1
        end if
```

If using `2**32` is a trick to handle negative integers, I need another solution to compile stdlib using NAG Fortran compiler.

degawa:
32-bit integer overflows in `stdlib_io_npy_save.fypp` and `stdlib_io_npy_load.fypp` are fixed from `2**32` to `2**24`, to become the power of base-number (256) at the 4th digit a base-256 number.
To improve readability and understandability, the powers of base-number are refactored from `2**8`, `2**16`, `2**24` to `256**1`, `256**2`, `256**3`, respectively.

- fixed and refactored `stdlib_io_npy_save.fypp`
- fixed and refactored `stdlib_io_npy_load.fypp`
- built with gfortran 10.3, ifort 2021.1, and nagfor 7.1 on Windows 10 with cmake 3.20.3
- executed `cmake --build build --target test` to run the test npy and confirmed it passed

closes #647 

Note that `config/DefaultFlags.cmake` is locally changed to specify compiler flags for nagfor as follows:
```cmake
elseif(CMAKE_Fortran_COMPILER_ID MATCHES "NAG")
  set(
    CMAKE_Fortran_FLAGS_INIT
    "-fpp"
    "-f2018"
    "-ieee=full"
  )
  set(
    CMAKE_Fortran_FLAGS_RELEASE_INIT
  )
  set(
    CMAKE_Fortran_FLAGS_DEBUG_INIT
    "-g"
    "-nan"
  )
```

### DIFF
--- a/src/stdlib_io_npy_load.fypp
+++ b/src/stdlib_io_npy_load.fypp
@@ -134,12 +134,12 @@ contains
 
         if (major > 1) then
             header_len = ichar(buf(1)) &
-                &      + ichar(buf(2)) * 2**8 &
-                &      + ichar(buf(3)) * 2**16 &
-                &      + ichar(buf(4)) * 2**24
+                &      + ichar(buf(2)) * 256**1 &
+                &      + ichar(buf(3)) * 256**2 &
+                &      + ichar(buf(4)) * 256**3
         else
-            header_len = ichar(buf(1)) &
-                &      + ichar(buf(2)) * 2**8
+            header_len = ichar(buf(1))  &
+                &      + ichar(buf(2)) * 256**1
         end if
         allocate(character(header_len) :: dict, stat=stat)
         if (stat /= 0) return
--- a/src/stdlib_io_npy_load.fypp
+++ b/src/stdlib_io_npy_load.fypp
@@ -138,7 +138,7 @@ contains
                 &      + ichar(buf(3)) * 256**2 &
                 &      + ichar(buf(4)) * 256**3
         else
-            header_len = ichar(buf(1))  &
+            header_len = ichar(buf(1)) &
                 &      + ichar(buf(2)) * 256**1
         end if
         allocate(character(header_len) :: dict, stat=stat)
--- a/src/stdlib_io_npy_load.fypp
+++ b/src/stdlib_io_npy_load.fypp
@@ -136,7 +136,7 @@ contains
             header_len = ichar(buf(1)) &
                 &      + ichar(buf(2)) * 2**8 &
                 &      + ichar(buf(3)) * 2**16 &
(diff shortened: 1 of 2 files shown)


sakamoti:
I think this patch looks fine. 
stdlib_io_npy_save and stdlib_io_npy_load have passed compilation phaze with nagfor7.1 (Linux version).

sakamoti (approved):
I think this patch looks fine.
stdlib_io_npy_save and stdlib_io_npy_load have passed compilation phaze with nagfor7.1 (Linux version).

awvwgk (approved):

milancurcic (approved):
Looks good, thank you!

PR #657, opened 2022-05-07
Remove support for manual make builds

### LINKED ISSUE #526: Should we continue maintaining the manual Makefiles?
With stdlib now being also an fpm package, as well as CMake support, should we continue maintaining the manual Makefiles?

I argued for manual Makefiles early on (before fpm) due to the possible absence of CMake or a sufficient version of CMake on many systems. However, the cost of maintaining manual Makefiles is quite high relative to the fpm or CMake build. For example, adding a new module and tests requires editing multiple Makefiles in the source tree. And it gets messier when trying to merge some older PRs, as the manual Makefiles would often be in conflict with the target branch.

We could run a poll on Discourse and see what users there think. I think this may be a good time to remove manual Makefile support. What do you think?

milancurcic:
This PR removes the manual Makefiles from directories, CI, and instructions docs (README and WORKFLOW). Its purpose is to remove the burden from PR contributors and stdlib maintainers when adding features to stdlib. The obvious downside is that people wouldn't be able to build stdlib with Make anymore.

Even if there is wide approval, I'd like this PR to stay open a while (at least a few weeks) to provide an opportunity for people to voice objections.

I think I removed all Make instances (files and instructions), but it's possible that I missed something.

Because this is a such a high-impact and project-wide change that affects users, I request review from many people. While I am in favor of this change, I am not sure that it's the best thing to do.

Discussion and poll on Discourse: https://fortran-lang.discourse.group/t/how-do-you-intend-to-build-the-fortran-stdlib/1862

Closes #526.


Carltoffel:
If we agree to remove the make builds, this PR would be a good opportunity to move the fpm section in README.md to the top. There was a discussion on that, but I forgot where.

milancurcic:
@Carltoffel I agree but suggest to do it in a separate PR to make this easier to review.

awvwgk (approved):
Happy to see them gone from the repo.

ghbrown (commented):
Thanks for making this PR, I agree it will reduce clutter and streamline the process for contributors.

I have two suggested changes, but unfortunately I can't make them myself using the built-in diff/comment/suggestions system because they are outside the lines that your proposed changes target.

- on line 66 of `Workflow.md` is still advertises support for make
- The "Using stdlib in your project" section of `README.md` should be slightly changed. For example, line 191 implies that there are non-CMake builds of the stdlib (which there won't be after this PR). I might change it something like "If your project uses CMake, the local installation of stdlib can be found using  `<same CMake snippet>`".

Besides these, everything looks good to me, and I'm happy to approve!

14NGiestas (approved):
It looks fine.
`find . -name "Make*"` and `find . -name "make*"` returns nothing so they are gone for good.

MuellerSeb:
I am in favor of dropping the makefiles to unify the build process. Seeing it with netcdf-fortran for example where the different install/build methods are out of sync and can hamper the usage of the lib as a dependency. Also, relying on CMake, it enables the use of [CPM](https://github.com/cpm-cmake/CPM.cmake) which I really love for dependency management.

MuellerSeb (approved):
I am in favor of dropping the makefiles to unify the build process. Seeing it with netcdf-fortran for example where the different install/build methods are out of sync and can hamper the usage of the lib as a dependency. Also, relying on CMake, it enables the use of [CPM](https://github.com/cpm-cmake/CPM.cmake) which I really love for dependency management.

jvdp1 (approved):
I support this PR and change. Thanks.

everythingfunctional (approved):
I don't have time to look too closely at the changes, but as they're all deletions I'd say this is probably good unless you missed some. I also think the idea of reducing the number of build systems being maintained is a good one.

milancurcic:
If there are no objects, I will apply the outstanding fixes and merge on Monday, May 23.

certik (approved):
I think you can merge this now! Let's get it done. Just fix up things if anything breaks after the merge.

milancurcic:
@certik, great, thanks for the encouragement. Will merge now.

ivan-pi:
Just wanted to mention this discussion from https://github.com/Reference-LAPACK/lapack/issues/488. They have similar problems as we had, keeping two build systems functioning and up to date. For the time being they are keeping both CMake and Make, but there seems to be a slight preference for using CMake.

Issue #614, opened 2022-01-03
Cmake, optimization and README.md


gareth-nx:
### Description

README.md instructs users to control compiler flags in CMake builds like this (e.g. to get "-O3"):
```
export FFLAGS="-O3"
cmake -B build -G Ninja -DCMAKE_MAXIMUM_RANK:String=7 -DCMAKE_INSTALL_PREFIX=$HOME/.local
```

However this does not have the intended effect; by default there are other options that override this optimization.

To see that this is the case, we can insert the line `set( CMAKE_VERBOSE_MAKEFILE on )` near the top of `CMakeLists.txt`, and build with the above commands. Then the compiler commands are printed out. Copying one (for illustration) we have:
``` 
/usr/bin/gfortran  -I/home/gareth/Code_Experiments/fortran/stdlib/my_branch/stdlib/build/src/mod_files  -O3 -O2 -g -DNDEBUG -Jmod_files/   -fimplicit-none -ffree-line-length-132 -Wall -Wextra -Wimplicit-procedure -pedantic-errors -std=f2018 -c /home/gareth/Code_Experiments/fortran/stdlib/my_branch/stdlib/build/src/stdlib_bitsets_64.f90 -o CMakeFiles/fortran_stdlib.dir/stdlib_bitsets_64.f90.o

```
Notice how right after the `"-O3"`, we have `"-O2 -g -DNDEBUG"`. This is overriding the `"-O3"` optimization which was intended. Our documentation doesn't make this clear, and this has [confused some efforts to try optimising code](https://github.com/fortran-lang/stdlib/issues/606#issuecomment-1003709678).

### Expected Behaviour

I expected the code to have been compiled with `-O3`, not `-O2`. 

The reason I noticed this is that my code was slower when linking to Cmake builds, vs Makefile builds (and hence I was getting [weird performance results](https://github.com/fortran-lang/stdlib/issues/606#issuecomment-1003709678)).

### Version of stdlib

15f3270e30f164bf2f63828a3f715e31e0077632

### Platform and Architecture

All

### Additional Information

It seems that by default, `stdlib` is having CMake add the flags for a `CMAKE_BUILD_TYPE=RelWithDebInfo` build. 

One way to work-around this issue is to have the user make up some ad-hoc BUILD_TYPE (so that the defaults are not invoked). However presumably there are some options that we really do want by default -- so the questions are:
*  What should we suggest in README.md?
*  Should any of the CMake defaults be changed? 

[...]

PR #659, opened 2022-05-21
Readme update

gareth-nx:
This pull request explains some CMake options that affect stdlib builds. 
* It is designed to address [these issues](https://github.com/fortran-lang/stdlib/issues/614) in a simple way.

The first issue is that user-specified compiler optimizations in `FFLAGS` are overridden by the default the stdlib CMake setup. This means that, in the original README, the use of FFLAGS was not having any effect. 
* To fix this, the updated README discusses `-DCMAKE_BUILD_TYPE`. This can be setup to avoid default optimizations (while retaining flags that are essential to compile stdlib). This allows optimization flags in `FFLAGS` to be used.

The second issue is that by default, the user does not see what commands CMake is using to compile the code. This makes it difficult to notice problems like those above. 
* The solution is to set `-DCMAKE_VERBOSE_MAKEFILE=On`, which prints the compiler options.

I am a CMake novice, so it would be good to have this checked by an expert like @awvwgk .

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,13 +134,13 @@ Important options are
 - `-DBUILD_SHARED_LIBS` set to `on` in case you want link your application dynamically against the standard library (default: `off`).
 - `-DBUILD_TESTING` set to `off` in case you want to disable the stdlib tests (default: `on`).
 - `-DCMAKE_VERBOSE_MAKEFILE` is by default set to `Off`, but if set to `On` will show commands used to compile the code.
-- `-DCMAKE_BUILD_TYPE` is set to `NoConfig` by default. Set to `Release` to compile with standard optimizations, or `RelWithDebInfo` for standard code development options. Beware these options may override compiler flags that are specified via `FFLAGS`. To avoid this, use an unknown value such as `-DCMAKE_BUILD_TYPE=UnknownBuildType`.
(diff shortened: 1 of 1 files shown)


awvwgk:
What happens if you set `-DCMAKE_BUILD_TYPE=Release` instead of using `NoConfig` and setting `FFLAGS` manually?

jvdp1 (approved):
It sounds good to me. IMO it could be merged, pending @awvwgk 's comments.

awvwgk:
The Release flags are currently empty

https://github.com/fortran-lang/stdlib/blob/3028a36fe59952ef576176f41ca20465e826be91/config/DefaultFlags.cmake#L7-L9

You could add the preferred options there and we can rely on `-DCMAKE_BUILD_TYPE=Release` rather than reverting to the `NoConfig` which is the usual CMake default.

jvdp1:
> The Release flags are currently empty
> 
> https://github.com/fortran-lang/stdlib/blob/3028a36fe59952ef576176f41ca20465e826be91/config/DefaultFlags.cmake#L7-L9
> 
> You could add the preferred options there and we can rely on `-DCMAKE_BUILD_TYPE=Release` rather than reverting to the `NoConfig` which is the usual CMake default.

I would propose to use the same compiler options as for the release option of `fpm`:
```
 -O3 -funroll-loops -Wimplicit-interface -fPIC -fmax-errors=1 -fcoarray=single
```

jvdp1:
@awvwgk Are you good with these changes? Can this PR be merged?

jvdp1:
Thank you all. I will merge it.

PR #662, opened 2022-06-19
Extraction of the demo programs from the specs

jvdp1:
This PR proposes to create small demo programs of the different features and to include them in the specs through the statement `include` in `ford`.
The main advantage is that the demo programs can be compiled and tested.

Using a small Fortran program, I extracted all demo programs from the specs, and generated small files for each of them. By doing that, I noticed that most of them still contained typos, missing statements, undeclared variables, ... I fixed most of them but still a few demo programs must be fixed.

So, the main advantage of this approach is that all demo programs that will be shown in the specs can be compiled (an issue that has been often mentioned).

__Questions__:
* What do you think about this idea?
* Is the structure OK (`test/examples`)?
* Currently the demo programs are added to the tests. Is this ok (I don't think so)? what should it be done?

__TO DO__:

* [x] Fix remaining programs (currently commented in the CMakefile)
* [x] Check the generated website by `ford`
* [x] Add support of the demo programs to `fpm`


urbanjost:
The intrinsics descriptions follow a simple pattern  (example program starts with "program demo_$TOPIC" and ends with "end demo_$TOPIC")  for a similar purpose  (as well as do almost all my own repositories).  I am interested in what develops and how it is implemented, as I have found that type of approach useful.   fpm(1) is still lacking any supported documentation approach other than that ford(1) seems to be used frequently.  Wondering if something might emerge that could be used by all the projects, like an option for ford(1) that could automatically generate man-pages as well as extract demo pages from source code and place them in the fpm example page if certain conventions are used. That is, place the documentation for a routine in the source code, and when ford(1) is run (or some simpler fpm-supported preprocessor) you would get the ford documentation, man-pages as real man-pages and as separate (HTML, LaTex, markdown, ...?) documents and any demo program in the documentation would be extracted and show in the ford(1) listing and in the example/ directory of fpm? 

So if you follow some ford-like convention that has a man-page style using simple plain text in your source code

 +  your code is well documented
 +  demo programs are automatically generated and easily testable via fpm
 +  the documentation for a user guide is automatically generated 
 +  and at least for Unix/GNU-Linux/WLS users you would have a CLI version of the procedures

CMakeLists.txt
```diff
@@ -68,6 +68,8 @@ endif()
 
 add_subdirectory(src)
 
+add_subdirectory(test)
```
awvwgk, line 71: Should we put this under just `example`. Fpm can pick up demo programs and run them as well, means we have to deploy this directory with fpm (maybe for a follow-up PR).

test/example/array/demo_falseloc.f90
```diff
@@ -1,8 +1,8 @@
...
-real, allocatable :: array(:)
-allocate(array(-200:200))
-call random_number(array)
-array(falseloc(array < 0.5), lbound(array)) = 0.0
+    use stdlib_array, only: falseloc
+    implicit none
+    real, allocatable :: array(:)
+    allocate (array(-200:200))
+    call random_number(array)
+    array(falseloc(array < 0.5), lbound(array)) = 0.0
```
14NGiestas, line 7:
```suggestion
    array(falseloc(array < 0.5, lbound(array,1))) = 0.0
```

This is failing in the CI, [here's the corresponding test](https://github.com/fortran-lang/stdlib/blob/master/src/tests/array/test_logicalloc.f90#L210)

test/example/stats_distribution_normal/demo_normal_rvs.f90
```diff
@@ -0,0 +1,45 @@
...
+! -0.633261681
+
+    print *, norm(0.0, 1.0, 10)       !an array of 10 standard norml random variates
+
+! -3.38123664E-02  -0.190365672  0.220678389  -0.424612164  -0.249541596
+!  0.865260184  1.11086845  -0.328349441  1.10873628  1.27049923
+
+    a(:, :, :) = 1.0
+    b(:, :, :) = 1.0
+    print *, norm(a, b)         ! a rank 3 random variates array
```
14NGiestas, line 29:
```
$ gh pr checkout 662
$ ./ci/fpm-deployment.sh && cd stdlib-fpm
$ fpm test
demo_normal_rvs.f90                    failed.
[   0%]Compiling...
example/demo_normal_rvs.f90:29:12:

   29 |     print *, norm(a, b)         ! a rank 3 random variates array
      |            1
Error: There is no specific function for the generic ‘rvs_normal’ at (1)
<ERROR> Compilation failed for object " example_demo_normal_rvs.f90.o "
<ERROR>stopping due to failed compilation
STOP 1
```

I got this error while trying the PR with fpm, and it seems the `rvs_normal` is missing some `elemental` statements.
The obvious patch below fixes it, should I open another PR just for this one?

```diff
diff --git a/src/stdlib_stats_distribution_normal.fypp b/src/stdlib_stats_distribution_normal.fypp
index e21b7af..5a02805 100644
--- a/src/stdlib_stats_distribution_normal.fypp
+++ b/src/stdlib_stats_distribution_normal.fypp
@@ -160,6 +160,7 @@ contains
 
 
     #:for k1, t1 in REAL_KINDS_TYPES
+    impure elemental &
     function rvs_norm_${t1[0]}$${k1}$(loc, scale) result(res)
     !
     ! Normal random variate (loc, scale)
@@ -178,6 +179,7 @@ contains
 
 
     #:for k1, t1 in CMPLX_KINDS_TYPES
+    impure elemental &
     function rvs_norm_${t1[0]}$${k1}$(loc, scale) result(res)
     !
     ! Normally distributed complex. The real part and imaginary part are       &
@@ -197,6 +199,7 @@ contains
 
 
     #:for k1, t1 in REAL_KINDS_TYPES
+    impure elemental &
     function rvs_norm_array_${t1[0]}$${k1}$(loc, scale, array_size) result(res)
         ${t1}$, intent(in) :: loc, scale
         integer, intent(in) :: array_size
@@ -250,6 +253,7 @@ contains
 
 
     #:for k1, t1 in CMPLX_KINDS_TYPES
+    impure elemental &
     function rvs_norm_array_${t1[0]}$${k1}$(loc, scale, array_size) result(res)
         ${t1}$, intent(in) :: loc, scale
         integer, intent(in) :: array_size
```

jvdp1: > The obvious patch below fixes it, should I open another PR just for this one?

Good catch. Yes, sure!

test/example/strings/demo_padl.f90
```diff
@@ -0,0 +1,15 @@
+program demo_padl
+    use stdlib_string_type, only: string_type, assignment(=)
```
14NGiestas, line 2: This demo is falling in the CI, because is missing a `write(formatted)`.

```suggestion
    use stdlib_string_type, only: string_type, assignment(=), write(formatted)
```

test/example/strings/demo_replace_all.f90
```diff
@@ -0,0 +1,16 @@
+program demo_replace_all
+    use stdlib_string_type, only: string_type, assignment(=)
```
14NGiestas, line 2: Same as [before](https://github.com/fortran-lang/stdlib/pull/662/files#r907616962)
```suggestion
    use stdlib_string_type, only: string_type, assignment(=), write(formatted)
```

test/example/strings/demo_padr.f90
```diff
@@ -0,0 +1,15 @@
+program demo_padr
+    use stdlib_string_type, only: string_type, assignment(=)
```
14NGiestas, line 2:
```suggestion
    use stdlib_string_type, only: string_type, assignment(=), write(formatted)
```

test/example/strings/demo_padr.f90
```diff
@@ -0,0 +1,15 @@
...
+    use stdlib_strings, only: padr
+    implicit none
+    string_type :: string
+
+    string = "right pad this string"
+! string <-- "right pad this string"
+
+    print *, padr(string, 25, "$") ! "right pad this string$$$$"
```
14NGiestas, line 10:
```suggestion
    print '(a)', padr(string, 25, "$") ! "right pad this string$$$$"
```
Or you know, just `use ..., only: ..., write(unformatted)`

test/example/stats_distribution_normal/demo_normal_rvs.f90
14NGiestas: I'm investigating further. My patch didn't work it just skipped this specific error and created another one :)
the generated ones with *array* in the interface shouldn't have elemental.

test/example/logger/demo_log_text_error.f90
```diff
@@ -0,0 +1,24 @@
...
+    character(128)          :: line
+    character(*), parameter :: message = 'Bad text found.'
+
+    open (newunit=lun, file=filename, status='old', &
+          form='formatted')
+    line_no = 0
+    do
+        read (lun, fmt='(a)', end=900) line
+        line_no = line_no + 1
+        call check_line(line, status, col_no)
```
14NGiestas, line 15: This function `check_line` doesn't exist in the program scope neither the `status` variable

test/example/logger/demo_log_text_error.f90
```diff
@@ -0,0 +1,24 @@
+program demo_log_text_error
+    use stdlib_logger
+
+    character(*), parameter :: filename = 'dummy.txt'
+    integer                 :: col_no, line_no, lun
```
14NGiestas, line 5:
```suggestion
    integer                 :: col_no, line_no, lun, status
```
`status` is not in scope

milancurcic:
Thanks all for this great effort. I agree with it. Two suggestions for now:

* Use minimum reasonable number of spaces for indentation. I think this is 2. Rationale: These will go into API docs. While the FORD-generated pages are mostly responsive (mobile-friendly), the source code listings are monospace and wrap around with no indication whether the line is continued or not, making the examples more difficult to read. For example, on my phone (Moto G Power) I get 46 columns in a listing before a line wraps around. Switching from indentation of 4 spaces to 2 would decrease, though not completely eliminate, the wrapped lines.
* Prefix the programs and sources with `example_` rather than `demo_`. This is very minor IMO, but goes toward consistency.

jvdp1:
> Thanks all for this great effort. I agree with it. Two suggestions for now:
> 
> * Use minimum reasonable number of spaces for indentation. I think this is 2. Rationale: These will go into API docs. While the FORD-generated pages are mostly responsive (mobile-friendly), the source code listings are monospace and wrap around with no indication whether the line is continued or not, making the examples more difficult to read. For example, on my phone (Moto G Power) I get 46 columns in a listing before a line wraps around. Switching from indentation of 4 spaces to 2 would decrease, though not completely eliminate, the wrapped lines.
> * Prefix the programs and sources with `example_` rather than `demo_`. This is very minor IMO, but goes toward consistency.

Thank you @milancurcic for your comments. The commits https://github.com/fortran-lang/stdlib/pull/662/commits/6d7f88961ccbafdf55c4e72ab95d44a73a9b3a41 and https://github.com/fortran-lang/stdlib/pull/662/commits/bf00288098dea430e6c7bb86fc1e2b7ff50a3ac5 address both comments.

14NGiestas:
> * Should all the examples be added to the CMake tests? Or is it possible to run them without including them in the tests, as with `fpm`?

It seems it [is possible](https://stackoverflow.com/questions/28707559/two-sets-of-tests-in-cmake) to create a custom target (let's say `add_custom_target(examples ...)`, so you can run the tests and the examples in separate, with a `make examples`.

I don't know if it's applicable to stdlib.


> * Is the current structure (two different test directories (./test/ and ./src/tests)) fine?

I guess moving it into `./src/examples/` would be less surprising than two `test` and `tests` folders around the codebase.

jvdp1:
> > * Should all the examples be added to the CMake tests? Or is it possible to run them without including them in the tests, as with `fpm`?
> 
> It seems it [is possible](https://stackoverflow.com/questions/28707559/two-sets-of-tests-in-cmake) to create a custom target (let's say `add_custom_target(examples ...)`, so you can run the tests and the examples in separate, with a `make examples`.

I am not aware of that option. I will look at it. Thanks.
 
> I don't know if it's applicable to stdlib.
> 
> > * Is the current structure (two different test directories (./test/ and ./src/tests)) fine?
> 
> I guess moving it into `./src/examples/` would be less surprising than two `test` and `tests` folders around the codebase.

Indeed, `.src/examples/` makes more sense.

awvwgk:
> > It seems it [is possible](https://stackoverflow.com/questions/28707559/two-sets-of-tests-in-cmake) to create a custom target (let's say `add_custom_target(examples ...)`, so you can run the tests and the examples in separate, with a `make examples`.
> 
> I am not aware of that option. I will look at it. Thanks.

Let's not create a custom target for this purpose. Instead prefix the test names and use `ctest -R '^example/'` to select tests based on a regular expression.

jvdp1:
> > > It seems it [is possible](https://stackoverflow.com/questions/28707559/two-sets-of-tests-in-cmake) to create a custom target (let's say `add_custom_target(examples ...)`, so you can run the tests and the examples in separate, with a `make examples`.
> > 
> > 
> > I am not aware of that option. I will look at it. Thanks.
> 
> Let's not create a custom target for this purpose. Instead prefix the test names and use `ctest -R '^example/'` to select tests based on a regular expression.

@awvwgk This sounds a good idea. But I am not sure how to organize that. Any hints?

jvdp1:
@fortran-lang/stdlib  This PR is ready for reviewing and merging.

14NGiestas (approved):
LGTM I've tested with both `cmake` and `fpm` (in which seems there is no way to run all of them at once AFAIK you need to provide a name).
Maybe it is worth mentioning in the README how to run only the examples.

jvdp1:
> LGTM I've tested with both `cmake` and `fpm` (in which seems there is no way to run all of them at once AFAIK you need to provide a name). Maybe it is worth mentioning in the README how to run only the examples.

Thank you. I just added instructions to run the examples with `fpm`.

jvdp1:
@14NGiestas @awvwgk I moved the directory `src/examples` to `example`, as discussed during the last Fortran Monthly call.
I believe this is ready to be merged.

awvwgk (approved):

14NGiestas (approved):
LGTM. One step closer to fpm-ization of fortran & stdlib.

jvdp1:
Thank you for the review. I will merge this PR.

PR #665, opened 2022-06-27
Fixes issue with `rvs_normal` where it lacks the elemental property

14NGiestas:
As noted in [PR#662](https://github.com/fortran-lang/stdlib/pull/662#discussion_r907615242) there's an error in the demo program given for `rvs_normal` where it lacks the shown elemental property.

```
example/demo_normal_rvs.f90:29:12:

   29 |     print *, norm(a, b)         ! a rank 3 random variates array
      |            1
Error: There is no specific function for the generic ‘rvs_normal’ at (1)
<ERROR> Compilation failed for object " example_demo_normal_rvs.f90.o "
<ERROR>stopping due to failed compilation
```

### DIFF
--- a/src/stdlib_stats_distribution_normal.fypp
+++ b/src/stdlib_stats_distribution_normal.fypp
@@ -160,6 +160,7 @@ contains
 
 
     #:for k1, t1 in REAL_KINDS_TYPES
+    impure elemental &
     function rvs_norm_${t1[0]}$${k1}$(loc, scale) result(res)
     !
     ! Normal random variate (loc, scale)
@@ -178,6 +179,7 @@ contains
 
 
     #:for k1, t1 in CMPLX_KINDS_TYPES
+    impure elemental &
     function rvs_norm_${t1[0]}$${k1}$(loc, scale) result(res)
     !
     ! Normally distributed complex. The real part and imaginary part are       &


14NGiestas:
I think it's just a matter of batch calling the function over an array without having to deal with `do` loops, if you take a look at [L185](https://github.com/fortran-lang/stdlib/blob/master/src/stdlib_stats_distribution_uniform.fypp#L185) you will see the pattern, 2 arguments -> `impure elemental`.

But I can see there is a problem with reproducible "randomness": if the order of the computations is changed (by doing it in parallel it's known that it happens), and since it depends on some stateful routines (there are some variables with the `save` attribute) I guess the result will be different, but I don't think the distribution will be changed.

jvdp1:
> But I can see there is a problem with reproducible "randomness": if the order of the computations is changed (by doing it in parallel it's known that it happens), and since it depends on some stateful routines (there are some variables with the `save` attribute) I guess the result will be different, but I don't think the distribution will be changed.

Indeed, I refered to the problem of reproducible randomness. I agree with you that it won't affect the generated distribution.
Furthermore, the equivalent function in `stdlib_stats_distribution_uniform` is already `impure elemental`. Therefore, it would be logical that the equivalent fiunctions are `impure elemental` in both `stdlib_stats_normal` and in `stdlib_stats_distribution_exponential`.

The specs should be adapted to reflect these changes too.

jvdp1 (approved):
LGTM. Thank you.

jvdp1:
Thank you. I will merge it to move further wiuth other PRs.

Issue #673, opened 2022-08-12
Determine the FORD version number for FORD CI document deployment


zoziha:
### Motivation

As far as I know, in both [#624](https://github.com/fortran-lang/stdlib/pull/624#issuecomment-1030510430) and [#278](https://github.com/fortran-lang/stdlib/pull/278), the failure of FORD document deployment occurs, which is often caused by the features and version updates of FORD.
Determining the appropriate FORD version for stdlib in stages will help new contributors are less likely to face FORD CI failures due to FORD version updates.

### Prior Art

https://github.com/fortran-lang/stdlib/blob/c6b54e1eee1be10ea144b430a673e20b843f50de/.github/workflows/doc-deployment.yml#L37

We can use:

```sh
pip install -v ford==6.1.10
```

### Additional Information

- https://github.com/Fortran-FOSS-Programmers/ford/issues/446

awvwgk:
There is some historical baggage on the documentation workflow, see https://github.com/fortran-lang/stdlib/issues/289. It might be time to revisit the workflow and use a different setup strategy.

zoziha:
Currently stdlib publishes FORD documents based on macOS and brew, while fpm publishes FORD documents based on Ubuntu and pip, it seems to use pip to install ford and publish docs is easier, maybe we can update stdlib's docs CI with [fpm's CI](https://github.com/fortran-lang/fpm/blob/1a8a96be600cc9f0bf99ec7a1c2aa710e889c0bb/.github/workflows/docs.yml#L1).

jvdp1:
I think it makes sense to use the same strategy in all Fortran-lang projects regarding `ford`.
@zoziha  would you like to open a PR with a new strategy?

PR #675, opened 2022-09-06
refactor: change in structure of stdlib in order to compiler it using fypp script

arteevraina:
This Pull Request makes few structural changes in `stdlib` to make get it compiled using the compiler wrapper script [here](https://github.com/fortran-lang/fpm/pull/729).

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,21 @@ To use `stdlib` within your `fpm` project, add the following lines to your `fpm.
 stdlib = { git="https://github.com/fortran-lang/stdlib", branch="stdlib-fpm" }
 ```
 
+### Build with [fypp-script](https://github.com/fortran-lang/fpm/pull/729)
+
+Fortran Package Manager has a compiler wrapper `fypp-script` that can be used to compile stdlib without 
+switching to `stdlib-fpm` branch.
+
+```sh
+git clone https://github.com/fortran-lang/fpm.git
+```
+
+```
+fpm run -- build --compiler "$PWD/fypp-gfortran.py" -C stdlib
+```
+
+**Note** : At this stage, building using `stdlib-fpm` branch should be the preferred way of using `stdlib` with `fpm`.
+
 ## Using stdlib in your project
 
 The stdlib project exports CMake package files and pkg-config files to make stdlib usable for other projects.


jvdp1:
To solve the issues, the following should be done:
* In `CMakeLists.txt`: add  `"-I${PROJECT_SOURCE_DIR}/include"` in the definition of the `fyppFlags`
* In `test/CMakeLists.txt`: remove the lines:
```
list(
  APPEND fyppFlags
  "-I${PROJECT_SOURCE_DIR}/src"
)
```

jvdp1:
Did you also solve the issues with:
* `test_mean_f03.f90`
*  `nmash.c`
?

arteevraina:
> Did you also solve the issues with:
> 
>     * `test_mean_f03.fypp`
> 
>     * `nmash.c`
>       ?

For `nmhash.c`, I have fixed the issues. Will push it.
For `test_mean_f03.fypp`, Still having the same issue. I have not been able to resolve it. Still working on it.

CMakeLists.txt
```diff
@@ -50,7 +50,7 @@ endif()
 
 # Custom preprocessor flags
 if(DEFINED CMAKE_MAXIMUM_RANK)
-  set(fyppFlags "-DMAXRANK=${CMAKE_MAXIMUM_RANK}")
+  set(fyppFlags "-DMAXRANK=${CMAKE_MAXIMUM_RANK} -I${PROJECT_SOURCE_DIR}/include")
 elseif(f03rank)
   set(fyppFlags)
```
awvwgk, line 55: This is the wrong place to add this option. Please append the option later or initialize the `fyppFlags` with the include directory and append the `MAXRANK` macro here.

arteevraina:
@awvwgk @jvdp1 

CMake build is failing since, it is not able to find `test_always_fail.f90` & `test_always_skip.f90` as I have deleted those files.

awvwgk:
> CMake build is failing since, it is not able to find `test_always_fail.f90` & `test_always_skip.f90` as I have deleted those files.

Than you also have to remove the declaration of the tests from the CMake build files.

test/stats/test_mean_f03.fypp
```diff
@@ -65,25 +65,36 @@ contains
...
         type(error_type), allocatable, intent(out) :: error
 
+        #:if MAXRANK > 7
         call check(error, mean(d8_${k1}$), sum(real(d8_${k1}$, dp))/real(size(d8_${k1}$), dp)&
                     , 'mean(d8_${k1}$): uncorrect answer'&
                     , thr = dptol)
         if (allocated(error)) return
+
+        #:else
+        call skip_test(error, "Rank 7 is not supported")
```
jvdp1, line 75: Here and other places: replace "Rank 7" by "Rank >7" or by "Rank 8 or higher"
```suggestion
        call skip_test(error, "Rank >7 is not supported")
```

arteevraina:
I remember we moved the include file from `src` to `include` directory. But seems like fpm-deployment script was not updated to handle that. 

@awvwgk @jvdp1

jvdp1:
> Should I change the include path of `common.fypp` in the `.fypp` files from which `common.fypp` not found error is coming in the fpm-deployment ?
No, I don't think so. Instead the variable `fyflags` in the fpm-deployement should be updated with the flag `-I` (similar to for the CMake files)

fpm.toml
```diff
@@ -0,0 +1,16 @@
+name = "stdlib"
+
+[dev-dependencies]
+test-drive.git = "https://github.com/fortran-lang/test-drive.git"
+
+[build]
+auto-examples = false
+auto-tests = true
+
+[preprocess.cpp]
```
jvdp1, line 10: Should it be 
```suggestion
[preprocess.fypp]
```
?
As far as I remember, CPP is not used within stdlib.

jvdp1:
To solve the issue with the `Build-API-Docs`, the directory `include` should be added/mentioned in the file `API-doc-FORD-file.md` (probably under the directive `include:`).

fpm.toml
arteevraina: Yes, that is not used. But, we have not enabled ``fypp`` inside ``fpm`` as of now. So, it cannot identity it as a valid preprocessor. Considering ``fypp-gfortran.py`` as a temporary solution, I think when we have ``fypp`` integrated inside fpm then I think we should rename it as fypp instead of cpp.

arteevraina:
Hi @jvdp1, I still seem to face some issues with docs ci. Can you help me in resolving this?

zoziha:
This should be related to the version update of FORD, see #673 .

jvdp1:
> This should be related to the version update of FORD, see #673 .

@arteevraina Indeed it seems to be related to #673

jvdp1:
@arteevraina Could you rebase your PR such that @zoziha chnages (#681) are also integrated into this PR, please?
Hopefully this will solve the issue with the CI.

fpm.toml
```diff
@@ -0,0 +1,16 @@
...
+macros = [
+  "PROJECT_VERSION_MAJOR=0",
+  "PROJECT_VERSION_MINOR=1",
+  "PROJECT_VERSION_PATCH=0",
```
awvwgk, line 14: Should we support extracting those values automatically in fpm?

arteevraina: Yes, I think we should add support to extract them for the version number ?

arteevraina: PR created for this https://github.com/fortran-lang/fpm/pull/766

README.md
```diff
@@ -199,6 +199,21 @@ To use `stdlib` within your `fpm` project, add the following lines to your `fpm.
...
 ```
 
+### Build with [fypp-script](https://github.com/fortran-lang/fpm/pull/729)
+
+Fortran Package Manager has a compiler wrapper `fypp-script` that can be used to compile stdlib without 
+switching to `stdlib-fpm` branch.
+
+```sh
+git clone https://github.com/fortran-lang/fpm.git
+```
```
jvdp1, line 209: Why is that provided? IMO a link to `fpm` should be sufficient.

PR #683, opened 2022-10-27
Euler 37 repmat

Euler-37:
Fortran 2d array repmat


src/stdlib_linalg_repmat.fypp
```diff
@@ -0,0 +1,30 @@
...
+contains
+
+  #:for k1, t1 in RCI_KINDS_TYPES
+    pure module function repmat_${t1[0]}$${k1}$(a, m, n) result(res)
+      ${t1}$, intent(in) :: a(:,:)
+      integer, intent(in) :: m,n
+      ${t1}$ :: res(m*size(a,1),n*size(a,2))
+      integer ::i,j,k,l
+      associate(ma=>size(a,1),na=>size(a,2))
+         do j=1,n
```
milancurcic, line 16: @fortran-lang/stdlib what's our latest consensus on the use of `do concurrent` in stdlib? The 4-nested loop can be written more concisely with `do concurrent`, but I recall an argument against this is that it could interfere if `repmat` were called from within an OpenMP directive. Is `do concurrent` in stdlib still discouraged?

milancurcic (changes requested):
Thank you for this PR. II couldn't find a discussion about `repmat` in the issues but it's in scope for stdlib. This still needs a spec entry in the [stdlib_linalg](https://github.com/fortran-lang/stdlib/blob/master/doc/specs/stdlib_linalg.md) spec doc. Let me know if you need my help with that.

test/linalg/test_linalg.fypp
```diff
@@ -3,7 +3,7 @@
 module test_linalg
     use testdrive, only : new_unittest, unittest_type, error_type, check, skip_test
     use stdlib_kinds, only: sp, dp, xdp, qp, int8, int16, int32, int64
-    use stdlib_linalg, only: diag, eye, trace, outer_product
+    use stdlib_linalg, only: diag, eye, trace, outer_product,repmat
```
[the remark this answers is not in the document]
Euler-37: Sorry,It is my first commit,I will add the new issue.

src/stdlib_linalg_repmat.fypp
Euler-37: gfortran will do loop-interchange in `do concurrent`, But it may not the best order.

milancurcic: OK, let's leave it as is then.

jvdp1: I can't retrieve the discussion, neither remember what was decided. I believe @awvwgk was also involve in it.

src/stdlib_linalg.fypp
```diff
@@ -92,6 +93,21 @@ module stdlib_linalg
     #:endfor
   end interface outer_product
 
+  ! repeat matrix (of 2d array)
+  interface repmat
+    !! version: experimental
+    !!
+    !! Creates large matrices from a small array, `repmat()` repeats the given values of the array to create the large matrix.
+    !! ([Specification](../page/specs/stdlib_linalg.html#
+    !! repmat-creates-large-matrices-from-a-small-array))
```
jvdp1, line 102: What is the difference with the intrinsic `spread` (i.e. "replicates a source array ncopies times along a specified dimension")?

14NGiestas: > What is the difference with the intrinsic `spread` (i.e. "replicates a source array ncopies times along a specified dimension")?

"replicates a source **ARRAY** ncopies times along a specified dimension", this doesn't quite work for matrices (rank-2) as you will need to do some complicated trickery to reshape it.

Euler-37: Actually ,Fortran's intrinsic array functions have some strange optimizations. It may generate temporary array and `copy`. And ifort will be slow if you do some elemental-wise array operation.

PR #689, opened 2022-12-22
New zfill function to left-pad a string with zeros

ecasglez:
Adds a `zfill` function to left-pad a string with zeros. I have implemented it becuase I think it is a simple but useful function.

This would close issue #394.

It consists on an interface `zfill` pointing to functions `zfill_string` and `zfill_char` to use it with strings and chars respectively. The implementation is equivalent to the one in functions `padl_string_default` and `padl_char_default`.

I have prepared the documentation, examples and tests. I think I haven't miss anything but feel free to tell me so I can make any correction.


urbanjost:
I have a zero padding function and thought I would just mention some of the possibly different behaviors for consideration. In some cases I am not sure what this one does yet, but even if the same they might be points to consider (I hope to try this one soon, but have not yet):

o allow integer input arguments as well as strings?
o should the input string be cropped and then padded or should it retain original right-hand spaces?
o does it handle signed values by moving the sign to the beginning of the string?
o should the length be optional, in which case it defaults to the length of input string when the input is a string?

Not saying it should behave one way or the other, but if not already described, the documentation should make sure to clarify the behavior, based on the experience I had with something similar. 

I did not get permission to make another one public, but in addition to left, right, center, and zero padding one other padding function I have found useful is to pad "numeric" strings so that the decimal character is lined up.

doc/specs/stdlib_strings.md
```diff
@@ -421,6 +421,45 @@ The result is a scalar of integer type or an integer array of rank equal to the
 {!example/strings/example_count.f90!}
 ```
 
+
+### `zfill`
+
+#### Description
+
+Returns a string of length `output_length` left padded with zeros.
```
milancurcic, line 430:
```suggestion
Returns a string of length `output_length` left-padded with zeros.
```

milancurcic (approved):
Thank you, I think this is good to go, I'll just re-run the CI just in case.

milancurcic:
Thanks, @urbanjost, for all these ideas. Here's my opinion:

> o allow integer input arguments as well as strings?

I don't think so, we already have solid capability to go from an integer to string, so I think `zfill(string_type(123))` is a decent solution for this.

> o should the input string be cropped and then padded or should it retain original right-hand spaces?
> o does it handle signed values by moving the sign to the beginning of the string?

IMO, I don't think so, to keep the behavior simple, and I think the current description is appropriate in this regard (i.e. we describe what the function does, but not what it doesn't do).

> o should the length be optional, in which case it defaults to the length of input string when the input is a string?

I think that would be a good addition, let's evaluate it for a separate PR. I'll open an issue.

Issue #701, opened 2023-03-15
Push stdlib version to the stdlib-fpm branch


perazz:
### Motivation

Hello stdlib developers, 

apologies I'm no stdlib expert yet.

I think it would be fantastic to soon see tighter/full integration of stdlib in fpm, meaning, that fpm could enable stdlib as a meta-package with basically no user input (like any C/C++ compiler does for their stdlibs). 

I see that the CI does some [minimal cleanup of the sources](https://github.com/fortran-lang/stdlib/blob/master/ci/fpm-deployment.sh) before pushing to the stdlib-fpm branch. I would like to: 

- ask whether there is any interest in it? 

- ask for feedback about whether any of those pre-processing steps could instead be replaced with more advanced source handling in fpm;

- fpm.toml does not store a `VERSION`, is that by design? otherwise, I'd like to suggest a PR that pushes it on deployment: 

https://github.com/fortran-lang/stdlib/blob/261a6274477ec8e23f94bd8be384158149ed6628/fpm.toml#L1-L10

jvdp1:
I totally agree with you.
During the GSoC 2022, @arteevraina, together with @awvwgk, worked on supporting preprocessing with `fpm` which would enable `fpm` to compile `stdlib`. See this [PR](https://github.com/fortran-lang/fpm/pull/729) for more details. Once preprocessing is supported by `fpm`, it should be able to fully compile `stdlib`.

arteevraina:
> ### Motivation
> Hello stdlib developers,
> 
> apologies I'm no stdlib expert yet.
> 
> I think it would be fantastic to soon see tighter/full integration of stdlib in fpm, meaning, that fpm could enable stdlib as a meta-package with basically no user input (like any C/C++ compiler does for their stdlibs).
> 
> I see that the CI does some [minimal cleanup of the sources](https://github.com/fortran-lang/stdlib/blob/master/ci/fpm-deployment.sh) before pushing to the stdlib-fpm branch. I would like to:
> 
> * ask whether there is any interest in it?
> * ask for feedback about whether any of those pre-processing steps could instead be replaced with more advanced source handling in fpm;
> * fpm.toml does not store a `VERSION`, is that by design? otherwise, I'd like to suggest a PR that pushes it on deployment:
> 
> https://github.com/fortran-lang/stdlib/blob/261a6274477ec8e23f94bd8be384158149ed6628/fpm.toml#L1-L10
> 
> ### Prior Art
> _No response_
> 
[...]

PR #702, opened 2023-03-15
fpm-deployment: update `VERSION` field in fpm.toml with actual version

perazz:
See #701, this update in the fpm deployment script updates the `VERSION` placeholder in fpm.toml so an actual version is printed. 

I don't know how to test the deployment on my fork.

### DIFF
--- a/ci/fpm-deployment.sh
+++ b/ci/fpm-deployment.sh
@@ -20,7 +20,6 @@ fi
 
 # Additional files to include
 include=(
-  "ci/fpm.toml"
   "LICENSE"
   "VERSION"
 )
@@ -53,6 +52,9 @@ find example -name "example_*.f90" -exec cp {} "$destdir/example/" \;
 
 # Include additional files
 cp "${include[@]}" "$destdir/"
+
+# Update version in the manifest (not all systems support sed -i)
+sed -E "s,version = \"VERSION\",version = \"$major.$minor.$patch\",g" ci/fpm.toml > "$destdir/fpm.toml"
 
 # Source file workarounds for fpm; ignore missing files
 rm "${prune[@]}"


jvdp1 (approved):
LGTM. Thanks @perazz for this PR.

arteevraina (approved):

arteevraina:
> I don't know how to test the deployment on my fork.

You can try running this script locally and try to generate the fpm compatible ``stdlib`` ? and look if the changes are reflected in ``fpm.toml``.

awvwgk (changes requested):
Why is this change necessary? We already deploy the `VERSION` file correctly for the stdlib-fpm branch.

perazz:
Currently, fpm.toml in the deployed branch reports `version = "VERSION"`, is this a feature? This proposed change updates that field with the actual version, i.e., `version = "0.2.1"`

awvwgk:
> Currently, fpm.toml in the deployed branch reports `version = "VERSION"`; this proposed change updates that field with the actual version, i.e., `version = "0.2.1"`

Maybe, I misunderstand something here, but this is exactly what is expected. The version is already deployed correctly in [`VERSION`](https://github.com/fortran-lang/stdlib/blob/stdlib-fpm/VERSION) on the `stdlib-fpm` branch. It will be read dynamically by the fpm binary. So why do we need to change the manifest at all?

If I recall correctly, I actually implemented this feature of reading the version from a file in fpm for stdlib to have a single place to specify the version for both build systems (fpm and CMake).

perazz:
Oh [I see](https://fpm.fortran-lang.org/en/spec/manifest.html?highlight=versio#project-version), version can be an external file. If we were going to support stdlib as a metapackage, I thought it was intuitive for the fpm-specific deployment to have a manifest file that contains all relevant pieces of information.

awvwgk:
That was unfortunately never true, especially due to the default names for the source directories as well as the automatic discovery of executables, the TOML manifest alone is usually incomplete. Running fpm will generate an internal representation which is complete, the two options are to have fpm dump the internal model or reimplements the whole logic supported by fpm.

jvdp1:
@perazz @awvwgk Is this PR still needed? Or could it be closed without merging it?

perazz:
Please close without merging @jvdp1, thank you.

jvdp1:
Ok. Thank you.

Issue #450, opened 2021-07-02
Interfaces and wrappers: BLAS


awvwgk:
**Description**

To support linear algebra operations we want to have interfaces/wrappers for BLAS and LAPACK, maybe even to BLACS, PBLAS and ScaLAPACK at some point. Before we start working on a high-level API, there are additional stages to consider. For this issue I want to focus on BLAS only, because the size of the problem is much more limited than the other linear algebra libraries and we can flesh out the workflow

1. How to include BLAS in stdlib?
    - building (reference) BLAS from source with fpm and/or CMake sounds like a bad idea (performance), but might be a fallback
    - linking against system BLAS and optimized BLAS libraries like MKL or OpenBLAS
2. How due we provide the BLAS functionality?
    - interface modules like BLAS95 are available
    - do we reproduce the actual reference BLAS interface (doesn't include `intent` in the dummy arguments)
    - do we overload the names to drop the s/d/c/z prefixes
    - are the interfaces part of stdlib or bundled from a separate project
3. How do we abstract the operations available in BLAS?
    - how many layers of abstractions do we want?
    - just simple wrapper to infer dimensions
    - overloaded operators

arjenmarkus:
One additional thing: as the source code does not use modules, we are
"polluting" the namespace with the names of the BLAS routines. So an
additional step might be to put the individual routines into modules to
avoid that.

Another aspect is the use of routines such as DMACH1 (not in LAPACK and
BLAS, but similar routines are). Some of these venerable packages require
you to adjust the code for specific machines. With Fortran in its current
incarnation we can rely on  intrinsic functions for most of these.

Op vr 2 jul. 2021 om 13:52 schreef Sebastian Ehlert <
***@***.***>:

awvwgk:
@arjenmarkus I think the namespace pollution can't be helped, at least if we want to rely on optimized BLAS libraries. Putting the routines into modules would produce ABI issues unless we export them with bind(C), which kind of defeats the purpose of cleaning up the global namespace with modules in the first place.

ivan-pi:
It will be good to study how NumPy embeds BLAS: [Accelerated BLAS/LAPACK libraries](https://numpy.org/doc/stable/user/building.html?highlight=blas#accelerated-blas-lapack-libraries)

[...]

PR #710, opened 2023-05-06
Add solve based on OpenBLAS

### LINKED ISSUE #709: The first step in facilitating BLAS integration into stdlib: add a solve function
### Motivation

Fortran arrays have a natural advantage in linear algebra, but unfortunately, stdlib currently does not integrate any high-level function implementation based on BLAS, LAPACK interfaces, such as `det`, `solve`, `inv`. For Fortran users, this is frustrating. Obviously we can use LAPACK directly, but stdlib's motivation is to be the math library that can catch up with `numpy`, and linear algebra is essential.
(I understand that this may not be a purely technical issue, it may involve consensus and norms.)

I'd like to start by trying to link `openblas` in stdlib (`numpy` gives preference to `openblas` and `mkl`) and show you that we are enthusiastic about BLAS. Maybe it won't succeed directly, but I'll still try to implement `linalg.solve` like `numpy`.

`solve` is not computationally efficient: There are two array assignments before `_gesv` is called, which takes a bit of time for large arrays, but is good for ease of use.
Also, since this is the first time LAPACK is encapsulated, there is no assertion on the return value `info`, so discussion is welcome.

### Prior Art

- [numpy.linalg.solve](https://numpy.org/doc/stable/reference/generated/numpy.linalg.solve.html)
- [scipy.linalg.solve](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.solve.html#scipy.linalg.solve)

### Additional Information

_No response_

zoziha:
- [x] Use `pkg-config` to find `openblas` and define `USE_OPENBLAS`;
- [x] `BUILD_OPENBLAS` is off by default because stdlib has few BLAS-based functions yet.
- [x] Routines based on BLAS and LAPACK often only support single or double precision floating-point types;
- [x] Add `BLAS_common.fypp` to set the BLAS related common variables;
- [x] Add a `solve` function to `stdlib_linalg`, and test it;
- [ ] Discusses and handles the `_gesv` return value `info`.
- [ ] Test it in CI (MINGW✔, Linux-gfortran✔, Linux-ifort, macOS-gfortran);

This PR is an attempt to irrationally generate enthusiasm for BLAS integration into stdlib. Close #709 .

### Syntax

```fortran
x = [[stdlib_linalg(module):solve(interface)]](a,b)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,7 @@ Important options are
   Compiling with maximum rank 15 can be resource intensive and requires at least 16 GB of memory to allow parallel compilation or 4 GB memory for sequential compilation.
 - `-DBUILD_SHARED_LIBS` set to `on` in case you want link your application dynamically against the standard library (default: `off`).
 - `-DBUILD_TESTING` set to `off` in case you want to disable the stdlib tests (default: `on`).
+- `-DBUILD_OPENBLAS` set to `on` in case you want to use routines based on OpenBLAS (default: `off`).
 - `-DCMAKE_VERBOSE_MAKEFILE` is by default set to `Off`, but if set to `On` will show commands used to compile the code.
 - `-DCMAKE_BUILD_TYPE` is by default set to `RelWithDebInfo`, which uses compiler flags suitable for code development (but with only `-O2` optimization). Beware the compiler flags set this way will override any compiler flags specified via `FFLAGS`. To prevent this, use `-DCMAKE_BUILD_TYPE=NoConfig` in conjunction with `FFLAGS`.
 
--- a/doc/specs/stdlib_linalg.md
+++ b/doc/specs/stdlib_linalg.md
@@ -6,6 +6,17 @@ title: linalg
 
 [TOC]
 
+The stdlib linear algebra functions rely on BLAS and LAPACK to provide
+efficient low level implementations of standard linear algebra algorithms.
+
+If possible, highly optimized libraries that take advantage of specialized processor functionality are preferred.
+Examples of such libraries are [OpenBLAS][1], [oneAPI MKL(TM)][2].
+Because these libraries are multithreaded and processor dependent,
+environment variables may be required to control the number of threads or to specify the processor architecture.
+
+[1]: https://www.openblas.net/
+[2]: https://www.intel.com/content/www/us/en/developer/tools/oneapi/onemkl.html
+
 ## `diag` - Create a diagonal array or extract the diagonal elements of an array
 
 ### Status
@@ -424,4 +435,44 @@ Specifically, upper Hessenberg matrices satisfy `a_ij = 0` when `j < i-1`, and l
 
 ```fortran
 {!example/linalg/example_is_hessenberg.f90!}
+```
+
+## `solve` - Solve a-linear-matrix-equation
+
+### Status
+
+Experimental
+
+### Description
+
+Solve a linear matrix equation, or system of linear scalar equations.
+
+Note: `solve` supports only single and double precision floating point types.
+
+### Class
+
+Impure function.
+
+### Synatx
+
+`x = [[stdlib_linalg(module):solve(interface)]](a,b)`
+
+### Arguments
+
+`a`: Shall be a rank-2 array of either `real` or `complex` type. This argument is `intent(in)`. Coefficient matrix.
+
+`b`: Shall be a rank-2 array of either `real` or `complex` type. This argument is `intent(in)`. Ordinate or “dependent variable” values.
+
+Note: `a` and `b` shall be of the same type.
+
+### Return value
+
+Returns a rank-2 array. Solution to the system `ax = b`. Returned shape is identical to `b`.
+
+### Example
+
+Solve the system of equations `x1 + 2 * x2 = 1` and `3 * x1 + 5 * x2 = 2`.
+
+```fortran
+{!example/linalg/example_solve.f90!}
 ```
--- a/src/stdlib_linalg.fypp
+++ b/src/stdlib_linalg.fypp
@@ -1,4 +1,5 @@
 #:include "common.fypp"
+#:include "BLAS_common.fypp"
 #:set RCI_KINDS_TYPES = REAL_KINDS_TYPES + CMPLX_KINDS_TYPES + INT_KINDS_TYPES
 module stdlib_linalg
   !!Provides a support for various linear algebra procedures
@@ -23,6 +24,10 @@ module stdlib_linalg
   public :: is_hermitian
   public :: is_triangular
   public :: is_hessenberg
+
+  #:if defined("USE_OPENBLAS")
+  public :: solve
+  #:endif
 
   interface diag
     !! version: experimental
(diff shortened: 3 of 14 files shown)


src/stdlib_linalg.fypp
```diff
@@ -214,6 +219,25 @@ module stdlib_linalg
...
+    !! version: experimental
+    !!
+    !! Solve a linear matrix equation, or system of linear scalar equations
+    !! ([Specification](../page/specs/stdlib_linalg.html#
+    !! solve-solve-a-linear-matrix-equation))
+    #:set REAL_KINDS_TYPES_GESV = list(zip(BLAS_REAL_KINDS, REAL_TYPES, BLAS_REAL_GESV))
+    #:set CMPLX_KINDS_TYPES_GESV = list(zip(BLAS_CMPLX_KINDS, CMPLX_TYPES, BLAS_CMPLX_GESV))
+    #:set RC_KINDS_TYPES_GESV = REAL_KINDS_TYPES_GESV + CMPLX_KINDS_TYPES_GESV
+    #:for k1, t1, g1 in RC_KINDS_TYPES_GESV
+    module function solve_${t1[0]}$${k1}$(a, b) result(x)
```
St-Maxwell, line 233: I would prefer a subroutine-style interface, e.g.
```fortran
subroutine solve(X, A, B)
```

README.md
```diff
@@ -133,6 +133,7 @@ Important options are
   Compiling with maximum rank 15 can be resource intensive and requires at least 16 GB of memory to allow parallel compilation or 4 GB memory for sequential compilation.
 - `-DBUILD_SHARED_LIBS` set to `on` in case you want link your application dynamically against the standard library (default: `off`).
 - `-DBUILD_TESTING` set to `off` in case you want to disable the stdlib tests (default: `on`).
+- `-DBUILD_OPENBLAS` set to `on` in case you want to use routines based on OpenBLAS (default: `off`).
```
St-Maxwell, line 136: `BUILD_WITH_OPENBLAS` may be better

jvdp1, line 136: `fpm` should be also able to compile the project.

doc/specs/stdlib_linalg.md
```diff
@@ -6,6 +6,17 @@ title: linalg
 
 [TOC]
 
+The stdlib linear algebra functions rely on BLAS and LAPACK to provide
+efficient low level implementations of standard linear algebra algorithms.
+
+If possible, highly optimized libraries that take advantage of specialized processor functionality are preferred.
+Examples of such libraries are [OpenBLAS][1], [oneAPI MKL(TM)][2].
```
jvdp1, line 13:
```suggestion
Examples of such libraries are [OpenBLAS][1], [Intel oneAPI MKL(TM)][2].
```

doc/specs/stdlib_linalg.md
```diff
@@ -425,3 +436,43 @@ Specifically, upper Hessenberg matrices satisfy `a_ij = 0` when `j < i-1`, and l
 ```fortran
 {!example/linalg/example_is_hessenberg.f90!}
 ```
+
+## `solve` - Solve a-linear-matrix-equation
```
jvdp1, line 440:
```suggestion
## `solve` - Solve a linear matrix equation
```

doc/specs/stdlib_linalg.md
```diff
@@ -425,3 +436,43 @@ Specifically, upper Hessenberg matrices satisfy `a_ij = 0` when `j < i-1`, and l
...
+
+### Status
+
+Experimental
+
+### Description
+
+Solve a linear matrix equation, or system of linear scalar equations.
+
+Note: `solve` supports only single and double precision floating point types.
```
jvdp1, line 450: This "note" will be implicitely mentioned through the description of the arguments. Therefore, it could be removed

doc/specs/stdlib_linalg.md
```diff
@@ -425,3 +436,43 @@ Specifically, upper Hessenberg matrices satisfy `a_ij = 0` when `j < i-1`, and l
...
+
+### Class
+
+Impure function.
```
jvdp1, line 454: For such procedures, I would prefera subroutine over a function, especially for efficienct with large vectors/matrices.

doc/specs/stdlib_linalg.md
```diff
@@ -425,3 +436,43 @@ Specifically, upper Hessenberg matrices satisfy `a_ij = 0` when `j < i-1`, and l
...
+
+### Synatx
+
+`x = [[stdlib_linalg(module):solve(interface)]](a,b)`
```
jvdp1, line 458: Could `b` be a vector too?

doc/specs/stdlib_linalg.md
```diff
@@ -425,3 +436,43 @@ Specifically, upper Hessenberg matrices satisfy `a_ij = 0` when `j < i-1`, and l
...
 {!example/linalg/example_is_hessenberg.f90!}
 ```
+
+## `solve` - Solve a-linear-matrix-equation
+
+### Status
+
+Experimental
+
+### Description
```
jvdp1, line 446: It might be good to add somewhere to which BLAS procedure it is linked with

jvdp1 (commented):
Thank you. I think it will be an important step for stdlib. My main concern is about the API. Should it be a simple wrapper, or something more complete (similar to @awvwgk [BLAS interfaces](https://github.com/awvwgk/blas-interface)).
See also #450 for some discussions.

doc/specs/stdlib_linalg.md
zoziha: Subroutines are usually faster than functions, functions are usually easier to use but may cause performance degradation. If you need high performance, you should use subroutines. If you need simpler code, you should use a function.

Intuitively, solving a system of `AX=B` linear equations, the normal form `X=solve(A, B)` will naturally be understood by the average user and will be surprised by `call solve(A, B)`. My personal recommendation is to think of `solve` as a function dedicated to solving systems of linear equations, rather than for performance, which can be done with `_gesv` routines.

- (Not recommended) Perhaps we could provide a `solve_` **subroutine** that implements the in-place version of `solve`, like this:
```fortran
subroutine solve_(a, b)
    real, intent(inout) :: a(:, :), b(:, :)
    integer :: ipiv(size(a, 1)), info
    call sgesv(size(a, 1), size(b, 2), a, size(a, 1), ipiv, b, size(b, 1), info)
end subroutine solve_
```

- (Recommended) Or for performance-sensitive users, it is recommended to just use the low-level interface `_gesv` for programming by borrowing the implementation of `solve`?

zoziha:
> Should it be a simple wrapper, or something more complete (similar to @awvwgk [BLAS interfaces](https://github.com/awvwgk/blas-interface)).

I'm not worried about this, and it shouldn't be too complicated to switch if the full BLAS interfaces library is available later.

src/stdlib_linalg.fypp
cval26: I agree with the above. It makes more sense to use a subroutine, so that the code stays performant when large matrices are involved. 

Also, would it make sense to call the procedure something like `linsolve`, instead of just `solve`, anticipating that in the future a nonlinear solver will be added as well?

jvdp1:
Related to #749 (ping @perazz )

jvdp1:
This PR should be reviewed considering latest developments of @perazz on BLAS/LAPACK

jvdp1:
Closed in favour of #806 .
Thank you to all contributors to this PR.

PR #714, opened 2023-05-14
change stdlib-random function `rtol` to `ishftc`

Euler-37:
``` c
static inline uint64_t rotl(const uint64_t x, int k) {
   return (x << k) | (x >> (64 - k));
}
```
fortran has the intrinsic bit function `ishftc(x,k)` ,it is same as rtol


awvwgk (approved):

PR #715, opened 2023-06-02
Add warning for breaking change in fpm

Carltoffel:
Fpm 0.8.2 introduced stdlib as a metapackage.
The previous way to include stdlib in a fpm project produces following error:
```
<ERROR> *cmd_run* Package error: Could not retrieve version string for metapackage key <stdlib>. Check syntax
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,7 +205,7 @@ stdlib = { git="https://github.com/fortran-lang/stdlib", branch="stdlib-fpm" }
 > To include the standard library metapackage, change the dependency to:
 > `stdlib = "*"`.
 > 
-> see also https://fpm.fortran-lang.org/en/spec/metapackages.html
+> [see also](https://fpm.fortran-lang.org/en/spec/metapackages.html)
 
 ## Using stdlib in your project
 
--- a/README.md
+++ b/README.md
@@ -199,6 +199,14 @@ To use `stdlib` within your `fpm` project, add the following lines to your `fpm.
 stdlib = { git="https://github.com/fortran-lang/stdlib", branch="stdlib-fpm" }
 ```
 
+> **Warning**
+> 
+> Fpm 0.8.2 and later implements stdlib as a *metapackage*.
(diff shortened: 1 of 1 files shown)


Carltoffel:
Maybe someone should disable all these CI checks if the only thing that has changed is the documentation. I don't know much about CI, just found [this](https://stackoverflow.com/questions/73298609/suppress-gitlab-ci-stage-if-push-only-changes-readme-md). Hope it helps.

README.md
```diff
@@ -199,6 +199,14 @@ To use `stdlib` within your `fpm` project, add the following lines to your `fpm.
 stdlib = { git="https://github.com/fortran-lang/stdlib", branch="stdlib-fpm" }
 ```
 
+> **Warning**
+> 
+> Fpm 0.8.2 and later implements stdlib as a *metapackage*.
```
perazz, line 204: Please use `0.9.0`, it’s the release metapackages went officially into

perazz: We have user documentation at https://fpm.fortran-lang.org/en/spec/metapackages.html 

can I ask you to use that link? I think it’s better than referring to a developer page

perazz (commented):
Thanks a lot @Carltoffel for catching this -   Yes this part of the docs definitely needs your suggested update

README.md
awvwgk: Would it be better to have the option in from to use stdlib from a branch as advertised here and have it as metapackage as well. This way we don't break backwards compatibility.

perazz: https://github.com/fortran-lang/fpm/issues/926

perazz (approved):
Great @Carltoffel, thank you. lgtm.

jvdp1:
@perazz @awvwgk @Carltoffel 
Is this still revelant since @fortran-lang/fpm#928 is merged? Can it be merged as is? I think so, but I prefer checking first.

perazz:
Thank you for asking @jvdp1 - yes I agree that this is OK to be merged.

PR #717, opened 2023-06-08
Extension of PR#679 to `stdlib_stats_distribution_exponential`

HugoMVale:
Please refer to #679 for an explanation about the issue to solve and approach used. As discussed there, I've extended the approach to the exponential module.

IMO, this module still has other opportunities for improvement. For instance, I don't understand why `rvs_exp_array_` does not call the corresponding scalar procedure `rvs_exp_`, but instead _repeats_ the code in `rvs_exp_0_`. I will have a better look at it and, eventually, suggest another PR.

### DIFF
--- a/doc/specs/stdlib_stats_distribution_exponential.md
+++ b/doc/specs/stdlib_stats_distribution_exponential.md
@@ -14,16 +14,17 @@ Experimental
 
 ### Description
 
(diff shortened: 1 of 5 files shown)


jvdp1 (commented):
LGTM.
Are the "new" behaviours ("NAN" instead of "error stop") reflected in the specs?

HugoMVale:
For the sake of clarity, the question raised above by @jvdp1 has been addressed with the last batch of commits. From my perspective, this PR is ready.

PR #718, opened 2023-06-08
Improve doc + examples of `stdlib_stats_distribution_normal`

HugoMVale:
As a complement to PR #679, I've made a couple of improvements on the doc and related examples.
* Improve readability of specs
* Add examples with sigma<=0 to show expected behavior (NaN, not error).
* Add examples with calls inside do concurrent to show it is (now) possible.
* Rename `example_norm_cdf.f90` to `example_normal_cdf.f90` for consistency with other files.

PR #721, opened 2023-06-24
fix doc stats-distribution-normal and stats-distribution-exponential

HugoMVale:
This PR aims to fix some residual/undetected issues with PR #717 and #718 , namely:
* inline math was not rendering due to use of incorrect symbol.
* footnotes were not rendering due to missing colon.
* outdated example filename.

I was not able to detect these issues earlier because, in my local windows env, FORD in unable to parse certain Fortran source files (encoding issue) and aborts. I still have this issue (and assume other may as well), but I am going around it by excluding the problematic files.

### DIFF
--- a/doc/specs/stdlib_stats_distribution_exponential.md
+++ b/doc/specs/stdlib_stats_distribution_exponential.md
@@ -14,11 +14,12 @@ Experimental
 
 ### Description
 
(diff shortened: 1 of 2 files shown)


jvdp1 (approved):
thank you @HugoMVale . I'll merge it.

PR #723, opened 2023-06-26
Support sorting arrays of bitsets

### LINKED ISSUE #720: Support sorting arrays of bitsets
### Motivation

Currently, stdlib does not support the sorting of bitset arrays. Support for sorting arrays of bitsets may allow stdlib to apply to modern numerical simulations.

In recent years, adaptive mesh refinement (AMR) has attracted much attention in fluid simulations. In some research aimed at the efficient implementation of AMR, a bit sequence like `11010010` representing the grid position and refinement level is used as an ID for a grid, and the bitset is used to store IDs. 
The IDs represented by bitsets are sorted for load balancing in parallel computations. stdlib has bitsets but cannot perform sorting. This prevents the implementation of AMR using stdlib.
Thus, it is worth supporting sorting arrays of bitsets.

### Prior Art

- [Hasbestan and Senocak](https://doi.org/10.1016/j.jcp.2018.04.039) and [Liu et al.](https://doi.org/10.1016/j.cma.2022.114662) use `std::bitset` to represent IDs in their implementation of AMR using C++.

### Additional Information

stdlib already implements the assignment operator (`=`) and comparison operators (`>, >=, <, <=`) for bitsets. So it is easy to support  sorting arrays of bitsets by adding a few lines to stdlib_sorting.fypp and related files:
```diff
+#! Derived type bitsets
+#:set BITSET_KINDS = ["bitset_64", "bitset_large"]
+
+#! Bitset types to be considered during templating
+#:set BITSET_TYPES = ["type({})".format(k) for k in BITSET_KINDS]
```
```diff
+#:set BITSET_TYPES_ALT_NAME = list(zip(BITSET_TYPES, BITSET_TYPES, BITSET_KINDS))

#! For better code reuse in fypp, make lists that contain the input types,
#! with each having output types and a separate name prefix for subroutines
#! This approach allows us to have the same code for all input types.
-#:set IRSC_TYPES_ALT_NAME = INT_TYPES_ALT_NAME + REAL_TYPES_ALT_NAME + STRING_TYPES_ALT_NAME + CHAR_TYPES_ALT_NAME
+#:set IRSC_TYPES_ALT_NAME = INT_TYPES_ALT_NAME + REAL_TYPES_ALT_NAME + STRING_TYPES_ALT_NAME + CHAR_TYPES_ALT_NAME + BITSET_TYPES_ALT_NAME
```

degawa:
To support sorting arrays of bitsets, `type(bitset_64)` and `type(bitset_large)` are added as arguments of procedures in the `stdlib_sorting`.

- Added bitset types to `ALT_NAME` in the `stdlib_sorting` modules.
- Added some tests.
- Updated and fixed API-doc and spec related to the `stdlib_sorting` module.
  - The overview of sorting was not updated to respect the original intent.
    > The Fortran Standard Library therefore provides a module, `stdlib_sorting`, with procedures to sort arrays of simple intrinsic numeric types
  - Performance benchmarks were not updated because the same computer environment was not available.
- Added an example.
- Changed the argument's `intent` from `out` to `inout`  in the procedure for overloading the assignment operator for bitset types.
  - This change avoids assignment failure when an assignment such as `array(1)=array(1)` is performed. 
  - Executed examples that were using the bitset types and confirmed those finished successfully.

closes #720


src/stdlib_bitsets.fypp
```diff
@@ -1170,8 +1170,8 @@ module stdlib_bitsets
 !! Version: experimental
 !!
 !! Used to define assignment for `bitset_large`.
-            type(bitset_large), intent(out) :: set1
-            type(bitset_large), intent(in)  :: set2
+            type(bitset_large), intent(inout) :: set1
```
jvdp1, line 1173: Why is this needed?
And if it is needed, I guess the corresponding interfaces should be also modified for `bitset_64`.
It might be wise to open another PR with these changes only related to `stdlib_bitset`

src/stdlib_bitsets_large.fypp
```diff
@@ -91,12 +91,11 @@ contains
...
 !     Used to define assignment for bitset_large
-        type(bitset_large), intent(out) :: set1
-        type(bitset_large), intent(in)  :: set2
+        type(bitset_large), intent(inout) :: set1
+        type(bitset_large), intent(in)    :: set2
 
         set1 % num_bits = set2 % num_bits
-        allocate( set1 % blocks( size( set2 % blocks, kind=bits_kind ) ) )
-        set1 % blocks(:) = set2 % blocks(:)
+        set1 % blocks = set2 % blocks(:)
```
jvdp1, line 98: This might restrict `stdlib`  to the most updated compilers. Why did you change it?

src/stdlib_bitsets.fypp
jvdp1: Should the corresponding specs be modified too?

jvdp1 (commented):
Thank you. On overall LGTM. Here are a few comments.

example/sorting/example_sort_bitsetl.f90
```diff
@@ -0,0 +1,47 @@
+program example_sort_bitsetl
```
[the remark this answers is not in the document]
degawa: I added the ending `l`  to indicate that it is an example of `type(bitset_large)`.
As in the other sorting examples, I remove the `l`.
The filename `example_sort_bitsetl.f90` will be changed to `example_sort_bitset.f90`

example/sorting/example_sort_bitsetl.f90
```diff
@@ -0,0 +1,47 @@
...
+           bitset_l("0100"), & ! 4
+           bitset_l("0011"), & ! 3
+           bitset_l("0001"), & ! 1
+           bitset_l("1010"), & ! 10
+           bitset_l("0100"), & ! 4
+           bitset_l("1001")]   ! 9
+
+  call sort(array)
+
+  block
```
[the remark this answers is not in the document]
degawa: Thank you for your suggestion. The `block` structure is not used in the other examples. Following your suggestion, I will remove the `block` structure.

src/stdlib_bitsets.fypp
degawa: This change and the change you mentioned below have been done to fix unexpected behavior when assigning the `bitset_large` type.

Without this fix, the tests for sorting `bitset_large` in the following environment will fail:
- msys2-build (MSYS, x86_64) in ci_windows.yml
- msys2-build (MINGW32, i686) in ci_windows.yml
- msys2-build (MINGW32, i686) in ci_windows.yml
- Build (macos-latest, 11, cmake) in CI.yml
- Build (macos-latest, 12, cmake) in CI.yml

As you said, opening a new issue for further discussion is better.

---
I share the reason for those modification.

The component `blocks` of `set1` with the `intent(out)` attribute is not allocated at the beginning of the assignment procedure. Therefore, when `size( set2 % blocks, kind=bits_kind )` is executed, an error occurs that `blocks` have not been allocated. To avoid this error, I modified the `intent` attribute from `intent(out)` to `intent(inout)`.

Although I could not find a direct description of this situation in the Fortran Standard, the following statements are relevant:

>The INTENT attribute for an allocatable dummy argument applies to both the allocation status and the
definition status. An actual argument that corresponds to an INTENT (OUT) allocatable dummy argument
is deallocated on procedure invocation (9.7.3.2). 

> When a variable of derived type is deallocated, any allocated allocatable subobject is deallocated. If an error condition occurs during deallocation, it is processor dependent whether an allocated allocatable subobject is deallocated.

> The INTENT (OUT) attribute for a nonpointer dummy argument specifies that the dummy argument becomes undefined on invocation of the procedure, except for any subcomponents that are default-initialized (7.5.4.6).

An error still occurred in the assignment procedure after the modification.
`allocate( set1 % blocks( size( set2 % blocks, kind=bits_kind ) ) )` gives the error "`set1%blocks` is already allocated" when the same variable of the `bitset_large` type was specified on both sides of the assignment operator. This sometimes happens when the array indices for the bitset array to be sorted are the same.
Since `set1` and `set2` are the same when the same array element is now specified on both sides of the assignment operator, it is natural that this error occurs.
Therefore, the automatic allocation functionality is used instead of explicit allocation.

Similar modifications are not required for `type(bitset_64)`. This is because `type(bitset_64)` uses an integer corresponding `integer(int64)` for storing bitset and thus uses a built-in assignment operation.

jvdp1: Ok. Thank you for the explanation. A new PR with a test replicating this issue would be useful.
Otherwise, this PR looks good to me.

degawa: I followed your comment and opened another issue #726 with replication instructions because discussing it in more detail would be better.

degawa:
I removed the user-defined assignment operator for the `bitset_large` type. Then I included the test @jvdp1 added in the pull request #727 and confirmed that all the tests passed.
These commits solve the issue of the self-assignment for the `bitset_large` type.

degawa:
> Nice that it solves the issue.
To keep "one idea, one PR", this change should be merged in https://github.com/fortran-lang/stdlib/pull/727 first.

I understand the review policy, but I have already merged the changes made in #727. How should I handle those?
Should I ask to remove the `assign_large` procedure in #727 for solving the self-assignment problem?
Many thanks for your advice.

jvdp1:
> I understand the review policy, but I have already merged the changes made in #727. How should I handle those? Should I ask to remove the `assign_large` procedure in #727 for solving the self-assignment problem? Many thanks for your advice.

Sorry for the confusion. #727 should be merged first, with the remove `assign_large` procedure (I just included it in the PR. Could you review it, please?). Then you can rebase your branch against the master branch.

src/stdlib_bitsets.fypp
```diff
@@ -1167,6 +1167,7 @@ module stdlib_bitsets
 !!```
 
 
+
```
jvdp1, line 1170:
```suggestion
```

src/stdlib_bitsets_large.fypp
```diff
@@ -90,6 +90,8 @@ contains
 
 
 
+
+
```
jvdp1, line 94:
```suggestion
```

jvdp1 (approved):
thank you @degawa the PR #727 has been merged. I will now merge your PR, which became much simpler and with a single aim.

Issue #729, opened 2023-08-14
Incorrect result for gamma functions of pure imaginary argument


banana-bred:
### Description

The gamma function implemented in `cgamma_csp` and `cgamma_cdp` in module `stdlib_specialfunctions_gamma` produce the wrong output when the argument is a pure imaginary. They return the complex conjugate of the output instead.

### Expected Behaviour

An example where the stdlib implementation of `gamma` returns the complex conjugate of the expected result. The expected result here is calculated (with worse accuracy) by the user-defined function `gammaWeierstrass`, which implements

https://en.wikipedia.org/wiki/Gamma_function#19th_century:_Gauss,_Weierstrass_and_Legendre ,

and by `exp(log_gamma(z))`. The latter two agree up to some low accuracy, but are both different than `gamma(z)`.
This is not expected, as `exp(log_gamma(z))` should be the same as `gamma(z)`.

test.f90: 
```
program test

  use iso_fortran_env,               only: stdout => output_unit
  use stdlib_specialfunctions_gamma, only: gamma, log_gamma

  implicit none

  integer, parameter :: sp = selected_real_kind(6)
  integer, parameter :: dp = selected_real_kind(15)

  complex(sp), parameter :: a = cmplx(0, 1, kind = sp)
  complex(dp), parameter :: b = cmplx(0, 1, kind = dp)

  write(stdout,'(9X," -- Stdlib Γ(z) -- ")',advance='no')
  write(stdout,'(10X,"|"," -- Weierstrass Γ(z) -- ")', advance='no')
  write(stdout,'(8X,"|"," -- exp(ln(Γ(z)))  -- ")')

  write(stdout,'(A,X,2E15.7," | ",2E15.7," | ",2E15.7)') "Γ(i): ", gamma(a),  gammaWeierstrass(b),  exp(log_gamma(b))
  write(stdout,'(A,X,2E15.7," | ",2E15.7," | ",2E15.7)') "Γ(-i):", gamma(-a), gammaWeierstrass(-b), exp(log_gamma(-b))
  write(stdout,'(A,X,2E15.7," | ",2E15.7," | ",2E15.7)') "Γ(i): ", gamma(b),  gammaWeierstrass(b),  exp(log_gamma(b))
  write(stdout,'(A,X,2E15.7," | ",2E15.7," | ",2E15.7)') "Γ(-i):", gamma(-b), gammaWeierstrass(-b), exp(log_gamma(-b))
  write(stdout,*)

  contains

[...]

PR #730, opened 2023-08-14
PR related to #729, gamma function of pure imaginary returns correct value

banana-bred:
PR opened to fix the issue reported in #729.


jvdp1 (approved):
Thank you for this fix. LGTM. I'll approce it and merge it later, if there are no other reviews.

jvdp1:
Thank you @banana-bred for this fix.

PR #737, opened 2023-09-03
Addition of GCC 13 to CI

jvdp1:


awvwgk (approved):
Thanks. Looks good to me

PR #739, opened 2023-09-19
Fixed incorrect intent of the variable "conflict" on hashmaps in document

Daisuke897:
In the source code, the intent of the variable "conflict" in the procedure "map_entry" is "out", whereas in the documentation it is "in".
I corrected it from "in" to "out" on the documentation.

https://github.com/fortran-lang/stdlib/blob/82f35a57063f5db39ac8c469867d560327ff7cd0/src/stdlib_hashmaps.f90#L417-L431


jvdp1 (approved):
Good catch! Thank you @Daisuke897 for this fix.

PR #741, opened 2023-10-13
Added a procedure for getting all the keys in a hashmap

### LINKED ISSUE #732: Adding a procedure getting all keys in the hashmaps
### Motivation

The hashmap in the stdlib does not provide a way to retrieve all keys in the map. As shown in the Prior Art section, a way to retrieve all keys in a hashmap is required in other programming languages, and it would significantly expands the use of the hashmap.

### Prior Art

The [Map](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html) in Java has the `SetKey()` method that returns the keys in this map.
The `list()` function for [Python's `dict`](https://docs.python.org/3/library/stdtypes.html#typesmapping) returns a list of all the keys used in the dictionary.
The `unordered_map` in C++ does not have a similar method, but a way to get all the keys is required. (for example, see https://stackoverflow.com/questions/8483985/obtaining-list-of-keys-and-values-from-unordered-map)

### Additional Information

The hashmaps in the stdlib, including `open_hash_map_type` and `chaining_hash_map_type`, store both the entered keys and values to the component `inverse`. So it is not challenging work to retrieve the keys as a `key_type` array. 
The conversion from the `key_type` to the original type is left to the users.

degawa:
To add a procedure for getting all the keys in a hashmap, the subroutine `get_all_keys` is added to the hashmap types.

`get_all_keys` is implemented as a subroutine rather than a function to maintain consistency with the other procedures, especially `get_other_data`.

- Added a deferred procedure to `hashmap_type`
- Implemented the type-bound procedure for `chaining_hashmap_type` and `open_hashmap_type`.
- Added some tests and an example.
- Updated the specification document for hashmaps.

closes #732


doc/specs/stdlib_hashmaps.md
```diff
@@ -1148,6 +1154,9 @@ Procedures to modify the content of a map:
 
 Procedures to report the content of a map:
 
+* `map % get_all_keys( all_keys )` - Returns all the keys
+  presented in the map;
```
jvdp1, line 1158: Should it be "present" or "presented"?

example/hashmaps/example_hashmaps_get_all_keys.f90
```diff
@@ -0,0 +1,53 @@
...
+  do i = 1, size(keys)
+    print '("Value of key ", I0, " = ", A)', i, key_to_char(keys(i))
+  end do
+  !Value of key 1 = initial key
+  !Value of key 2 = second key
+  !Value of key 3 = last key
+
+contains
+  !Converts key type to character type
+  function key_to_char(key) result(str)
```
jvdp1, line 44: Could this function a `pure` function?

jvdp1 (commented):
Thank you @degawa for this PR. LGTM! I have only a few minor suggestions.

example/hashmaps/example_hashmaps_get_all_keys.f90
```diff
@@ -0,0 +1,53 @@
+program example_set_other_data
```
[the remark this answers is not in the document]
degawa: Thanks for your correction.
The program name `example_set_other_data` at the `end program` must also be corrected to `example_hashmaps_get_all_keys` , so I will fix it along with the other corrections.

example/hashmaps/example_hashmaps_get_all_keys.f90
degawa: Yes, this function can be `pure` function. I will add pure attribute to it.

doc/specs/stdlib_hashmaps.md
degawa: I use "presented in" to mean "located in," "existing in," or "contained in" here.
I cannot answer whether "present" or "presented" is correct due to my poor English skills, so I need help to create a natural expression, especially a phrase referring to "keys in a hash map."

jvdp1: @milancurcic Any idea how it should be written?

degawa: @jvdp1 
The following expression is used in the [documentation](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html) of Map in Java:
> keySet()
>Returns a Set view of the keys contained in this map.

So, I would like to change "presented" to "contained" to follow this expression. I would appreciate your comments.
Thank you.

jvdp1: @degawa "contained" sounds good to me. Please, go ahead!

degawa: Thank you, @jvdp1, for giving me a supportive push forward.
I fixed the documentation for the `get_all_keys` and pushed it.
I would appreciate a review.

jvdp1 (approved):
LGTM. Thank you for this PR @degawa !

PR #742, opened 2023-10-29
Fypp call update for windows compatability

chuckyvt:
This is a small change I made to the fypp call in CMake so that it can run fypp and build on Windows.  If CMake can't find fypp the location can be specified via '-DCMAKE_INCLUDE_PATH" variable when configuring project.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,9 @@ find_package(Python3 3.5 REQUIRED COMPONENTS Interpreter)
 
 # Find fypp file
 find_file(FYPP fypp REQUIRED)
+if(NOT FYPP)
+  message(FATAL_ERROR "Preprocessor fypp not found! Please install fypp following the instructions in https://fypp.readthedocs.io/en/stable/fypp.html#installing.")
+endif()
 
 # Cmd line string to run fypp
 SET(FYPP ${Python3_EXECUTABLE} ${FYPP})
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -42,12 +42,18 @@ if(NOT DEFINED CMAKE_MAXIMUM_RANK)
   set(CMAKE_MAXIMUM_RANK 4 CACHE STRING "Maximum array rank for generated procedures")
 endif()
 
-# --- find preprocessor
-find_program(FYPP fypp)
+# Find python.  Python 3.5 or higher required for ftypp
```
jvdp1, line 45:
```suggestion
# Find python.  Python 3.5 or higher required for fypp
```

jvdp1 (commented):
Thank you for adding this check. However, it breaks the CI. Any idea why?

CMakeLists.txt
chuckyvt: CMake wasn't find fypp.   I had assumed (incorrectly) that the find_file function would find the fypp package if it had been installed.   Doing some trials, its seem it will be difficult to reliably find fypp across the various installation options, without the user manually specifying the path via -DCMAKE_INCLUDE_PATH or similar.   

Based on that, I think its worth considering including fypp in the stdlib package.  Its a single py file, so relatively compact.  and the fypp license allows it.   The pull request has been updated to reflect this.   

Understand if don't want to add, in that case I can close the pull request.

jvdp1:
@awvwgk any ideas on how to solve this issue? Maybe a solution like for `test-drive` (with CMake files)?

I am not in favour of adding the fypp file into the stdlib project.

awvwgk (commented):
I would prefer to download fypp from the repo or pypi rather than committing it to our repository.

PR #743, opened 2023-11-01
Proposal for a reference string to number conversion facility in stdlib

jalvesz:
The following PR contains an adaptation of this project https://github.com/jalvesz/Fortran-String-to-Num which was developed after the discussions about how to do fast string to double conversion here https://fortran-lang.discourse.group/t/faster-string-to-double/2208/85

I have provided unit tests for string to real32 and real64, as well as a short example.

Might need some help/guidance to finalize a proper automatic documentation, and totally open to change the naming convention if needed.


klausler:
I ran an exhaustive test over the full range of 32-bit IEEE-754 numbers and observed many discrepancies in the real values read by `str2float` vs. the values read via internal input.

jalvesz:
Thanks for pointing it out and the test @klausler, I saw the problem with the small values!!

So, with the current strategy, this value for instance:
0.140129846432481707092372958328991613128026194187651577175706828388979108268586060148663818836212158203125E-44

Would be read first with an integer containing:
int_wp = 14012984
the variable resp = 9, contains the decimal places that should be applied first to get 0.14012984 and the exponent integer
sige*max(0,i_exp) = -44
But, since the total exponent is computed from
`expbase(nwnb-1+resp-sige*max(0,i_exp))`
which can not hold 1e-51 that gives an underflow the strategy doesn't work properly here...

Thinking how to make sure to manage those limit cases without degrading the performance

jalvesz:
For the edge cases, something like this could work:
```Fortran
exp_aux = nwnb-1+resp-sige*max(0,i_exp)
if( exp_aux>0 .and. exp_aux<=nwnb+nfnb) then
         v = sign*int_wp*expbase(exp_aux)
 else if(exp_aux>nwnb+nfnb) then
         v = sign*int_wp*fractional_base(exp_aux-(nwnb+nfnb))*expbase(nwnb+nfnb)
 end if
```
Tried on
"0.462428493227189633404830762485672323322486440819250204679832533683631057286333998490590602159500122070312E-43"

With ifort and ifx the error is below acceptable tolerance, with gfortran I get:
formatted read : 0.462428493E-43
str2real       : 0.448415509E-43

klausler:
> For the edge cases, something like this could work:
> 
> ```fortran
> exp_aux = nwnb-1+resp-sige*max(0,i_exp)
> if( exp_aux>0 .and. exp_aux<=nwnb+nfnb ) then
>          v = sign*int_wp*expbase(exp_aux)
>  else if(exp_aux>nwnb+nfnb) then
>          v = sign*int_wp*fractional_base(exp_aux-(nwnb+nfnb))*expbase(nwnb+nfnb)
>  end if
> ```
> 
> Tried on "0.462428493227189633404830762485672323322486440819250204679832533683631057286333998490590602159500122070312E-43"
> 
> With ifort and ifx the error is below acceptable tolerance, with gfortran I get: formatted read : 0.462428493E-43 str2real : 0.448415509E-43

This test is reading the results of binary->decimal conversions that have enough decimal digits to be exact.  The values read back must match the original binary values exactly; there is no "acceptable tolerance".

jalvesz:
Well, the problem being that even this is not exact
```Fortran
use iso_fortran_env
real :: x
x = 10._real32**(-38)
print *, x
```
gfortran : 9.99999935E-39
ifx : 9.9999994E-39

and it get worst with 
```Fortran
x = 10._real32**(-44)
```
gfortran : 9.80908925E-45
ifx : 9.8090893E-45

Anyhow, fixed the problem. It was coming from the table of exponents which did not accurately defined the 10 basis. Changed for the following:
```Fortran
integer(kind=ikind), parameter :: nwnb = 39 !> number of whole number factors
integer(kind=ikind), parameter :: nfnb = 37 !> number of fractional number factors
integer :: e
real(wp), parameter :: whole_number_base(nwnb)   = [(10._wp**(nwnb-e),e=1,nwnb)]
real(wp), parameter :: fractional_base(nfnb)   = [(10._wp**(-e),e=1,nfnb)]
real(wp), parameter :: expbase(nwnb+nfnb) = [whole_number_base, fractional_base]
```
Now I get:

gfortran
input          : "0.462428493227189633404830762485672323322486440819250204679832533683631057286333998490590602159500122070312E-43"
formatted read : 0.462428493E-43
str2real              : 0.462428493E-43

ifort
input          : "0.462428493227189633404830762485672323322486440819250204679832533683631057286333998490590602159500122070312E-43"
formatted read : .4624285E-43
str2real             : .4624285E-43

klausler:
If you're using negative powers of ten in your algorithm, it's never going to be exact.

jalvesz:
I know, but I haven't found a better approach to be fast and accurate. In any case you can see in the last results that both the formatted read and the proposed algorithm give the same output to the last decimal place.

Previously I was defining the exponents table with "1eN",  this thread https://stackoverflow.com/questions/47337262/defining-exponent-in-fortran hinted me to using "10._wp**N" which did indeed improve the precision.

I'll update when finishing more tests.

jalvesz:
Changed the interface according to the discussions in the forum asking for something like "r = to_num( s , mold = r )" (the mold argument is there just to disambiguate the function call).

@klausler I'm rerunning your tests loosening the error check by
```Fortran
if (abs(rval - rcheck)<=10*epsilon(0.0) * abs(rcheck)) then
```
It has been running for quite a while on 16threads and so far no error messages, don't know when it will finish, I'll update fully then... a relative error to 10*epsilon(0._wp) is already quite good I would say. If there is a different method that would enable both speed and better accuracy I would gladly try to adapt it.

klausler:
How much faster is this inexact algorithm than an internal `READ` statement with a good Fortran compiler?

jalvesz:
Between 30x to 70x depending on the hardware and compiler (using -O3 -march=native)

klausler:
Using a variant of my test from above, sweeping over all the ca. 4B valid 32-bit IEEE-754 numbers, I measured that it takes 4795 total CPU seconds to just do all the internal writes, 2107 additional CPU seconds to also do all the internal reads, or 526 additional CPU seconds to also run `str2num` instead of using internal reads.  That's a speed-up of 4x, which is not 70x but would still be great if the results were the same.  (This is with f18 ("llvm flang"), which has pretty fast binary<->decimal conversions.)

jalvesz:
Just saw the tests were finish, the only errors prompted were 
```console
str2float failure: rval 7F800000 rcheck        0
 want                                                                                                         Infinity
  got  0.000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000E+00
str2float failure: rval FF800000 rcheck        0
 want                                                                                                        -Infinity
  got  0.000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000E+00
           4278190080  total str2float passes out of            4278190082  tests
                    0  total str2float failures out of            4278190082  tests
```

I did get those speed ups with str2double (now just to_num) ... I'll double check speedups for real32

One idea for coming across the precision issue would be to store the exponents in double precision. As the multiplication is done just once at the end, it should no add much penalty. I'll try that out.

jalvesz:
Mixing the precisions gets the error down to epsilon(0.0) for real32

I'm kind of puzzled by the NaN assignment. I did define a parameter `real(wp), parameter :: rNaN = real(z'7fc00000',wp)` which is assigned if "N" if found
```Fortran
else if( iachar(s(p:p)) == NaN ) then
            v = rNaN; return
```
But a simple print returns 0.00000000 instead (same for Inf) ... any clues ?

jalvesz:
@klausler reran your tests with the latest version and got the error down to zero. No prompts other than the Infinity and -Infinity with the condition on `if (rval == rcheck) then` ... have yet to understand why the parameters are not being correctly assigned. Would you mind cross-validating? With the new interface just change
```Fortran
use stdlib_str2num, only: to_num
...
rcheck = to_num(buff,rcheck)
```

jvdp1:
> Mixing the precisions gets the error down to epsilon(0.0) for real32
> 
> I'm kind of puzzled by the NaN assignment. I did define a parameter `real(wp), parameter :: rNaN = real(z'7fc00000',wp)` which is assigned if "N" if found
> 
> ```fortran
> else if( iachar(s(p:p)) == NaN ) then
>             v = rNaN; return
> ```
> 
> But a simple print returns 0.00000000 instead (same for Inf) ... any clues ?

If possible I suggest to use the ones provided by `ieee_arithmetic` as in other `stdlib` module (e.g. in the [stats mean module](https://github.com/fortran-lang/stdlib/blob/b8fbb3ce1aaf2fd1729897739fcdae8f6b3c4ef4/src/stdlib_stats_mean.fypp#L6) )

jalvesz:
@jvdp1 by any chance do you know why there are those two fails in the build? when I look at the logs, they do not seem to be related to the sources.

jvdp1:
> @jvdp1 by any chance do you know why there are those two fails in the build? when I look at the logs, they do not seem to be related to the current PR.

No idea. But it seems unrelated to this PR.

jalvesz:
@klausler after verifying that your tests runs successfully without any errors with strict equivalence, I adapted it to check for the speedups. I saw values oscillating between 6x and 13x stabilizing around 8x for the real32 reader (to_num) compared with the formatted read.
Compiler: gfortran 13.2
Options: -O3 -march=native
OS: WSL Ubuntu 20.04
CPU: Intel Xeon Gold @ 2.9GHz

jalvesz:
@jvdp1 regarding this:
>If possible I suggest to use the ones provided by ieee_arithmetic as in other stdlib module (e.g. in the [stats mean module](https://github.com/fortran-lang/stdlib/blob/b8fbb3ce1aaf2fd1729897739fcdae8f6b3c4ef4/src/stdlib_stats_mean.fypp#L6) )

I have just one remark:
Defining a parameter variable should in principle offer better optimization compared to having to call the function `ieee_value(1._wp, ieee_positive_inf)` which is not an intrinsic function. Thus one cannot do something like:
`real(wp), parameter :: Inf = ieee_value(1._wp, ieee_positive_inf)` (at least not with gfortran or ifort). One can only evaluate within the implementation. Then, I'm not sure how this would behave with all the compilers out there, and if it is an actual concern?

jvdp1:
> I have just one remark: Defining a parameter variable should in principle offer better optimization compared to having to call the function `ieee_value(1._wp, ieee_positive_inf)` which is not an intrinsic function. Thus one cannot do something like: `real(wp), parameter :: Inf = ieee_value(1._wp, ieee_positive_inf)` (at least not with gfortran or ifort). One can only evaluate within the implementation. Then, I'm not sure how this would behave with all the compilers out there, and if it is an actual concern?

There have been many discussions about NaN and Inf, see e.g., #128 . As I remember the main issue was portability.

src/stdlib_str2num.f90
```diff
@@ -0,0 +1,532 @@
...
+
+    elemental subroutine to_int_32(s,v,p,stat)
+        !> Return an unsigned 32-bit integer
+        ! -- In/out Variables
+        character(*), intent(in) :: s !> input string
+        integer, intent(inout)  :: v !> Output real value
+        integer(1), intent(out)  :: p !> position within the number
+        integer(1), intent(out)  :: stat !> status upon succes or failure to read
+        ! -- Internal Variables
+        integer(1)  :: val
```
jvdp1, line 196:
```suggestion
        integer(int8), intent(out)  :: p !> position within the number
        integer(int8), intent(out)  :: stat !> status upon succes or failure to read
        ! -- Internal Variables
        integer(int8)  :: val 
```
For portability across compilers

src/stdlib_str2num.f90
```diff
@@ -0,0 +1,532 @@
...
+    end subroutine
+
+    elemental subroutine to_real_sp(s,v,p,stat)
+        integer, parameter :: wp    = sp
+        !> Sequentially unroll the character and get the sub integers composing the whole number, fraction and exponent
+        ! -- In/out Variables
+        character(*), intent(in) :: s !> input string
+        real(wp), intent(inout)  :: v !> Output real value
+        integer(1), intent(out)  :: p !> last position within the string
+        integer(1), intent(out)  :: stat !> status upon success or failure to read
```
jvdp1, line 222:
```suggestion
        integer(int8), intent(out)  :: p !> last position within the string
        integer(int8), intent(out)  :: stat !> status upon success or failure to read
```

src/stdlib_str2num.f90
```diff
@@ -0,0 +1,532 @@
...
+
+        ! -- Internal Variables
+        integer(kind=ikind), parameter :: nwnb = 39 !> number of whole number factors
+        integer(kind=ikind), parameter :: nfnb = 37 !> number of fractional number factors
+        integer :: e
+        real(dp), parameter :: whole_number_base(nwnb)   = [(10._dp**(nwnb-e),e=1,nwnb)]
+        real(dp), parameter :: fractional_base(nfnb)   = [(10._dp**(-e),e=1,nfnb)]
+        real(dp), parameter :: expbase(nwnb+nfnb) = [whole_number_base, fractional_base]
+
+        integer(1)  :: sign, sige !> sign of integer number and exponential
```
jvdp1, line 232:
```suggestion
        integer(int8)  :: sign, sige !> sign of integer number and exponential
```

src/stdlib_str2num.f90
```diff
@@ -0,0 +1,532 @@
...
+        integer, parameter :: maxdpt = 11 !> Maximum depth to read values on int_wp
+        integer(dp) :: int_wp !> long integer to capture fractional part
+        integer     :: i_exp !> integer to capture whole number part
+        integer     :: exp_aux
+        integer(1)  :: i, pP, pE, val , resp
```
jvdp1, line 237:
```suggestion
        integer(int8)  :: i, pP, pE, val , resp
```

src/stdlib_str2num.f90
```diff
@@ -0,0 +1,532 @@
...
+    end subroutine
+
+    elemental subroutine to_real_dp(s,v,p,stat)
+        integer, parameter :: wp    = dp
+        !> Sequentially unroll the character and get the sub integers composing the whole number, fraction and exponent
+        ! -- In/out Variables
+        character(*), intent(in) :: s !> input string
+        real(wp), intent(inout)  :: v !> Output real value
+        integer(1), intent(out)  :: p !> last position within the string
+        integer(1), intent(out)  :: stat !> status upon success or failure to read
```
jvdp1, line 316:
```suggestion
        integer(int8), intent(out)  :: p !> last position within the string
        integer(int8), intent(out)  :: stat !> status upon success or failure to read
```

src/stdlib_str2num.f90
```diff
@@ -0,0 +1,532 @@
...
+
+        ! -- Internal Variables
+        integer(kind=ikind), parameter :: nwnb = 40 !> number of whole number factors
+        integer(kind=ikind), parameter :: nfnb = 64 !> number of fractional number factors
+        integer :: e
+        real(wp), parameter :: whole_number_base(nwnb)   = [(10._wp**(nwnb-e),e=1,nwnb)]
+        real(wp), parameter :: fractional_base(nfnb)   = [(10._wp**(-e),e=1,nfnb)]
+        real(wp), parameter :: expbase(nwnb+nfnb) = [whole_number_base, fractional_base]
+
+        integer(1)  :: sign, sige !> sign of integer number and exponential
```
jvdp1, line 326:
```suggestion
        integer(int8)  :: sign, sige !> sign of integer number and exponential
```

src/stdlib_str2num.f90
```diff
@@ -0,0 +1,532 @@
...
+        integer, parameter :: maxdpt = 19 !> Maximum depth to read values on int_wp
+        integer(wp) :: int_wp !> long integer to capture fractional part
+        integer     :: i_exp !> integer to capture whole number part
+        integer     :: exp_aux
+        integer(1)  :: i, pP, pE, val , resp
```
jvdp1, line 331:
```suggestion
        integer(int8)  :: i, pP, pE, val , resp
```

src/stdlib_str2num.f90
```diff
@@ -0,0 +1,532 @@
...
+
+    subroutine to_real_qp(s,v,p,stat)
+        integer, parameter :: wp    = qp
+        !> Sequentially unroll the character and get the sub integers composing the whole number, fraction and exponent
+        ! -- In/out Variables
+        character(*), intent(in) :: s !> input string
+        real(wp), intent(inout)  :: v !> Output real value
+        integer(1), intent(out)  :: p !> last position within the string
+        integer(1), intent(out)  :: stat !> status upon success or failure to read
+
```
jvdp1, line 411:
```suggestion
        integer(int8), intent(out)  :: p !> last position within the string
        integer(int8), intent(out)  :: stat !> status upon success or failure to read

```

src/stdlib_str2num.f90
```diff
@@ -0,0 +1,532 @@
...
+        ! -- Internal Variables
+        integer(kind=ikind), parameter :: nwnb = 50 !> number of whole number factors
+        integer(kind=ikind), parameter :: nfnb = 64 !> number of fractional number factors
+        integer :: e
+        real(wp), parameter :: whole_number_base(nwnb)   = [(10._wp**(nwnb-e),e=1,nwnb)]
+        real(wp), parameter :: fractional_base(nfnb)   = [(10._wp**(-e),e=1,nfnb)]
+        real(wp), parameter :: expbase(nwnb+nfnb) = [whole_number_base, fractional_base]
+
+        integer(1)  :: sign, sige !> sign of integer number and exponential
```
jvdp1, line 420:
```suggestion
        integer(int8)  :: sign, sige !> sign of integer number and exponential
```

src/stdlib_str2num.f90
```diff
@@ -0,0 +1,532 @@
...
+        integer, parameter :: maxdpt = 19 !> Maximum depth to read values on int_dp
+        integer(dp) :: int_dp1, int_dp2 !> long integers to capture whole and fractional part
+        integer     :: i_exp !> integer to capture exponent number
+        integer     :: exp_aux
+        integer(1)  :: i, pP, pE, val , resp, icount, aux
```
jvdp1, line 425:
```suggestion
        integer(int8)  :: i, pP, pE, val , resp, icount, aux
```

src/stdlib_str2num.f90
```diff
@@ -0,0 +1,532 @@
...
+    end subroutine
+    
+    !---------------------------------------------
+    ! Internal Utility functions
+    !---------------------------------------------
+    
+    elemental function mvs2nwsp(s) result(p)
+        !> move string to position of the next non white space character
+        character(*),intent(in) :: s !> character chain
+        integer(1) :: p !> position
```
jvdp1, line 513:
```suggestion
        integer(int8) :: p !> position
```

doc/specs/stdlib_str2num.md
```diff
@@ -0,0 +1,105 @@
...
+### Return value
+
+Return a scalar of numerical type (`integer`, `real32` or `real64`).
+
+### Example
+
+```fortran
+program example_string_to_number
+  use stdlib_kinds, only: dp
+  use stdlib_str2num
```
jvdp1, line 40:
```suggestion
  use stdlib_str2num, only: to_num
```

doc/specs/stdlib_str2num.md
```diff
@@ -0,0 +1,105 @@
...
+  implicit none
+  character(:), allocatable :: txt
+  real(dp) :: x
+
+  txt = ' 8.8541878128e−12 '
+  x = to_num( txt , x )
+end program example_random_seed
```
jvdp1, line 47:
```suggestion
end program example_string_to_number
```

doc/specs/stdlib_str2num.md
```diff
@@ -0,0 +1,105 @@
...
+### Return value
+
+Return a scalar of numerical type (`integer`, `real32` or `real64`).
+
+### Example
+
+```fortran
+program example_str2num
+    use stdlib_kinds, only: dp
+    use stdlib_str2num
```
jvdp1, line 79:
```suggestion
    use stdlib_str2num, only: to_num_p
```

doc/specs/stdlib_str2num.md
```diff
@@ -0,0 +1,105 @@
...
+
+Convert a stream of values in a string to an array of values.
+
+### Syntax
+
+`number = [[stdlib_str2num(module):to_num_p(interface)]](string, mold)`
+
+### Arguments
+
+`string`: argument has `intent(in)` and is of type `character(*), pointer`.
```
jvdp1, line 66:
```suggestion
`string`: argument has `intent(in)` and is of type `character(:), pointer`.
```

doc/specs/stdlib_str2num.md
```diff
@@ -0,0 +1,105 @@
...
+`mold`: argument has `intent(in)` and is of numerical type. currently: `integer`, `real32`, `real64` or `real128`. **Note**: the mold argument is included to help compilers chose the correct implementation at compile-time. Currently, compilers are not able to disambiguate functions with respect to the left-hand-side of an assignment.
+
+### Return value
+
+Return a scalar of numerical type (`integer`, `real32` or `real64`).
+
+### Example
+
+```fortran
+program example_str2num
```
jvdp1, line 77:
```suggestion
program example_to_num_p
```

doc/specs/stdlib_str2num.md
```diff
@@ -0,0 +1,105 @@
...
+    real(dp), allocatable :: r(:)
+    integer :: i 
+
+    chain = " 1.234   1.E1 1e0     0.1234E0  12.21e+001 -34.5E1"
+    allocate( r(6) )
+    cptr => chain
+    do i =1, 6
+        r(i) = to_num_p( cptr , r(i) ) !> the cptr pointer is shifted within the function
+    end do
+end program
```
jvdp1, line 91:
```suggestion
end program example_to_num_p
```

doc/specs/stdlib_str2num.md
```diff
@@ -0,0 +1,105 @@
...
+program example_str2num
+    use stdlib_kinds, only: dp
+    use stdlib_str2num
+    character(:), allocatable, target :: chain
+    character(len=:), pointer :: cptr
+    real(dp), allocatable :: r(:)
+    integer :: i 
+
+    chain = " 1.234   1.E1 1e0     0.1234E0  12.21e+001 -34.5E1"
+    allocate( r(6) )
```
jvdp1, line 86:
```suggestion
    real(dp) :: r(6)
    integer :: i 

    chain = " 1.234   1.E1 1e0     0.1234E0  12.21e+001 -34.5E1"
```

doc/specs/stdlib_str2num.md
```diff
@@ -0,0 +1,105 @@
...
+    cptr => chain
+    do i =1, 6
```
jvdp1, line 88:
```suggestion
    do i = 1, 6
```

doc/specs/stdlib_str2num.md
```diff
@@ -0,0 +1,105 @@
...
+        r(i) = to_num_p( cptr , r(i) ) !> the cptr pointer is shifted within the function
```
jvdp1, line 89: Why do you use a `c_ptr`, and just provide the `character(*)` as an argument (and potentially use a pointer inside the function? This would facilitate its use by the user.

doc/specs/stdlib_str2num.md
```diff
@@ -0,0 +1,105 @@
...
+
+Experimental
+
+### Description
+
+Convert a stream of values in a string to an array of values.
+
+### Syntax
+
+`number = [[stdlib_str2num(module):to_num_p(interface)]](string, mold)`
```
jvdp1, line 62: Suggestion:

transform it to a subrouine and hide the pointer and the loop inside the subroutine:
```fortran
call to_num_p(string, array)
...
subroutine to_num_p(string, array)
character(*), intent(in), target :: string
real(dp), intent(out) :: array(:)
....
cptr => string

do i = 1, size(array)
...
enddo

end subroutine
```

doc/specs/stdlib_str2num.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Arguments
+
+`string`: argument has `intent(in)` and is of type `character(*), pointer`.
+
+`mold`: argument has `intent(in)` and is of numerical type. currently: `integer`, `real32`, `real64` or `real128`. **Note**: the mold argument is included to help compilers chose the correct implementation at compile-time. Currently, compilers are not able to disambiguate functions with respect to the left-hand-side of an assignment.
```
jvdp1, line 68: What about  `integer(int8)`, `integer(int64)`?

doc/specs/stdlib_str2num.md
```diff
@@ -0,0 +1,105 @@
...
+    allocate( r(6) )
+    cptr => chain
+    do i =1, 6
+        r(i) = to_num_p( cptr , r(i) ) !> the cptr pointer is shifted within the function
+    end do
+end program
+```
+
+## Note
+The accuracy of the conversion is implementation dependent; it is recommend that implementers guarantee precision down to the last 3 bits.
```
jvdp1, line 95:
```suggestion
The accuracy of the conversion is implementation dependent; it is recommended that implementers guarantee precision down to the last 3 bits.
```

How can the implementers guarantee that?

src/stdlib_str2num.f90
```diff
@@ -0,0 +1,532 @@
...
+
+    interface to_num_p
+        module procedure to_int_p
+        module procedure to_float_p
+        module procedure to_double_p
+        module procedure to_quad_p
+    end interface
+
+    interface to_num_base
+        module procedure to_int_32
```
jvdp1, line 60: Could it be extended to `int8`, and others?

src/stdlib_str2num.f90
```diff
@@ -0,0 +1,532 @@
...
+
+    !---------------------------------------------
+    ! String To Number Implementations
+    !---------------------------------------------
+
+    elemental subroutine to_int_32(s,v,p,stat)
+        !> Return an unsigned 32-bit integer
+        ! -- In/out Variables
+        character(*), intent(in) :: s !> input string
+        integer, intent(inout)  :: v !> Output real value
```
jvdp1, line 192:
```suggestion
        integer(int32), intent(inout)  :: v !> Output real value
```

Would the same algorithm be used for e.g., `int8` or `int64`? If yes, then it could be easily extended with `fypp`.

src/stdlib_str2num.f90
```diff
@@ -0,0 +1,532 @@
...
+        ! -- In/out Variables
+        character(*), intent(in) :: s !> input string
+        real(wp), intent(inout)  :: v !> Output real value
+        integer(1), intent(out)  :: p !> last position within the string
+        integer(1), intent(out)  :: stat !> status upon success or failure to read
+
+        ! -- Internal Variables
+        integer(kind=ikind), parameter :: nwnb = 40 !> number of whole number factors
+        integer(kind=ikind), parameter :: nfnb = 64 !> number of fractional number factors
+        integer :: e
```
jvdp1, line 321: For all these internal implementations, I always prefer to be explicit. For example, if the user compiles the library with `int64` as default (instead of `int32`, would the implementation still work?

src/stdlib_str2num.f90
```diff
@@ -0,0 +1,532 @@
...
+        p = 1
+        do while( p<len(s) .and. (iachar(s(p:p))==WS.or.iachar(s(p:p))==LF.or.iachar(s(p:p))==CR) ) 
+            p = p + 1
+        end do
+    end function
+    
+    elemental function mvs2wsp(s) result(p)
+        !> move string to position of the next white space character
+        character(*),intent(in) :: s !> character chain
+        integer(1) :: p !> position
```
jvdp1, line 524:
```suggestion
        integer(int8) :: p !> position
```

test/string/test_string_to_number.f90
```diff
@@ -0,0 +1,343 @@
...
+                write(*,"('to_num         : ' g0)") to_num_out
+                write(*,"('difference abs : ' g0)") abs_err
+                write(*,"('difference rel : ' g0 '%')") rel_err * 100
+                ucheck = .false.
+            end if
+        end function
+    end subroutine
+
+    subroutine test_to_quadruple(error)
+        use stdlib_str2num
```
jvdp1, line 219:
```suggestion
```
It could be moved to the top of the module, and be called only once.

test/string/test_string_to_number.f90
```diff
@@ -0,0 +1,343 @@
...
+                write(*,"('formatted read : ' g0)") formatted_read_out
+                write(*,"('to_num         : ' g0)") to_num_out
+                write(*,"('difference abs : ' g0)") abs_err
+                write(*,"('difference rel : ' g0 '%')") rel_err * 100
+                ucheck = .false.
+            end if
+        end function
+    end subroutine
+
+    subroutine test_to_quadruple(error)
```
jvdp1, line 218: It seems to be 3 copy-paste. If so, then a `fypp` loop could be implemented.

jvdp1 (commented):
Thank you @jalvesz, and sorry for the delay of my review.

Here are some suggestions to improve the portability across compilers and potentially its use.

src/stdlib_str2num.f90
```diff
@@ -0,0 +1,532 @@
...
+        real(qp) :: r    !> Output real value
+        integer(1),intent(inout), optional :: stat
+        ! -- Internal Variables
+        integer(1) :: p !> position within the number
+        integer(1) :: err
+        !----------------------------------------------
+        call to_num_base(s,r,p,err)
+        p = min( p , len(s) )
+        s => s(p:)
+        if(present(stat)) stat = err
```
jvdp1, line 181: Note: Not all compilers support `qp`. Therefore, this should be preprocessed with `fypp`.
Using `stdlib_kinds` insead of `iso_fortran_env` will help to get the correctkind.

src/stdlib_str2num.f90
```diff
@@ -0,0 +1,532 @@
...
+!> Conversion to quadruple precision can deviate at about 200*epsilon(0.0_qp)
+!>   example:
+!>   input          : 0.140129846432481707092372958328991613128026194187651577175706828388979108268586060148663818836212158203125E-443
+!>   formatted read : 0.140129846432481707092372958328991608E-443
+!>   to_num         : 0.140129846432481707092372958328996233E-443
+!>   difference abs :                                 0.4625E-475 
+!>   difference rel :                                 0.3300E-029%
+
+module stdlib_str2num
+    use iso_fortran_env, only: sp => real32, dp => real64, qp => real128
```
jvdp1, line 25: Would it be possible to use the ones defined in `stdlib_kinds` instead of redefining them here (mainly for portability)?

src/stdlib_str2num.f90
jalvesz: Yes it could, we could indeed extended with fypp

src/stdlib_str2num.f90
jalvesz: I see, you are right! For these parameters what I have seen is that it should be a number not too big in order to not saturate the stack. The values I suggested for the number of powers of ten to keep in the stack were found empirically, in any case I think that having ikind = int8 should work as the values are smaller than 127.

src/stdlib_str2num.f90
jalvesz: Which preprocessing macro would you suggest here? I know that nvfortran does not support qp, but gfortran and ifort/ifx do.

src/stdlib_str2num.f90
jalvesz: Yes sure, I used this because even though I launched the preprocessing for gfortran and intel I saw that qp was defined as -1 when doing my tests, so I used iso_fortran_env to ensure having it defined. This can be changed, But I guess the best option would be that the qp code is not written if the selected compiler does not support qp. Could you guide me on how to do that in this scope with fypp?

jalvesz:
@jvdp1 I saw that thes files test_*.fypp are not being preprocessed. Should a line like this be included in the .sh ?
```shell
# Preprocess stdlib sources
find test -maxdepth 2 -iname "*.fypp" \
  | cut -f1 -d. | xargs -P "$njob" -I{} "$fypp" "{}.fypp" "$destdir/{}.f90" $fyflags
```
after line 46

src/stdlib_str2num.f90
jalvesz: Solved it with "#:if WITH_QP"

doc/specs/stdlib_str2num.md
jalvesz: I wrote this after a discussion with @ivan-pi as a kind of recommendation, but it is difficult to provide a hard rule on how to guarantee it. This is more or less the current upper-bound of the error for qp. Maybe someone can come-up with a better implementation down the road to improve this :)

src/stdlib_str2num.f90
jvdp1: I open a [PR](https://github.com/jalvesz/stdlib/pull/1) in your project with a suggestion with `fypp`. don't hesitate to ask questions there or here.

src/stdlib_str2num.f90
jvdp1: See suggestions in [this PR](https://github.com/jalvesz/stdlib/pull/1)

jvdp1:
> @jvdp1 I saw that thes files test_*.fypp are not being preprocessed. Should a line like this be included in the .sh ?
> 
> ```shell
> # Preprocess stdlib sources
> find test -maxdepth 2 -iname "*.fypp" \
>   | cut -f1 -d. | xargs -P "$njob" -I{} "$fypp" "{}.fypp" "$destdir/{}.f90" $fyflags
> ```
> 
> after line 46 ... I tried that, saw there is an issue with one of the test_*.fypp files but I can't see exactly whichone as the WSL closes due to some early exit

Good point! I always use `CMake` but indeed, the `test_*.fypp` should be preprocessed too.
Could you open another PR with this suggestion? I will try it on my Fedora computer.

jvdp1:
> @jvdp1 I saw that thes files test_*.fypp are not being preprocessed. Should a line like this be included in the .sh ?
> 
> ```shell
> # Preprocess stdlib sources
> find test -maxdepth 2 -iname "*.fypp" \
>   | cut -f1 -d. | xargs -P "$njob" -I{} "$fypp" "{}.fypp" "$destdir/{}.f90" $fyflags
> ```
> 
> after line 46 ... I tried that, saw there is an issue with one of the test_*.fypp files but I can't see exactly whichone as the WSL closes due to some early exit

@jalvesz  Here is a solution (not the most elegant, but it works):
```bash
# Preprocess stdlib sources
find test -name "test_*.fypp" -exec cp {} "$destdir/test/" \;
fyflags="${fyflags} -I src"
find $destdir/test -maxdepth 1 -iname "*.fypp" \
  | cut -f1 -d. | xargs -P "$njob" -I{} "$fypp" "{}.fypp" "{}.f90" $fyflags
find $destdir/test -name "test_*.fypp" -exec rm {} \;
```

The issues are that:
 1) it cannot find the file `common.fypp`
 2) the file structure in the original directory and in the stdlib-fpm directory are different.

jalvesz:
Thanks @jvdp1, so tried with this, but I still have a problem with one of the test files, with a preprocessing option WITH_CBOOL (I think), I can't see exactly where as the WSL crashes and I can't catch exactly which file, do you know if there is a log somewhere?

jvdp1:
Ho, right. the following must be added on the top of the file
test/string/test_string_assignment.fypp:
```
#: include "common.fypp"
```

example/strings/example_str2num.f90
```diff
@@ -0,0 +1,22 @@
+program example_str2num
+    use stdlib_kinds, only: dp
+    use stdlib_str2num
```
jvdp1, line 3:
```suggestion
    use stdlib_str2num, only: to_num_p
```

test/string/test_string_to_number.fypp
```diff
@@ -0,0 +1,348 @@
...
+            new_unittest("to_float", test_to_float), &
+            new_unittest("to_double", test_to_double) &
+#:if WITH_QP
+            , new_unittest("to_quadruple", test_to_quadruple) &
+#:endif
+            ]
+    end subroutine collect_string_to_number
+
+    subroutine test_to_float(error)
+        use stdlib_str2num
```
jvdp1, line 24:
```suggestion
```

test/string/test_string_to_number.fypp
```diff
@@ -0,0 +1,348 @@
...
+                write(*,"('to_num         : ' g0)") to_num_out
+                write(*,"('difference abs : ' g0)") abs_err
+                write(*,"('difference rel : ' g0 '%')") rel_err * 100
+                ucheck = .false.
+            end if
+        end function
+    end subroutine
+
+    subroutine test_to_double(error)
+        use stdlib_str2num
```
jvdp1, line 123:
```suggestion
```

test/string/test_string_to_number.fypp
```diff
@@ -0,0 +1,348 @@
...
+                write(*,"('difference abs : ' g0)") abs_err
+                write(*,"('difference rel : ' g0 '%')") rel_err * 100
+                ucheck = .false.
+            end if
+        end function
+    end subroutine
+
+#:if WITH_QP
+    subroutine test_to_quadruple(error)
+        use stdlib_str2num
```
jvdp1, line 223:
```suggestion
```

src/stdlib_str2num.fypp
```diff
@@ -0,0 +1,553 @@
...
+    
+    function to_quad_p(s,mold,stat) result(r)
+        ! -- In/out Variables
+        character(len=:), pointer :: s !> input string
+        real(qp), intent(in) :: mold !> dummy argument to disambiguate at compile time the generic interface
+        real(qp) :: r    !> Output real value
+        integer(1),intent(inout), optional :: stat
+        ! -- Internal Variables
+        integer(1) :: p !> position within the number
+        integer(1) :: err
```
jvdp1, line 192:
```suggestion
        integer(int8),intent(inout), optional :: stat
        ! -- Internal Variables
        integer(int8) :: p !> position within the number
        integer(int8) :: err
```

doc/specs/stdlib_str2num.md
jalvesz: I've been giving some thought to your idea, on one side I think it does facilitate things, but at the same time I feel that it would introduce some particular cases to handle (multiple dimension arrays, stopping on some condition other than the end of the string, etc) providing the low level interface gives the user the freedom to adapt on top of it. What do you think?

doc/specs/stdlib_str2num.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Syntax
+
+`number = [[stdlib_str2num(module):to_num_p(interface)]](string, mold)`
+
+### Arguments
+
+`string`: argument has `intent(in)` and is of type `character(*), pointer`.
+
+`mold`: argument has `intent(in)` and is of numerical type. currently: `integer`, `real32`, `real64` or `real128`. **Note**: the mold argument is included to help compilers chose the correct implementation at compile-time. Currently, compilers are not able to disambiguate functions with respect to the left-hand-side of an assignment.
```
jvdp1, line 68:
```suggestion
`mold`: argument has `intent(in)` and is of numerical type. currently: `int8`, `int16`,`int32`, `int64`, `real32`, `real64` or `real128`. **Note**: the mold argument is included to help compilers chose the correct implementation at compile-time. Currently, compilers are not able to disambiguate functions with respect to the left-hand-side of an assignment.
```

doc/specs/stdlib_str2num.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Return value
+
+Return a scalar of numerical type (`integer`, `real32` or `real64`).
```
jvdp1, line 72:
```suggestion
Return a scalar of numerical type (`int8`, int16`, int32`, int64`, `real32`, `real64` or `real128`).
```
I wonder if we should use e.g., `sp` instead of `real32`.

jalvesz: Wonder the same thing. For the doc I think the most explicit would be the best.

src/stdlib_str2num.fypp
```diff
@@ -0,0 +1,553 @@
...
+!>   example:
+!>   input          : 0.140129846432481707092372958328991613128026194187651577175706828388979108268586060148663818836212158203125E-443
+!>   formatted read : 0.140129846432481707092372958328991608E-443
+!>   to_num         : 0.140129846432481707092372958328996233E-443
+!>   difference abs :                                 0.4625E-475 
+!>   difference rel :                                 0.3300E-029%
+
+module stdlib_str2num
+    use stdlib_kinds, only: sp, dp, qp, int8, int16, int32, int64
+    use ieee_arithmetic
```
jvdp1, line 27:
```suggestion
    use, intrinsic:: ieee_arithmetic, only: ieee_value, ieee_positive_inf, ieee_quiet_nan
```

src/stdlib_str2num.fypp
```diff
@@ -0,0 +1,553 @@
...
+        ${t1}$, intent(in) :: mold !> dummy argument to disambiguate at compile time the generic interface
+        ${t1}$ :: v !> Output integer 32 value
+        ! -- Internal Variables
+        integer(int8) :: p !> position within the number
+        integer(int8)  :: stat !> error status
+        !----------------------------------------------
+        call to_num_base(s,v,p,stat)
+    end function
+    
+    function to_${k1}$_p(s,mold,stat) result(v)
```
jvdp1, line 98: could this function be pure?

src/stdlib_str2num.fypp
```diff
@@ -0,0 +1,553 @@
...
+    #:endfor
+
+    elemental function to_float(s,mold) result(r)
+        ! -- In/out Variables
+        character(*), intent(in) :: s !> input string
+        real(sp), intent(in) :: mold !> dummy argument to disambiguate at compile time the generic interface
+        real(sp) :: r !> Output real value
+        ! -- Internal Variables
+        integer(1) :: p !> position within the number
+        integer(1) :: stat ! error status
```
jvdp1, line 122:
```suggestion
        integer(int8) :: p !> position within the number
        integer(int8) :: stat ! error status
```

src/stdlib_str2num.fypp
```diff
@@ -0,0 +1,553 @@
...
+    
+    function to_float_p(s,mold,stat) result(r)
+        ! -- In/out Variables
+        character(len=:), pointer :: s !> input string
+        real(sp), intent(in) :: mold !> dummy argument to disambiguate at compile time the generic interface
+        real(sp) :: r    !> Output real value
+        integer(1),intent(inout), optional :: stat
+        ! -- Internal Variables
+        integer(1) :: p !> position within the number
+        integer(1) :: err
```
jvdp1, line 135:
```suggestion
        integer(int8), intent(inout), optional :: stat
        ! -- Internal Variables
        integer(int8) :: p !> position within the number
        integer(int8) :: err
```

src/stdlib_str2num.fypp
```diff
@@ -0,0 +1,553 @@
...
+    end function
+
+    elemental function to_double(s,mold) result(r)
+        ! -- In/out Variables
+        character(*), intent(in) :: s !> input string
+        real(dp), intent(in) :: mold !> dummy argument to disambiguate at compile time the generic interface
+        real(dp) :: r    !> Output real value
+        ! -- Internal Variables
+        integer(1) :: p !> position within the number
+        integer(1) :: stat ! error status
```
jvdp1, line 150:
```suggestion
        integer(int8) :: p !> position within the number
        integer(int8) :: stat ! error status
```

src/stdlib_str2num.fypp
```diff
@@ -0,0 +1,553 @@
...
+    
+    function to_double_p(s,mold,stat) result(r)
+        ! -- In/out Variables
+        character(len=:), pointer :: s !> input string
+        real(dp), intent(in) :: mold !> dummy argument to disambiguate at compile time the generic interface
+        real(dp) :: r    !> Output real value
+        integer(1),intent(inout), optional :: stat
+        ! -- Internal Variables
+        integer(1) :: p !> position within the number
+        integer(1) :: err
```
jvdp1, line 163:
```suggestion
        integer(int8),intent(inout), optional :: stat
        ! -- Internal Variables
        integer(int8) :: p !> position within the number
        integer(int8) :: err
```

doc/specs/stdlib_str2num.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Syntax
+
+`number = [[stdlib_str2num(module):to_num(interface)]](string, mold)`
+
+### Arguments
+
+`string`: argument has `intent(in)` and is of type `character(*)`.
+
+`mold`: argument has `intent(in)` and is of numerical type. currently: `integer`, `real32`, `real64` or `real128`. **Note**: the mold argument is included to help compilers chose the correct implementation at compile-time. Currently, compilers are not able to disambiguate functions with respect to the left-hand-side of an assignment.
```
klausler, line 29: All interpretation of expressions in Fortran and most other programming languages is context-free.  This is not a temporary inability of compilers to perform an analysis, and there are far more possible contexts than being the entire right-hand side of an expression.

src/stdlib_str2num.fypp
```diff
@@ -0,0 +1,553 @@
...
+        real(sp), intent(in) :: mold !> dummy argument to disambiguate at compile time the generic interface
+        real(sp) :: r !> Output real value
+        ! -- Internal Variables
+        integer(1) :: p !> position within the number
+        integer(1) :: stat ! error status
+        !----------------------------------------------
+        call to_num_base(s,r,p,stat)
+    end function
+    
+    function to_float_p(s,mold,stat) result(r)
```
jvdp1, line 127: the functions `to_float`, `to_double` and `to_quad`, and similarly the `to_*_p` could be simplified to one single copy in a fypp preprocessor

src/stdlib_str2num.fypp
```diff
@@ -0,0 +1,553 @@
...
+        integer(int8), intent(out)  :: p !> last position within the string
+        integer(int8), intent(out)  :: stat !> status upon success or failure to read
+
+        ! -- Internal Variables
+        integer(kind=ikind), parameter :: nwnb = 39 !> number of whole number factors
+        integer(kind=ikind), parameter :: nfnb = 37 !> number of fractional number factors
+        integer :: e
+        real(dp), parameter :: whole_number_base(nwnb)   = [(10._dp**(nwnb-e),e=1,nwnb)]
+        real(dp), parameter :: fractional_base(nfnb)   = [(10._dp**(-e),e=1,nfnb)]
+        real(dp), parameter :: expbase(nwnb+nfnb) = [whole_number_base, fractional_base]
```
jvdp1, line 249: should these be `real(dp)` or `real(sp)`?

doc/specs/stdlib_str2num.md
jvdp1: Actually, because all types of `integer` and `real` can be now supported (through fypp preprocessing) the following should be sufficient (similarly as in other specs):
```suggestion
`mold`: argument has `intent(in)` and is of numerical type (that is of `integer` or of `real32`). **Note**: the mold argument is included to help compilers chose the correct implementation at compile-time. Currently, compilers are not able to disambiguate functions with respect to the left-hand-side of an assignment.
```

doc/specs/stdlib_str2num.md
jvdp1: See comment below: Shortly, the following could be written (similarly to other specs):
```suggestion
Return a scalar of numerical type (i.e., `integer` or `real`).
```

this will be not fypp/compiler-dependent

doc/specs/stdlib_str2num.md
jvdp1: Yes, it makes sense to provide the low-level interface only. Keep it as is for now

doc/specs/stdlib_str2num.md
jalvesz: Agreed, I felt some comment was needed in order to justify the requirement of a mold argument. What would you suggest as a modification to this comment?

jvdp1: Additionally, with the latest changes, kinds are defined in `stdlib_kinds` and not in `iso_fortran_env`, where `stdlib_kinds` defines its own `sp`, `dp` and `qp`.

klausler: "The type of the `mold` argument defines the type of the result."

doc/specs/stdlib_str2num.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Return value
+
+Return a scalar of numerical type (`integer`, `real32` or `real64`).
```
jvdp1, line 33:
```suggestion
Return a scalar of numerical type (i.e., `integer`, or `real`).
```

doc/specs/stdlib_str2num.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Syntax
+
+`number = [[stdlib_str2num(module):to_num_p(interface)]](string, mold)`
+
+### Arguments
+
+`string`: argument has `intent(in)` and is of type `character(*), pointer`.
+
+`mold`: argument has `intent(in)` and is of numerical type. currently: `int8`, `int16`,`int32`, `int64`, `real32`, `real64` or `real128`. **Note**: the mold argument is included to help compilers chose the correct implementation at compile-time. Currently, compilers are not able to disambiguate functions with respect to the left-hand-side of an assignment.
```
jvdp1, line 68:
```suggestion
`mold`: argument has `intent(in)` and is of numerical type (currently of `integer` or `real`). **Note**: the mold argument is included to help compilers chose the correct implementation at compile-time. Currently, compilers are not able to disambiguate functions with respect to the left-hand-side of an assignment.
```

doc/specs/stdlib_str2num.md
```diff
@@ -0,0 +1,105 @@
...
+## Note
+The accuracy of the conversion is implementation dependent; it is recommended that implementers guarantee precision down to the last 3 bits.
+
+**The current implementation has been tested to provide for** :
+
+`real32`  : exact match
+
+`real64`  : precision up-to epsilon(0.0_real64)
+
+`real128` : precision around 200*epsilon(0.0_real128)
```
jvdp1, line 103:
```suggestion
`sp`  : exact match

`dp`  : precision up-to epsilon(0.0_dp)

`qp` : precision around 200*epsilon(0.0_qp)
```

test/string/test_string_to_number.fypp
```diff
@@ -0,0 +1,345 @@
+module test_string_to_number
+    use stdlib_kinds, only: sp, dp, qp
+    use stdlib_str2num
```
jvdp1, line 3:
```suggestion
    use stdlib_str2num, only: to_num
```

src/stdlib_str2num.fypp
jalvesz: Using dp here is what enabled obtaining exact precision for sp, otherwise we'll comeback to the initial issue pointed out by @klausler, see this comment above for reference: https://github.com/fortran-lang/stdlib/pull/743#issuecomment-1791953430 for

jvdp1: Ok. Thank you. Therefore I suggest to add a comment to prevent a future change, e.g.:
```suggestion
        ! We use dp here to obtain exact precision for sp.
        ! Otherwise errors may appear in comparison to formatted read.
        ! See  https://github.com/fortran-lang/stdlib/pull/743#issuecomment-1791953430 for more details

        real(dp), parameter :: whole_number_base(nwnb)   = [(10._dp**(nwnb-e),e=1,nwnb)]
        real(dp), parameter :: fractional_base(nfnb)   = [(10._dp**(-e),e=1,nfnb)]
        real(dp), parameter :: expbase(nwnb+nfnb) = [whole_number_base, fractional_base]
```

jvdp1: Providing tests that would fail if `dp` is changed to `sp` would be useful too (if it is not the case yet).

src/stdlib_str2num.fypp
jalvesz: It can't because of this `s => s(p:)` where the pointer is shifted from within.

src/stdlib_str2num.fypp
jalvesz: Did a commit to reformat this, but now I have to include an implementation for xqp which I did not foresaw

src/stdlib_str2num.fypp
jalvesz: Added the comment in the last commit. 
The test for sp in that cause should fail as for sp the check is on a strict equality `if(abs(rel_err) > 0.0_wp)` while for dp is on a softer condition `if(abs(rel_err) > epsilon(0.0_wp))`

src/stdlib_str2num.fypp
jalvesz: By any chance do you know how could I remove xdp from this loop ? :
`#:for k1, t1 in (INT_KINDS_TYPES + REAL_KINDS_TYPES)`
I could also duplicate the implementation for qp, changing wp = xdp ... but for intel compilers, these two are the same, so this would create a conflict at the generic interface.

jalvesz: For the moment opted for avoiding pre-processing xdp with:
```Fortran
#:for k1, t1 in (INT_KINDS_TYPES + REAL_KINDS_TYPES)
#:if k1 != "xdp"
```

jvdp1: In the `config/CMakeLists.txt` there is a test to ensure that `xdp` is not used if `qp` is defined to true.

jvdp1: Note that the same "issue" appears for all `stdlib` modules that use such a `fypp` loop (e.g., in `stdlib_stats_mean.fypp`). So it should not be a problem.

jvdp1: > For the moment opted for avoiding pre-processing xdp with:
> 
> ```fortran
> #:for k1, t1 in (INT_KINDS_TYPES + REAL_KINDS_TYPES)
> #:if k1 != "xdp"
> ```

If you want to avoid `xdp` for all compilers, then I think it is the best option currently.

jalvesz: So from here
>In the config/CMakeLists.txt there is a test to ensure that xdp is not used if qp is defined to true.

I should understand that stdlib shall always be built with either xdp or qp, but never both, right? In that case another option would be to rename the qp implementation an "recycle it"

jvdp1: If the compiler supports both `xdp` and `qp`, then both versions of the procedure will be generated. However, they will have differnet names with the suffix `xdp` or `qp`.

I am not sure I understand well your issue. Is it because there is no implementation for `xdp`?

test/string/test_string_to_number.f90
jalvesz: Done

src/stdlib_str2num.fypp
jalvesz: My question is wether we should guarantee simultaneous support for both xdp and qp ? with gfortran these two have different identifiers (xdp = 10, qp=16) with intel both are (xdp = 16, qp=16). For the latter this creates an issue as for the interfaces it can not distinguish between both. (In the case WITH_QP and WITH_XDP are both set to True)

jvdp1: > My question is wether we should guarantee simultaneous support for both xdp and qp ? 

I am in favour to support both simulataneously, as for other procedures in `stdlib`.

> with gfortran these two have different identifiers (xdp = 10, qp=16) with intel both are (xdp = 16, qp=16). 

> For the latter this creates an issue as for the interfaces it can not distinguish between both. (In the case WITH_QP and WITH_XDP are both set to True)

there should be no issue with Intel when using `cmake`, because `WITH_XDP` is set to `false` in the config/CmakeList.txt. Or do I miss something?

jalvesz: OK, just added support for xdp based on the qp implementation.

jvdp1 (approved):
Here are some final? comments, mainly related to `FORD` and the generation of the specs.

ci/fpm-deployment.sh
```diff
@@ -9,7 +9,7 @@ destdir="${DESTDIR:-stdlib-fpm}"
 fypp="${FYPP:-$(which fypp)}"
 
 # Arguments for the fypp preprocessor
-fyflags="${FYFLAGS:--DMAXRANK=4}"
+fyflags="${FYFLAGS:--DMAXRANK=8}"
```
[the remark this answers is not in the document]
jalvesz: I had an issue with one of the tests which has up to 8 dimensions when testing. If I remember correctly is the file test_mean_f03.fypp that when processed creates tests up to dimension 8 for the mean function. Maybe is this file that needs fixing.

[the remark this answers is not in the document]
jvdp1: right, this file is only compiled if `CMAKE_MAXIMUM_RANK` > 7. 
A similar condition should be added in the `fpm-deployment.sh`, e.g. (first and last lines are additions):
```bash
# Arguments for the fypp preprocessor
maxrank=4
fyflags="${FYFLAGS:--DMAXRANK=$maxrank}"

# Number of parallel jobs for preprocessing
if [ $(uname) = "Darwin" ]; then
  njob="$(sysctl -n hw.ncpu)"
else
  njob="$(nproc)"
fi

# Additional files to include
include=(
  "ci/fpm.toml"
  "LICENSE"
  "VERSION"
)

# Files to remove from collection
prune=(
  "$destdir/test/test_always_fail.f90"
  "$destdir/test/test_always_skip.f90"
  "$destdir/test/test_hash_functions.f90"
  "$destdir/src/common.f90"
  "$destdir/src/f18estop.f90"
)

if [[ $maxrank < 8 ]]; then
  prune+=("$destdir/test/stats/test_mean_f03.fypp")
fi
```

Note: I did not test this suggestion.

doc/specs/stdlib_str2num.md
```diff
@@ -0,0 +1,104 @@
...
+program example_string_to_number
+  use stdlib_kinds, only: dp
+  use stdlib_str2num, only: to_num
+  implicit none
+  character(:), allocatable :: txt
+  real(dp) :: x
+
+  txt = ' 8.8541878128e−12 '
+  x = to_num( txt , x )
+end program example_string_to_number
```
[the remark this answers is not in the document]
jalvesz: Done in the last two commits: https://github.com/fortran-lang/stdlib/pull/743/commits/fbdf87033ccdd1825e9035344a9d8637a92fff73 and https://github.com/fortran-lang/stdlib/pull/743/commits/cfb95daf00807d11970e2e5fd334fc56459197fc

ci/fpm-deployment.sh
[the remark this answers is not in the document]
jalvesz: I'll commit your suggestion as this is unrelated to the current PR. But I think it should be looked at as indeed in that file there is a global variable `${t1}$ :: d8_${k1}$(2, 3, 4, 2, 3, 4, 2, 3) = reshape(d1_${k1}$, [2, 3, 4, 2, 3, 4, 2, 3], [${t1}$:: 3])` that causes problems if the MAXRANK=4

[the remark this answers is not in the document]
jalvesz: I'll test and come back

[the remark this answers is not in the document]
jvdp1: That is on purpose. The maximum rank before F03 was 7 and since F03 is 15. Some compilers do (did?) not support up to 15 ranks. Therefore, the compilation of this file depends on maximum rank requested by the user, and supported by the compiler.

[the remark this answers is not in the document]
jalvesz: Done: just had to change [[ $maxrank -lt 8 ]] for [ ${maxrank} -lt 8 ]

ci/fpm-deployment.sh
```diff
@@ -34,6 +35,10 @@ prune=(
   "$destdir/src/f18estop.f90"
 )
 
+if [ ${maxrank} -lt 8 ]; then
+  prune+=("$destdir/test/test_mean_f03.f90")
+fi
+
```
jvdp1, line 41: @jalvesz with #758 these lines should not be needed anymore. Could you check it, please (and remove them if needed)?

jalvesz: Great, it worked locally, I just pushed the rollback on these lines

jalvesz:
Is it possible to re-launch the build checks without a commit? I've seen a few times here that the macos-latest build fails after staying idle for hours. It doesn't seem to be related to the changes in the PR.

jvdp1:
> Is it possible to re-launch the build checks without a commit? I've seen a few times here that the macos-latest build fails after staying idle for hours. It doesn't seem to be related to the changes in the PR.

Done! Indeed, it seems related to the CI, instead of `stdlib` itself

jalvesz:
Hi, is there something else that should be done before the PR can be merged?

jvdp1:
IMO this PR sounds good and ready to be merged. But it would be good to have at least one additional reviewer (maybe @klausler @milancurcic @gnikit @awvwgk ?)

src/stdlib_str2num.fypp
```diff
@@ -0,0 +1,569 @@
...
+        ${t1}$, intent(in) :: mold !> dummy argument to disambiguate at compile time the generic interface
+        ${t1}$ :: v !> Output ${t1}$ value
+        ! -- Internal Variables
+        integer(int8) :: p !> position within the number
+        integer(int8)  :: stat !> error status
+        !----------------------------------------------
+        call to_num_base(s,v,p,stat)
+    end function
+    
+    function to_${k1}$_p(s,mold,stat) result(v)
```
milancurcic, line 91: Does `_p` here refer to pointer? If yes, `_ptr` or `_pointer` will be clearer for the user.

But "pointer" is also something implementation specific, rather than describing what the function does.

Should it be `to_num_from_stream` or similar?

src/stdlib_str2num.fypp
```diff
@@ -0,0 +1,569 @@
...
+        integer(int8) :: p !> position within the number
+        integer(int8) :: err
+        !----------------------------------------------
+        call to_num_base(s,v,p,err)
+        p = min( p , len(s) )
+        s => s(p:)
+        if(present(stat)) stat = err
+    end function
+
+    #:endfor
```
milancurcic, line 107: It would be nice to have an interface to a shallow wrapper that takes `string_type` as input. Better integration with the rest of stdlib.

src/stdlib_str2num.fypp
```diff
@@ -0,0 +1,569 @@
...
+    elemental function mvs2wsp(s) result(p)
+        !> move string to position of the next white space character
+        character(*),intent(in) :: s !> character chain
+        integer(int8) :: p !> position
+        !----------------------------------------------
+        p = 1
+        do while( p<len(s) .and. .not.(iachar(s(p:p))==WS.or.iachar(s(p:p))==LF.or.iachar(s(p:p))==CR) ) 
+            p = p + 1
+        end do
+    end function
```
milancurcic, line 567: I understand these are internal, but full names (e.g. `move_string_to_nonwhitespace` and `move_string_to_whitespace`) would be clearer in places where these are used.

milancurcic (approved):
Thanks for this PR, looks good. Only minor nit-picks in comments.

Not a big deal, but it would be nice if the style was consistent within the module, e.g. the whitespaces around operators, some are there, some aren't.

src/stdlib_str2num.fypp
jalvesz: Yes, the `_p` if related to the function taking as input a pointer to a character. This `_ptr` looks good,  this also `to_num_from_stream`, why not. I like to find the balance between explicit names and not-too-long but this is fine for me. If no objection from others, I'll change then for your last suggestion @milancurcic.

src/stdlib_str2num.fypp
jalvesz: Haven't used the string_type from stdlib so didn't think about that. Maybe after merging the PR we could take a look at that?

src/stdlib_str2num.fypp
jalvesz: Ok, I'll rename those. What about: `shift_to_nonwhite_space` and `shift_to_white_space` ?

milancurcic: I like it but white space, when referred to as a character in computer programming, is a single word "whitespace".

src/stdlib_str2num.fypp
milancurcic: Definitely, since it's an additional feature, can be a future PR.

src/stdlib_str2num.fypp
jalvesz: done

src/stdlib_str2num.fypp
jalvesz: done

jvdp1:
> I would be keen to merge this and make a minor stdlib release, unless others have any objections.

@gnikit I have no objections to that.

ivan-pi:
@jvdp1, I think you can go ahead and merge.

jvdp1:
thank you @jalvesz for  this PR, and all participants for useful discussions and reviews. As suggested I will merge it this PR.

jalvesz:
Thank you @jvdp1 @klausler @milancurcic @gnikit @ivan-pi for you excellent guidance and reviews! hope this PR will be useful for the Fortran community!

PR #753, opened 2023-12-25
Fix `bits()` in `bitset_type` before initialization

### LINKED ISSUE #746: `bits()` in `bitset_type` returned a non-zero value even though no value is set.
### Description

The following code, built with gfortran, caught an error because a type-bound procedure `bits()` returns a non-zero value.

```Fortran
use testdrive, only : error_type, check
use stdlib_bitsets, only: bitset_large
type(error_type), allocatable :: error

block
    type(bitset_large) :: set6
    call check(error,set6%bits(),0, &
    'set6 % bits() returned non-zero value '//&
    'even though set6 was not initialized.')
    if (allocated(error)) return
end block
```

### Expected Behaviour

I expected `bits()` to return 0 since no value is set for `set6`.

### Version of stdlib

b8fbb3ce1aaf2fd1729897739fcdae8f6b3c4ef4

### Platform and Architecture

Windows 10 22H2 64bit, gfortran 11.2 bundled with quickstart Fortran on Windows

### Additional Information

This error occurs when child types of `bitset_type`, `bitset_large` and `bitset_64`, are declared in a `block` construct, and the error is caused by the fact that the component `num_bits` of `bitset_type` is not initialized to `0` as the default value. 
```Fortran
    type, abstract :: bitset_type
!! version: experimental
!!
!! Parent type for bitset_64 and bitset_large ([Specification](../page/specs/stdlib_bitsets.html#the-stdlib_bitsets-derived-types))

        private
        integer(bits_kind) :: num_bits

    contains
```

The error can be fixed by setting the default value of `num_bits` to `0`.

jvdp1:
Fixes #746 

+ Addition of tests that trigger the bug with GFortran 11 and 12

Issue #746 reported by @degawa

### DIFF
--- a/src/stdlib_bitsets.fypp
+++ b/src/stdlib_bitsets.fypp
@@ -100,7 +100,7 @@ module stdlib_bitsets
 !! Parent type for bitset_64 and bitset_large ([Specification](../page/specs/stdlib_bitsets.html#the-stdlib_bitsets-derived-types))
 
         private
-        integer(bits_kind) :: num_bits
+        integer(bits_kind) :: num_bits = 0_bits_kind
 
     contains
 
--- a/src/stdlib_bitsets_large.fypp
+++ b/src/stdlib_bitsets_large.fypp
@@ -1051,7 +1051,7 @@ contains
     pure module subroutine set_range_large(self, start_pos, stop_pos)
 !
 !     Sets all valid bits to 1 from the START_POS to the STOP_POS positions
-!     in SELF. If STOP_POA < START_POS no bits are changed. Positions outside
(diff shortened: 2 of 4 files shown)


degawa (approved):
I confirmed that the issue #746 has been resolved in my environment.

I would like to thank you, @jvdp1, for fixing the CI/CD environment. I have been trying to resolve this issue, but it has been beyond my ability.

gnikit (approved):
Looks good thanks @jvdp1

gnikit:
Let's wait a couple of days for any remaining reviews and I there are none I will merge it.

jvdp1:
I'll merge this PR. Thank you all for your contributions

Issue #724, opened 2023-07-01
bitset count performance


PierUgit:
In the bitsets module, the `bit_count_large()` function is scanning bit by bit with `btest()` in the main loop:
```
        bit_count = 0
        do block_ = 1_bits_kind, size(self % blocks, kind=bits_kind) - 1
            do pos = 0, block_size-1
                if ( btest( self % blocks(block_), pos ) ) &
                    bit_count = bit_count + 1
            end do
        end do
```

Using chunks of blocks and the fact that `btest()` is elemental, one can get significant performance improvement (on a similar implementation I obtain a 5x speed-up witch a chunk size of 1024):

```
        bit_count = 0
        do block_ = 1_bits_kind, size(self % blocks, kind=bits_kind) - 1, chunk
            block__ = min(block_+chunk-1,size(self % blocks, kind=bits_kind))
            do pos = 0, block_size-1
                bit_count = bit_count + count( btest(self % blocks(block_:block__), pos ) )
            end do
        end do
```

(edit) of course, this makes sense only for a large number of bits

Euler-37:
fortran intrinsic function `popcnt` looks better. https://gcc.gnu.org/onlinedocs/gfortran/POPCNT.html

PierUgit:
Looks like, but then why is it not used in `bit_count_large()`?

jvdp1:
If I remember well, `bitset_large` was developed as an externsion of `bitset_64`. For `bitset_64`, `popcnt` cannot be used because some bits may not be used. The solution with `btest()` is therefore legitimate. The extension of `bit_count` for `bitset_large`  from `bitset_64` was straighfoward.

However, I agree that @PierUgit 's solution or using `popcnt` in this loop would be a more efficient implementation.

@PierUgit , do you still have such an implementation (i recognize that my answer is quite late!)? If so, would you like to open a PR with your solution/optimization?

PierUgit:
@jvdp1 I don't have a code ready in Github, I did some tests independently on a code of mine... To be honest I have first to learn how to use fpm and build the stdlib before being able to submit PRs here...

jvdp1:
> @jvdp1 I don't have a code ready in Github, I did some tests independently on a code of mine... 

Is it ok for you if I submit a PR and ask you for reviewing/testing it?

> To be honest I have first to learn how to use fpm and build the stdlib before being able to submit PRs here...

[...]

PR #756, opened 2024-01-03
Replaced btest() by popcnt() in bit_count_large()

PierUgit:
Related to issue #724 

Test case also added

### DIFF
--- a/src/stdlib_bitsets_large.fypp
+++ b/src/stdlib_bitsets_large.fypp
@@ -146,7 +146,7 @@ contains
 
         integer(bits_kind) :: nblocks, pos
 
-        nblocks = size( self % blocks )
+        nblocks = size( self % blocks, kind=bits_kind )
         bit_count = sum( popcnt( self % blocks(1:nblocks-1) ) )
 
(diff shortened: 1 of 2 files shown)


src/stdlib_bitsets_large.fypp
```diff
@@ -144,19 +144,13 @@ contains
...
 
-        bit_count = 0
-        do block_ = 1_bits_kind, size(self % blocks, kind=bits_kind) - 1
-            do pos = 0, block_size-1
-                if ( btest( self % blocks(block_), pos ) ) &
-                    bit_count = bit_count + 1
-            end do
-
-        end do
+        nblocks = size( self % blocks )
```
jvdp1, line 149:
```suggestion
        nblocks = size( self % blocks, kind=bits_kind )
```

In case that `bit_kinds` is changed to `int64`

jvdp1 (approved):
Thank you. LGTM. I only have a minor suggestion (which might be overkill).

jvdp1 (approved):
LGTM.

I wil wait a couple of days, and merge it if there are no other comments.

jvdp1:
Thank you @PierUgit . As it is a minor chane, I will merge it.

Issue #734, opened 2023-08-29
Improve fpm build in README for faster getting started


jalvesz:
Hello,

I wanted to make a suggestion regarding the README instructions for building with fpm that might seem evident but I felt that could have avoided me some head scratching.

After forking and downloading the branch locally, the 'git checkout stdlib-fpm' command did not work directly, so I had to do the following which seems a bit more explicit:
`source ./ci/fpm-deployment.sh`
`cd stdlib-fpm/`

I would suggest adding these two lines to the readme but I wanted to ask if there is any reason not to?

(side question: would it be worth it to have a companion fpm-deployment.bat for windows?)

jvdp1:
Sorry for the late reply.

> After forking and downloading the branch locally, the 'git checkout stdlib-fpm' command did not work directly, so I had to do the following which seems a bit more explicit: `source ./ci/fpm-deployment.sh` `cd stdlib-fpm/`
> 
> I would suggest adding these two lines to the readme but I wanted to ask if there is any reason not to?

No, there is nor reason to not mention it. Ideally `fpm` should support `fypp`. But meanwhile I support your idea. Would you like to open a PR with your suggestion?
 
> (side question: would it be worth it to have a companion fpm-deployment.bat for windows?)

I am not familiar with Windows. However, if it helpfuls, please propose such a script.

PR #757, opened 2024-01-05
Add CLI option for building with fpm

jalvesz:
Related to issue https://github.com/fortran-lang/stdlib/issues/734

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,7 +184,8 @@ git checkout stdlib-fpm
 fpm build --profile release
 ```
 
-Or alternatively:
+**Alternative**: as fpm does not currently support fypp natively, building stdlib is done in two steps (a) launch the preprocessor through the fpm-deployment.sh script, which creates a subfolder `stdlib-fpm` (b) build the project using the processed files within the latter subfolder. Which can be done with the following commands:
+
 ```sh
 source ./ci/fpm-deployment.sh
 cd stdlib-fpm/
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ git checkout stdlib-fpm
 fpm build --profile release
 ```
 
-**Alternative**: as fpm does not currently support fypp natively, building stdlib is done in two steps (a) launch the preprocessor through the fpm-deployment.sh script, which creates a subfolder `stdlib-fpm` (b) build the project using the processed files within the latter subfolder. Which can be done with the following commands:
+**Alternative**: as `fpm` does not currently support `fypp` natively, building `stdlib` with `fpm` can be done in two steps: a) launch the preprocessor through the `fpm-deployment.sh` script, which creates a subfolder `stdlib-fpm` and b) build the project using the processed files within the latter subfolder. This process can be done with the following commands:
 
 ```sh
 source ./ci/fpm-deployment.sh
--- a/README.md
+++ b/README.md
@@ -184,6 +184,13 @@ git checkout stdlib-fpm
 fpm build --profile release
 ```
 
+Or alternatively:
+```sh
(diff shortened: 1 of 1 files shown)


awvwgk:
This approach is meant for the continuous deployment of the fpm version and not for building stdlib locally. I would prefer to not advertise this way.

jvdp1:
More and more people (including me) use this way for developping stdlib. So
i think it is good to mention it, until fpm can fully support fypp. Maybe
addind a warning about its use would clarify it?

Le sam. 6 janv. 2024, 10:37, Sebastian Ehlert ***@***.***> a
écrit :

jalvesz:
I couldn't build it locally with the proposed approach, that's why I opened the issue #734.

Maybe an additional note can be included to clarify?

jvdp1:
I agree. I would also suggest to make the difference between stdlib-fpm
(for users) and stdlib (with the script for development).

At least this issue shows that the Readme must be clarify.

jalvesz:
What about something in the lines of:
"NOTE: as fpm does not currently support fypp natively, building stdlib is done in two steps (a) launch the preprocessor through the fpm-deployment.sh script, which creates a subfolder stdlib-fpm (b) build the project using the processed files within the latter subfolder. Which can be done with the following commands:" ?

README.md
```diff
@@ -184,6 +184,14 @@ git checkout stdlib-fpm
 fpm build --profile release
 ```
 
+**Alternative**: as fpm does not currently support fypp natively, building stdlib is done in two steps (a) launch the preprocessor through the fpm-deployment.sh script, which creates a subfolder `stdlib-fpm` (b) build the project using the processed files within the latter subfolder. Which can be done with the following commands:
+
+```sh
+source ./ci/fpm-deployment.sh
```
jvdp1, line 190:
```suggestion
git checkout stdlib
source ./ci/fpm-deployment.sh
```

The note sounds good to me. 

If @awvwgk (and others) is/are ok, it is good for me

jalvesz: Question: is the "git checkout stdlib" command necessary here? So far I have not needed this line. Here my working env:
Windows+WSL+VSCode
So, with VS Code and git installed locally, I can track changes and swap branches. I use the WSL for launching the build script, and also compiling under Linux with the compilers installed in the WSL. Then I can use the same folder to recompile with my Windows compilers.

jvdp1: Yes, you are right. 
I meant `git clone stdlib` or `git checkout master`. 
But it might be to specific to my environment. So you can ignore this comment. Sorry for the  noise.

jalvesz: I see, I kind of assumed that whomever will use this, would have already done `git clone stdlib`. No problem about it

README.md
```diff
@@ -184,6 +184,14 @@ git checkout stdlib-fpm
 fpm build --profile release
 ```
 
+**Alternative**: as fpm does not currently support fypp natively, building stdlib is done in two steps (a) launch the preprocessor through the fpm-deployment.sh script, which creates a subfolder `stdlib-fpm` (b) build the project using the processed files within the latter subfolder. Which can be done with the following commands:
```
jvdp1, line 187:
```suggestion
**Alternative**: as `fpm` does not currently support `fypp` natively, building `stdlib` with `fpm` can be done in two steps: a) launch the preprocessor through the `fpm-deployment.sh` script, which creates a subfolder `stdlib-fpm` and b) build the project using the processed files within the latter subfolder. This process can be done with the following commands:
```

jvdp1 (approved):
The note and command lines look good to me.
Hopefully this may help new developers to start to develop `stdlib` with `fpm`.

PR #758, opened 2024-01-07
Minor changes related to `fypp` prepocessing

jvdp1:
Minor changes inspired by #675 and #743

@jalvesz : I believe that this would be a cleaner solulion to the issue your faced with `fpm-deployment.sh` and `test_mean_f03.fypp`.


awvwgk (approved):
Thanks for sharing. Looks like a reasonable change to include in stdlib.

jvdp1:
Thank you @awvwgk. As there are minor changes, I will merge it.

Issue #749, opened 2023-12-04
Support for linear algebra


perazz:
Hello stdlib developers, 
I'm opening this issue to summarize and coalesce upcoming efforts to integrate linear algebra operations in stdlib, in particular: 

1) Accessible interfaces for common linear algebra operations

We want stdlib to be able to solve linear algebra tasks by wrapping against libraries like BLAS, LAPACK, SCALAPACK, with a user-friendly interface. I think this means:
- To get the best possible user-level API (almost no inputs at all), maintaining familiarity the syntax that other libraries also provide (scipy, etc.)  but have an expert interface with all control knobs to tune the algorithms (e.g. via settings stored in derived types). Thanks to Fortran arrays, If this is done well, I believe stdlib could have the best linear algebra API out there;
- Wrapping against BLAS and LAPACK means that the low-level API should probably not be changed (so that down the road, stdlib could just link against external BLAS libraries or platform-dependent frameworks), but a reference implementation must be provided anyways. Should we aim at maintaining our own Modernized BLAS/LAPACK at fortran-lang, or just automate the download process from netlib? I would personally like the idea to develop a Modernized version once and forall (those reference implementations will almost never change anymore), although those are huge repos that would require significant work.  

2) Support for common IO formats for matrices and tensors

- To allow efficient working with multidimensional arrays, we should support easy serialization/deserialization and conversion among formats. This could include formats like NPZ, matrixmarket, etc. An easily extendible API should be provided to have plugins for other formats which might or might not fit in the scope of stdlib.

- I believe an equally important task is to define derived types that are capable of storing temporary information which is not strictly matrix data (e.g. matrix factorization, or working arrays for the matrix solvers), to avoid unnecessary overhead in case of repeated algebra operations. This means that "simple" array storage may need to be replaced with matrix derived types in those cases. 

Because there are plenty of options in defining these APIs, It is crucial to the success of this task that as much feedback as possible is given, so I would like to encourage all ideas - and criticisms - to be discussed on this issue, so that we can come up with the best possible version. I am also opening a discussion page on the [Fortran-lang Discourse](https://fortran-lang.discourse.group/t/linear-algebra-support-in-stdlib/6917) that we can use for more verbose discussions. 

Thank you,
Federico

cc @certik @awvwgk @fortran-lang/stdlib @fortran-lang/admins

### Linked issues 

Linear algebra (BLAS/LAPACK)
#1
#10
#67 
#450
#476 

[...]

PR #760, opened 2024-01-10
Sparse algebra support with OOP API

jalvesz:
PR related to:
https://github.com/fortran-lang/stdlib/issues/38
https://github.com/fortran-lang/stdlib/issues/749
https://github.com/fortran-lang/stdlib/pull/189

Here I try to propose the OOP API upfront as a means to centralize data-containment format within stdlib, which I feel would be a (the) big added value. I'm migrating the proposal started here [FSPARSE](https://github.com/jalvesz/FSPARSE) to adapt it to stdlib.

Current proposal design:

* Derived Types: 
`sparse_type` (abstract base type)
`COO_type`, `CSR_type`, `CSC_type`, `ELL_type`, `SELLC_type` (DTs with no data buffer)
`COO_?p_type`, `CSR_?p_type`, `CSC_?p_type`, `ELL_?p_type`, `SELLC_?p_type` (DTs with data buffer, where `?` kind precision including complex values)
* matrix-vector product:
```Fortran
call spmv( matrix, x, y [,alpha,beta,op] ) !> y = alpha * op(matrix) * x + beta * y
```
`op`: `sparse_op_none='N'` (default), `sparse_op_transpose='T'` or `sparse_op_hermitian='H'` 

* Data accessors:
```Fortran
val = matrix%at( i , j ) !> to get the value at (i,j) position, if (i,j) not in the sparse pattern, value = 0, if out-of-bounds val = NaN
call matrix%add( i , j, val ) !> to set the value at (i,j) position, if (i,j) not in the current structure, skip
!> OR add a data block
call matrix%add( dofs_i(:), dofs_j(:), mat(:,:) ) !>
```

* Conversions :
```Fortran
call coo2ordered(COO [, sort_data]) !> sort and remove duplicates from the COO data structure.
 
call dense2coo( dense , COO )  !> dense is a plain 2d array
call coo2dense( COO  , dense )

call coo2csr( COO , CSR ) !> assumes ordered and unique indexes for COO ... to implement a sorting algorithms before transferring data
call csr2coo( CSR , COO ) !> assumes ordered and unique indexes for CSR 

call coo2csc( COO , CSC ) !> assumes ordered and unique indexes for COO 
call csc2coo( CSC , COO ) !> assumes ordered and unique indexes for CSC

call csr2ell( CSR , ELL [, num_nz_rows]  ) 
call csr2sellc( CSR , SELLC [, chunk ]  ) !> chunk sizes for spmv [4, 8, 16]

call diag( <sparse> , diag ) !> extract the diagonal components of the sparse matrix
```


zerothi:
Wouldn't it be useful if `sparse_t` has the interface for `mat%to_coo|csr|...` etc instead of using a function call. Since this is going OOP, lets keep it like that.

jalvesz:
This could be done even shorter with a `%to(...)` interface as the types are declared before calling the conversion. I've done that before but did not propose it here upfront as I wanted to hear some opinions on what the OOP API in stdlib should look like.

zerothi:
Agreed, `%to` would be even better, or `%convert` or clarity? hmm...

ftucciarone:
Is it a good idea to have matvec product that do not initialize the result vector? I have to say that I have spent a good hour debugging a code and the source of divergence in the solver was the fact that I was not initializing the result vector before passing it to the subroutine. While I understand that this is a great way to perform Axpy operation, I would advise to have separate matvec (with initialization) and Axpy (without) routines. 
`
    subroutine matvec_csr_1d_sp(vec_y,matrix,vec_x)
        type(CSR_sp), intent(in) :: matrix
        real(sp), intent(in)    :: vec_x(:)
        real(sp), intent(inout) :: vec_y(:)
        integer :: i, j
        real(sp) :: aux

        associate( data => matrix%data, col => matrix%col, rowptr => matrix%rowptr, &
            & nnz => matrix%nnz, nrows => matrix%nrows, ncols => matrix%ncols, sym => matrix%sym )
            if( sym == k_NOSYMMETRY) then
                do concurrent(i=1:nrows)
                    do j = rowptr(i), rowptr(i+1)-1
                        vec_y(i) = vec_y(i) + data(j) * vec_x(col(j))
                    end do
                end do

            else if( sym == k_SYMTRIINF )then
                do i = 1 , nrows
                    aux  = 0._sp
                    do j = rowptr(i), rowptr(i+1)-2
                        aux = aux + data(j) * vec_x(col(j))
                        vec_y(col(j)) = vec_y(col(j)) + data(j) * vec_x(i)
                    end do
                    aux = aux + data(j) * vec_x(i)
                    vec_y(i) = vec_y(i) + aux
                end do

            else if( sym == k_SYMTRISUP )then
                do i = 1 , nrows
                    aux  = vec_x(i) * data(rowptr(i))
                    do j = rowptr(i)+1, rowptr(i+1)-1
                        aux = aux + data(j) * vec_x(col(j))
                        vec_y(col(j)) = vec_y(col(j)) + data(j) * vec_x(i)
                    end do
                    vec_y(i) = vec_y(i) + aux
                end do

            end if
        end associate
    end subroutine
`

jalvesz:
Hi @ftucciarone, thanks for checking out the current draft and sorry for the inconvenience.

Yes, I have intentionally designed the API to not initialize the vector, in the current version within FSPARSE https://github.com/jalvesz/FSPARSE?tab=readme-ov-file#sparse-matrix-vector-product I updated the README to be clear about it but forgot to update here.

The reason for that choice is because it allows maximum flexibility for preprocessing linear systems needed operations with the same matrix operator that will be used for the solver. So it basically places the "responsability" on the user to place a `y=0` just before if one does indeed need a clean vector.

This is of course totally open to debate, I just proposed following my experience reworking solvers. My feeling is that doing that can avoid having to manage different implementations or optionals, when the solution can be to explicitly state that the interface is updating (y = y + M * x) instead of overwriting (y = M  * x).

ftucciarone:
Hi @jalvesz, it was actually a pleasure to look into this draft, I am learning a lot and I look forward to discuss it with you if possible. 

I have to say that I have particularly strong feelings about the "non-initialization" problem, mostly twofold. 
First, if y = y + M*x is the chosen way to go, then it must be written very large, at the very beginning of the documentation, to make sure everyone sees that. Test example should also take care of that, showing that the result is wrong if not initialized before. 
Second, I have the feeling that doing first y=0_dp and then call matvec(y, A, x) might add an overhead due to the initialization of y for very large problems, while initializing the component y(i) while doing the matvec might be faster, but I have to test  because I'm not 100% sure about this. 

However, I think that separating matvec and Axpy could be feasible at this stage (I can take care of that with the correct guidance) and potentially doable by preprocessing as well (I'm thinking at a fypp if).

Cheers, Francesco

jalvesz:
@ftucciarone thanks for the idea! I will also bring parts of a discussion I had with @ivan-pi on exactly this topic, where he brought to my attention the following:

> The MKL library offers, y := beta y + alpha A x, and then you need to pick either beta = 0, or beta = 1, depending on what you want.

> In Apple's Accelerate Framework on the other hand, they offer two functions:
```
SparseMultiply(A,x,y)           // y = A x
SparseMultiplyAdd(A,x,y)     // y += A x
```

> In PSBLAS, spmm covers both vectors and matrices (rank-2 dense arrays):
```
call psb_spmm(alpha, a, x, beta, y, desc_a, info)
call psb_spmm(alpha, a, x, beta, y, desc_a, info, trans, work)
```
https://psctoolkit.github.io/psblasguide/userhtmlse4.html#x18-550004

Taking all those ideas together, I could imagine using optionals to cover:
```
call matvec( A , vec_x, vec_y ) !> vec_y = vec_y + A * y
call matvec( A , vec_x, vec_y , overwrite = .true. ) !> vec_y = A * y
call matvec( A , vec_x, vec_y , alpha=value, beta=value ) !> vec_y = beta * vec_y + alpha * A * y
```
Or the default to be overwrite and an optional for addition, or simply with beta =1 `if(present(beta))` would automatically determine that.

Let me know your thoughts on that

zerothi:
> Taking all those ideas together, I could imagine using optionals to cover:
> 
> ```
> call matvec( A , vec_x, vec_y ) !> vec_y = vec_y + A * y
> call matvec( A , vec_x, vec_y , overwrite = .true. ) !> vec_y = A * y
> call matvec( A , vec_x, vec_y , alpha=value, beta=value ) !> vec_y = beta * vec_y + alpha * A * y
> ```
To follow how BLAS handles things, if `beta=0` then `overwrite = true`. There should be no need to explicitly add `overwrite=True`, it could be difficult for end users to understand `matvec(..., beta=2, overwrite=True)`, using `beta=0` would be simpler, and since it is zero it should be *ok*.

jalvesz:
Done in the last commit. I added as default: beta=0 and alpha=1.

jalvesz:
@ftucciarone thanks for the suggestion! This is indeed a needed feature, though some comments:

* I would say that the very next step would be to get the current PR improved in terms of code structure and the specs in order for it to be merged such that these kind of proposals can be worked on incrementally on top of it. Any help on that would be honestly very welcomed. If we add more features before polishing and merging the PR, it would become unmanageable and I fear it won't get merged any time soon. ( smaller PRs are easier to review and merge :) and considering the specs are not ready, this one will get too big soon )

* Regarding the triangular solver, I had added the symmetry definition in the DTs for this kind of usages, the idea being that the symmetry defines the pre-supposed storage scheme. So, if `sym == k_SYMTRIINF` (SYMmetric TRIangular INFerior) we imply the matrix represents a lower triangular storage, so on and so forth. Maybe a different notation can be used, let me know.

* It is quite common for preconditioning iterative solvers to factorize the matrix by $A=LDL^T$ (including the diagonal term), meaning that the preconditioning step $LDL^T x = b$ needs a forward $L z = b$, diagonal $y = D^{-1}z$, and backward $L^T x = y$ passes. So I would suggest to keep them as explicit individual steps in order to compose as pleased depending on the factorization scheme.

zerothi:
Does the author here and other interested know about the GraphBlas initiative? https://graphblas.org/

It seems it would be valuable to adhere to the API there for a greater overlap? There could be some inspiration at least.

doc/specs/stdlib_sparse.md
```diff
@@ -0,0 +1,243 @@
...
+ELL%index(4,1:3) = [1,3,4]
+```
+<!-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -->
+### `SELL-C`: The Sliced ELLPACK with Constant blocks format
+#### Status
+
+Experimental
+
+#### Description
+The Sliced ELLPACK format `SELLC` is a variation of the `ELLPACK` format. This modification reduces the storage size compared to the `ELLPACK` format but maintaining its efficient data access scheme. It can be seen as an intermediate format between `CSR` and `ELLPACK`. For more details read [here](https://arxiv.org/pdf/1307.6209v1)
```
perazz, line 136:
```suggestion
The Sliced ELLPACK format `SELLC` is a variation of the `ELLPACK` format. This modification reduces the storage size compared to the `ELLPACK` format but maintaining its efficient data access scheme. It can be seen as an intermediate format between `CSR` and `ELLPACK`. For more details read [the reference](https://arxiv.org/pdf/1307.6209v1)
```

doc/specs/stdlib_sparse.md
```diff
@@ -0,0 +1,243 @@
...
+
+Experimental
+
+### Description
+
+This module provides facility functions for converting between storage formats.
+
+### Syntax
+
+`call ` [[stdlib_sparse_conversion(module):coo2ordered(interface)]] `(coo)`
```
perazz, line 180: I like having subroutines to maximize performance and avoid the issues with function assignment. I wonder if we should also provide 'clean' function interfaces to generate new objects? Just giving it a thought. 

They would probably pair better with overloaded operators for example, but also in other ways. For example, one could have: 

```fortran
type(COO_sp) :: mat
[...]

! Convert to CSR using interface CSR_sp and then do stuff
call my_CSR_kernel( CSR_sp(mat) , x , y, z )
```

This would require to add an interface
```Fortran
interface CSR_sp
   module procedure coo2csr_fun
end interface

elemental type(CSR_sp) function coo2csr_fun(COO) result(CSR)
    type(COO_sp), intent(in) :: COO
    call coo2csr( COO, CSR )
end function
```

doc/specs/stdlib_sparse.md
```diff
@@ -0,0 +1,243 @@
...
+
+### Arguments
+
+`COO`, `intent(inout)`: Shall be any `COO` type. The same object will be returned with the arrays reallocated to the correct size after removing duplicates.
+
+`sort_data`, `logical(in)`, `optional`:: Boolean optional to determine whether to sort the data in the COO graph while sorting the index array, defult `.false.`.
```
perazz, line 186:
```suggestion
`sort_data`, `logical(in)`, `optional`:: Shall be an optional `logical` argument to determine whether data in the COO graph should be sorted while sorting the index array, defult `.false.`.
```

src/stdlib_sparse_conversion.fypp
```diff
@@ -0,0 +1,452 @@
...
+    interface coo2dense
+        #:for k1, t1, s1 in (KINDS_TYPES)
+        module procedure coo2dense_${s1}$
+        #:endfor
+    end interface
+
+    !> Conversion from coo to csr
+    !> 
+    !> Enables transferring data from a COO matrix to a CSR matrix
+    !> under the hypothesis that the COO is already ordered.
```
perazz, line 41: What are the implications of an unordered COO matrix? Would that break any other manipulation routines? 
If so, I would recommend to always run a sort internally whenever these operations take place (it would be even better to keep track of it via an `is_sorted` flag inside the derived type, if it is not too much work). 

In other words, I would suggest to go for safety first, and performance later. Another way to do this may be to use an optional flag, that assumes *not* sorted by default: 

```Fortran
subroutine coo2dense(COO, is_sorted=.false.) 
```

src/stdlib_sparse_conversion.fypp
```diff
@@ -0,0 +1,452 @@
...
+                
+                end do
+            end do
+         end block
+        end associate
+    end subroutine
+
+    #:endfor
+
+    recursive subroutine quicksort_i(a, first, last)
```
perazz, line 243: Could the routines from `stdlib_sorting` be used instead?

src/stdlib_sparse_kinds.fypp
```diff
@@ -0,0 +1,420 @@
...
+      integer, allocatable  :: index(:,:) !> Matrix coordinates index(2,nnz)
+    contains
+      procedure :: malloc => malloc_coo
+    end type
+
+    #:for k1, t1, s1 in (KINDS_TYPES)
+    type, public, extends(COO_type) :: COO_${s1}$
+        ${t1}$, allocatable :: data(:) 
+    contains
+        procedure :: get => get_value_coo_${s1}$
```
perazz, line 54: I would put these interfaces also in the corresponding `abstract` type

src/stdlib_sparse_kinds.fypp
```diff
@@ -0,0 +1,420 @@
...
+        integer, intent(in) :: ik, jk
+        integer :: k
+        ! naive implementation
+        do k = 1,self%nnz
+            if( ik == self%index(1,k) .and. jk == self%index(2,k) ) then
+                val = self%data(k)
+                return
+            end if
+        end do
+        val = zero_${s1}$
```
perazz, line 305: I like the fact that get/set should all be `pure` subroutines. I suggest we add one more option, that is when the required indices are out of bounds. I think the best case scenario would be: 

- `(i,j)` in matrix bounds -> return value, or 0.0 if not part of the sparse pattern
- `(i,j)` out of bounds -> return `ieee_quiet_nan` 

I think this would give more safety to the users, as it is easy to spot whether wrong inputs were provided. Sparse matrices are fast but not very flexible, so I think the option that matrix size (nrows, ncols) grows dynamically with user-data setting should not be considered.

test/linalg/test_sparse_spmv.fypp
```diff
@@ -0,0 +1,290 @@
...
+        !> Error handling
+        type(error_type), allocatable, intent(out) :: error
+        #:for k1, t1, s1 in (KINDS_TYPES)
+        block
+            integer, parameter :: wp = ${k1}$
+            type(CSC_${s1}$) :: CSC
+            ${t1}$, allocatable :: vec_x(:)
+            ${t1}$, allocatable :: vec_y(:)
+
+            call CSC%malloc(4,5,10)
```
perazz, line 137: As I think I've commented earlier, I would really love to see a unified and consistent matrix setting interface for all matrices that would take in 3 arrays: `i, j, v` containing row, column, and value of each non-zero value. 

I think the internals should be left `public` for the advanced user, but this simple API would be very useful to reduce potential errors. For example: automatically remove zeros, ensure correct storage pointers, etc.

perazz (commented):
Thank you @jalvesz, I think it looks very good already. 
I left quite a few comments that can be summarized as follows: 
- would be great to see some simpler interfaces, e.g. function-based (`m = CSR_sp(n)`) but also for setting/getting values including matrix initialization a la Matlab, i.e., `m = CSR_sp(i, j, v [, nrows] [, ncols])` or `call m%new(i, j, v [, nrows] [, ncols])`
- By default, never assume that the internal representation is not ready to be used. In other words, all functions should always work i.e. by putting in automatically the necessary sorting. To save time, the user should state it with an optional, non-default argument. Further optimizations can come next
- Return `NaN`s for anything that's outside of the matrix size. This will signal to the user that the wrong inputs were used.

doc/specs/stdlib_sparse.md
```diff
@@ -183,7 +183,7 @@ This module provides facility functions for converting between storage formats.
 
 `COO`, `intent(inout)`: Shall be any `COO` type. The same object will be returned with the arrays reallocated to the correct size after removing duplicates.
 
-`sort_data`, `logical(in)`, `optional`:: Boolean optional to determine whether to sort the data in the COO graph while sorting the index array, defult `.false.`.
+`sort_data`, `logical(in)`, `optional`:: Shall be an optional `logical` argument to determine whether data in the COO graph should be sorted while sorting the index array, defult `.false.`.
```
ivan-pi, line 186:
```suggestion
`sort_data`, `logical(in)`, `optional`:: Shall be an optional `logical` argument to determine whether data in the COO graph should be sorted while sorting the index array, default `.false.`.
```

src/stdlib_sparse_kinds.fypp
jalvesz: Resolved in https://github.com/fortran-lang/stdlib/pull/760/commits/21a8547d2eec103fe04a6a825b8994a4cac71c5d and https://github.com/fortran-lang/stdlib/pull/760/commits/2fb4e83c3835c4a2ef67c15dc0a6a7112c132c59

src/stdlib_sparse_conversion.fypp
jalvesz: For the case of sorting only the index array that was easy. I'm still thinking how to do it for the case when wanting to make the data follow the sorting of the index without creating temporal arrays.

ivan-pi: Salvatore Filippone has written elsewhere that quicksort doesn't tend to be the most suitable for sorting indexes. (This can of course vary from matrix to matrix.) Not to block this nice PR, just saying that in the future this can be revisited for better performance. stdlib is a reference implementation first and foremost.

perazz: > how to do it for the case when wanting to make the data follow the sorting of the index without creating temporal arrays.

To sort a list, and swap data according to the sort, this is the way I do it, you're welcome to use this template Jose:

```Fortran
    ! Quicksort, and swap list accordingly
    #:for lt,ln in LIST_TYPES_NAMES
    pure recursive subroutine ${name}$_quicksort_andlist_${ln}$(array,list,down)

      ${type}$, dimension(:), intent(inout) :: array
      ${lt}$, dimension(size(array)), intent(inout) :: list
      logical, optional, intent(in) :: down

      integer(ISIZE), parameter :: SMALL_SIZE = 20
      integer(ISIZE) :: i, j, n
      ${type}$  :: chosen
      logical :: descending

      descending = .false.; if (present(down)) descending = down
      n = size(array,kind=ISIZE)

      choose_sorting_algorithm: if (n>1 .and. n <= SMALL_SIZE) then

         ! Use interchange sort for small lists
         do i = 1, n - 1
            do j = i + 1, n
               if (toBeSwapped(array(i),array(j),.false.)) then
                   call swap_data_${name}$(array(i),array(j))
                   call swap_list(list(i),list(j))
               end if
            end do
         end do

      elseif (n>SMALL_SIZE) then
         ! Use partition (quick) sort if the array is big
         chosen = array(int(n/2))
         i = 0
         j = n + 1

         scan_lists: do
            ! Scan array from left end
            ! until element >= chosen is found
            scan_from_left: do
              i = i + 1
              if (toBeSwapped(array(i),chosen,.true.) .or. i==n) exit scan_from_left
            end do scan_from_left

            ! Scan array from right end
            ! until element <= chosen is found

            scan_from_right: do
               j = j - 1
               if (toBeSwapped(chosen,array(j),.true.) .or. j==1) exit scan_from_right
            end do scan_from_right

            swap_element: if (i < j) then
                ! Swap two out of place elements
                call swap_data_${name}$(array(i),array(j))
                call swap_list(list(i),list(j))
            else if (i == j) then
                i = i + 1
                exit
            else
                exit
            endif swap_element

         end do scan_lists

         if (1 < j) call ${name}$_quicksort_andlist_${ln}$(array(:j),list(:j),down)
         if (i < n) call ${name}$_quicksort_andlist_${ln}$(array(i:),list(i:),down)

      end if choose_sorting_algorithm ! test for small array

      contains
      
         elemental subroutine swap_list(a,b)
            ${lt}$, intent(inout) :: a,b
            ${lt}$ :: tmp
            tmp = a
            a = b
            b = tmp
         end subroutine swap_list

         elemental subroutine swap_data_${name}$(a,b)
            ${type}$, intent(inout) :: a, b
            ${type}$                :: tmp
            tmp = a
            a   = b
            b   = tmp
         end subroutine swap_data_${name}$

         elemental logical function toBeSwapped(a,b,orEqual)
            ${type}$, intent(in) :: a,b
            logical, intent(in) :: orEqual

            toBeSwapped = merge(is_before(a,b),is_after(a,b),descending)
            if (orEqual .and. a==b) toBeSwapped = .true.

         end function toBeSwapped

    end subroutine ${name}$_quicksort_andlist_${ln}$
    
    #:endfor
```

In this template: 
- `${type}$` is the array type (in your case, the integer column indices) 
- `${name}$` its template name
- `${lt}$` is the list type (in your case, `real(sp)` if you have real matrix data, etc.)
- we're working with numbers, so you can replace templates `is_before` with simpler comparisons `a<b` etc

jalvesz: @ivan-pi Yes, that's why I used a combined sort: count_sort for the first level, then quicksort for the smaller chuncks for each row. Is the most efficient approach I have found so far. 
Both: in the sorting interfaces from stdlib I see that for making a secondary data be reordered following the sorting of something else, it is needed to create a temporal index mapping https://stdlib.fortran-lang.org/page/specs/stdlib_sorting.html "Sorting a related rank one array". This would be detrimental as is for large sparse matrices. That's why a included a quicksort that takes in the array to be sorted and an extra data buffer. @perazz the code you propose is very neat, the question is: should this be made available as a sorting interface elsewhere from stdlib and use it here as a particular application? (coming to your original comment of using stdlib facilities instead of locally cooked ones)

perazz: @jalvesz I agree that no temporaries should be made. 
The way we dealt with this for algebra error control was to define a local error handler, in a way that can later be promoted to the whole stdlib. 

So one way to do this could be to write good code (e.g. fully templated for integer/real arguments of either array or list) here, so that it can nearly automatically be promoted to be an `stdlib_sorting` method later.

src/stdlib_sparse_kinds.fypp
jalvesz: I've been scratching my head around this one: if these are deferred procedures in the abstract type, then it would be necessary to declare a different procedure for each kind in the abstract type in order to define also the abstract interfaces, which I feel would blotter the implementation. Also, for the base type with no data buffer, my impression is that there is no point on having these functions as their purpose is to interact with the data buffer. Any suggestions?

note: changed the name of "set" -> "add" (following the notation of [UFC](https://arxiv.org/pdf/1205.3039) suggested by @ivan-pi ) and "get" -> "at" following your suggestion @perazz .

jalvesz:
@perazz I tried to adress all your comments as best as possible in the latest commits:

Added a `from_ijv` interface subroutine to initialize `COO`, `CSR`, `ELL` and `SELLC` types. It might not be the most performant implementation but I think it gives a base interface that can be improved.

The COO matrix does have an attribute to indicate whether or not the data is sorted. I agree that for some operations it should be called, the initialization using the `from_ijv` interface does enforces sorting the matrix. But giving that sorting operations can be expensive, I also think that the user should have the control and responsability of calling it just when extrictly necessary.

The data accessor Type-Bound procedures are now called `add` (it can take an ijv scalar triplet or a v(:,:) block matrix with i(:), j(:) indices) and `at` (only scalar tuple ij).

perazz:
Looks great @jalvesz. If I'm not asking too much, I think that it would be valuable to have a second example program, that instead of starting from a dense matrix, starts from `i, j, v` data and showcases matrix initialization and the `at` / `add` interfaces. Otherwise, LGTM!

perazz (approved):

jalvesz:
Added both examples @perazz, let me know what do you think.

src/stdlib_sparse_conversion.fypp
```diff
@@ -0,0 +1,426 @@
...
+                end do
+            end do
+         end block
+        end associate
+    end subroutine
+
+    #:endfor
+
+    #:for k1, t1, s1 in (KINDS_TYPES)
+    recursive subroutine quicksort_i_${s1}$(a, b, first, last)
```
jvdp1, line 244: would it be an idea to add such a procedure in `stdlib_sorting`, based on `sort_index`? `sort_index` is clearly limited for such simpler cases, and making the argument `index` as `intent(inout)` would already facilitate such cases.

jalvesz: @jvdp1 I'm not sure if I got it right, you mean to say to add a quick sort within sort_index? Or to add it as a different method within the sort module?... I took a look at sort index but my first impression is that changing the index array to `inout` won't be enough. The internals seems to be thought for handling an array to sort and the work index arrays. Here, we need to sort one array and have a secondary array be sorted together with the first, simultaneously. From what I read in the specs, non of the current sorting interfaces enable this directly, but I just had quick glance so I might be missing something. I wouldn't mind trying to generalize the quicksort, or if you help me out figuring out how to adapt any other of the sorting procedures to enable sorting two arrays simultaneously (avoiding extra allocations), I could give it a try.

jvdp1: I was thinking to extend `sort_index` such that a provided vector (other than just `integer(int64)`) is sorted in the same way as the sorted vector (if I am right, it is what you need). The kind of the array `index` is already generated by `fypp`. So I think that it should be possible (but I may be wrong). I will have a look.

jvdp1: @jalvesz Could you maye test the `sort_index` in [this branch](https://github.com/jvdp1/stdlib/tree/extend_sort_index)? I did some dirty modifications such that the array `index` van be of any `integer` or `real` type. If it works, we will need to discuss names, and approaches, before getting a nice implementation.
Here is an example I tested in my branch:
```fortran
program example_sort_index_real
  use stdlib_sorting, only: sort_index
  implicit none
  integer, allocatable :: array(:)
  real, allocatable :: index(:)

  array = [5, 4, 3, 1, 10, 4, 9]
  index = real(array)

  call sort_index(array, index)

  print *, array   !print [1, 3, 4, 4, 5, 9, 10]
  print *, index   !print [1., 3., 4., 4., 5., 9., 10.]

end program example_sort_index_real
```

jalvesz: Wow thanks! I'll test this out and let you know.

jalvesz: @jvdp1 I tested your modifications to sort a COO matrix representing a 2millions nodes mesh, having 14millions of NNZ before sorting/removing duplicates:

Sorting index + data using quicksort (3 runs in release profile)
elapsed time (cpu_time) [s]: 1.4375559 / 1.4128339 / 1.1456219
Sorting index + data using sort_index:
elapsed time (cpu_time) [s]: 1.4756519 / 1.5304509 / 1.1852899

If I do not sort the data (only the index(1:2,1:nnz) array) then the stdlib `sort` is about 9% faster than the quicksort.

I would say these are more than satisfying results!

Now, I included support for complex sparse matrices and the sorting interfaces do not support them. So we should discuss what to do here to extend for complex data ( ? )

jvdp1: > I would say these are more than satisfying results!

Nice. Thank you for testing it.

> Now, I included support for complex sparse matrices and the sorting interfaces do not support them. So we should discuss what to do here to extend for complex data ( ? )

It seems that complex data is not supported at all by the `stdlib` sorting procedures. It should be quite easy to add to all of them by extending the `fypp` variables to the complex  definitions.
Note that my changes broke the original `stdlib_index`. Now that I know that it works, I will have a closer look for a proper implementation.

jalvesz: So I'm guessing this would deserve a nice PR on its own. Should we wait for that before this PR? or can we keep reviewing this knowing that this particular point can be improved down the road?

jvdp1: You are right, it deserves its own PR. I suggest that you continue with your internal quicksort subroutine. We can review it later when the PR related to the sort procedures will be ready.

certik:
Thanks for the PR! I think it looks in the right direction.

In order for me to start using these routines (for parallel sparse matmul, various conversions, etc.), I would need a version that does not use the stdlib's custom derived type, since I have my own derived type on my code. Is there a way to do that?

The only way I know is to split the API into two parts:

* low level that does not use derived types
* high level, OO style (this PR)

Maybe there is another way, but the above way seems robust. Then I can start using the low level API, and just give it the CSR arrays that I have in my code from my own derived type.

Why not structure this PR in this low level / high level style? It's almost there, but not quite.

perazz:
I'm biased toward object-oriented APIs, but I like the way the current PR is structured: each sparse structure has a derived type with the same API, but the expert user can also access the internal array structures because they've all been made `public`. 

So one way to use the `CSR` arrays would be to just refer to them: 
```fortran
call my_csr_subroutine(mat%row_ptr, mat%columns, mat%data)
```

Another option to make the API even more friendly would be to define two `move_alloc` subroutines per data type, that the user could use (at zero performance cost) to go back and forth between an array implementation and the derived type: 

```Fortran
call mat%move_alloc_to(IA, JA, A) ! move allocation to IA, JA, A
call mat%move_alloc_from(IA, JA, A) ! move allocation from arrays into matrix

pure subroutine csr_move_alloc_to(mat,IA,JA,A)
   class(CSR_type), intent(inout) :: mat
   integer, intent(out), allocatable :: IA(:), JA(:)
   real(${rk}$), intent(out), allocatable :: A(:)
   call move_alloc(from=mat%row_ptr,to=IA)
   call move_alloc(from=mat%column,to=JA)
   call move_alloc(from=mat%data,to=A)
end subroutine csr_move_alloc_to
pure subroutine csr_move_alloc_from(mat,IA,JA,A)
   class(CSR_type), intent(out) :: mat
   integer, intent(inout), allocatable :: IA(:), JA(:)
   real(${rk}$), intent(inout), allocatable :: A(:)
   call move_alloc(from=IA,to= mat%row_ptr)
   call move_alloc(from=JA,to= mat%column)
   call move_alloc(from=A,to= mat%data)
end subroutine csr_move_alloc_from
```

@jalvesz @certik @jvdp1 do you think it would be worthwhile?

jvdp1:
> Thanks for the PR! I think it looks in the right direction.
> 
> In order for me to start using these routines (for parallel sparse matmul, various conversions, etc.), I would need a version that does not use the stdlib's custom derived type, since I have my own derived type on my code. Is there a way to do that?
> 
> The only way I know is to split the API into two parts:
> 
> * low level that does not use derived types
> * high level, OO style (this PR)
> 
> Maybe there is another way, but the above way seems robust. Then I can start using the low level API, and just give it the CSR arrays that I have in my code from my own derived type.
> 
> Why not structure this PR in this low level / high level style? It's almost there, but not quite.

@certik Thank you for your comment. I agree with you that these features should be splitted into 2 parts (I have also my own sparse library, and I might be interested to call some of the stdlib features) :
* low level procedures (e.g, spmv)
* high level for OO

However, @jalvesz started with the OO approach, and I think it is quite close to be merged as is, pending a few "minor" changes. Restructuring this PR might be a huge task.

Therefore, I advice to review and merge this PR as is, maybe after implementing some suggestions where we think the API level could be already lowered a bit.
The merged procedures could then be reviewed in a second stage based on users' feedbacks and with having in mind this 2-level approach. The high-level OO approach would not be modified for the user. Only the low-level approach would be added based on waht is already existing.

I am afraid that if we go for the 2-API level approach in one go, we will have nothing in `stdlib` at the end
due to too many discussions, lack of time, @jalvesz motivation,.... This topic of sparse matrices is on the table for a long time, with many unsuccessful attempts. I think @jalvesz 's implementation is the closest one to be merged. At this stage, I prefer to have something usable in stdlib, even if it only includes OO-API. Users may start to use it, provide feedbacks, and hopefully contribute to stdlib.
Such an approach has been used for other stdlib features (e.g., sorting, hash maps, loggers) with some success.
I will be happy to hear your feedback.

jvdp1 (commented):
Here are some comments for the specs only (note: I did not look to the code yet; so some comments might just be dum).

src/stdlib_sparse_kinds.fypp
```diff
@@ -0,0 +1,894 @@
...
+    enum, bind(C)
+        enumerator :: sparse_full  !> Full Sparse matrix (no symmetry considerations)
+        enumerator :: sparse_lower !> Symmetric Sparse matrix with triangular inferior storage
+        enumerator :: sparse_upper !> Symmetric Sparse matrix with triangular supperior storage
+    end enum
+    public :: sparse_full, sparse_lower, sparse_upper
+
+    #:for k1, t1, s1 in (R_KINDS_TYPES)
+    ${t1}$, parameter, public :: zero_${s1}$ = 0._${k1}$
+    ${t1}$, parameter, public :: one_${s1}$ = 1._${k1}$
```
jvdp1, line 24: Could you keep them private? setting them public might generate some issues on the user 's side, as many of us define these values in our own code.

src/stdlib_sparse_kinds.fypp
```diff
@@ -0,0 +1,894 @@
...
+    #:endfor
+
+    ! -- Classes
+
+    !> Base sparse type holding the meta data related to the storage capacity of a matrix.
+    type, public, abstract :: sparse_type
+      integer :: nrows = 0 !> number of rows
+      integer :: ncols = 0 !> number of columns
+      integer :: nnz   = 0 !> number of non-zero values
+      integer :: storage = sparse_full !> assumed storage symmetry
```
jvdp1, line 38:
```suggestion
      integer :: nrows = 0 !! number of rows
      integer :: ncols = 0 !! number of columns
      integer :: nnz   = 0 !! number of non-zero values
      integer :: storage = sparse_full !! assumed storage symmetry
```

src/stdlib_sparse_kinds.fypp
```diff
@@ -0,0 +1,894 @@
...
+    end type
+
+    !> COO: COOrdinates compresed format 
+    type, public, extends(sparse_type) :: COO_type
+      logical               :: is_sorted = .false. !> whether the matrix is sorted or not
+      integer, allocatable  :: index(:,:) !> Matrix coordinates index(2,nnz)
```
jvdp1, line 44:
```suggestion
      logical               :: is_sorted = .false. !! whether the matrix is sorted or not
      integer, allocatable  :: index(:,:) !! Matrix coordinates index(2,nnz)
```

src/stdlib_sparse_kinds.fypp
```diff
@@ -0,0 +1,894 @@
...
+        procedure, non_overridable :: add_value => add_value_coo_${s1}$
+        procedure, non_overridable :: add_block => add_block_coo_${s1}$
+        generic :: add => add_value, add_block
+    end type
+    #:endfor
+
+    !> CSR: Compressed sparse row or Yale format
+    type, public, extends(sparse_type) :: CSR_type  
+      integer, allocatable  :: col(:)    !> matrix column pointer
+      integer, allocatable  :: rowptr(:) !> matrix row pointer
```
jvdp1, line 63:
```suggestion
      integer, allocatable  :: col(:)    !! matrix column pointer
      integer, allocatable  :: rowptr(:) !! matrix row pointer
```

src/stdlib_sparse_kinds.fypp
```diff
@@ -0,0 +1,894 @@
...
+
+    !> Conversion from dense to coo
+    !> 
+    !> Enables extracting the non-zero elements of a dense 2D matrix and
+    !> storing those values in a COO format. The coo matrix is (re)allocated on the fly.
+    interface dense2coo
+        #:for k1, t1, s1 in (KINDS_TYPES)
+        module subroutine dense2coo_${s1}$(dense,COO)
+            ${t1}$, intent(in) :: dense(:,:)
+            type(COO_${s1}$), intent(inout) :: COO
```
jvdp1, line 144: Could it be:
```suggestion
            type(COO_${s1}$), intent(out) :: COO
```

src/stdlib_sparse_kinds.fypp
```diff
@@ -0,0 +1,894 @@
...
+    !> COO: COOrdinates compresed format 
+    type, public, extends(sparse_type) :: COO_type
+      logical               :: is_sorted = .false. !> whether the matrix is sorted or not
+      integer, allocatable  :: index(:,:) !> Matrix coordinates index(2,nnz)
+    contains
+      procedure :: malloc => malloc_coo
+    end type
+
+    #:for k1, t1, s1 in (KINDS_TYPES)
+    type, public, extends(COO_type) :: COO_${s1}$
```
jvdp1, line 50: Just to be consistent with rest of `stdlib`:
```suggestion
    type, public, extends(COO_type) :: COO_${s1}$_type
```

However, I will agree that it will become a bit long.

src/stdlib_sparse_kinds.fypp
```diff
@@ -0,0 +1,894 @@
...
+        ${t1}$, allocatable :: data(:) 
+    contains
+        procedure, non_overridable :: at => at_value_csc_${s1}$
+        procedure, non_overridable :: add_value => add_value_csc_${s1}$
+        procedure, non_overridable :: add_block => add_block_csc_${s1}$
+        generic :: add => add_value, add_block
+    end type
+    #:endfor
+  
+    !> Compressed ELLPACK
```
jvdp1, line 98: For all, info on version and link to specs should be added.

src/stdlib_sparse_kinds.fypp
```diff
@@ -0,0 +1,894 @@
...
+        ${t1}$, allocatable :: data(:,:)
+    contains
+        procedure, non_overridable :: at => at_value_sellc_${s1}$
+        procedure, non_overridable :: add_value => add_value_sellc_${s1}$
+        procedure, non_overridable :: add_block => add_block_sellc_${s1}$
+        generic :: add => add_value, add_block
+    end type
+    #:endfor
+
+    !> Conversion from dense to coo
```
jvdp1, line 136: I suggest to move all these conversion procedures to its own module.

jvdp1 (commented):
here are some additional quick comments.

doc/specs/stdlib_sparse.md
```diff
@@ -0,0 +1,305 @@
...
+#### Description
+The parent `sparse_type` is as an abstract derived type holding the basic common meta data needed to define a sparse matrix, as well as shared APIs. All sparse matrix flavors are extended from the `sparse_type`.
+
+```Fortran
+type, public, abstract :: sparse_type
+    integer :: nrows   !! number of rows
+    integer :: ncols   !> number of columns
+    integer :: nnz     !> number of non-zero values
+    integer :: storage !> assumed storage symmetry
+end type
```
jvdp1, line 30: Just a quick thought: it might be of interest to have two DTs with different kinds for the integers, e.g.,
```fortran
type,...... :: sparse_int32_type
 ....
 integer(int32) :: nnz
 ...
end type
type,...... :: sparse_int64_type
 ....
 integer(int64) :: nnz
 ...
end type
````

File: src/stdlib_sparse_conversion.fypp (hunk 10 above)
Comment (jvdp1) [reply src/stdlib_sparse_conversion.fypp:244]: See #849 for such an implementation

File: doc/specs/stdlib_sparse.md (hunk 19 above)
Comment (perazz) [reply doc/specs/stdlib_sparse.md:30]: I'm not sure how hard it would be to accomplish this. Maybe in the current PR it would be enough to generalize the index `kind` from default `integer` to `integer(ik)`, so that later this extension will be easier. 

What you're suggesting is probably that there should be two `abstract` sparse types, one per length `kind`: 

```Fortran
type, public, abstract :: sparse_type

end type
#:for ik,it in INT_KINDS_TYPES
type, public, abstract, extends(sparse_type) :: sparse_${ik}$_type
    ${it}$ :: nrows   !! number of rows
    ${it}$ :: ncols   !> number of columns
    ${it}$ :: nnz     !> number of non-zero values
    ${it}$ :: storage !> assumed storage symmetry
end type sparse_${ik}$_type
!:endfor
```

jvdp1: Indeed, I had something similar to your proposition in mind. However, I agree that it should be for a following PR.

doc/specs/stdlib_sparse.md
```diff
@@ -0,0 +1,305 @@
...
+### `SELL-C`: The Sliced ELLPACK with Constant blocks format
+#### Status
+
+Experimental
+
+#### Description
+The Sliced ELLPACK format `SELLC` is a variation of the `ELLPACK` format. This modification reduces the storage size compared to the `ELLPACK` format but maintaining its efficient data access scheme. It can be seen as an intermediate format between `CSR` and `ELLPACK`. For more details read [the reference](https://arxiv.org/pdf/1307.6209v1)
+
+<!-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -->
+## `add`/`at` - Sparse Matrix data accessors
```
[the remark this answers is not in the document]
jalvesz: Splitted it into two subsections, one for `add` and another for `at`

doc/specs/stdlib_sparse.md
```diff
@@ -0,0 +1,305 @@
...
+
+### Status
+
+Experimental
+
+### Description
+Type-bound procedures to enable adding or requesting data in/from a sparse matrix.
```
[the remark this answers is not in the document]
jalvesz: It just adds values into pre-existing entries in a given sparse pattern.

doc/specs/stdlib_sparse.md
```diff
@@ -0,0 +1,305 @@
...
+
+### Syntax
+
+`call matrix%add(i,j,v)` or
+`call matrix%add(i(:),j(:),v(:,:))`
+
+### Arguments
+
+`i`, `intent(in)`: Shall be an integer value or rank-1 array.
```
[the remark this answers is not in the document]
jalvesz: done in https://github.com/fortran-lang/stdlib/pull/760/commits/dde88a732d67a0b0bd4425d7734ffeccc53e28d0

doc/specs/stdlib_sparse.md
```diff
@@ -0,0 +1,305 @@
...
+`j`, `intent(in)`: Shall be an integer value or rank-1 array.
+`v`, `intent(in)`: Shall be a `real` or `complex` value or rank-2 array. The type shall be in accordance to the declared sparse matrix object.
+
+### Syntax
```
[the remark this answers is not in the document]
jalvesz: done in https://github.com/fortran-lang/stdlib/pull/760/commits/dde88a732d67a0b0bd4425d7734ffeccc53e28d0

doc/specs/stdlib_sparse.md
```diff
@@ -0,0 +1,305 @@
...
+
```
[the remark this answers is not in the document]
jalvesz: `at` returns a value from the sparse pattern. Just like from a dense matrix one would request a value by `val=dense(i,j)`, here we propose `val = sparse%at(i,j)` to mimic a dense notation.

doc/specs/stdlib_sparse.md
```diff
@@ -0,0 +1,305 @@
...
+`v = matrix%at(i,j)`
+
+### Arguments
+
+`i`, `intent(in)` : Shall be an integer value.
+`j`, `intent(in)` : Shall be an integer value.
+`v`, `result`     : Shall be a `real` or `complex` value in accordance to the declared sparse matrix object. If the `ij` tuple is within the sparse pattern, `v` returns the value in the data buffer. If the `ij` tuple is outside the sparse pattern, `v` returns `0`. If the `ij` tuple is outside the matrix pattern `(nrows,ncols)`, `v` returns `NaN`.
```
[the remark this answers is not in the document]
jalvesz: done in https://github.com/fortran-lang/stdlib/pull/760/commits/dde88a732d67a0b0bd4425d7734ffeccc53e28d0

doc/specs/stdlib_sparse.md
```diff
@@ -0,0 +1,305 @@
...
+
+### Syntax
+
+`call ` [[stdlib_sparse_conversion(module):dense2coo(interface)]] `(dense,coo)`
+
+### Arguments
+
+`dense`, `intent(in)`: Shall be a rank-2 array of `real` or `complex` type.
+
+`coo`, `intent(inout)`: Shall be a `COO` type of `real` or `complex` type.
```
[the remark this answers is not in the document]
jalvesz: change to `intent(out)` in https://github.com/fortran-lang/stdlib/pull/760/commits/dde88a732d67a0b0bd4425d7734ffeccc53e28d0 for all conversion procedures except for coo2ordered.

src/stdlib_sparse_kinds.fypp
jalvesz: done in https://github.com/fortran-lang/stdlib/pull/760/commits/dde88a732d67a0b0bd4425d7734ffeccc53e28d0

src/stdlib_sparse_kinds.fypp
jalvesz: Added the references to the specs in https://github.com/fortran-lang/stdlib/pull/760/commits/dde88a732d67a0b0bd4425d7734ffeccc53e28d0

src/stdlib_sparse_kinds.fypp
jalvesz: These are just module interfaces, the procedures are defined in their own submodule

doc/specs/stdlib_sparse.md
jalvesz: I get the feeling that adding this `COO_${s1}$_type` plus then the different integer sizes would end up with declarations such as:
```Fortran
type(CSR_int32_dp_type) :: mymatrix
```
wich I find very verbose. The target design I was striving for is to have these two things being syntactically close from a hierarchical point of view:
```Fortran
real(dp), allocatable :: dense(:,:)
type(CSR_dp) :: sparse
```
Is there a possibility to find a middle ground?

doc/specs/stdlib_sparse.md
```diff
@@ -0,0 +1,305 @@
...
+<!-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -->
+## `sparse_conversion` - Sparse matrix to matrix conversions
+
+### Status
+
+Experimental
+
+### Description
+
+This module provides facility functions for converting between storage formats.
```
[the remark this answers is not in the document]
jalvesz: There are several conversion procedures within the module

jalvesz:
@certik I understand and agree that a low level API is also desirable. Ideally we should eventually be able to link agains MKL following their interfaces which are quite similar but not exactly to Sparse BLAS syntax: https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-fortran/2024-0/sparse-blas-level-2-and-level-3-routines-002.html

Now, this is a huge work. My feeling is that it is doable by changing the low-level back-ends without having to modify the high-level API. I tried to bring this proposal to kickoff the interest and see if that could happen eventually and progressively once the DTs are in shape. Also, the idea of the DTs is to have sparse objects at the same level as a dense array such that higher-level interfaces can then be designed to work on either dense or sparse matrices.

doc/specs/stdlib_sparse.md
jalvesz: In https://github.com/fortran-lang/stdlib/pull/760/commits/6ae038bd54ab9927aec20bb0787ba89fa5769c17 I added a `ilp` parameter following the linalg module to enable recompiling using int64. Rightnow it is set to int32, we could add a preprocessing macro to change it.

doc/specs/stdlib_sparse.md
```diff
@@ -0,0 +1,305 @@
...
+#### Description
+The parent `sparse_type` is as an abstract derived type holding the basic common meta data needed to define a sparse matrix, as well as shared APIs. All sparse matrix flavors are extended from the `sparse_type`.
+
+```Fortran
+type, public, abstract :: sparse_type
+    integer :: nrows   !> number of rows
+    integer :: ncols   !> number of columns
+    integer :: nnz     !> number of non-zero values
+    integer :: storage !> assumed storage symmetry
+end type
```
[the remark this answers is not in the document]
jalvesz: I didn't see the need as the DTs only have simple integer scalars or allocatable arrays as members. My understanding is that such types are normally well destroyed by the compilers.

doc/specs/stdlib_sparse.md
perazz: > possibility to find a middle ground

I agree with you @jalvesz. @jvdp1's idea is correct, but maybe, who needs two index types when you can have the largest one already? In other words, could we just require that the index type for sparse matrices be a 64-bit one and define  

```
integer, parameter :: index_t = int64
```

To support 32-bit index types, one idea could be to just define the interface on the abstract type: 

```
type, abstract :: sparse
 ...
contains
   procedure(get_elem), deferred, abstract :: get_int64
   procedure :: get_int32 ! implemented, just calls get_int64
   generic :: get => get_int32, get_int64
end type
...
elemental real(dp) function get_int32(this,i,j)
    class(sparse), intent(in) :: this
    integer(int32), intent(in) :: i,j
    get_int32 = this%get(int(i,int64),int(j,int64))
end function
```

jalvesz: >In other words, could we just require that the index type for sparse matrices be always a 64-bit

I'm very hesitant about this. given that with int32 one can have `nnz = 2,147,483,647` this means that, for a COO matrix of double precision values of such size, one would have 2147483647 *( 2 * 4 + 8 ) / (1024)^3 = 23 Gb of memory. Transforming the index arrays to int64 => 2147483647 *( 2 * 8 + 8 ) / (1024)^3 = 39 Gb ... And all the loops will be done by searching on larger sized indexing arrays, thus plummetting performance (do not know to which extent). Also, MKL uses int32 for the kind definition of index arrays. Shall we set a default to int64 this would make it more complicated to eventually bind against it.

Thinking about this makes me more inclined for either allowing recompilation by changing the `ilp` parameter, or indeed adding support for both sizes with separate kinds. I would in that case let the int32 with a simplified syntax and the int64 one with something like `long` added as a suffix.

perazz: I agree with you @jalvesz: at this stage, probably the best thing to do is just to generalize the integer kind: 

> from default `integer` to `integer(ik)`

so that the implementation already supports any integer `kind`s, but a more thoughtful decision about what should be used and how flexible should that be, is deferred.

src/stdlib_sparse_kinds.fypp
jvdp1: Could these module interfaces be removed from `stdlib_sparse_kinds`, and just be in their own module (and not submodule)? What is the advantage of having them in a submodule?

doc/specs/stdlib_sparse.md
jvdp1: > could we just require that the index type for sparse matrices be always a 64-bit one

The default kind should be `integer(int32)`, at least for linking them or using other libraries (like MKL Sparse BLAS).

At this stage, using `integer(ik)` is fine. There is enough WIP without this feature.

src/stdlib_sparse_kinds.fypp
jalvesz: rolled back on the use of submodules [c8d94a3](https://github.com/fortran-lang/stdlib/pull/760/commits/c8d94a3e2544da0f1d40f21967ac424c751c159a) ... I used the submodules to share the constants, they are now in a separated module.

jalvesz:
Updates: added in-place transpose (and conjugate transpose) for the spmv kernels

perazz (approved):
Thank you @jalvesz for the update. LGTM as the implementation is very clean and it looks like you've addressed all previous comments. I would like this PR to be merged before new features are added, as it's becoming harder to track the new additions.

So I suggest any others that have comments/edits to please do so, and maybe we should think of merging soon if no further comments arise.

jalvesz:
Thanks @perazz for the feedback. I did a few last cleanups and enhanced the sellc format support. On my side I also think the PR is now ready to be merged if there are no further comments. Future features/improvements could be addressed with separate PRs having this base available.

jvdp1 (approved):
Thank you @jalvesz for this PR. LGTM. I agree woth @perazz that this first version is ready to be merged

perazz:
Sounds good @jvdp1 @jalvesz - let's wait another couple of days, and then I will merge.

jalvesz:
In https://github.com/fortran-lang/stdlib/pull/760/commits/22cd23e88d57007fa2dc24421c7bae21531f54d5 I rollback the use of local associate construct for the SELLC spmv because after discussion with @ivan-pi I realized that it hinders optimizations. I saw drops in performance of about 40-60% compared to using the internal subroutine kernels.

jalvesz:
@liaoweiyang2017 I invite you to look at #1082 and give your feedback.

liaoweiyang2017:
@jalvesz Thank you for the guidance!

I'll prepare a PR for `spmm` following the `gemm`/`spmv` pattern.

**Planned implementation:**
- `spmm(A, B, C, [alpha], [beta], [op_A], [op_B])` interface
- CSR×CSR (Gustavson's algorithm), CSR×dense, dense×CSR variants
- Comprehensive test suite
- Documentation and examples

I'll have a draft PR ready within the next few days.

Thanks!

PR #773, opened 2024-03-14
Fix aliased `string` `move` interface

### LINKED ISSUE #744: `string_intrinsic (Subprocess aborted)` failed
### Description

After building when I run tests using `cmake --build build --target test`

```console
% cmake --build build --target test
.....
99% tests passed, 1 tests failed out of 286

Label Time Summary:
quadruple_precision    =   0.13 sec*proc (2 tests)

Total Test time (real) =   3.13 sec

	  1 - always_skip (Skipped)
	101 - check4 (Skipped)

The following tests FAILED:
	 45 - string_intrinsic (Subprocess aborted)
Errors while running CTest
Output from these tests are in: /Users/pranavchiku/repos/stdlib-fortran-lang/build/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
make: *** [test] Error 8
```

### Expected Behaviour

Tests shall pass.

### Version of stdlib

https://github.com/fortran-lang/stdlib/commit/b8fbb3ce1aaf2fd1729897739fcdae8f6b3c4ef4

### Platform and Architecture

MacOS

### Additional Information

_No response_

perazz:
Fixes #744 (previous fix: 

The following line is causing a segfault on macOS, gfortran 13.2.0, hence failed test: 

https://github.com/fortran-lang/stdlib/blob/b8fbb3ce1aaf2fd1729897739fcdae8f6b3c4ef4/test/string/test_string_intrinsic.f90#L717

This line is _invalid Fortran_ because it violates the no-alising rule (that is one of the few reasons Fortran can optimize better than C): two dummy arguments passed by reference cannot point to the same address. So probably, gfortran optimizes out the temporary variable which then causes trouble. 

My proposed solution is to check the address first, and avoid the dangerous operation if the variable is the same.

### DIFF
--- a/src/stdlib_string_type.fypp
+++ b/src/stdlib_string_type.fypp
@@ -17,6 +17,7 @@ module stdlib_string_type
        & to_title_ => to_title, to_sentence_ => to_sentence, reverse_ => reverse
     use stdlib_kinds, only : int8, int16, int32, int64, lk, c_bool
     use stdlib_optval, only: optval
+    use iso_c_binding, only: c_loc,c_associated
     implicit none
     private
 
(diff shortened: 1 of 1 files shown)


awvwgk (approved):
That's a great usage of pointer association and targets. I think we can work with that.

jvdp1 (approved):
Nice solution! Thank you @perazz

perazz:
Thank you Sebastian and Jeremie, I'm going to wait another few days, then merge if there are no more comments.

PR #778, opened 2024-03-27
Update to include int32 hashmap keytype

chuckyvt:
This is a fairly minor update to include int32 key types to hash maps.    I have tested this a good bit in my own code.  It has worked well and is a nice addition to the hashmap capabilities.


src/stdlib_hashmap_wrappers.f90
```diff
@@ -101,7 +101,8 @@ end function hasher_fun
 
         module procedure set_char_key,   &
                          set_int8_key,   &
-                         set_other
+                         set_other,      &
+                         set_int32_key
```
jvdp1, line 105:
```suggestion
                         set_int32_key, &
                         set_other
```

src/stdlib_hashmap_wrappers.f90
```diff
@@ -308,6 +309,21 @@ subroutine set_other( other, value )
     end subroutine set_other
 
 
+    subroutine set_int32_key( key, value )
```
jvdp1, line 312: Could it be moved afet `set_int_8_key`?

jvdp1 (commented):
Thank you @chuckyvt. This is indeed a useful procedure to avoid multiple `...transfer(value, 1_int8)....` with the default `integer` on the user side.
Could you also udpate the specs, please?
Could you also add a test for this specific procedure, please?
Should it be a corresponding `get` function (quite a long time I did use `hashmap`, so this question might be irrelevant)?

chuckyvt:
In progress code updates.   Added a 'get_int32_key' function to match set function.

src/stdlib_hashmap_wrappers.f90
```diff
@@ -277,6 +279,21 @@ subroutine get_int8_key( key, value )
     end subroutine get_int8_key
 
 
+    subroutine get_int32_key( key, value )
+!! Version: Experimental
+!!
+!! Gets the contents of the key as an INTEGER(INT8) vector
```
gareth-nx, line 285: Should this comment have `INTEGER(INT32)` rather than `INTEGER(INT8)` ?

src/stdlib_hashmap_wrappers.f90
```diff
@@ -323,6 +340,21 @@ subroutine set_int8_key( key, value )
     end subroutine set_int8_key
 
 
+    subroutine set_int32_key( key, value )
+!! Version: Experimental
+!!
+!! Sets the contents of the key from an INTEGER(INT8) vector
```
gareth-nx, line 346: Should this comment have `INTEGER(INT32)`?

gareth-nx (commented):
Thanks for this. I agree that some tests are needed and understand this is in progress. But it looks like a nice addition.

src/stdlib_hashmap_wrappers.f90
chuckyvt: This should be updated per latest revision

src/stdlib_hashmap_wrappers.f90
chuckyvt: This should be updated per latest revision.

chuckyvt:
Requested code updates are complete.  Also updated the specs document.   The testing item is a bit more work however.  I could use some help or suggestions on the best approach.  All the current tests are based around native int8 keys.   For example, there are no test cases I can find for the existing character strings?   My thought was it might make sense to add an extract test file dedicated to non int8 keys.  Character key, int32 key, and could be extended in the future.

gareth-nx:
Thanks @chuckyvt. The addition of extra tests sounds good. Indeed, we will have to add tests for other reasons too (not necessarily in this pull request), see https://github.com/fortran-lang/stdlib/issues/785.

chuckyvt:
Have pushed an updated to test_maps.fypp that includes tests for int32 and character key types, in addition to the existing base int8 keys.   I'm not very familiar with the test drive framework, so I tried to stay as close as possible to the existing testing layout.

test/hashmaps/test_maps.fypp
```diff
@@ -53,10 +54,10 @@ contains
             type(error_type), allocatable, intent(out) :: error
 
             type(chaining_hashmap_type)   :: map
-            integer(int8)             :: test_8_bits(test_size)
+            integer(int8)             :: test_8_bits(test_size,key_types)
 
             call generate_vector(test_8_bits)
-
+
```
jvdp1, line 60:
```suggestion

```

test/hashmaps/test_maps.fypp
```diff
@@ -53,10 +54,10 @@ contains
             type(error_type), allocatable, intent(out) :: error
 
             type(chaining_hashmap_type)   :: map
-            integer(int8)             :: test_8_bits(test_size)
+            integer(int8)             :: test_8_bits(test_size,key_types)
 
             call generate_vector(test_8_bits)
```
jvdp1, line 59: `test_8_bits` is not a vector anymore.
Could you explain why `test_8_bits`  is changed to an array? I am not sure to understand the aim here.

test/hashmaps/test_maps.fypp
```diff
@@ -108,32 +111,54 @@ contains
...
+        do index2=1, test_size, test_block
+            call set( key, test_8_bits( index2:index2+test_block-1, 1 ) )
+            call map % key_test( key, present )
+            call check(error, present, "Int8 KEY not found in map KEY_TEST.")
+
+            call set( key, transfer( test_8_bits( index2:index2+test_block-1, 2 ), [0_int32] ) )
+            call map % key_test( key, present )
+            call check(error, present, "Int32 KEY not found in map KEY_TEST.")
+
+            call set( key, to_string( transfer( test_8_bits( index2:index2+test_block-1, 3 ), 0_int32 ) ) )
```
jvdp1, line 158: Could you add a comment to explain the different tests (and their purposes)? It might help reviewers and future developers.

test/hashmaps/test_maps.fypp
```diff
@@ -142,34 +167,50 @@ contains
...
         integer(int_index), intent(in)          :: test_block
         integer :: index2
         type(key_type) :: key
         type(other_type) :: other
         logical :: exists
 
-        do index2=1, size(test_8_bits), test_block
-            call set( key, test_8_bits( index2:index2+test_block-1 ) )
+        do index2=1, test_size, test_block
+            call set( key, test_8_bits( index2:index2+test_block-1, 1 ) )
```
jvdp1, line 178: Why did you use different columns of `test_8_bits`? Is it not possible to use the same values for all tests?

jvdp1 (commented):
Thank you @chuckyvt for adding the tests. I am a bit puzzled by your approach (but it does not mean it is wrong). Could you add a few comments to help the readers and future developers, please? MY main question is to understand why `test_8_bits` was changed to an array (see comments below).

test/hashmaps/test_maps.fypp
chuckyvt: The approach is to create a unique int8 vector for each key type that can then be used via transfer or other mechanism to generate a unique key hash for the test program.   With this approach, using a single int8 vector will result in duplicate keys and mapping conflicts, so that must be managed.   There are other approaches, but this seemed to be the easiest additional and extensible to future key types.  For example, we could add a dedicated int32 vector generation function, but not sure how that improves on using transfer with an int8 array to create an int32 array, and would require a dedicated function for each key type.  (So for example if there is want to add int64 or real key types, additional functions would need to be written).

chuckyvt: Its also worth noting that the structure of test_maps.fypp is not ideal with the code base being repeated twice for open and chaining maps.  I believe refactoring this code into a single source approach should be done before any additional function additions.

test/hashmaps/test_maps.fypp
chuckyvt: I've added some additional comments per discussion above.   Feel free to add more.   I've also removed the character key type test.  The code worked fine on my computer, but failed the CI checks for reasons I don't understand.  It was a nice to have, but out of scope for this pull request which is for int32 key types.   

Also, per the note above, I think test_maps.fypp should be refactored to remove duplicate code prior to adding other key type checks to make code addition less error prone.

src/stdlib_hashmap_wrappers.f90
```diff
@@ -277,6 +279,21 @@ subroutine get_int8_key( key, value )
...
+!! Version: Experimental
+!!
+!! Gets the contents of the key as an INTEGER(INT32) vector
+!! Arguments:
+!!     key   - the input key
+!!     value - the contents of key mapped to an INTEGER(INT32) vector
+        type(key_type), intent(in)              :: key
+        integer(int32), allocatable, intent(out) :: value(:)
+        
+        value = transfer( key % value, value )
```
gareth-nx, line 292: Here I wonder whether it matters that `value` is unallocated in the RHS call to `transfer` (since it will be deallocated on input if allocated)?

src/stdlib_hashmap_wrappers.f90
```diff
@@ -323,6 +340,21 @@ subroutine set_int8_key( key, value )
...
 
 
+    subroutine set_int32_key( key, value )
+!! Version: Experimental
+!!
+!! Sets the contents of the key from an INTEGER(INT32) vector
+!! Arguments:
+!!     key   - the output key
+!!     value - the input INTEGER(INT32) vector
+                        type(key_type), intent(out) :: key
```
gareth-nx, line 350: Remove extra indentation here

src/stdlib_hashmap_wrappers.f90
```diff
@@ -323,6 +340,21 @@ subroutine set_int8_key( key, value )
...
+                        integer(int32), intent(in)   :: value(:)
+                
+                        key % value = transfer(value, key % value)
```
gareth-nx, line 353: See question about `transfer` above.

test/hashmaps/test_maps.fypp
```diff
@@ -24,7 +24,7 @@ module test_stdlib_chaining_maps
     integer, parameter        :: test_size = rand_size*4
     integer, parameter        :: test_16 = 2**4
     integer, parameter        :: test_256 = 2**8
-
+    integer, parameter        :: key_types = 2
```
gareth-nx, line 27: Here could you add a comment along the lines of `! The key types are int8 and int32`. I think that's the idea?

test/hashmaps/test_maps.fypp
gareth-nx: Hopefully this is straightforward with `fypp`.

test/hashmaps/test_maps.fypp
```diff
@@ -249,6 +275,7 @@ module test_stdlib_open_maps
     integer, parameter        :: test_size = rand_size*4
     integer, parameter        :: test_16 = 2**4
     integer, parameter        :: test_256 = 2**8
+    integer, parameter        :: key_types = 2
```
gareth-nx, line 278: Please add a comment here along the lines of `! The tested key types are int8 and int32`

gareth-nx (commented):
This looks good. I have one query though about `transfer` when the `mold` argument is not allocated -- is this standard? I wonder if this could be related to the issue with strings as keys that you mentioned earlier.

src/stdlib_hashmap_wrappers.f90
gareth-nx: I've asked this question at fortran discourse [here](https://fortran-lang.discourse.group/t/using-transfer-with-unallocated-mold/7867).

chuckyvt: Thanks for asking.  Intuitively it seems like it shouldn't matter since its only pulling type info, not size.  But I also did a quick search and didn't come up with anything.   We could hardcode it to [0_int8], but I don't like that quite as much.

src/stdlib_hashmap_wrappers.f90
chuckyvt: Included in latest submit.

test/hashmaps/test_maps.fypp
chuckyvt: Tried to add in comments to explain this.

test/hashmaps/test_maps.fypp
chuckyvt: Also added.

chuckyvt:
Note with the latest submit I added pure to the int32 get and set subroutines.   Eventually I think adding this to the other interfaces makes sense but is probably a separate pull request.  (Also can make sure there are no issues with pure with the int32 type)

test/hashmaps/test_maps.fypp
jvdp1: > Its also worth noting that the structure of test_maps.fypp is not ideal with the code base being repeated for open and chaining maps. I believe refactoring this code into a single source approach should be done before any additional function additions.

Could you open an issue for this, please?

test/hashmaps/test_maps.fypp
jvdp1: > I've added some additional comments per discussion above.

Great. Thank you.

> I've also removed the character key type test. The code worked fine on my computer, but failed the CI checks for reasons I don't understand. It was a nice to have, but out of scope for this pull request which is for int32 key types.

I agree with you. Let's keep this PR focused on `int32`

jvdp1 (approved):
LGTM. Pending the issue with `transfer`, it can be merged IMO. Thank you @chuckyvt for this PR.

gareth-nx (approved):

gareth-nx:
I'm also happy for this to be merged. It looks like the use of `transfer` with an unallocated mold is OK. 

Do we have a record of what usage was causing segfaults? Might come in handy.

jvdp1:
I'll merge this PR. Thank you @chuckyvt

PR #780, opened 2024-04-02
add implicit none to example programs that do not have it

gareth-nx:
This adds `implicit none` to a number of programs in the `example` folder that otherwise don't have it. 

While not essential, the programs are used in library documentation, and IMO it's better to demonstrate good practice.

I have run the tests using `cmake --build build --target test` and they all appear to pass.

### DIFF
--- a/example/bitsets/example_bitsets_all.f90
+++ b/example/bitsets/example_bitsets_all.f90
@@ -1,5 +1,6 @@
 program example_all
   use stdlib_bitsets
+  implicit none
   character(*), parameter :: &
     bits_all = '111111111111111111111111111111111'
   type(bitset_64) :: set0
--- a/example/bitsets/example_bitsets_and.f90
+++ b/example/bitsets/example_bitsets_and.f90
@@ -1,5 +1,6 @@
 program example_and
   use stdlib_bitsets
+  implicit none
   type(bitset_large) :: set0, set1
   call set0%init(166)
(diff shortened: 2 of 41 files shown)


jvdp1 (approved):
Thank you @gareth-nx . I agree that it is better to demonstrate good practice.
I wonder if it might be even better to use `implicit none (type, external)` rather `implicit none`. Or would it exclude too many compilers that do not support this more recent feature?

gareth-nx:
Thanks for the review @jvdp1.

I actually haven't moved to using `implicit none(type, external)` in my own programs.  

From [this link](https://fortran-lang.discourse.group/t/what-does-implicit-none-type-external-do/497) I understand that the main difference relates to external functions and subroutines. 

In the context of the example programs, these are probably moot points.  But so was the absence of `implicit none`, as it turned out. 

The bigger question is whether we should uniformly encourage `implicit none (type, external)` as the default, rather than just `implicit none`.  

I'm not sure. But maybe if there are doubts about compiler support, we can just leave it for now?

jvdp1:
> I'm not sure. But maybe if there are doubts about compiler support, we can just leave it for now?

OK. We can leave it as it is now.

gareth-nx:
No worries. I'll wait a while in case there are other suggestions. But likely this can be merged as is.

perazz (approved):

PR #781, opened 2024-04-02
Add brief explanation on how to include stdlib in a Makefile

jannisteunissen:
The current documentation on how to use stdlib in a project focuses on Cmake, see https://github.com/fortran-lang/stdlib?tab=readme-ov-file#using-stdlib-in-your-project

It can seem as is you actually need CMAKE to include stdlib (I got this impression). I think it would also be helpful to describe how to include stdlib in a regular makefile. This small change includes such documentation. I was not 100% sure about the paths: is it correct that the library is always in <build_folder>/src?

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,6 +217,8 @@ stdlib = { git="https://github.com/fortran-lang/stdlib", branch="stdlib-fpm" }
 
 ## Using stdlib in your project
 
+### Using stdlib with CMake
+
 The stdlib project exports CMake package files and pkg-config files to make stdlib usable for other projects.
 The package files are located in the library directory in the installation prefix.
 
@@ -234,6 +236,18 @@ target_link_libraries(
 
 To make the installed stdlib project discoverable add the stdlib directory to the ``CMAKE_PREFIX_PATH``.
 The usual install location of the package files is ``$PREFIX/lib/cmake/fortran_stdlib``.
+
+### Using stdlib with a regular Makefile
+
+The library can be included in a regular Makefile, for example with
+```make
+STDLIB_PATH := path/to/build
+LIBDIRS := $(STDLIB_PATH)/src
+INCDIRS := $(STDLIB_PATH)/src/mod_files
+...
+```
+Here it is assumed that the compiler will look for libraries in `LIBDIRS` and for module files in `INCDIRS`.
+A flag `-lfortran_stdlib` should be included to link against `libfortran_stdlib.a` or `libfortran_stdlib.so`.
 
 ## Documentation
 
--- a/README.md
+++ b/README.md
@@ -242,18 +242,47 @@ stdlib = { git="https://github.com/fortran-lang/stdlib", branch="stdlib-fpm" }
 ### Using stdlib with a regular Makefile
 
 After the library has been built, it can be included in a regular Makefile.
+The recommended way to do this is using the [pkg-config](https://www.freedesktop.org/wiki/Software/pkg-config/) tool, for which an example is shown below.
+```make
+# Necessary if the installation directory is not in PKG_CONFIG_PATH
+install_dir := path/to/install_dir
+export PKG_CONFIG_PATH := $(install_dir)/lib/pkgconfig:$(PKG_CONFIG_PATH)
+
+STDLIB_CFLAGS := `pkg-config --cflags fortran_stdlib`
+STDLIB_LIBS := `pkg-config --libs fortran_stdlib`
+
+# Example definition of Fortran compiler and flags
+FC := gfortran
+FFLAGS := -O2 -Wall -g
+
+# Definition of targets etc.
+...
+
+# Example rule to compile object files from .f90 files
+%.o: %.f90
+    $(FC) -c -o $@ $< $(FFLAGS) $(STDLIB_CFLAGS)
+
+# Example rule to link an executable from object files
+%: %.o
+    $(FC) -o $@ $^ $(FFLAGS) $(STDLIB_LIBS)
+
+```
+
+The same can also be achieved without pkg-config.
 If the library has been installed in a directory inside the compiler's search path,
-only a flag `-lfortran_stdlib` is required to link against `libfortran_stdlib.a` or `libfortran_stdlib.so`.
+only a flag `-lfortran_stdlib` is required.
 If the installation directory is not in the compiler's search path, one can add for example
 ```make
-libdir=${install_dir}/lib
-moduledir=${install_dir}/include/fortran_stdlib/<compiler name and version>
+install_dir := path/to/install_dir
+libdir := $(install_dir)/lib
+moduledir := $(install_dir)/include/fortran_stdlib/<compiler name and version>
 ```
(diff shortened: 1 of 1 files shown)


jvdp1 (approved):
Thank you @jannisteunissen . LGTM. It might be good to move also the lines 204-216 in this section, such that the section "Using stdlib in your project" contains the directives for CMake, Make, and fpm.

perazz:
Apologies I'm no pkg-config expert. However, I think the reader may think that stdlib can be built directly using a makefile with this PR. This is not true: the standard procedure is: 
1) build it with CMake <- necessary anyways
2) install it to a desired folder (cmake --install . --prefix=/path/to/my/installation)
3) use/link against it 

This PR is about step 3), but it reports valid paths as after step 2), i.e., it points to the build folder. 
I think this may be confusing although probably correct (I haven't tested it). After install, we have a different scenario: 

A) after a CMake build+install, folders are like 
https://github.com/fortran-lang/stdlib/blob/e88daa336865b0a4a087b3ea2a15c1d1e94debc5/config/template.pc#L2-L4

i.e., for example: 
```pkg-config
libdir=${prefix}/lib
includedir=${prefix}/include
moduledir=${prefix}/include/fortran_stdlib/GNU-13.2.0
```

B) an stdlib-fpm install is not currently supported and I think it's a good idea to have, I will open a PR for this.

perazz:
I would not worry about the general include folder, as I don't think that stdlib is ever going to include C headers

README.md
```diff
@@ -235,6 +223,39 @@ target_link_libraries(
 To make the installed stdlib project discoverable add the stdlib directory to the ``CMAKE_PREFIX_PATH``.
 The usual install location of the package files is ``$PREFIX/lib/cmake/fortran_stdlib``.
 
+### Using stdlib with with fpm
```
perazz, line 226:
```suggestion
### Using stdlib with fpm
```

perazz (approved):
LGTM. Thank you @jannisteunissen

jvdp1 (approved):
LGTM. Thank you @jannisteunissen .

Let wait a few days, and if there are no more comments, we could merge it.

gareth-nx:
It might be useful to have more explicit instructions for people who get confused compiling things (like me). I didn't find it immediately obvious how go from the current draft content to compiling programs. 

For what it's worth I would tend to do something like this, which is probably derived from since-deleted stdlib documentation. 

```
export PKG_CONFIG_PATH=/home/gareth/Code_Experiments/fortran/stdlib/stdlib/local_install_gfortran/lib/pkgconfig:$PKG_CONFIG_PATH
export STDLIB_CFLAGS=$(pkg-config --cflags fortran_stdlib)
export STDLIB_LIBS=$(pkg-config --libs fortran_stdlib)

export FC=gfortran
export FFLAGS="-O3 -flto"

# Compile
$FC $FFLAGS -c $STDLIB_CFLAGS my_program.f90
# Link
$FC $FFLAGS -o my_program my_program.o $STDLIB_LIBS 
```

jvdp1:
> from since-deleted stdlib documentation.

Which doc do you refer to?

gareth-nx:
I just dug out an old varient of the README -- the relevant section is copied below. 

## Using stdlib in your project

The stdlib project exports CMake package files and pkg-config files to make stdlib usable for other projects.
The package files are located in the library directory in the installation prefix.

For CMake builds of stdlib you can find a local installation with

```cmake
find_package(fortran_stdlib REQUIRED)
...
target_link_libraries(
  ${PROJECT_NAME}
  PRIVATE
  fortran_stdlib::fortran_stdlib
)
```

To make the installed stdlib project discoverable add the stdlib directory to the ``CMAKE_PREFIX_PATH``.
The usual install location of the package files is ``$PREFIX/lib/cmake/fortran_stdlib``.

For non-CMake build systems (like make) you can use the exported pkg-config file by setting ``PKG_CONFIG_PATH`` to include the directory containing the exported pc-file.
The usual install location of the pc-file is ``$PREFIX/lib/pkgconfig``.
In make you can obtain the required compile and link arguments with

```make
STDLIB_CFLAGS := $(shell pkg-config --cflags fortran_stdlib)
STDLIB_LIBS := $(shell pkg-config --libs fortran_stdlib)
```

README.md
```diff
@@ -235,6 +223,39 @@ target_link_libraries(
...
+libdir=${install_dir}/lib
+moduledir=${install_dir}/include/fortran_stdlib/<compiler name and version>
+```
+Here it is assumed that the compiler will look for libraries in `libdir` and for module files in `moduledir`.
+The library can also be included from a build directory without installation:
+```make
+libdir=$(build_dir)/src
+moduledir=$(build_dir)/src/mod_files
+```
+
```
gareth-nx, line 258: Should we add a something like below, to cover the case of compiling in a shell script or naively from the command line? I don't have much expertise on build systems, and often write small scripts to build programs.

### Using stdlib when compiling within a script, or on the command line
This example uses [pkg-config](https://www.freedesktop.org/wiki/Software/pkg-config/) to manage the compiler flags.
```
export PKG_CONFIG_PATH=${install_dir}/lib/pkgconfig:$PKG_CONFIG_PATH
export STDLIB_CFLAGS=$(pkg-config --cflags fortran_stdlib)
export STDLIB_LIBS=$(pkg-config --libs fortran_stdlib)

export FC=gfortran
export FFLAGS="-O3 -flto"

# Compile
$FC $FFLAGS -c $STDLIB_CFLAGS my_program.f90
# Link
$FC $FFLAGS -o my_program my_program.o $STDLIB_LIBS 
```

README.md
```diff
@@ -235,6 +223,39 @@ target_link_libraries(
...
+```make
+libdir=${install_dir}/lib
+moduledir=${install_dir}/include/fortran_stdlib/<compiler name and version>
+```
+Here it is assumed that the compiler will look for libraries in `libdir` and for module files in `moduledir`.
+The library can also be included from a build directory without installation:
+```make
+libdir=$(build_dir)/src
+moduledir=$(build_dir)/src/mod_files
+```
```
gareth-nx, line 257: Here it would be good to see an example of how `libdir` and `moduledir` would be used inside the makefile to affect the compiler flags.

gareth-nx (commented):
Thanks for following this up, it could help make stdlib more accessible.

README.md
jannisteunissen: I think the information on pkg-config would certainly be nice to include as well, I can do that. I am a little hesitant to include actual compile rules, since these will depend on the particular compiler used and perhaps also on other things. To explain this clearly to people who are not familiar with Makefiles and compilation rules could take too much space. Instead, it could be nice to include a few worked-out usage examples with Makefiles in a separate location.

jalvesz:
Just a small cosmetic suggestion: maybe use this: https://gist.github.com/pierrejoubert73/902cc94d79424356a8d20be2b382e1ab to add collapsible regions in the readme would be a good way to keep the view of the online version compact yet enable adding complete examples for each use case!?

Question: does pkg-config work on Windows? if so, a batch equivalent of the shell example could be useful.

gareth-nx (approved):

Issue #783, opened 2024-04-02
Massive slow down in docs generation


ZedThree:
Just a heads up in case it wasn't noticed already: generating the docs went from taking 1m30s to ~1h10m with #772 (see previous [action runs](https://github.com/fortran-lang/stdlib/actions/workflows/doc-deployment.yml)).

I think this is at least partly due to the larger number of entities that need processing (93454 vs 19548 previously), but there does also seem to be a general slow down which I've not investigated further. One workaround would be to explicitly exclude the new files.

As a separate issue, the [formatting on these docstrings](https://stdlib.fortran-lang.org/module/stdlib_linalg_lapack.html) is probably not as intended. Ford ignores the first line of most of the new docstrings because they're in the form `subroutine: start of doc comment` which ford interprets (and then ignores as unknown) as a `key: value` config setting.

jvdp1:
Thank you @ZedThree for reporting this issue.

>I think this is at least partly due to the larger number of entities that need processing (93454 vs 19548 previously), but there does also seem to be a general slow down which I've not investigated further. One workaround would be to explicitly exclude the new files.

The issue is that `ford` ignores the `cpp` directives as it uses `fypp` as preprocessor.

>As a separate issue, the [formatting on these docstrings](https://stdlib.fortran-lang.org/module/stdlib_linalg_lapack.html) is probably not as intended. Ford ignores the first line of most of the new docstrings because they're in the form subroutine: start of doc comment which ford interprets (and then ignores as unknown) as a key: value config setting.

Thank you. I will have a look.

jvdp1:
> As a separate issue, the [formatting on these docstrings](https://stdlib.fortran-lang.org/module/stdlib_linalg_lapack.html) is probably not as intended. Ford ignores the first line of most of the new docstrings because they're in the form `subroutine: start of doc comment` which ford interprets (and then ignores as unknown) as a `key: value` config setting.

@perazz : could you consider this issue (even if those files will be currently ignored by ford), maybe by moving these docs inside the `interface` blocks (like for other stdlib modules)?

ZedThree:
> The issue is that ford ignores the cpp directives as it uses fypp as preprocessor.

One option could be to add a wrapper script that runs both preprocessors appropriately, and then use that script as the preprocessor for ford?

PR #784, opened 2024-04-02
Ford documentation: improvements

jvdp1:
Related to #783 by @ZedThree

Proposition:
* Ford: upgrade from `6.1.10` to `7.0.5`
* exclude all BLAS/LAPACK source files from the `ford` processing (2 reasons: 1) too many procedures; 2) no support of both `fypp` and `cpp` by `ford`)
* `stdlib_str2num.fypp`: small fixes for `ford` processing


jvdp1:
@zedthree it seems that this latest version of `ford` (`7.0.5`) does not generate the links between the specs to the code, e.g., for `call [[stdlib_sorting(module):ord_sort(interface)]]( array[, work, reverse ] )`. Has the notation been changed between `6.1.10` and `7.0.5`?

Also, it seems that it is not possible to exclude files in `6.1.10`. Hence, the need to upgrade `ford` to `7.0.5`.

gareth-nx:
Thanks for this.

IMO it would be good to include documentation for the BLAS/LAPACK interfaces, if not the individual routines for each precision. Otherwise I think it will be very easy for users to become confused about what is available.

I'm not sure whether your pull request keeps the interfaces or not?

jvdp1:
> IMO it would be good to include documentation for the BLAS/LAPACK interfaces, if not the individual routines for each precision. Otherwise I think it will be very easy for users to become confused about what is available.
> 
> I'm not sure whether your pull request keeps the interfaces or not?

A description of BLAS/LAPACK is available in the [specs](https://stdlib.fortran-lang.org/page/specs/stdlib_linalg.html#blas-and-lapack). It was decided that only a link would be provided to the individual LAPACK interfaces.

However, I will try to find a way to process only one of the precisions. It could be indeed useful.

perazz (approved):

perazz:
These changes look great to me. 
@jvdp1 @gareth-nx @ZedThree: In PR #786 I've addressed the FORD syntax issues, so, FORD can now process all BLAS and LAPACK sources just fine. However, I do confirm that it is a lengthy process (at least on my Mac). 

I think that the best thing to do would be to exclude the sources, and have FORD only parse the interface blocks. However, FORD will fail in this case (because it cannot find the module procedures in the interface)

ZedThree:
@jvdp1 The issue with that particular example in `stdlib_sorting.md` is that it is within backticks:

```markdown
##### Syntax

`call [[stdlib_sorting(module):ord_sort(interface)]]( array[, work, reverse ] )`
```

and in Ford v7 the priority has changed such that backticks now (correctly IMO) are always formatted verbatim, so this needs to be something like:

```markdown
##### Syntax

`call` [[stdlib_sorting(module):ord_sort(interface)]] `( array[, work, reverse ] )`
```

A quick search shows about 270 examples of links within code blocks like this. A minimal change to ford to restore the old behaviour is:

```diff
modified   ford/_markdown.py
@@ -271,7 +271,7 @@ class FordLinkExtension(Extension):
     def extendMarkdown(self, md: MetaMarkdown):  # type: ignore[override]
         project = self.getConfig("project")
         md.inlinePatterns.register(
-            FordLinkProcessor(md, project=project), "ford_links", 174
+            FordLinkProcessor(md, project=project), "ford_links", 194
         )
 
```

but this would really need to be user-configurable, as otherwise it's not possible to write links verbatim (in general, nesting of inline markup is a bit of a nightmare of conflicting syntax, so this will always be a compromise).

If there's a strong desire for this, I can add this option and get it out in 7.0.6

jvdp1:
Thank you @ZedThree for your explanations.

If the following syntax works with Ford v7, I can modify the specs within this PR to support it

> ```
> ##### Syntax
> 
> `call` [[stdlib_sorting(module):ord_sort(interface)]] `( array[, work, reverse ] )`
> ```

> If there's a strong desire for this, I can add this option and get it out in 7.0.6

No need to revert it IMO. We can adapt these notations.

doc/specs/stdlib_math.md
```diff
@@ -156,7 +156,7 @@ Returns a logarithmically spaced rank 1 array from [`base`^`start`, `base`^`end`
 
 #### Syntax
 
-`res = [[stdlib_math(module):logspace(interface)]] (start, end [, n [, base]])`
+`res = ` [[stdlib_math(module):logspace(interface)]] ` (start, end [, n [, base]] `)`
```
jvdp1, line 159:
```suggestion
`res = ` [[stdlib_math(module):logspace(interface)]] ` (start, end [, n [, base]])`
```

doc/specs/stdlib_stats.md
```diff
@@ -78,7 +78,7 @@ The scaling can be changed with the logical argument `corrected`. If `corrected`
 
 ### Syntax
 
-`result = [[stdlib_stats(module):cov(interface)]](array, dim [, mask [, corrected]])`
+`result = ` [[stdlib_stats(module):cov(interface)]] `(array, dim [, mask [, corrected]] `)`
```
jvdp1, line 81:
```suggestion
`result = ` [[stdlib_stats(module):cov(interface)]] `(array, dim [, mask [, corrected]])`
```

doc/specs/stdlib_stats.md
```diff
@@ -248,9 +248,9 @@ The _k_-th order moment about `center` is defined as :
 
 ### Syntax
 
-`result = [[stdlib_stats(module):moment(interface)]](array, order [, center [, mask]])`
+`result = ` [[stdlib_stats(module):moment(interface)]] `(array, order [, center [, mask]] `)`
```
jvdp1, line 251:
```suggestion
`result = ` [[stdlib_stats(module):moment(interface)]] `(array, order [, center [, mask]])`
```

doc/specs/stdlib_stats.md
```diff
@@ -248,9 +248,9 @@ The _k_-th order moment about `center` is defined as :
...
 
-`result = [[stdlib_stats(module):moment(interface)]](array, order, dim [, center [, mask]])`
+`result = ` [[stdlib_stats(module):moment(interface)]] `(array, order, dim [, center [, mask]] `)`
```
jvdp1, line 253:
```suggestion
`result = ` [[stdlib_stats(module):moment(interface)]] `(array, order, dim [, center [, mask]])`
```

doc/specs/stdlib_stats.md
```diff
@@ -306,9 +306,9 @@ The use of the term `n-1` for scaling is called Bessel 's correction. The scalin
 
 ### Syntax
 
-`result = [[stdlib_stats(module):var(interface)]](array [, mask [, corrected]])`
+`result = ` [[stdlib_stats(module):var(interface)]] `(array [, mask [, corrected]] `)`
```
jvdp1, line 309:
```suggestion
`result = ` [[stdlib_stats(module):var(interface)]] `(array [, mask [, corrected]])`
```

doc/specs/stdlib_stats.md
```diff
@@ -306,9 +306,9 @@ The use of the term `n-1` for scaling is called Bessel 's correction. The scalin
...
 
-`result = [[stdlib_stats(module):var(interface)]](array, dim [, mask [, corrected]])`
+`result = ` [[stdlib_stats(module):var(interface)]] `(array, dim [, mask [, corrected]] `)`
```
jvdp1, line 311:
```suggestion
`result = ` [[stdlib_stats(module):var(interface)]] `(array, dim [, mask [, corrected]])`
```

jvdp1:
@perazz @ZedThree @gareth-nx :

Currently, this PR does:
* `Ford`: upgrade from `6.1.10` to `7.0.5`
* Fix all the specs to comply new `Ford` rules
* Exclude all LAPACK source files from the `Ford` processing (reason: 1) too many procedures => too long)

Processing the BLAS files does not significantly influence the time of the CI.

I propose that we merge this PR as is, even if the LAPACK files are excluded for `ford`. The main reason is to reduce the time of the CI of other PR.

perazz:
LGTM @jvdp1

perazz:
If there was a way to include the LAPACK interfaces (`stdlib_linalg_lapack.f90`), without FORD looking for all `module procedure` implementations, it would be great, but to the best of my knowledge, it is not possible.

gareth-nx (approved):
Thanks @jvdp1, this looks good. I agree it makes sense to exclude Lapack until we can address the doc build times.

jvdp1:
Thank you for your reviews.

> If there was a way to include the LAPACK interfaces (`stdlib_linalg_lapack.f90`), without FORD looking for all `module procedure` implementations (`stdlib_linalg_lapack_*.f90`), it would be great, but to the best of my knowledge, it is not possible.

I tried several things similar to your proposition, but all failed! @ZedThree do you have maybe an idea? If not, I will wait a few more days and then merge it.

ZedThree:
There is the (slightly unintuitive) `display: none` option that stops the rendering of any children of an entity -- except that it seems we still actually *process* the docstring, we just don't *display* it, which doesn't help.

It should be possible to fix this, unfortunately I'm at a conference basically all this week, so it might take me awhile to look into it. A quick hack does show some improvement:

```py
modified   ford/sourceform.py
@@ -612,6 +612,10 @@ class FortranBase:
     def filter_display(self, collection: Sequence) -> List:
         """Remove items from collection if they shouldn't be displayed"""
 
+        for obj in collection:
+            if not self._should_display(obj):
+                self.source_file._to_be_markdowned.remove(obj)
+
         return [obj for obj in collection if self._should_display(obj)]
```

but there still seems to be a pretty big slowdown.

jvdp1:
Thank you @ZedThree for your comment. There is no rush for this issue IMO. I will merge this PR as is. 
You can look at this issue later if you want, and when you have time.

PR #786, opened 2024-04-03
BLAS, LAPACK: Refactor comments for FORD compatibility

perazz:
In this PR I've: 
- moved all comment blocks to after `interface`/`subroutine`/`function`
- removed heading `!> NAME:` comments that would otherwise be parsed by FORD as commands

FORD can now process all BLAS/LAPACK files properly (tested locally). 
However, the process is still lengthy due to the large number of procedure (and ~0.6M lines of code added).
So, I think the idea of PR #784 still applies


src/stdlib_linalg_blas_aux.fypp
```diff
@@ -28,9 +28,9 @@ module stdlib_linalg_blas_aux
 
      contains
 
-     !> DCABS1: computes |Re(.)| + |Im(.)| of a double complex number
 
      pure real(dp) function stdlib_dcabs1(z)
+     !> DCABS1 computes |Re(.)| + |Im(.)| of a double complex number
```
jvdp1, line 33: I think it should be `!!` instead of `!>` [when the docstrings are inside blocks](https://forddocs.readthedocs.io/en/latest/user_guide/writing_documentation.html) (at least it is how it was done in other files). `!>` should be used for pre-documentation.

```suggestion
     !! DCABS1 computes |Re(.)| + |Im(.)| of a double complex number
```

jvdp1 (commented):
Thank you @perazz for looking at that. I have just a minor comment.

gareth-nx (approved):
Looks good, thanks

src/stdlib_linalg_blas_aux.fypp
perazz: Thanks @jvdp1, completed in [5735af9](https://github.com/fortran-lang/stdlib/pull/786/commits/5735af92cdf9aeb683a9b7939f487138cf33be4e)

jvdp1 (approved):
Thank you @perazz . IMO this could be merged.

perazz:
Thanks @jvdp1 @gareth-nx for the approvals: I will merge, as this is just a cosmetic change.

PR #787, opened 2024-04-05
fpm branch: .f90 -> .F90 if preprocessed

perazz:
Cf.  https://github.com/fortran-lang/stdlib/pull/772#issuecomment-2039265289
@fortran-lang/stdlib

### DIFF
--- a/ci/fpm-deployment.sh
+++ b/ci/fpm-deployment.sh
@@ -35,6 +35,27 @@ prune=(
   "$destdir/src/f18estop.f90"
 )
 
+# Files that need preprocessed Fortran extension -> .F90 
+preprocessed=(
+   "$destdir/src/stdlib_linalg_constants" 
+   "$destdir/src/stdlib_linalg_blas" 
+   "$destdir/src/stdlib_linalg_blas_aux"
+   "$destdir/src/stdlib_linalg_blas_s"
+   "$destdir/src/stdlib_linalg_blas_d"
+   "$destdir/src/stdlib_linalg_blas_q"
+   "$destdir/src/stdlib_linalg_blas_c"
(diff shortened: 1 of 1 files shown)


jvdp1 (approved):

perazz:
Thanks @jvdp1, let's wait one or two days and merge, if there are no comments.

perazz:
Merged.

jalvesz:
This command is not supported in WSL because the system is case insensitive https://superuser.com/questions/1587020/rename-file-on-wsl-using-mv-produced-error-are-the-same-file.
```sh
# Capitalize .f90 -> .F90 for preprocessed files
for pp_source in "${preprocessed[@]}"
do 
   mv "$pp_source.f90" "$pp_source.F90" 
done
```
Now  I'm not able to preprocess the files using the shell script as I get:
```sh
++ for pp_source in "${preprocessed[@]}"
++ mv stdlib-fpm/src/stdlib_linalg_constants.f90 stdlib-fpm/src/stdlib_linalg_constants.F90
mv: 'stdlib-fpm/src/stdlib_linalg_constants.f90' and 'stdlib-fpm/src/stdlib_linalg_constants.F90' are the same file

[processus terminé avec le code 1 (0x00000001)]
```
and the process just exits.

Is there a way to either prune the list directly in the first command to then apply twice the fypp command? once for those without preprocessing and once for those with, instead of renaming them afterwards?

Maybe if instead of a shell script, a python script is used it make things easier? I've just started to apply a similar idea here https://github.com/jalvesz/FSPARSE/blob/main/deployement.py (extremely basic) that I think we could extend here to make an easier fypp preprocessing process and also OS agnostic.

PR #797, opened 2024-04-13
Update linalg.md and example_blas_gemv.f90

zoziha:
## Description

1. Linking relying on `fpm.toml` requires adding the `link` keyword within the `[build]` section.
2. An error in the `example_blas_gemv` has been corrected.
image

### DIFF
--- a/doc/specs/stdlib_linalg.md
+++ b/doc/specs/stdlib_linalg.md
@@ -41,7 +41,7 @@ These can be enabled during the build process. For example, with CMake, one can 
 The same is possible from the `fpm` branch, where the `cpp` preprocessor is enabled by default. For example, the macros can be added to the project's manifest:
 
 ```toml
-# Link to blas and lapack libraries; or `link = "openblas"` for openblas
+# Link against appropriate external BLAS and LAPACK libraries, if necessary
 [build]
 link = ["blas", "lapack"]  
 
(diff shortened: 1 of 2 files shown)


doc/specs/stdlib_linalg.md
```diff
@@ -41,6 +41,10 @@ These can be enabled during the build process. For example, with CMake, one can
 The same is possible from the `fpm` branch, where the `cpp` preprocessor is enabled by default. For example, the macros can be added to the project's manifest:
 
 ```toml
+# Link to blas and lapack libraries; or `link = "openblas"` for openblas
```
perazz, line 44:
```suggestion
# Link against appropriate external BLAS and LAPACK libraries, if necessary
```

perazz (approved):
LGTM with the suggested edit, thanks @zoziha. 

Now that BLAS/LAPACK support is enabled with our local implementation, we will better address how to link against external libraries. 
For CMake builds it will be easy (`FindBLAS`, `FindLAPACK`), but for `fpm`, there is no support yet and I think the best way to do it is to search for said libraries via metapackage. 

This PR is good and I will merge if @jvdp1 also has no further comments.

jvdp1 (approved):
Thank you @zoziha for this PR. LGTM. It can be merged after @perazz 's suggestion has been considered

Issue #791, opened 2024-04-09
python preprocessor


jalvesz:
### Motivation

This proposal is meant to start a discussion on a replacement for the current `fpm` ci for stdlib. The idea would be to use a python script to preprocess stdlib before building with fpm or CMake. While CMake already has a customized fypp preprocessor, such script could serve as a replacement using a `PRE_BUILD` action with `add_custom_command`. Also, currently the `fpm` branch lacks the means for a flexible way of adding dependencies in the toml file. This proposal would try to remedy these shortcomings.

### Additional Information

#790 #787 

### Proposal
Say a `fypp_deployement.py` file is at the root of stdlib:
```python
import os
import fypp
import argparse
from joblib import Parallel, delayed

def pre_process_toml(kargs):
    """
    Pre-process the fpm.toml
    """
    from tomlkit import table, dumps
    data = table()
    data.add("name", "stdlib")
    data.add("version", str(kargs.vmajor)+
                    "."+str(kargs.vminor)+
                    "."+str(kargs.vpatch) )
    data.add("license", "MIT")
    data.add("author", "stdlib contributors")
    data.add("maintainer", "@fortran-lang/stdlib")
    data.add("copyright", "2019-2021 stdlib contributors")

    if(kargs.with_blp):
        build = table()
        build.add("link", ["lapack", "blas"] )
        data.add("build", build)

    dev_dependencies = table()
    dev_dependencies.add("test-drive", {"git" : "https://github.com/fortran-lang/test-drive", 
                                        "tag" : "v0.4.0"})
    data.add("dev-dependencies", dev_dependencies)

    preprocess = table()
    preprocess.add("cpp", {} )
    preprocess['cpp'].add("suffixes", [".F90", ".f90", ".fypp"] )
    preprocess['cpp'].add("macros", ["MAXRANK="+str(kargs.maxrank), 
                                 "PROJECT_VERSION_MAJOR="+str(kargs.vmajor),
                                 "PROJECT_VERSION_MINOR="+str(kargs.vminor),
                                 "PROJECT_VERSION_PATCH="+str(kargs.vpatch)] )
    data.add("preprocess", preprocess)

    with open("fpm.toml", "w") as f:
        f.write(dumps(data))

C_PREPROCESSED = (
    "stdlib_linalg_constants" ,
    "stdlib_linalg_blas" ,
    "stdlib_linalg_blas_aux",
    "stdlib_linalg_blas_s",
    "stdlib_linalg_blas_d",
[...]

PR #802, opened 2024-04-21
fpm deployment enhancement

jalvesz:
#791 #790 #787 

Enable building stdlib with fpm directly from the root folder
> python config/fypp_deployment.py
> fpm build

Deploy the `stdlib-fpm` from the preprocessing created in the root folder. This deployment could be rendered unnecessary by building directly stdlib from the root.


jalvesz:
@perazz maybe you know if there is a way to avoid launching the tests `test_always_fail` and `test_always_skip` when doing `fpm test` ? The other solution I have in mind is to explicitly add the tests to the manifest and deactivate the auto-test option?

perazz:
Yes I fear that would be the only option, though it's clearly sub-optimal (for each new test, it will have to be added to `fpm.toml`). I think the question is why those tests are there in the first place for? In other words, is there a valid reason those dummy tests should be kept into `stdlib`, or they could be safely removed? At the end of the day they are testing `test-drive` rather than stdlib.

jalvesz:
>I think the question is why those tests are there in the first place for? In other words, is there a valid reason those dummy tests should be kept into stdlib, or they could be safely removed?

I was wondering the same thing!

jvdp1:
> > I think the question is why those tests are there in the first place for? In other words, is there a valid reason those dummy tests should be kept into stdlib, or they could be safely removed?
> 
> I was wondering the same thing!

These tests were introduced for testing the procedure `check` in `stdlib_error`. Later we started to use `test-drive`.
If `check` is not used anymore, I guess these tests could be removed.

jalvesz:
Thanks @jvdp1, removing those tests enables indeed testing from the root build using fpm!

Before going further I wanted to ask about the `stdlib-fpm` branch. Given that this mechanism enables preprocessing and building from root, should this branch be maintained nonetheless? or shall we envision deprecating it progressively? Just to know "how far" should we push this build mechanism.

I'll propose a modification to the README.md before moving from draft to ready for review, but wanted to have some feedback on this point.

README.md
```diff
@@ -179,19 +179,38 @@ earlier builds do not affect the new build.
 Fortran Package Manager (fpm) is a package manager and build system for Fortran.   
 You can build `stdlib` using provided `fpm.toml`:
 
+**Option 1**: From root folder
+
+As `fpm` does not currently support `fypp` natively, `stdlib` now proposes a python script to preprocess and build. This script enables modification of the different `fypp` macros available in `stdlib`. The preprocessed files will be dumped at `<current_folder>/temp/*.f90` or `*.F90`. You can use the following command line:
```
jvdp1, line 184:
```suggestion
As `fpm` does not currently support `fypp` natively, `stdlib` now proposes a python script to preprocess and build it.
This script enables modification of the different `fypp` macros available in `stdlib`. The preprocessed files will be dumped at `<current_folder>/temp/*.f90` or `*.F90`.
You can use the following command line:
```

README.md
```diff
@@ -179,19 +179,38 @@ earlier builds do not affect the new build.
...
 ```
 
-**Alternative**: as `fpm` does not currently support `fypp` natively, building `stdlib` with `fpm` can be done in two steps: a) launch the preprocessor through the `fpm-deployment.sh` script, which creates a subfolder `stdlib-fpm` and b) build the project using the processed files within the latter subfolder. This process can be done with the following commands:
+or the short-cut
 
 ```sh
-source ./ci/fpm-deployment.sh
-cd stdlib-fpm/
-fpm build --profile release
+python config/fypp_deployment.py --build 1
```
jvdp1, line 194: Why is the `1` needed?
could it be:
```suggestion
python config/fypp_deployment.py --build
```

config/requirements.txt
```diff
@@ -0,0 +1,2 @@
+fypp
```
jvdp1, line 1: Why is this file needed? What is its aim?

jvdp1 (commented):
LGTM. Thank you @jalvesz . I just have a few last comments.
A nice improvement is the support of the cpp directives in the branch `stdlib-fpm`.

README.md
jalvesz: Since I defined the parameter as a boolean, it expects a value (an int that can be 0 or 1) after the key word, otherwise it throws an error, I can check if there is a way to facilitate that.

config/requirements.txt
jalvesz: Since we need here python for the preprocessing, I added the requirements file for installation of dependencies. Maybe I should add that to the README

README.md
jalvesz: found it https://stackoverflow.com/questions/8259001/python-argparse-command-line-flags-without-arguments and solved with this commit https://github.com/fortran-lang/stdlib/pull/802/commits/95ee064fa3d61d5b6d4107bface789af6d6ae462

config/fypp_deployment.py
```diff
@@ -0,0 +1,165 @@
+import os
+import fypp
+import argparse
+from joblib import Parallel, delayed
```
jvdp1, line 4: I am just trying the script now on a different machine. It could be good to mention somewhere the these should be installed first (if it is not already the case, of course).

jalvesz: Oh yes, that was the main purpose of the requirements.txt file, to simply do `pip install -r requirements.txt`. I'll add those and the comment in the README.

jalvesz: Done on commit https://github.com/fortran-lang/stdlib/pull/802/commits/a5978340d587361e6ffded32f8f5715740fbc20b

jvdp1 (approved):
I played a bit with the script, and I think it adds nice features (especially that `cpp` directives are still kept in `stdlib-fpm`).

Thank you @jalvesz!

perazz (approved):
Looks great to me in its current form. 
Down the road, I think we may want this script to be able to pass whatever defined preprocessor macros to the `fpm.toml`: 

```toml
[preprocess]
[preprocess.cpp]
macros = ["added","by","deployment","script"]
```
So, let's keep that in mind for now.

jalvesz:
> Down the road, I think we may want this script to be able to pass whatever defined preprocessor macros to the fpm.toml:

Totally! While in the current PR I intentionally striped out that function to make things concise, in the script I posted in #791 that function is still available and can be easily added.

jvdp1:
@jalvesz @perazz It seems that this PR is ready. I will merge it in a couple of days.

PR #805, opened 2024-04-25
Loadtxt real format update to list directed

chuckyvt:
This is a minor change to the loadtxt function to be more robust based on my experience in 'real world' use.   Currently loadtxt specifies a format for the real read function, which makes it 'brittle'.   If the text format is too far off, the read fails.  Changing to list directed read is more robust in my use over the last few months in the version I build and use.  The existing test for loadtxt is a good functional test but doesn't do a good job of simulating actual use, as by definition the text input and read format are defined to be identical.


milancurcic:
Thanks for tagging me. It was a long time ago, but I think we added the explicit edit descriptors to allow savetxt and loadtxt to do a bit-for-bit roundtrip of data. That is important for some cases, and obviously it's limiting in some others, as @chuckyvt points out. I think it's a design choice in the end, and stdlib should support a broad range of use cases. How about adding an optional argument? Not sure what's the best solution.

perazz:
FYI I see the `loadtxt` test often fail in the CI, see i.e. https://github.com/fortran-lang/stdlib/actions/runs/8844659963/job/24286968620?pr=801:

```
152/301 Test #152: loadtxt ................................***Failed    0.01 sec
At line 159 of file /Users/runner/work/stdlib/stdlib/build/src/stdlib_io.f90
Fortran runtime error: Bad value during floating point read

Error termination. Backtrace:
#0  0x104945a8e
#1  0x104946735
#2  0x10494731b
#3  0x104b66528
#4  0x104b69f19
#5  0x104b6b5cf
#6  0x104b683dd
#7  0x104513351
#8  0x104509db6
#9  0x104509df2
```

chuckyvt:
I looked at the original pull request referenced above.  It added format specifier to both write and read, which were both list directed at the time.   Specifying a write format makes sense, and no changes to that in this pull request.  But as far as I know a list directed read would not introduce additional rounding error or loss of precision, and is the simplest and best approach for loadtxt imho.

With that said, I updated the pull request to add an optional 'fmt' field to see how that would look and work.   There currently isn't a way to specify list directed via character variable, so it has to be handled via if statements.   This update uses '*' to specify list directed, but could change to anything.  The thread below has some discussion on this and proposes "(LD)" for example.  

https://fortran-lang.discourse.group/t/format-string-corresponding-to-list-directed-i-o/522/10

src/stdlib_io.fypp
```diff
@@ -145,11 +146,35 @@ contains
...
-          read(s, "(*"//FMT_REAL_${k1}$(1:len(FMT_REAL_${k1}$)-1)//",1x))") d(i, :)
+        if ( present( fmt ) ) then
+          if ( fmt == '*' ) then
+            read (s,*) d(i, :)
+          else
+            read (s,fmt) d(i, :)
+          endif
+        else 
+          read (s,"(*"//FMT_REAL_${k1}$(1:len(FMT_REAL_${k1}$)-1)//",1x))") d(i, :)
+        end if
```
jvdp1, line 157: Checking the same `if` statement at every iteration would be quite slow. So I suggest to move this if statement outside the loop.
Could something like this work:
```fortran
fmt_ = optval('*('//fmt//',1x)', '*', fmt)

do i = 1, max_rows_
  read (s, fmt_) d(i,:)
enddo
```

test/io/test_loadtxt.f90
```diff
@@ -34,17 +34,18 @@ subroutine test_loadtxt_int32(error)
         integer(int32), allocatable :: input(:,:), expected(:,:)
         real(sp), allocatable :: harvest(:,:)
         integer :: n
-
+
```
jvdp1, line 37:
```suggestion

```

test/io/test_loadtxt.f90
```diff
@@ -55,17 +56,22 @@ subroutine test_loadtxt_sp(error)
         !> Error handling
         type(error_type), allocatable, intent(out) :: error
         real(sp), allocatable :: input(:,:), expected(:,:)
+        character(len=*), parameter :: FMT_REAL_SP = '(es15.8e2)'
         integer :: n
 
         allocate(input(10,10))
         allocate(expected(10,10))
-
+
```
jvdp1, line 64:
```suggestion

```

src/stdlib_io.fypp
```diff
@@ -81,7 +81,7 @@ module stdlib_io
 contains
 
   #:for k1, t1 in KINDS_TYPES
-    subroutine  loadtxt_${t1[0]}$${k1}$(filename, d, skiprows, max_rows)
+    subroutine  loadtxt_${t1[0]}$${k1}$(filename, d, skiprows, max_rows, fmt)
```
jvdp1, line 84: [to_string](https://stdlib.fortran-lang.org/page/specs/stdlib_strings.html#to_string) actually uses `format` instead of `fmt` (I prefer `fmt`, but let see what others think about this).

src/stdlib_io.fypp
```diff
@@ -143,15 +145,52 @@ contains
...
+        ! Use list directed read if user has specified fmt='*'
+        do i = 1, max_rows_
+          read (s,*) d(i, :)
+        enddo
+      else
+        ! Otherwise pass default or user specified fmt string.  
+        do i = 1, max_rows_
+          read (s,fmt_) d(i, :)
+        enddo
+      endif
```
jvdp1, line 162:
```suggestion
        ! Use list directed read if user has specified fmt='*'
        ! Otherwise pass default or user specified fmt string.  
        do i = 1, max_rows_
          read (s,fmt_) d(i, :)
        enddo
```

This should worl too, right?

src/stdlib_io.fypp
```diff
@@ -143,15 +145,52 @@ contains
...
+        ! Use list directed read if user has specified fmt='*'
+        do i = 1, max_rows_
+          read (s,*) d(i, :)
+        enddo
+      else
+        ! Otherwise pass default or user specified fmt string.
+        do i = 1, max_rows_
+          read (s,fmt_) d(i, :)
+        enddo
+      endif
```
jvdp1, line 176:
```suggestion
        ! Use list directed read if user has specified fmt='*'
        ! Otherwise pass default or user specified fmt string.
        do i = 1, max_rows_
          read (s,fmt_) d(i, :)
        enddo
```

src/stdlib_io.fypp
```diff
@@ -143,15 +145,52 @@ contains
...
+      if ( fmt_ == '*' ) then
+        do i = 1, max_rows_
+          read (s,*) d(i, :)
+        enddo
+      else
+        ! Otherwise pass default user specified fmt string.
+        do i = 1, max_rows_
+          read (s,fmt_) d(i, :)
+        enddo
+      endif
```
jvdp1, line 190:
```suggestion
      ! Use list directed read if user has specified fmt='*'
     ! Otherwise pass default user specified fmt string.
        do i = 1, max_rows_
          read (s,fmt_) d(i, :)
        enddo
```

jvdp1 (commented):
LGTM. I have only some minor suggestions.

Could you also update the specs, please?

src/stdlib_io.fypp
chuckyvt: No, sadly that doesn't work.   * is not equal to "*"

Discussed in more detail here.  
https://fortran-lang.discourse.group/t/format-string-corresponding-to-list-directed-i-o/522/10

src/stdlib_io.fypp
chuckyvt: Since these functions are generally built to replicate the numpy functions of the same name, I thought it made sense to keep the naming convention used there.   savetxt uses 'fmt'.   (loadtxt doesnt' have an equivalent fmt field, only a delimiter field)

src/stdlib_io.fypp
jvdp1: Ok. Thank you for the explanation!

jvdp1:
Thank you @chuckyvt for the explanations. Please, before merging it, could you:
* rebase your branch to solve the CI issue?
* update the specs?

chuckyvt:
Ok, this should be complete.  Also added a line to the loadtxt example.

jvdp1 (approved):
Thank you @chuckyvt for this PR and improvement. LGTM. I will wait a couple of days, and if there are no additional comments, I will merge it.

jvdp1:
Thank you @chuckyvt . I will merge it.

PR #806, opened 2024-04-25
linalg: solve

perazz:
Linear system solution $Ax=b$, for both 1-rhs (`b(:)`) or n-rhs (`b(:,:)`) cases, based on LAPACK `*GESV` functions.

- [x] base implementation
- [x] tests
- [x] exclude unsupported `xdp` 
- [x] documentation  
- [x] submodule 
- [x] examples
- [x] `pure subroutine` interface

Prior art: 
- NumPy: `solve(a, b)`
- Scipy: `solve(a, b, lower=False, overwrite_a=False, overwrite_b=False, check_finite=True, assume_a='gen', transposed=False)`
- IMSL: `lu_solve(a, b, transpose=False)`
- #710 cc @zoziha 

Proposed implementation: 
- `pure` (only `intent(in)` arguments) 
- "expert" interface (`intent(out)` error handler, `a` can be overwritten)

Example calls:
- `x = solve(a,b)` -> simplest `pure` call
- `x = solve(a,b,overwrite_a=.false.,err)` -> expert call: 
    - `logical(lk), optional, intent(in) :: overwrite_a` = option to avoid internal allocation, but destroy input matrix `a`. Default: `.false.`
    - `type(linalg_state_type), intent(out) :: err` = return state variable

cc: @jvdp1 @zoziha @jalvesz @fortran-lang/stdlib: I believe this is ready for consideration.


doc/specs/stdlib_linalg.md
```diff
@@ -600,6 +600,56 @@ Specifically, upper Hessenberg matrices satisfy `a_ij = 0` when `j < i-1`, and l
...
+
+`x = ` [[stdlib_linalg(module):solve(interface)]] `(a, b, [, overwrite_a], err])`
+
+### Arguments
+
+Two 
+
+`a`: Shall be a rank-2 `real` or `complex` square array containing the coefficient matrix. It is normally an `intent(in)` argument. If `overwrite_a=.true.`, it is an `intent(inout)` argument and is destroyed by the call. 
+
+`b`: Shall be a rank-1 array of the same kind as `a`, containing the right-hand-side vector. It is an `intent(in)` argument.
```
zoziha, line 631:
```suggestion
`b`: Shall be a rank-1 or rank-2 array of the same kind as `a`, containing the right-hand-side vector(s). It is an `intent(in)` argument.
```

doc/specs/stdlib_linalg.md
```diff
@@ -600,6 +600,56 @@ Specifically, upper Hessenberg matrices satisfy `a_ij = 0` when `j < i-1`, and l
...
+
+### Syntax
+
+`Pure` interface:
+
+`x = ` [[stdlib_linalg(module):solve(interface)]] `(a, b)`
+
+Expert interface:
+
+`x = ` [[stdlib_linalg(module):solve(interface)]] `(a, b, [, overwrite_a], err])`
```
zoziha, line 623:
```suggestion
`x = ` [[stdlib_linalg(module):solve(interface)]] `(a, b [[, overwrite_a], err])`
```

zoziha (commented):
LGTM. @perazz, thanks for your efforts, `stdlib` gradually has a user-friendly, high-level BLAS API.

test/linalg/test_linalg_solve.fypp
```diff
@@ -0,0 +1,190 @@
...
+        if (allocated(error)) return
+        
+        call check(error, all(abs(x-res)<abs(res)*1.0e-3_${rk}$), 'results match expected')
+        if (allocated(error)) return        
+
+    end subroutine test_${ri}$_solve
+
+    !> 2x2 Complex linear system
+    !> https://math.stackexchange.com/questions/1996540/solving-linear-equation-systems-with-complex-coefficients-and-variables
+    subroutine test_2x2_${ri}$_solve(error)
```
jvdp1, line 133: Is the license of this content compatible with the MIT license? It should be checked before merging this PR.

jvdp1 (commented):
Thank you. Here are a few suggestions/comments.

test/linalg/test_linalg_solve.fypp
jalvesz: https://opensource.stackexchange.com/questions/3966/can-i-legally-use-code-i-found-on-a-forum#:~:text=In%20the%20case%20of%20all,manner%20any%20changes%20you%20make.

> If it's not simple and you can't/don't want to rewrite it, then you need to consider how it's licensed on the website or forum you found it on. In the case of all StackExchange sites, scroll to the bottom where you'll see a link to [CC BY-SA 3.0](https://creativecommons.org/licenses/by-sa/3.0/). You are free to use and change any content (code or text) that you find on SE sites, as long as you give credit and as long as you continue to share in the same manner any changes you make.

Seems like it is compatible.

jvdp1: This [post](https://opensource.stackexchange.com/a/7436) seems to suggest an answer a bit different. 

@perazz Could you consider a different reference that would avoid this potential license issue?

perazz: Sounds good @jvdp1 @jalvesz, I will create a brand new complex test system, so all license issues are avoided. Apologies, I did not think a linear system could be covered by copyright.

perazz: Test removed and replaced at [4458b88](https://github.com/fortran-lang/stdlib/pull/806/commits/4458b88248c2c6fb72f91e0843a98f445d852bb2)

perazz:
@jvdp1 @zoziha I've implemented a `pure subroutine` interface, that looks [like this](https://github.com/fortran-lang/stdlib/blob/316c44ad0a3a1868b9ead2596acdd6c4e4c9c231/src/stdlib_linalg.fypp#L283-L296):

```fortran
call solve_lu(a,b,x,pivot=ipiv,overwrite_a=.true.,err=err)
``` 

- Is `solve_lu` a good name? I've chosen this because it follows the same pattern as `solve_lstsq` in the least-squares interface
- Both the subroutine and the function interfaces are now `pure`

Please, let me know what you think of this approach, I will complete the documentation for this after we've defined the final version. Thanks!

jvdp1:
> @jvdp1 @zoziha I've implemented a `pure subroutine` interface

Great! Thanks @perazz. LGTM

> * Is `solve_lu` a good name? I've chosen this because it follows the same pattern as `solve_lstsq` in the least-squares interface

LGTM. it makes sense and gives some info on how it is solves.

doc/specs/stdlib_linalg.md
```diff
@@ -600,6 +600,56 @@ Specifically, upper Hessenberg matrices satisfy `a_ij = 0` when `j < i-1`, and l
...
+`b`: Shall be a rank-1 or rank-2 array of the same kind as `a`, containing the right-hand-side vector(s). It is an `intent(in)` argument.
+
+`overwrite_a` (optional): Shall be an input logical flag. if `.true.`, input matrix `a` will be used as temporary storage and overwritten. This avoids internal data allocation. This is an `intent(in)` argument.
+
+`err` (optional): Shall be a `type(linalg_state_type)` value. This is an `intent(out)` argument. The function is not `pure` if this argument is provided.
+
+### Return value
+
+Returns an array value that represents the solution to the linear system of equations.
+
```
jvdp1, line 640: What happens if A is singular?

jvdp1 (approved):
Thank you @perazz . LGTM, and it is almost ready to be merged. Nice addition!

doc/specs/stdlib_linalg.md
perazz: `LINALG_ERROR` is raised: 

https://github.com/fortran-lang/stdlib/blob/316c44ad0a3a1868b9ead2596acdd6c4e4c9c231/doc/specs/stdlib_linalg.md?plain=1#L641

I've improved the documentation at [5e0620c](https://github.com/fortran-lang/stdlib/pull/806/commits/5e0620c3297f0d86e6517805c70786cf9061b52d), thanks @jvdp1

perazz:
Dear all @jvdp1 @jalvesz @zoziha I've documented the subroutine interface and added a no-allocation example. 
I believe this should be ready now. `pivot` is the only pre-allocated array needed in this case. I suggest we wait until we have the full picture (what pre-allocated arrays do `svd`, `eig`, etc. need? ) so we can create a `linalg_storage` type a posteriori

doc/specs/stdlib_linalg.md
```diff
@@ -600,6 +600,107 @@ Specifically, upper Hessenberg matrices satisfy `a_ij = 0` when `j < i-1`, and l
...
+Result vector or array `x` returns the exact solution to within numerical precision, provided that the matrix is not ill-conditioned. 
+An error is returned if the matrix is rank-deficient or singular to working precision. 
+If all optional arrays are provided by the user, no internal allocations take place.
+The solver is based on LAPACK's `*GESV` backends.
+
+### Syntax
+
+Simple (`Pure`) interface:
+
+`x = ` [[stdlib_linalg(module):solve_lu(interface)]] `(a, b, x)`
```
jalvesz, line 672: Shouldn't the interface documentation here start with `call` instead of `x=`?

jalvesz (changes requested):

doc/specs/stdlib_linalg.md
perazz: Thank you @jalvesz: [a05809e](https://github.com/fortran-lang/stdlib/pull/806/commits/a05809eba2925e949527ff9cb19f737a05b592b2)

zoziha (approved):

jvdp1:
Thank you @perazz . I will merge it. 
As discussed [here](https://github.com/fortran-lang/stdlib/pull/801#issuecomment-2102588451), I will prepare a new release today/tomorrow.

perazz:
Thanks a lot @jvdp1

PR #807, opened 2024-05-01
macos-latest -> macos-12

perazz:
It looks like macOS-latest runners have arm architecture where gcc is not supported (at least yet).
For now, I suggest to select the macos version to 12.


jvdp1 (approved):
Thank you. LGTM

PR #813, opened 2024-05-12
Fix example starts with

### LINKED ISSUE #804: example_starts_with prints logical results as binary
This line of code
```
  print'(a)', starts_with("pattern", "pat")  ! T
```
in [`stdlib/example/strings/example_starts_with.f90`](https://github.com/fortran-lang/stdlib/blob/19f84ce94a72ba2c83a92bedcb3c790ecc8bf088/example/strings/example_starts_with.f90#L4)
prints binary because of the `a` format.

It's not a problem per se, but just checking if it's intentional.
```
% gfortran example_starts_with.f90
% a.out |& cat -v
^A^@^@^@
^@^@^@^@
%
```

jvdp1:
Fixes #804 

cc @sscalpone

### DIFF
--- a/example/strings/example_ends_with.f90
+++ b/example/strings/example_ends_with.f90
@@ -1,6 +1,6 @@
 program example_ends_with
   use stdlib_strings, only: ends_with
   implicit none
-  print'(l)', ends_with("pattern", "ern")  ! T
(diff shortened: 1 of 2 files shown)


perazz (approved):

perazz:
LGTM, thank you @jvdp1

PR #815, opened 2024-05-16
Fix: `copy_a` deallocation in `determinant`

perazz:
I noticed a typo in `det` that does not deallocate the internal pointer matrix properly: it should be when the temporary is created (`copy_a==.true.`):

 https://github.com/fortran-lang/stdlib/blob/206f84a0915d1e7ac31e7d2ad254c2b88ada2695/src/stdlib_linalg_determinant.fypp#L227

I don't know why an error was never printed in the tests, as it should be impossible to deallocate the target
of a pointer, if this is not `allocatable`. 
Anyways this is a fix for it, plus an additional test. 

cc: @jvdp1 @jalvesz


jvdp1 (approved):
LGTM. Thank you @perazz

perazz:
Thank you @jvdp1. I will merge this as it was a clear bug.

PR #818, opened 2024-05-16
`lstsq`: return correct array size

### LINKED ISSUE #817: Issue/Question about the output of `lstsq`
Hej,

I'm really happy that the `linalg` module is on its way to being production-ready. I have already played a bit with the latest additions in `v0.6.0` and observed something an undesirable (in my opinion) behaviour regarding the output of `lstsq`. Below is a MWE.

```
program main
  use stdlib_kinds, only: dp
  use stdlib_linalg, only: lstsq, solve_lstsq
  implicit none

  integer :: m, n
  real(dp), allocatable :: A(:, :), b(:), x(:), x_lstsq(:)

  ! Create the problem.
  m = 10 ; n = 5
  allocate(A(m, n)) ; call random_number(A)
  allocate(x(n)) ; call random_number(x)
  b = matmul(A, x)
 
  ! Solve the least-squares problem.
  x_lstsq = lstsq(A, b)

  write(output_unit, *) shape(A), shape(b), shape(x), shape(x_lstsq)
  write(output_unit, *) "Norm of the error :", norm2(x_lstsq(1:n)-x)

end program main
```

When running this code, the vector `x_lstsq` computed by `lstsq` is of size `m` rather than being of size `n`. I know that, by default, `*gelsd` takes `A` and `b` as input and overwrites the first `n` entries of `b` with the solution `x`. I believe this is quite misleading and can potentially cause issues if the user is directly using `solve_lstsq`. Calling `solve_lstsq` with `x` being an `n`-vector instead of an `m`-vector actually raises an error since the following test then fails

```
       if (lda<1 .or. n<1 .or. ldb<1 .or. ldb/=m .or. ldx/=m) then
          err0 = linalg_state_type(this,LINALG_VALUE_ERROR,'invalid sizes: a=',[lda,n], &
                                                           'b=',[ldb,nrhs],' x=',[ldx,nrhsx])
          call linalg_error_handling(err0,err)
          if (present(rank)) rank = 0
          return
       end if
```

in e.g. `stdlib_linalg_s_solve_lstsq_one`. Are there any reasons why we should keep the output vector as an `m`-vector rather than an `n`-vector ? From a mathematical and practical point of view, I would expect the following behaviour:
- `lstsq` takes an `m` x `n` matrix `A` and an `m`-vector `b` as entry and returns an `n`-vector.
- `solve_lstsq` takes as arguments `m` x `n` matrix `A`, an `m`-vector `b` and an `n`-vector `x`.

perazz:
Fix #817. 

for `Ax=b` with `shape(A) == [m, n]`:

- Subroutine interface: allow `size(x,1)>=n`. 
    - if also `size(x,1)>=m`: use as temporary storage and do not reallocate
    - otherwise: allocate temporary for `*GESV` interface
- Function interface: always return `size(x,1)==n`

### DIFF
--- a/doc/specs/stdlib_linalg.md
+++ b/doc/specs/stdlib_linalg.md
@@ -767,7 +767,7 @@ Result vector `x` returns the approximate solution that minimizes the 2-norm \( 
 
 `b`: Shall be a rank-1 or rank-2 array of the same kind as `a`, containing one or more right-hand-side vector(s), each in its leading dimension. It is an `intent(in)` argument. 
 
-`x`: Shall be an array of same kind and rank as `b`, containing the solution(s) to the least squares system. It is an `intent(inout)` argument.
+`x`: Shall be an array of same kind and rank as `b`, and leading dimension of at least `n`, containing the solution(s) to the least squares system. It is an `intent(inout)` argument.
 
 `real_storage` (optional): Shall be a `real` rank-1 array of the same kind `a`, providing working storage for the solver. It minimum size can be determined with a call to [[stdlib_linalg(module):lstsq_space(interface)]]. It is an `intent(inout)` argument.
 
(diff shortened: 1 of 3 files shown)


jvdp1 (approved):
LGTM. thank you

Issue #819, opened 2024-05-16
Building and testing stdlib on WSL1 with two fails


nsh-envs:
### Description

cmake version 3.22.1
ctest version 3.22.1
fypp version 3.2
gfortran version 11.4.0
gmake 4.3

Building the stdlib with no additional settings. When running "cmake -B build" and "cmake --build build" I succesfully build the library. 

Then when testing with "ctest --test-dir build/test" I get errors for "mean" and "string_to_number". 

### Expected Behaviour

No errors when testing. 

### Version of stdlib

42182b080d232fb0704ae1578a9235fe3cf57ef7

### Platform and Architecture

WSL1, Windows 11 enterprise

### Additional Information

OUTPUT when rerunning failed:
usr@host:~/Documents/Fortran/stdlib$ ctest --test-dir build/test --rerun-failed --output-on-failure
Internal ctest changing into directory: /home/niels/Documents/Fortran/stdlib/build/test
Test project /home/niels/Documents/Fortran/stdlib/build/test
    Start 35: mean
1/2 Test #35: mean .............................***Failed    0.02 sec
# Testing: stats_mean
  Starting test_stats_mean_all_int8 ... (1/73)
       ... test_stats_mean_all_int8 [PASSED]
  Starting test_stats_mean_all_int8 ... (2/73)
       ... test_stats_mean_all_int8 [PASSED]
  Starting test_stats_mean_all_optmask_int8 ... (3/73)
       ... test_stats_mean_all_optmask_int8 [PASSED]
  Starting test_stats_mean_int8 ... (4/73)
       ... test_stats_mean_int8 [PASSED]
  Starting test_stats_mean_optmask_int8 ... (5/73)
       ... test_stats_mean_optmask_int8 [PASSED]
  Starting test_stats_mean_mask_all_int8 ... (6/73)
       ... test_stats_mean_mask_all_int8 [PASSED]
  Starting test_stats_mean_mask_int8 ... (7/73)
       ... test_stats_mean_mask_int8 [PASSED]
  Starting test_stats_mean_all_int16 ... (8/73)
       ... test_stats_mean_all_int16 [PASSED]
  Starting test_stats_mean_all_optmask_int16 ... (9/73)
       ... test_stats_mean_all_optmask_int16 [PASSED]
  Starting test_stats_mean_int16 ... (10/73)
       ... test_stats_mean_int16 [PASSED]
  Starting test_stats_mean_optmask_int16 ... (11/73)
[...]

PR #821, opened 2024-05-18
Fixes for extended and quad precision checking. Add CI cheks with fpm

jalvesz:
After looking at #819 I realized that there were several issues:

1. in the `_specialfunctions_gamma` files, there were this kind of fypp macros:
```Fortran
#:set WITH_QP = False
#:set WITH_XDP = False
#:include "common.fypp"
```
Which implied that the different kind definitions were being redefined. In parallel, this means that some files were correctly preprocessed and others not, depending on when the problematic files were being preprocessed.
2. many tests are not using fypp "properly" so I refactored those that I catched after the modifications
3. Seems like there is no CI checking xdp, so I propose to use the fpm CI to check xdp and qp with gfortran. 
4. I  don't know if test-drive is not built with xdp and qp, but the `call check( ... )` using the tolerance argument `thr` had to be modified to avoid passing `xdp` or `qp` arguments.

 I think this is ready for consideration @perazz @jvdp1 @nsh-envs


src/stdlib_specialfunctions_gamma.fypp
```diff
@@ -1,7 +1,6 @@
-#:set WITH_QP = False
-#:set WITH_XDP = False
 #:include "common.fypp"
-#:set CI_KINDS_TYPES = INT_KINDS_TYPES + CMPLX_KINDS_TYPES
+#:set RC_KINDS_TYPES = REAL_KINDS_TYPES[0:2]
```
jvdp1, line 2: For robustness, is it not possible to use something like:
```suggestion
#:set RC_KINDS_TYPES = REAL_KINDS_TYPES.remove("qp").remove("xdp")
```
If the order in `REAL_KINDS_TYPES` chnages (for whatever reasons), the current approach might be brokem

jvdp1, line 2: For what stands the `C` in `RC_KINDSTYPES` (I was expecting the it includes `complex`)

jvdp1 (commented):
Thank you @jalvesz .
I only have a minor commnet, regarding the robustness of the definition of the variable `RC_KINDS_TYPES`.

src/stdlib_specialfunctions_gamma.fypp
jalvesz: I'll test this out, but not sure if it will work as is since `REAL_KINDS_TYPES` is the zipped varaible containing the 3 values INIT, KIND and TYPES. But it is true that if for whatever reason the order is changed that can be problematic.

jalvesz: Sorry, I just wanted to create a local copy of `REAL_KINDS_TYPES` with the subset, the name is indeed misleading, I'll change it.

perazz:
I agree with @jvdp1: great work @jalvesz, thank you for finding this out! I don't want to have you rewrite your PR but here is some thought. 
Elsewhere because LAPACK also does not support `xdp`, I've used this approach: 

https://github.com/fortran-lang/stdlib/blob/42182b080d232fb0704ae1578a9235fe3cf57ef7/src/stdlib_linalg_determinant.fypp#L17-L18

So here, it could be something like this? 

```
#:if rk not in ["xdp","qp"]
```

Also: the special functions just want to be for the "core" types (32- and 64-bit). 
So I'm not suggesting to do it in this PR, but would there be a way we could unify this once and forall? For example, by having some `CORE_REAL_KINDS` defined in `common.fypp` as opposed to _all_ `REAL_KINDS`? 

https://github.com/fortran-lang/stdlib/blob/42182b080d232fb0704ae1578a9235fe3cf57ef7/include/common.fypp#L21-L28

src/stdlib_specialfunctions_gamma.fypp
jalvesz: As I feared, this strategy does not work (it produces and error on fypp)

jalvesz: @jvdp1 this is what I managed:
```
#:set R_KINDS_TYPES = [KT for KT in REAL_KINDS_TYPES if KT[0] in ["sp","dp"]]
#:set C_KINDS_TYPES = [KT for KT in CMPLX_KINDS_TYPES if KT[0] in ["sp","dp"]]
#:set CI_KINDS_TYPES = INT_KINDS_TYPES + C_KINDS_TYPES
```
What do you think?

awvwgk:
For test-drive to accept extended double and quadruple precision the C preprocessor macros `WITH_XDP=1` and `WITH_QP=1` need to be set when compiling test-drive.

jvdp1 (commented):
Thank ypu @jalvesz .

test/specialfunctions/test_specialfunctions_gamma.fypp
```diff
@@ -1,7 +1,7 @@
-#:set WITH_QP = False
-#:set WITH_XDP = False
 #:include "common.fypp"
-#:set CI_KINDS_TYPES = INT_KINDS_TYPES + CMPLX_KINDS_TYPES
+#:set R_KINDS_TYPES = [KT for KT in REAL_KINDS_TYPES if KT[0] in ["sp","dp"]]
```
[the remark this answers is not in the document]
jalvesz: @jvdp1 I didn't have time last week to look at that. I guess this should be done by propagating the fypp macros backwards in the dependency hierarchy but right now I do not see clearly how to do that without touching the build of test-drive itself or transferring the python preprocessing script to fpm.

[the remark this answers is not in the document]
jvdp1: I guess it should be also in the CMake files. I will have a look.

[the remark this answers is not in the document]
jalvesz: Should we then have one of the cmake builds test xdp and qp instead of the fpm build?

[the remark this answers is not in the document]
jvdp1: Yes, good idea!
But the issue might still be there if someone try the tests with `fpm`, right? Or do I miss something?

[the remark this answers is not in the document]
jalvesz: Yes, it would still be there. The ideas on top of my head are: add the python preprocessing to test-drive as well or add it to fpm (it might be the most robust solution but I'm guessing not the easiest)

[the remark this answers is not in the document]
jalvesz: Passing the corresponding flags in the build `fpm test --profile release --flag '-DWITH_XDP -DWITH_QP'` are indeed enough to backpropagate the option.

test/quadrature/test_simps.fypp
```diff
@@ -25,135 +19,70 @@ contains
...
-            new_unittest("simps_even_dp", test_simps_even_dp), &
-            new_unittest("simps_even_qp", test_simps_even_qp), &
-            new_unittest("simps_weights_even_sp", test_simps_weights_even_sp), &
-            new_unittest("simps_weights_even_dp", test_simps_weights_even_dp), &
-            new_unittest("simps_weights_even_qp", test_simps_weights_even_qp), &
-            new_unittest("simps_six_sp", test_simps_six_sp), &
-            new_unittest("simps_six_dp", test_simps_six_dp), &
-            new_unittest("simps_six_qp", test_simps_six_qp) &
+            new_unittest("simps_sp", test_simps_sp) &
+            #:for k1, t1 in REAL_KINDS_TYPES[1:]
```
jvdp1, line 23: Why does `REAL_KINDS_TYPES` start at 1, instead of at 0? Is it safe (assume that the order changes for whatever reasons, some tests might be ignored)?
My suggestion would be to use the whole vector `REAL_KINDS_TYPES`, even if some tests are duplicated.

jvdp1 (approved):
Thank you @jalvesz . Pending a minor comment, I think this PR can be merged.

jvdp1:
@jalvesz It sounds it is ready to be merged. We will wait a few more days, and then merge it. Thank you.

jvdp1:
Thank you @jalvesz . I will merge it now.

Issue #822, opened 2024-05-21
Extend `sort_index` interface to allow `int32` index argument


Beliavsky:
The `index` argument of `sort_index` has type `int64`. I suggest that the `sort_index` interface be extended so that an `index` argument of type `int32` is also supported. I understand that `int64` was chosen to accommodate very large arrays, but it in most cases `int32` is sufficient, and many programmers use default integers. The current interface trips people up, for example [here](https://fortran-lang.discourse.group/t/error-there-is-no-specific-subroutine-for-the-generic-sort-index/1728/9) and [here](https://github.com/fortran-lang/stdlib/issues/719).

jvdp1:
thank you @Beliavsky for the heads up. Indeed, having both version could be useful.
I will have a look.

PR #824, opened 2024-05-21
Rename `int size` to `int_index`

jvdp1:
Related to [this Discouse discussion](https://fortran-lang.discourse.group/t/error-there-is-no-specific-subroutine-for-the-generic-sort-index/1728/14) and to #822


perazz (approved):
As a future extension, would it make sense to support both typical integer sizes i.e. `int32` and `int64` via fypp through generic interfaces? From a quick look at the code it seems like there would be no ambiguous arguments, so it could be a backward-compatible extension

jvdp1:
Thank you @perazz . I agree with you and I will submit a PR later today to support it.

PR #829, opened 2024-05-23
Support of `int32` `index` array in `sort_index`

### LINKED ISSUE #822: Extend `sort_index` interface to allow `int32` index argument

jvdp1:
As proposed by @Beliavsky 

Fixes #822 

To be discussed: names of `int_index` and `int_index_low`: propositions are welcome.

### DIFF
--- a/src/stdlib_sorting.fypp
+++ b/src/stdlib_sorting.fypp
@@ -6,6 +6,8 @@
 #:set CHAR_TYPES_ALT_NAME = list(zip(["character(len=*)"], ["character(len=len(array))"], ["char"]))
 #:set BITSET_TYPES_ALT_NAME = list(zip(BITSET_TYPES, BITSET_TYPES, BITSET_KINDS))
 
+#:set INT_INDEX_TYPES_ALT_NAME = list(zip(["int_index", "int_index_low"], ["integer(int_index)", "integer(int_index_low)"], ["default", "low"]))
+
 #! For better code reuse in fypp, make lists that contain the input types,
 #! with each having output types and a separate name prefix for subroutines
 #! This approach allows us to have the same code for all input types.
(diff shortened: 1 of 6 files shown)


src/stdlib_sorting.fypp
```diff
@@ -147,14 +151,16 @@ module stdlib_sorting
         max_merge_stack = int( ceiling( log( 2._dp**64 ) / &
                                log(1.6180339887_dp) ) )
 
-    type run_type
+#:for ki, ti, namei in INT_INDEX_TYPES_ALT_NAME
+    type run_type_${namei}$
```
perazz, line 155: Looks perfect @jvdp1, the only comment I have is about the `run_type`. Is this a private derived type? If not, renaming `run_type` to `run_type_default` would introduce a backward incompatible change. So, maybe, they could be `run_type` and `run_type_low` instead?

jvdp1: `run_type` is `private` . I renamed it as it was easier to do so with the `fypp` loop.

perazz: Great - good to go then!

perazz (approved):

jvdp1:
I added a couple of checks on the size of the arrays in `sort_index` through [this commit](https://github.com/fortran-lang/stdlib/pull/829/commits/1c385ffd140679079f4bbfdba75b66c19e14f384).

I will wait a few days, and if there are no addtional comments, I will merge this PR.

PR #830, opened 2024-05-27
linalg: fix typo in least-squares complex space allocation

### LINKED ISSUE #823: Bug in the complex least-squares solver
### Description

Consider the following code

```fortran
program main
  use stdlib_kinds, only: dp
  use stdlib_linalg, only: solve_lstsq
  implicit none

  ! Dimension of the problem.
  integer, parameter :: m = 100, n = m-1
  integer :: k
  complex(dp) :: A(m, n), b(m), x_true(n), x_lstsq(n)

  do k = 1, n
      ! Zero-out data.
      A = 0.0_dp ; b = 0.0_dp ; x_true = 0.0_dp ; x_lstsq = 0.0_dp

      ! Generate a random least-squares problem of size (k+1, k).
      call random_number(A(:k+1, :k)%re) ; call random_number(A(:k+1, :k)%im)
      call random_number(x_true(:k)%re) ; call random_number(x_true(:k)%im)
      b(:k+1) = matmul(A(:k+1, :k), x_true(:k))

      ! Solve the lstsq problem.
      call solve_lstsq(A(:k+1, :k), b(:k+1), x_lstsq(:k))

      ! Check correctness of the solution.
      write(output_unit, *) "Iteration k", k, norm2(abs(x_true - x_lstsq))
  end do

end program main
```

Although it looks quite contrived, it is adapted from a naïve implementation of a `gmres` solver where `A` actually is upper Hessenberg. It thus corresponds to the increasingly larger least-squares problem that needs to be solved at each iteration of the `gmres` solver. In this MWE, `A` is generated as a random matrix just to make sure the error I'll report does not depends on the structure of the matrix.

Initially, everything runs just fine. However, the code crashes for `k` equal 39 with the following error:

```
Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x7f147f8eb51f in ???
#1  0x7f147f94b838 in ???
#2  0x7f147f94e452 in ???
#3  0x559663cefe05 in __stdlib_linalg_MOD_stdlib_linalg_z_solve_lstsq_one
	at build/dependencies/stdlib/src/stdlib_linalg_least_squares.f90:1142
#4  0x559663cc6b47 in MAIN__
	at app/main.f90:115
#5  0x559663ccd799 in main
	at app/main.f90:3
Segmentation fault (core dumped)
<ERROR> Execution for object " MWE_stdlib_lstsq " returned exit code  139
<ERROR> *cmd_run*:stopping due to failed executions
STOP 139
```

Not that the error reported actually randomly switches between `double free or corruption (out)` (with `STOP 134`), `SIGSEGV` (with `STOP 139`), `munmap_chunk(): invalid pointer` (with `STOP 134`), or `free(): invalide size` (with `STOP 134`). If single precision is used instead of double precision, the code crashes at iteration `k = 39` instead.

### Expected Behaviour

The code should runs smoothly. Note that if `real` variables instead of `complex` ones are being used, no problem is encountered.

### Version of stdlib

0.6.1

### Platform and Architecture

Linux with Ubuntu 22.04

### Additional Information

The code is generated with `fpm` and the latest `stdlib` version is fetched directly by `fpm`. The code is compiled with `gfortran 13.2.0` installed with `conda`.

perazz:
Attempt to fix #823: there is a typo querying complex storage space in the lstsq solver: 

@jalvesz @loiseaujc if you have an x86 unix setup, can I ask you to check if this fixes your issue? 
It does look indeed like a needed fix.

### DIFF
--- a/src/stdlib_linalg_least_squares.fypp
+++ b/src/stdlib_linalg_least_squares.fypp
@@ -305,7 +305,7 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
          else
             allocate(cwork(lcwork))
          endif
-         ncs = size(iwork,kind=ilp)
+         ncs = size(cwork,kind=ilp)
          #:endif         
                     
          if (nrs<lrwork .or. nis<liwork#{if rt.startswith('c')}# .or. ncs<lcwork#{endif}# &


jalvesz:
I tested it with a WSL, it still segfaults with the change  you proposed.

perazz:
I believe I've found the issue (passing real space `nrs` instead of complex space `ncs` for the complex case). I apologize as for some reason this error was never triggered in my macOS development environment (no bounds checking due to `(*)` in LAPACK + macOS overprovisions allocations, I guess?)

jalvesz:
Yes, that solved the segfault!!

perazz:
Awesome! Thank you @loiseaujc and @jalvesz for reporting.

awvwgk (commented):
Thanks for fixing

test/linalg/test_linalg_lstsq.fypp
```diff
@@ -100,6 +102,39 @@ module test_linalg_least_squares
...
     #:endfor
+    
+    ! Test issue #823
+    subroutine test_issue_823(error)
+        type(error_type), allocatable, intent(out) :: error
+        
+        ! Dimension of the problem.
+        integer(ilp), parameter :: n = 42
+        ! Data for the least-squares problem.
+        complex(dp) :: A(n+1, n), b(n+1), x_true(n), x_lstsq(n+1)
```
jalvesz, line 113: x_lstsq should be of size `x_lstsq(n)`. Tested it locally with the proposed fix and it works ... had to use n+1 in Compiler Explorer as it was not working there

jalvesz (changes requested):

test/linalg/test_linalg_lstsq.fypp
perazz: Thanks @jalvesz: [89dc992](https://github.com/fortran-lang/stdlib/pull/830/commits/89dc99273cb3a86e901362a72154c39288a6eeda)

jvdp1 (approved):
LGTM. Thank you

perazz:
Thank you for the two approvals, merging now.

PR #833, opened 2024-06-04
Deactivate gcc-13 from CI

perazz:
Due to on-going issues with GCC-13 in the github runners, I suggest to temporarily deactivate GCC-13 from the CI. 
See: https://github.com/fortran-lang/fpm/pull/1043
cc: @jvdp1 @fortran-lang/stdlib

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -20,11 +20,8 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-12]
-        gcc_v: [10, 11, 12, 13, 14] # Version of GFortran we want to use.
+        gcc_v: [10, 11, 12] # Version of GFortran we want to use.
         build: [cmake]
-        exclude:
-          - os: ubuntu-latest
-            gcc_v: 13 
         include:
           - os: ubuntu-latest
             gcc_v: 10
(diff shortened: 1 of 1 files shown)


jalvesz:
I saw in one of the comments that someone bumped to ubuntu-24.04 and gcc14 to keep an up-to-date build. I tested it here https://github.com/jalvesz/FSPARSE/actions/runs/9366790111/job/25785056275 by explicitly setting the ubuntu version in the matrix and it works. Maybe it could be useful then to drop gcc13 from the `ubuntu-latest` which is currently pointing to 22.04 and add one option with ubuntu-24.04 and gcc 14 ?

perazz:
cc @jalvesz: 

- `ubuntu-latest` -> `ubuntu-24.04`
- `gcc-13` restored for all configurations, except `macos-latest`

jalvesz:
Should the same change be done for the fpm CI yml script? (Even though it is not failing, but to be consistent)

gnikit:
I would strongly advise against moving to Ubuntu 24, the image is still in Beta. 
image

The removal of GCC 13 was a side effect of GitHub wanting to remove dependencies to third-party PPAs, which is understandable.

If we wish to test against multiple versions of GCC we should use the [setup-fortran](https://github.com/marketplace/actions/setup-fortran) Action instead of manually installing compilers or swapping runner images.

PR #834, opened 2024-06-09
ci: use fortran-setup for GCC, Intel LLVM and Intel Classic

gnikit:
This PR replaces the manual setup of GCC, Intel LLVM and Intel Classic in our CI via the setup-fortran GA workflow.
The motivation behind this is to remove a lot of the maintenance cost + also pin the Intel compiler versions.
This PR should allow us to extend CI testing with the nvidia compilers and with more modern versions of macos on arm64.

## Additional notes

I have disabled the GCC-13 + Macos-12 run since there appears to be failing due to ld linker errors stemming from the gcc cask.


gnikit:
This PR would supersede https://github.com/fortran-lang/stdlib/pull/833

awvwgk (approved):
Thanks this is much cleaner now.

jvdp1 (approved):
Thank you @gnikit . This looks much cleaner and easier to maintain!

perazz (approved):

PR #836, opened 2024-06-15
`stdlib_*laset` add array bound checks in presence of hardcoded input address

### LINKED ISSUE #835: `svd` not working for matrices with a single column
### Description

Consider the following MWE

```fortran
program main
  use stdlib_kinds, only: dp
  use stdlib_linalg, only: svd
  implicit none

  integer, parameter :: m = 2, n = 1
  ! Sigular value decomposition of A.
  real(kind=dp) :: A(n, n)
  real(kind=dp) :: U(n, n), S(n), Vt(n, n)

  ! Random matrix.
  call random_number(A) 

  ! Call to stdlib.
  call svd(A, S, U, Vt)

end program main
```

Using `gfortran 13.2.0`, I get this run time error: 

```
At line 75528 of file build/dependencies/stdlib/src/stdlib_linalg_lapack_d.F90
Fortran runtime error: Index '2' of dimension 1 of array 'a' above upper bound of 1

Error termination. Backtrace:
#0  0x5621220d2c4a in __stdlib_linalg_lapack_d_MOD_stdlib_dgesdd
	at build/dependencies/stdlib/src/stdlib_linalg_lapack_d.F90:75528
#1  0x56212190442b in __stdlib_linalg_MOD_stdlib_linalg_svd_d
	at build/dependencies/stdlib/src/stdlib_linalg_svd.f90:483
#2  0x5621218f77c1 in MAIN__
	at app/main.f90:15
#3  0x5621218f780c in main
	at app/main.f90:2
<ERROR> Execution for object " MWE_stdlib_lstsq " returned exit code  2
<ERROR> *cmd_run*:stopping due to failed executions
```

### Expected Behaviour

Computing the SVD of a column vector is admittedly a contrived example but this piece of code is part of larger subroutine in `LightKrylov` iteratively computing the SVD of large-scale matrices using Lanczos Bidiagonalization. Note that if a row vector is considered instead of a column vector, the MWE runs perfectly.

Pinging @perazz, @jalvesz, @jvdp1.

### Version of stdlib

0.6.1

### Platform and Architecture

Linux with Ubuntu 22.04

### Additional Information

_No response_

perazz:
Fix #835. Calls like this
```fortran
call stdlib_slaset( 'U', nr-1,nr-1, zero,zero, a(1,2),lda )
```

cause bound checking errors in debug mode in case `nr==1` because `size(a,2)==1` (internally, nothing would be executed).

In several other places, LAPACK already had this fixed i.e., at: 
```fortran
                    ! zero out below r
                    if( n > 1 ) then
                       call stdlib_slaset( 'L', n-1, n-1, zero, zero, a( 2, 1 ),lda )
                    end if
```

i.e. by simply requiring `n>1`. This PR extends this safety checks to all applicable cases, i.e. where argument `a` address is hard-coded. A test is added too. 

cc: @jvdp1 @jalvesz @loiseaujc @fortran-lang/stdlib

### DIFF
--- a/src/stdlib_linalg_lapack_c.fypp
+++ b/src/stdlib_linalg_lapack_c.fypp
@@ -65732,7 +65732,7 @@ module stdlib_linalg_lapack_c
                     call stdlib_cgeqrf( m, n, a, lda, work( itau ), work( nwork ),lwork-nwork+1, &
                               ierr )
                     ! zero out below r
-                    call stdlib_claset( 'L', n-1, n-1, czero, czero, a( 2, 1 ),lda )
+                    if (n>1) call stdlib_claset( 'L', n-1, n-1, czero, czero, a( 2, 1 ),lda )
                     ie = 1
                     itauq = 1
                     itaup = itauq + n
@@ -65918,7 +65918,7 @@ module stdlib_linalg_lapack_c
                     call stdlib_cungqr( m, m, n, u, ldu, work( itau ),work( nwork ), lwork-nwork+&
                               1, ierr )
                     ! produce r in a, zeroing out below it
-                    call stdlib_claset( 'L', n-1, n-1, czero, czero, a( 2, 1 ),lda )
+                    if (n>1) call stdlib_claset( 'L', n-1, n-1, czero, czero, a( 2, 1 ),lda )
                     ie = 1
                     itauq = itau
                     itaup = itauq + n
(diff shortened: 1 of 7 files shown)


jvdp1 (approved):
LGTM! Thank you @perazz

perazz:
Thank you @jvdp1. Let's wait one or two days, and then merge if there are no further comments.

PR #611, opened 2021-12-28
Hash maps

wclodius2:
Now that hash functions have been added to the standard library, it is time to add hash maps. Currently this PR only includes the documentation, `doc/specs/stdlib_hash_maps.md`, giving the proposed API. The proposed API has three module files: `stdlib_key_data_wrapper.f90`, which provides wrappers for the key and data components of the entries and some of the hash function; `stdlib_chaining_hash_map.f90`, which implements a chaining hash map with linked lists; and `stdlib_open_hash_maps.f90`, which implements a linear addressing open hashing map. After the (modified) API  has been approved by two people I will add the required module files.

PR #843, opened 2024-07-02
Remove other derived type from hashmap

chuckyvt:
This is a proposed update to the hashmap routines to remove the 'other' derived type that is currently required to store values.   I reviewed the PR history related to the hashmap routines to try to understand the reason for using the other derived type.  I don't see a clear reason other than perhaps it was thought that directly using unlimited polymorphic type would require use of select type in the code?  I should note up front that this change is not backwards compatible with existing code, as the get_other_data function will now require as class(*), allocatable data type in the other field, which is discussed a bit more below.  But I do believe it simpler and a better long-term approach.        

For an example.   Currently to set a value, and then retrieve it, the code looks like:
```
type(other_type)      :: other
class(*), allocatable  :: data
type(chaining_hashmap_type) :: map

call set(other, 4.0)
call map%map_entry(key, other)

call map%get_other_data(key, other)
call get(other, data)

select type (data)
    type is (real)
        print *, 'Value = ', data
end select
```

This will simply too
```
class(*), allocatable  :: data
type(chaining_hashmap_type) :: map

call map%map_entry(key, 4.0)

call map%get_other_data(key, data)

select type (data)
    type is (real)
        print *, 'Value = ', data
end select
```

As mentioned above, this change is not backwards compatible with existing code that uses the other type.   The map_entry part will function correctly, however the get_other_data routine will require a class(*), allocatable type data.   I believe the hashmaps are still listed as experimental, and thus subject to change, so would like to think this isn't too detrimental.   

I am submitting this in draft form, as the specification still need updating.   

Note, would like to give credit to @LKedward, as his [fhash](https://github.com/LKedward/fhash) package was used as a reference in developing this PR.


jvdp1 (commented):
Thank you @chuckyvt for this proposition. I already thought about this complexity (see for example [here](https://github.com/fortran-lang/stdlib/pull/664)). But I think that your solution is more elegant.

A question is: why was the DT `other_type` considered as needed in this implementation, while it seems to add only one level of abstraction? If there are no major reasons, I am in favour of implementing your PR.

jvdp1:
@chuckyvt The `other_type` was also questionned during the implementation of the hashmap modules: see [here](https://github.com/fortran-lang/stdlib/pull/611#pullrequestreview-841042599), [here](https://github.com/fortran-lang/stdlib/pull/611#issuecomment-1002299547), [here](https://github.com/fortran-lang/stdlib/pull/611#issuecomment-1002326532), [here](https://github.com/fortran-lang/stdlib/pull/611#issuecomment-1002353076), [here](https://github.com/fortran-lang/stdlib/pull/611#issuecomment-1003242366), [here](https://github.com/fortran-lang/stdlib/pull/611#pullrequestreview-842287347), [here](https://github.com/fortran-lang/stdlib/pull/611#issuecomment-1003494314), [here](https://github.com/fortran-lang/stdlib/pull/611#issuecomment-1003729877), [here](https://github.com/fortran-lang/stdlib/pull/611#issuecomment-1003732681), [here](https://github.com/fortran-lang/stdlib/pull/611#issuecomment-1006062350), [here](https://github.com/fortran-lang/stdlib/pull/611#issuecomment-1006283971).

I probably missed some coments on this topic. But it seems that the main concern regarding `other_type` vs `class(*)` is the performance.

src/stdlib_hashmaps.f90
```diff
@@ -184,7 +184,7 @@ subroutine key_get_other_data( map, key, other, exists )
             import hashmap_type, key_type, other_type
```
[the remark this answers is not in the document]
chuckyvt: This has been removed with latest updates.

src/stdlib_hashmap_chaining.f90
```diff
@@ -568,8 +568,7 @@ module subroutine map_chain_entry(map, key, other, conflict)
                 new_ent % next => map % slots(hash_index) % target
                 map % slots(hash_index) % target => new_ent
                 call copy_key( key, new_ent % key )
-                if ( present(other) ) call copy_other( other, new_ent % other )
-
+                if ( present(other) ) new_ent % other = other
```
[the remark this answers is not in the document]
chuckyvt: So, initially I had no concerns with your suggestion as I had read similar feedback.  However when I submitted this change, all the Windows CI checks failed.   Not sure of the cause, I do wonder if its a Windows stack overflow issue.  But I have changed it back to the original `new_ent % other = other` line for now.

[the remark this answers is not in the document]
jvdp1: ok. this is strange...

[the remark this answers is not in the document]
chuckyvt: Did some more trials on this.  Two options, both pass all CI tests.    

Original in-line allocation:
```new_ent % other = other```

Or explicit allocation:
```if ( present( other ) ) then
    if ( allocated(new_ent % other) ) deallocate(new_ent % other)
    allocate(new_ent % other, source = other)
endif```

I'm not exactly sure how new_ent % other is already allocated under certain situations, but there are numerous gfortran failures without.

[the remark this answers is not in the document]
Comment (jvdp1) [reply src/stdlib_hashmap_chaining.f90]: If the `allocated` check is needed to avoid some bugs, then I prefer the original one because it is clearer (and also valid).

File: doc/specs/stdlib_hashmaps.md [hunk 3]
```diff
@@ -118,38 +116,28 @@ value, `int32`.
 
 ### The `stdlib_hashmap_wrappers`' module's derived types
 
-The `stdlib_hashmap_wrappers` module defines two derived types:
-`key_type`, and `other_type`. The `key_type` is intended to be used
-for the search keys of hash tables.  The `other_type` is intended to
-store additional data associated with a key. Both types are
-opaque. Their current representations are as follows
+The `stdlib_hashmap_wrappers` defines `key_type` which is intended to
+be used for the search keys of hash tables.  The tye is opaque.
```
jvdp1, line 120:
```suggestion
be used for the search keys of hash tables. The tye is opaque.
```

doc/specs/stdlib_hashmaps.md
```diff
@@ -375,13 +316,13 @@ This code does not pass any of the SMHasher tests, but the resulting
...
 expected to be minor compared to its faster hashing rate.
 
-
 ##### Example
 
 ```fortran
 {!example/hashmaps/example_hashmaps_fnv_1a_hasher.f90!}
 ```
 
+
```
jvdp1, line 325:
```suggestion
```

doc/specs/stdlib_hashmaps.md
```diff
@@ -1570,15 +1469,17 @@ and the entry was not entered into the map, a `.false.` value indicates
 that `key` was not present in the map and the entry was added to the
 map. 
 
-* If `key` is already present in `map` then the presence of `other` 
-is ignored.
+* If `key` is already present in `map` and the `conflict` argument has been
+provided then the presence of `other` is ignored.  If `conflict` has not
```
jvdp1, line 1473:
```suggestion
provided then the presence of `other` is ignored. If `conflict` has not 
```

doc/specs/stdlib_hashmaps.md
```diff
@@ -1762,23 +1666,26 @@ and access the entry's data.
...
-logical. It is an `intent(out)` argument. If present with the value
+`exists` (optional): shall be a scalar variable of type `logical`.
+It is an `intent(out)` argument. If present with the value
 `.true.` an entry with that `key` existed in the map and its `other`
-data was replaced, otherwise if `exists` is `.false.` the entry did
+data was replaced.  If `exists` is `.false.` the `key` did
 not exist and nothing was done.
 
+* If `key` is not already present in `map` and `exists` has not 
+been provided then the routine will error stop if.
```
jvdp1, line 1680:
```suggestion
been provided then the routine will error stop.
```

example/hashmaps/CMakeLists.txt
```diff
@@ -26,3 +24,5 @@ ADD_EXAMPLE(hashmaps_set)
 ADD_EXAMPLE(hashmaps_set_other_data)
 ADD_EXAMPLE(hashmaps_slots_bits)
 ADD_EXAMPLE(hashmaps_total_depth)
+
+
```
jvdp1, line 28:
```suggestion
```

example/hashmaps/example_hashmaps_get_other_data.f90
```diff
@@ -1,11 +1,11 @@
 program example_get_other_data
   use stdlib_kinds, only: int8, int64
   use stdlib_hashmaps, only: chaining_hashmap_type, int_index
-  use stdlib_hashmap_wrappers, only: fnv_1_hasher, key_type, other_type, set, get
+  use stdlib_hashmap_wrappers, only: fnv_1_hasher, key_type, set, get
   implicit none
   logical                     :: conflict
   type(key_type)              :: key
-  type(other_type)            :: other
+
```
jvdp1, line 8:
```suggestion
```

example/hashmaps/example_hashmaps_map_entry.f90
```diff
@@ -1,43 +1,54 @@
...
-  type(other_type)    :: other
   integer             :: int_scalar
   
+  type :: array_data_wrapper
+      integer, allocatable :: array(:)
+  end type
+  
+  type(array_data_wrapper) :: array_example
+  
+
```
jvdp1, line 17:
```suggestion
```

example/hashmaps/example_hashmaps_remove.f90
```diff
@@ -2,45 +2,41 @@ program example_remove
...
   integer             :: int_scalar
   
   ! Initialize hashmap with 2^10 slots.
   ! Hashmap will dynamically increase size if needed.
   call map%init(fnv_1_hasher, slots_bits=10)
-  
-  ! Initialize other type with data to store.
-  call set(other, 4.0)
-  
+
```
jvdp1, line 15:
```suggestion

```

example/hashmaps/example_hashmaps_set_other_data.f90
```diff
@@ -1,23 +1,38 @@
...
+  use stdlib_hashmap_wrappers, only: key_type, set, fnv_1_hasher
+  
   implicit none
   logical :: exists
-  type(open_hashmap_type) :: map
-  type(key_type)      :: key
-  type(other_type)    :: other
+  type(chaining_hashmap_type) :: map
+  class(*), allocatable   :: data
+
```
jvdp1, line 10:
```suggestion
```

jvdp1 (approved):
Thank you @chuckyvt . LGTM, pending some very minor comments.
Out of curiosity, did you do some benchmarks between the two versions?

chuckyvt:
I believe all requested changes in.  I did do some prelim benchmarking.   It appeared that mapping data may be slightly faster with this approach.   (which makes sense to me, it takes out an interface).  I haven't been able to compare getting data.  I will work to pull that together and upload that comparison.

chuckyvt:
First of all, should note this is on Windows, with Intel IFX 2024.1 compiler in CMake Release mode.    

I went through and ran the updated open and chaining test programs with this PR.  (ie the _pr files).  I also went back to the current version tests and made some updates to make it an equivalent comparison.  (Removed the redundant dummy value and associated allocation, added a get(other,data) command so that on the get tests the final step is with an unlimited polymorphic value). 

Long story short, it's difficult to make any judgements with any confidence.  Comparing the files, I think I could convince myself that the PR has slightly better performance?  But it's also possible it's just random.   Without a Python 'timeit' type function that will repeat the commands many times, it's difficult to gauge.   

[test_open_maps_pr.txt](https://github.com/user-attachments/files/16186572/test_open_maps_pr.txt)
[test_open_maps_baseline.txt](https://github.com/user-attachments/files/16186573/test_open_maps_baseline.txt)
[test_chaining_maps_pr.txt](https://github.com/user-attachments/files/16186575/test_chaining_maps_pr.txt)
[test_chaining_maps_baseline.txt](https://github.com/user-attachments/files/16186576/test_chaining_maps_baseline.txt)

chuckyvt:
Also, I have tried everything I can think of to get the one Windows CI 32 Bit GFortran build to run reliably, but no luck.  It fails on the set_other_data example, which is a relatively simple case.

jvdp1:
> Comparing the files, I think I could convince myself that the PR has slightly better performance? But it's also possible it's just random

Thank you @chuckyvt . These "benchmarks" ar eenough for me. It confirms that your changes do not reduce the performances. Thank you for running them.

doc/specs/stdlib_hashmaps.md
```diff
@@ -1478,8 +1376,8 @@ are examined.
 or `int32` array. It is an `intent(in)` argument. It is a `key` whose 
 presence in the `map` is being examined.
 
-`present` (optional): shall be a scalar variable of type default
-`logical`. It is an intent(out) argument. It is a logical flag where
+`present`: shall be a scalar variable of type `logical`.
+It is an intent(out) argument. It is a logical flag where
```
jvdp1, line 1380:
```suggestion
It is an `intent(out)` argument. It is a logical flag where
```

jvdp1:
> Also, I have tried everything I can think of to get the one Windows CI 32 Bit GFortran build to run reliably, but no luck. It fails on the set_other_data example, which is a relatively simple case.

@chuckyvt I tested several things. It seems it is related with the allocate statement of the other data. I could make things only worse. Could it be a compiler bug?

@jalvesz @perazz Any ideas on this issue?

perazz:
@jvdp1 @chuckyvt the use of `class(*), allocatable` is to enable hashing of any derived types beyond the standard intrinsics, so, I believe that removal of `other_type` would introduce a backward incompatible change? I found some details at the original discussion here: https://github.com/fortran-lang/stdlib/pull/611

Perhaps to increase performance when using intrinsic types, imho the best approach would be to template the hashmap derived type: 

```Fortran
type, abstract, extends(hashmap_type) :: chaining_hashmap_type
type, extends(chaining_hashmap_type) :: chaining_hashmap_type_generic
type, extends(chaining_hashmap_type) :: chaining_hashmap_type_int32
type, extends(chaining_hashmap_type) :: chaining_hashmap_type_int64
!....
```

but I understand that would be a significant deviation from the current PR.

chuckyvt:
> @jvdp1 @chuckyvt the use of `class(*), allocatable` is to enable hashing of any derived types beyond the standard intrinsics, so, I believe that removal of `other_type` would introduce a backward incompatible change? I found some details at the original discussion here: #611
> 
> Perhaps to increase performance when using intrinsic types, imho the best approach would be to template the hashmap derived type:
> 
> ```fortran
> type, abstract, extends(hashmap_type) :: chaining_hashmap_type
> type, extends(chaining_hashmap_type) :: chaining_hashmap_type_generic
> type, extends(chaining_hashmap_type) :: chaining_hashmap_type_int32
> type, extends(chaining_hashmap_type) :: chaining_hashmap_type_int64
> !....
> ```
> 
> but I understand that would be a significant deviation from the current PR.

Just for clarity, this PR only affects the 'other' or value part of the hashmaps, there is no change to the code associated with the key interface and hashing.  You are correct, this would be incompatible with existing "other_type" uses.   This is covered in the description of the PR.  Perhaps there is a way to publish a notice and/or simple transition guide along with this PR?

Finally, I've also thought about extending hashmaps to specific types as you mentioned.  That is out of scope of this PR but there is nothing in this PR preventing that.  Performance would improve and would not require use of 'select type' on retrieval which would be nice.  However, I believe we would still need a class(*) interface to store mixed data type maps, and also to store derived types.

perazz:
@chuckyvt I see that the mingW i686 setup uses gfortran-13.2.0. It's hard to guess without having the same setup in place, however I have non-trivial issues with gfortran>=13 with nested allocatable components and you may have hit the same compiler problem. Unfortunately that has to date no straightforward solution

perazz (approved):
LGTM, it is a nice update imho. I like in this approach that it enables the hashmap type to be templated in the future to static `value` types (replace `class(*)` with the actual type), which may be even faster.

jvdp1:
@chuckyvt @perazz @jalvesz Is this PR ready to be merged?

chuckyvt:
Yes, it is complete as far as I know.

jalvesz (approved):

jalvesz:
Thanks @chuckyvt for this update! I'll merge it since there have been no more comments.

PR #847, opened 2024-07-07
Rewrite fypp directive to avoid the use of removesuffix in stdlib_math_meshgrip

jvdp1:
`removesuffix` used in `stdlib_math_meshgrid` requires Python >3.9. 

Here is a small change to avoid the requirement of Python >3.9.

@igirault : was there a special need to use `removesuffix`? Do I overlook something? Or are my changes satisfying?

\cc @perazz @jalvesz

### DIFF
--- a/src/stdlib_math_meshgrid.fypp
+++ b/src/stdlib_math_meshgrid.fypp
@@ -7,7 +7,7 @@
   do i${j}$ = 1, size(x${j}$)
   #:endfor
   #:for j in indices
-  xm${j}$(${"".join(f"i{j}," for j in indices).removesuffix(",")}$) = &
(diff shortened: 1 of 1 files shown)


perazz (approved):
LGTM, thank you @jvdp1. Besides the greater compatibility, I also like your syntax better.

jalvesz:
LGTM @jvdp1 indeed it looks cleaner this way!!

igirault:
> @igirault : was there a special need to use removesuffix? Do I overlook something? Or are my changes satisfying?

not at all, this was due to my poor understanding of the `join` operator in python. This is indeed way nicer with your correction :)

jvdp1:
Thank you @igirault for your answer.

With all these approvals, I will merge this PR.

Issue #694, opened 2023-02-19
Tests on PPC: 3 tests failed out of 282


barracuda156:
### Description

```
--->  Testing fortran-stdlib
Executing:  cd "/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_lang_fortran-stdlib/fortran-stdlib/work/build" && ctest test 
Test project /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_lang_fortran-stdlib/fortran-stdlib/work/build
        Start   1: always_skip
  1/282 Test   #1: always_skip ............................***Skipped   0.01 sec
        Start   2: always_fail
  2/282 Test   #2: always_fail ............................   Passed    0.01 sec
        Start   3: test-drive/all-tests
  3/282 Test   #3: test-drive/all-tests ...................   Passed    0.01 sec
        Start   4: test-drive/check
  4/282 Test   #4: test-drive/check .......................   Passed    0.01 sec
        Start   5: test-drive/select
  5/282 Test   #5: test-drive/select ......................   Passed    0.01 sec
        Start   6: logicalloc
  6/282 Test   #6: logicalloc .............................   Passed    0.42 sec
        Start   7: ascii
  7/282 Test   #7: ascii ..................................   Passed    0.01 sec
        Start   8: stdlib_bitset_64
  8/282 Test   #8: stdlib_bitset_64 .......................   Passed    0.01 sec
        Start   9: stdlib_bitset_large
  9/282 Test   #9: stdlib_bitset_large ....................   Passed    0.01 sec
        Start  10: hash_functions
 10/282 Test  #10: hash_functions .........................***Failed    0.07 sec
        Start  11: 32_bit_hash_performance
 11/282 Test  #11: 32_bit_hash_performance ................   Passed    0.03 sec
        Start  12: 64_bit_hash_performance
 12/282 Test  #12: 64_bit_hash_performance ................   Passed    0.55 sec
        Start  13: chaining_maps
 13/282 Test  #13: chaining_maps ..........................   Passed    3.12 sec
        Start  14: open_maps
 14/282 Test  #14: open_maps ..............................   Passed    2.77 sec
        Start  15: maps
 15/282 Test  #15: maps ...................................   Passed    7.12 sec
        Start  16: loadtxt
 16/282 Test  #16: loadtxt ................................   Passed    0.25 sec
        Start  17: savetxt
 17/282 Test  #17: savetxt ................................   Passed    0.01 sec
[...]

PR #848, opened 2024-07-09
`sort_index`: use of only `int_index` iterators inside `sort_index`

jvdp1:
This PR aims to simplifiy the implementation of `sort_index`, without modifying the API. So, `sort_index` still accept an array `index` of kind `int_index` or `int_index_low`, but all iterations and counts are internally done using `int_index` integers.
This is a first step to extend `sort_index` to a procedure that accepts an array `index` of a intrinsic type `integer`, `real`, `complex`, ....

Note: these changes partly revert some changes done in #829 and can be compared against #824 (see [here](https://github.com/fortran-lang/stdlib/compare/3d283a7a70c8461496b870c4448f8d74f7b4563d...jvdp1:stdlib:split_sort_index) for a comparison with the original implementation)
Note2: these changes are in agreement with the implementation in `ord_sort`.

cc: @perazz @jalvesz

### DIFF
--- a/src/stdlib_sorting_sort_index.fypp
+++ b/src/stdlib_sorting_sort_index.fypp
@@ -80,31 +80,32 @@ contains
 ! a non-increasing sort. The logic of the determination of indexing largely
 ! follows the `"Rust" sort` found in `slice.rs`:
 ! https://github.com/rust-lang/rust/blob/90eb44a5897c39e3dff9c7e48e3973671dcd9496/src/liballoc/slice.rs#L2159
-! The Rust version is a simplification of the Timsort algorithm described
-! in https://svn.python.org/projects/python/trunk/Objects/listsort.txt, as
+! The Rust version in turn is a simplification of the Timsort algorithm
+! described in
+! https://svn.python.org/projects/python/trunk/Objects/listsort.txt, as
 ! it drops both the use of 'galloping' to identify bounds of regions to be
 ! sorted and the estimation of the optimal `run size`. However it remains
 ! a hybrid sorting algorithm combining an iterative Merge sort controlled
 ! by a stack of `RUNS` identified by regions of uniformly decreasing or
-! non-decreasing sequences that may be expanded to a minimum run size, with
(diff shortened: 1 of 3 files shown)


src/stdlib_sorting_sort_index.fypp
```diff
@@ -102,17 +102,17 @@ contains
         ${ti}$, intent(out), optional :: iwork(0:)
         logical, intent(in), optional            :: reverse
 
-        ${ti}$ :: array_size, i, stat
+        integer(int_index) :: array_size, i, stat
         ${t2}$, allocatable :: buf(:)
         ${ti}$, allocatable :: ibuf(:)
 
-        if ( size(array, kind=int_index) > huge(1_${ki}$) ) then
+        if ( size(array, kind=int_index) > huge(1_int_index) ) then
```
jvdp1, line 109:
```suggestion
        if ( size(array, kind=int_index) > huge(int(1, kind=i)) then
```

jvdp1: Or could this check be simply removed?

perazz (approved):
I like the fact of uniforming all internal `size` handling to 64-bit integes. The code is clearer, less need for pre-processing, and may be even more performing on some platforms. LGTM. 

I was wondering if you think there may be issues on some 32-bit platforms i.e. https://github.com/fortran-lang/stdlib/issues/694, but I see tests OK with 64-bit integers there, so, all good I think.

jvdp1:
> I was wondering if you think there may be issues on some 32-bit platforms i.e. #694, but I see tests OK with 64-bit integers there, so, all good I think.

Thank you @perazz . I see that `ord_sort` (that relies on `int64` integers) was OK on #694. So, I guess it is fine.

jvdp1:
As this PR does not modify the current API, and only slightly modifies the implementation of `sort_index` (by mainly cleaning it and coming back to its original version), I will merge this PR and open another one related to @jalvesz issue in #760

PR #849, opened 2024-07-09
Addition of `sort_adj` to sort an rank 1 array in the same order as an input array

jvdp1:
Addition of `sort_adj` to sort an rank 1 array in the same order as an input array.
It is basically an extension of the original `sort_index` procedure, by making the array `index` `intent(inout)`, instead of `intent(out)` only.

To be discussed:
* Name of the procedure: currently called `sort_adj`, but I think another (more appropriate) name should be found
* Name of the arguments in the API, especially of the associated array currently called `index`: suggestions of a more appropriate name?
* Is the current implementation of `sort_index` still valid/appropriate/useful? 

@jalvesz `sort_adj` could replace your internal sorting procedure in your sparse implementation. If not, I think it is still a nice addition to `stdlib`.


src/stdlib_sorting.fypp
```diff
@@ -13,6 +13,9 @@
 #! This approach allows us to have the same code for all input types.
 #:set IRSCB_TYPES_ALT_NAME = INT_TYPES_ALT_NAME + REAL_TYPES_ALT_NAME + STRING_TYPES_ALT_NAME + CHAR_TYPES_ALT_NAME &
     & + BITSET_TYPES_ALT_NAME
+#:set IR_INDEX_TYPES_ALT_NAME = INT_TYPES_ALT_NAME + REAL_TYPES_ALT_NAME
+
+
```
jvdp1, line 18:
```suggestion
```

src/stdlib_sorting.fypp
```diff
@@ -505,6 +578,44 @@ module stdlib_sorting
 
     end interface sort
 
+
```
jvdp1, line 581:
```suggestion
```

src/stdlib_sorting.fypp
```diff
@@ -535,12 +663,32 @@ module stdlib_sorting
             ${t2}$, intent(out), optional             :: work(0:)
             ${ti}$, intent(out), optional            :: iwork(0:)
             logical, intent(in), optional             :: reverse
+
```
jvdp1, line 666:
```suggestion
```

jalvesz:
What about naming the secondary array `array_adjoint` or `adjoint_array` to be super explicit about its role?

perazz:
Thank you for this PR @jvdp1. I agree with @jalvesz and like the `adjoint_array` name. 

Regarding your questions:
- I believe the original `sort_index` is appropriate because it represents an `argsort`, that is a widely used operation. So I think that the former API would better be maintained (just call `sort_adj` with `[(j,j=1,n)]` input for the list). No idea why the former name was chosen, but I think now it would be late to change it. 
- In my library, this function is named `sort_and_list`, I agree it's not a great name. A probably better option could be `sort_pair`, as [in other languages](https://stackoverflow.com/questions/279854/how-do-i-sort-a-vector-of-pairs-based-on-the-second-element-of-the-pair) you can do it with a custom comparison function. 

These arguments: 
https://github.com/fortran-lang/stdlib/blob/28b6f719a5eaa9d815c7b801ed13a46f300247ef/src/stdlib_sorting_sort_adj.fypp#L98-L99

could become
```Fortran
         ${t1}$, intent(inout)         :: sorted_array(0:) 
         ${ti}$, intent(inout)         :: adjoint_list(0:) 
```

src/stdlib_sorting.fypp
```diff
@@ -505,6 +576,43 @@ module stdlib_sorting
...
 
+    interface sort_adj
+!! Version: experimental
+!!
+!! The generic subroutine interface implementing the `SORT_ADJ` algorithm,
+!! based on the `"Rust" sort` algorithm found in `slice.rs`
+!! https://github.com/rust-lang/rust/blob/90eb44a5897c39e3dff9c7e48e3973671dcd9496/src/liballoc/slice.rs#L2159
+!! but modified to return an array of indices that would provide a stable
+!! sort of the rank one `ARRAY` input.
+!! ([Specification](../page/specs/stdlib_sorting.html#sort_adj-creates-an-array-of-sorting-indices-for-an-input-array-while-also-sorting-the-array))
```
jvdp1, line 587: to be updated

jvdp1:
Thank you @jalvesz and @perazz for the suggestions.
Based on them I called the new procedure `sort_adjoint` and the associated array `adjoint_array`. I kept the name `array` for the sorted array, as it is used for the other procedures.

`sort_index` just calls `sort_adjoint` with an initialised `index` array.

example/sorting/example_sort_adjoint.f90
```diff
@@ -0,0 +1,15 @@
+program example_sort_adjoint
+  use stdlib_sorting, only: sort_adjoint
+  implicit none
+  integer, allocatable :: array(:)
+  real, allocatable :: adjoint(:)
+
+  array = [5, 4, 3, 1, 10, 4, 9]
```
perazz, line 7: This is a great example @jvdp1. Could it be made into a test as well? The test would check that `array(i)>=array(i-1)`, and that `nint(adjoint(i),kind=${ik}$)==array(i)`

doc/specs/stdlib_sorting.md
```diff
@@ -46,6 +46,9 @@ data:
 * `ORD_SORT` is intended to sort simple arrays of intrinsic data
   that have significant sections that were partially ordered before
   the sort;
+* `SORT_ADJOINT` is based on `ORD_SORT`, but in addition to sorting the
+  input array, it returns a related array re-ordered in the
```
perazz, line 50:
```suggestion
  input array, it re-orders a second array of the same size
```

doc/specs/stdlib_sorting.md
```diff
@@ -46,6 +46,9 @@ data:
...
+  same way;
```
perazz, line 51:
```suggestion
  according to the same permutations;
```

perazz (approved):
LGTM. Pending to add one test program, I think this is ready to merge.

jalvesz:
@perazz @jvdp1 for the test program, maybe something like one of the tests I wrote for the sparse matrices could work?:
```Fortran
    subroutine test_coo2ordered(error)
        !> Error handling
        type(COO_sp_type) :: COO
        integer :: row(12), col(12)
        real    :: data(12)

        row = [1,1,1,2,2,3,2,2,2,3,3,4]
        col = [2,3,4,3,4,4,3,4,5,4,5,5]
        data = 1.0

        call from_ijv(COO,row,col,data)

        call coo2ordered(COO,sort_data=.true.)
        call check(error, COO%nnz < 12 .and. COO%nnz == 9 )
        if (allocated(error)) return

        call check(error, all(COO%data==[1,1,1,2,2,1,2,1,1])  )
        if (allocated(error)) return

        call check(error, all(COO%index(1,:)==[1,1,1,2,2,2,3,3,4])  )
        if (allocated(error)) return

        call check(error, all(COO%index(2,:)==[2,3,4,3,4,5,4,5,5])  )
        if (allocated(error)) return

    end subroutine 
```
Applying the sorting to `row` and having `col` follow. For the sparse matrices I also remove duplicates after sorting, here I think that step is not required.

doc/specs/stdlib_sorting.md
```diff
@@ -140,6 +143,23 @@ argument or allocated internally on the stack.
 Arrays can be also sorted in a decreasing order by providing the argument `reverse
 = .true.`.
 
+#### The `SORT_ADJOINT` subroutine
+
+The `SORT` and `ORD_SORT` subroutines can sort rank 1 isolated
+arrays of intrinsic types, but do nothing for the coordinated sorting
+of related data, e.g., a related rank 1 array. Therefore the module
+provides a subroutine, `SORT_ADJOINT`, that re-order such a rank 1 array
+in the same way as the input array based on the `ORD_SORT` algorithm,
```
jvdp1, line 152:
```suggestion
according to the same permutations as for the input array based on the `ORD_SORT` algorithm,
```

doc/specs/stdlib_sorting.md
```diff
@@ -385,6 +405,84 @@ element of `array` is a `NaN`.
 {!example/sorting/example_radix_sort.f90!}
 ```
 
+#### `sort_adjoint` - sorts an associated array in the same way as the input array, while also sorting the array.
```
jvdp1, line 408:
```suggestion
#### `sort_adjoint` - sorts an associated array 
according to the same permutations as for the input array.
```

doc/specs/stdlib_sorting.md
```diff
@@ -385,6 +405,84 @@ element of `array` is a `NaN`.
...
+
+##### Status
+
+Experimental
+
+##### Description
+
+Returns the input `array` sorted in the direction requested while
+retaining order stability, and an associated array whose elements are
+sorted in the same way as the input `array`.
```
jvdp1, line 418:
```suggestion
sorted according to the same permutations as for the input `array`.
```

example/sorting/example_sort_adjoint.f90
jvdp1: Good suggestion. I added the tests "test_real_sort_adjointes_" based on it.

Issue #853, opened 2024-08-01
CI fails trying to install `pip` package on `msys2`


minhqdao:
### Description

An error recently appeared while trying to install `fypp` using `pip` on `msys2`. The [CI fails](https://github.com/fortran-lang/stdlib/actions/runs/10178599692/job/28152543155?pr=760) with the following error message:

```
Run pip install fypp
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try 'pacman -S
    python-xyz', where xyz is the package you are trying to
    install.
    
    If you wish to install a non-MSYS2-packaged Python package,
    create a virtual environment using 'python -m venv path/to/venv'.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip.
```

### Expected Behaviour

`fypp` is installed without failing.

### Version of stdlib

master

### Platform and Architecture

Windows-MSYS2

zoziha:
@minhqdao, this seems to be caused by changes in `Python` and `Pip`. If we insist on using the system-level installation, we can pass the `--break-system-packages` parameter, such as:
```sh
pip install --break-system-packages fypp
```

Or use the virtual environment `venv`:
```sh
python3 -m venv path/to/venv
source <venv>/bin/activate
pip install fypp
```

minhqdao:
That's correct. And if we take the `venv` route, we'd also need to make sure the `venv` is properly `source`d for every subsequent `run`.

But as you pointed out in https://github.com/fortran-lang/stdlib/pull/854#issuecomment-2267405530, it's probably alright to abandon `msystem: msys` altogether. I'll merge the PR so the CI stops failing.

PR #854, opened 2024-08-01
Remove `msys` from CI and install `fypp` via msys2 package

minhqdao:
- [x] Remove `msystem: msys` from CI.
- [x] Install `fypp` via msys2 package instead of `pip`.

Address https://github.com/fortran-lang/stdlib/issues/853.

### DIFF
--- a/.github/workflows/ci_windows.yml
+++ b/.github/workflows/ci_windows.yml
@@ -23,9 +23,8 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
-    - name: Setup MinGW native environment
+    - name: Setup environment
       uses: msys2/setup-msys2@v2
-      if: contains(matrix.msystem, 'MINGW')
       with:
         msystem: ${{ matrix.msystem }}
         update: false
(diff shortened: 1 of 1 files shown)


zoziha:
@minhqdao, thank you for your sharing. 

As far as I knowA, [`msystem:MSYS` does not have the fypp package](https://packages.msys2.org/base/mingw-w64-python-fypp). In this PR, `msystem:MSYS` will actually use the tools of the `msystem:MINGW64` environment, that is to say, `msystem:MSYS` has no actual effect for CI testing. Combined with the topic of #853, `msystem:MSYS` should be removed.

I have been a user of MSYS2 for few years. I believe that `msystem:MSYS` was originally only used to provide an initial Unix-like compilation environment and tools for other environments, and almost no one will use it as a development environment.

It is sufficient for `Stdlib` to use `msystem:MINGW64`, `msystem:UCRT64`, and `msystem:CLANG64`. `MSYS2` also plans to abandon the 32-bit environment, so we should remove the tests of `msystem:MSYS` and the 32-bit environment (maybe later); Since `Flang` and `LFortran` in the `msystem:CLANG64` environment are not ready yet, `CLANG64` does not need to be supported at present.

minhqdao:
I see, it was probably not the best idea to merge the two setups then – I can revert that. In fact, the idea for this PR was only to fix the setup for `msystem: mingw64`. So we can still do that and leave `msystem: msys` as is for now because it is still working. If there's a consensus on removing `msys` entirely, I'm happy to do that in a separate PR.

minhqdao:
Ok, I see it was `msystem: msys` which failed. Let's probably remove it then.

zoziha (approved):
@minhqdao, thanks, LGTM.

minhqdao:
Will merge this so the CI stops failing. If there's ever the need to bring back `msystem: msys`, we can do https://github.com/fortran-lang/stdlib/issues/853#issuecomment-2302798585.

PR #771, opened 2024-02-18
Add `minizip` and extract `.npz` files

minhqdao:
- [x] Interface with `minizip` to handle `zip` extraction (can be reused for other matrix formats).
- [x] Write tests for `zip` extraction.
- [x] Load `.npz` files.
- [x] Add tests for loading entire `.npz` files.
- [ ] Add examples.
- [ ] Polish code.
- [x] Build `minizip` from source.
- [ ] Add documentation.

Related Issues:
https://github.com/fortran-lang/stdlib/issues/486
https://github.com/fortran-lang/stdlib/issues/763

Future Endeavors:
- Compress `zip` files.
- Save `.npz` files.
- Use zip interface to load and save other matrix formats.

Issue #763, opened 2024-01-16
Support for I/O of standard formats


perazz:
### Motivation

As linear algebra support develops, it would be useful to have simple interfaces to access test matrices to/from standard formats into the stdlib formats. 

### Prior Art

- Matrix Market: https://math.nist.gov/MatrixMarket/ https://math.nist.gov/MatrixMarket/mmio/f/mmiof77.html
- Harwell-Boeing: https://math.nist.gov/MatrixMarket/formats.html#hb-code
- NumPy: https://numpy.org/doc/stable/reference/generated/numpy.load.html
- https://help.imsl.com/java/5.0.1/api/com/imsl/math/SparseMatrixEx2.html
- https://help.imsl.com/c/7.0/html/cmath/default.htm?turl=matrixstoragemodes.htm

### Additional Information

cc @minhqdao

jalvesz:
For reference, scipy IO API: https://docs.scipy.org/doc/scipy/reference/generated/scipy.io.mmwrite.html#scipy.io.mmwrite
the TACO library also offers an API of the sorts for C++ and Python: http://tensor-compiler.org/docs/scientific_computing.html

ivan-pi:
I've added #486 which is also relevant. I would be keen to have the SciPy `.npz` format supported, which is also used to pass sparse matrices.

ivan-pi:
> the TACO library also offers an API of the sorts for C++ and Python: http://tensor-compiler.org/docs/scientific_computing.html

The formats supported by TACO are:
* [Matrix Market (Coordinate) Format (.mtx)](http://math.nist.gov/MatrixMarket/formats.html#MMformat)
* [Rutherford-Boeing Format (.rb)](https://www.cise.ufl.edu/research/sparse/matrices/DOC/rb.pdf)
* [FROSTT Format (.tns)](http://frostt.io/tensors/file-formats.html)

ivan-pi:
There is also a naive text format that MATLAB uses, described under
- [spconvert](https://www.mathworks.com/help/matlab/ref/spconvert.html)
- [Importing Sparse Matrices](https://www.mathworks.com/help/matlab/math/constructing-sparse-matrices.html#f6-34158)

I've used it in the past, to quickly output a matrix and import it into MATLAB for inspection ([spy](https://www.mathworks.com/help/matlab/ref/spy.html)) and calculating the spectral radius.

jalvesz:
[...]

Issue #486, opened 2021-08-18
Save and load in binary, compatible with NumPy/Matlab and others


certik:
* NumPy: https://numpy.org/devdocs/reference/generated/numpy.lib.format.html
* Matlab: https://www.mathworks.com/help/matlab/import_export/mat-file-versions.html
* SciPy code to load/write MAT files: https://docs.scipy.org/doc/scipy/reference/generated/scipy.io.loadmat.html
* Julia library to read/write MAT files: https://github.com/JuliaIO/MAT.jl

First requested [here](https://fortran-lang.discourse.group/t/what-are-your-favorite-and-least-favorite-things-about-fortran/27/49).

milancurcic:
This seems useful and in scope. I often work with MAT files (of various versions) from colleagues and I use SciPy.io `loadmat` and `savemat`.

For my own interoperable binary data between Fortran and Python, I use NetCDF. I don't think any of the language-specific binary formats will beat it in terms of features, performance, or stability. Likewise for HDF5 which is suitable for unstructured data.

MarDiehl:
There is already an HDF5 writer/reader which looks promising: https://github.com/geospace-code/h5fortran. I uses the Fortran bindings of the C library.
I think it is reasonable to keep HDF5 support out of stdlib, it is neither part of the C nor the python standard library.

awvwgk:
I got the basic structure for reading and writing npy files implemented in https://github.com/fortran-lang/stdlib/pull/581. Needs some polishing, especially the reading, and much more unit tests to cover all possible errors the loading can encounter.

TejasAvinashShetty:
[libnpy](http://www.maths.unsw.edu.au/~mclean/libnpy-0.5.tgz) seems to be a  library that provides simple routines for saving a C or Fortran array to a data file using NumPy's own binary format. 
Please see https://scipy-cookbook.readthedocs.io/items/InputOutput.html
> Not my idea  See first [CAZT's](https://stackoverflow.com/users/8644947/cazt) comment on [CAZT's  stackoverflow answer](https://stackoverflow.com/a/46333419/7952027)

jvdp1:
> There is already an HDF5 writer/reader which looks promising: https://github.com/geospace-code/h5fortran. I uses the Fortran bindings of the C library. I think it is reasonable to keep HDF5 support out of stdlib, it is neither part of the C nor the python standard library.

[...]

PR #865, opened 2024-08-23
Handle `npz` files

minhqdao:
Adds handling (loading and saving of `npz` files).

As an `npz` file is nothing but `npy` files zipped together to a single file, we either need to include a library that handles `zip` file creation/extraction for us or call system programs via the command line. In contrast to a previous attempt (https://github.com/fortran-lang/stdlib/pull/771), this PR uses the shell to extract and compress npz files.

- [x] Load `npz` files.
- [x] Save `npz`.
- [x] Add tests.
- [x] Add examples.
- [x] Add documentation.

Related Issues:
https://github.com/fortran-lang/stdlib/issues/486
https://github.com/fortran-lang/stdlib/issues/763


awvwgk (changes requested):
Thank you for this PR, this introduces a lot of needed features in stdlib. However, it is hard to review because it introduces three new features to stdlib, which are dependent but each should have their own independent review.

It would be great if you could open a PR for the filesystems first, a PR for the zip support second and finally a PR for the npz support.

Having dependent PRs is a bit challenging for reviewing. One option would be to push the three branches to the stdlib repo and open PRs from within the stdlib repo, as this improves the experience of reviewing dependent branches. Alternatively, you can open PRs from all three branches from your fork and we try to focus the review only on the parts introduced by the respective PR.

Please let me know if you need any support with this.

doc/specs/stdlib_array.md
```diff
@@ -10,8 +10,58 @@ title: array
 
 Module for index manipulation and array handling tasks.
 
-## Procedures and methods provided
+## Derived types provided
```
jvdp1, line 13: Thank you for this PR. Could you open a PR specific for the additions related to `stdlib_array`, please? It will make the review of this PR easier. Let me know if you need help.

PR #869, opened 2024-09-08
Swap

jalvesz:
Proposal to add an elemental swap interface for base kinds

- [x] elemental subroutines
- [x] tests
- [x] examples
- [x] documentation

**Reference**
https://fortranwiki.org/fortran/show/swap by @urbanjost

cc: @jvdp1 @perazz


perazz:
I think this is a great useful addition @jalvesz. 
The only question I have is about the string version. Consider this example: 

```Fortran
program tswap
    character(5) :: a = '12345'
    character(8) :: b = 'abcdefgh'
    print *, a, ' <-> ', b
    call swap_str(a,b)
    print *, a, ' <-> ', b
contains
    elemental subroutine swap_str(lhs,rhs)
        character(*), intent(inout) :: lhs, rhs
        character(len=max(len(lhs),len(rhs))) :: temp
        temp = lhs ; lhs = rhs ; rhs = temp
    end subroutine
end
```

that [prints](https://godbolt.org/z/oW5bzsWvz): 

```
 12345 <-> abcdefgh
 abcde <-> 12345
```

An option to overcome this issue, would be to fix the string length: 
```Fortran
character(*), intent(inout) :: lhs
character(len=len(lhs)), intent(inout) :: res
```

and then provide an interface for the `stdlib_string_type` maybe?

awvwgk (changes requested):
Thanks for sharing. Can you explicitly test whether `swap(x, x)` works as expected?

jalvesz:
@perazz I think that the limitation here is that this swap interface is intended as a "swap values" under the hypothesis of equal kind and type `lhs` and `rhs`. In order to accomplish a dynamic swap, we would need to swap the memory address, something like:

```Fortran
program tswap
    character(:), allocatable :: a 
    character(:), allocatable :: b
    a = '12345'
    b = 'abcdefgh'
    print *, 'Swap values'
    print *, a, ' <-> ', b
    call swap_str(a,b)
    print *, a, ' <-> ', b

    a = '12345'
    b = 'abcdefgh'
    print *, 'Swap mem address'
    print *, a, ' <-> ', b
    call swap_str_a(a,b)
    print *, a, ' <-> ', b
contains
    elemental subroutine swap_str(lhs,rhs)
        character(*), intent(inout) :: lhs, rhs
        character(len=max(len(lhs),len(rhs))) :: temp
        temp = lhs ; lhs = rhs ; rhs = temp
    end subroutine
    subroutine swap_str_a(lhs,rhs)
        character(:), allocatable, intent(inout) :: lhs, rhs
        character(:), allocatable :: temp
        call move_alloc( lhs , temp )
        call move_alloc( rhs , lhs )
        call move_alloc( temp , rhs )
    end subroutine
end
```
Now, these two procedures cannot be interfaced under the same interface as it leads to an ambiguous interface. But a secondary interface can be added to handle this use-case, like `mem_swap` or something in that line? 

We could propose also a swap interface to the `stdlib_string_type` indeed.

jalvesz:
@awvwgk adding something like this ? :
```Fortran
subroutine test_swap_${k1}$(error)
        ${t1}$ :: x(3), y(3)
        
        x = [${t1}$ :: 1, 2, 3]
        y = [${t1}$ :: 4, 5, 6]

        call swap(x,y)
        
        call check(error, all( x == [${t1}$ :: 4, 5, 6] ) )
        if (allocated(error)) return
        call check(error, all( y == [${t1}$ :: 1, 2, 3] ) )
        if (allocated(error)) return

        ! check self swap
        call swap(x,x)
        
        call check(error, all( x == [${t1}$ :: 4, 5, 6] ) )
        if (allocated(error)) return
end subroutine test_swap_${k1}$
```

awvwgk (approved):
Looks good to me

perazz:
> We could propose also a swap interface to the `stdlib_string_type` indeed.

Exactly @jalvesz, so the stdlib user has one more reason to use the provided string type.

perazz (approved):
Looks great as far as I'm concerned @jalvesz, thank you.

doc/specs/stdlib_math.md
```diff
@@ -61,6 +61,38 @@ Here inputs are of type `real` and kind `sp`
 {!example/math/example_clip_real.f90!}
 ```
 
+### `swap` function
```
jvdp1, line 65:
```suggestion
### `swap` subroutine
```

or

```suggestion
### `swap` procedure
```

doc/specs/stdlib_math.md
```diff
@@ -61,6 +61,38 @@ Here inputs are of type `real` and kind `sp`
...
+
+`call` [[stdlib_math(module):swap(interface)]] ` (lhs, rhs)`
+
+#### Status
+
+Experimental
+
+#### Class
+
+Elemental function.
```
jvdp1, line 81:
```suggestion
Elemental subroutine.
```

doc/specs/stdlib_math.md
```diff
@@ -61,6 +61,38 @@ Here inputs are of type `real` and kind `sp`
...
+
+#### Argument(s)
+
+`lhs`: scalar or array of any of the intrinsic types `integer`, `real`, `complex`, `logical`, `character`, `string_type` type. This argument is `intent(inout)`.
```
jvdp1, line 85: Like for the sorting procedures, `bitset_type` could be also added.

src/stdlib_math.fypp
```diff
@@ -42,6 +42,18 @@ module stdlib_math
     #:endfor
     end interface clip
 
+    interface swap
```
jvdp1, line 45:
```suggestion
    !> Returns the greatest common divisor of two integers
    !> ([Specification](../page/specs/stdlib_math.html#swap_subroutine))
    !>
    !> Version: experimental
    interface swap
```

jvdp1 (commented):
Thank you @jalvesz . LGTM. Here are some minor comments.

doc/specs/stdlib_math.md
jalvesz: I added support for bitset types here https://github.com/fortran-lang/stdlib/pull/869/commits/1993d628402a8cb9fc90fca78da222f73fcfb393 but let me know what you think as I'm not familiar with this DT. An afterthought on that, should the swap interface for DTs from stdlib be here or within the module of the given DT?

src/stdlib_math.fypp
jalvesz: Maybe this as alternative: https://github.com/fortran-lang/stdlib/pull/869/commits/6a30b2ff974aaa9a0a27aa730e0ece321c8b13ac ?

src/stdlib_math.fypp
```diff
@@ -509,5 +525,52 @@ contains
...
+    #:endfor
+
+    elemental subroutine swap_bool(lhs, rhs)
+        logical, intent(inout) :: lhs, rhs
+        logical :: temp
+        temp = lhs; lhs = rhs; rhs = temp
+    end subroutine
+
+    elemental subroutine swap_str(lhs,rhs)
+        character(*), intent(inout) :: lhs, rhs
```
jvdp1, line 554: What happens if len(rhs) > len(lhs)?

src/stdlib_math.fypp
```diff
@@ -509,5 +525,52 @@ contains
...
+        character(len=max(len(lhs),len(rhs))) :: temp
```
jvdp1, line 555: could it be:
```suggestion
        character(len=:), allocatable :: temp
```

src/stdlib_math.fypp
```diff
@@ -509,5 +525,52 @@ contains
...
+        temp = lhs ; lhs = rhs ; rhs = temp
+    end subroutine
+
+    elemental subroutine swap_stt(lhs,rhs)
+        use stdlib_string_type
```
jvdp1, line 560:
```suggestion
        use stdlib_string_type, only: string_type
```

src/stdlib_math.fypp
```diff
@@ -509,5 +525,52 @@ contains
...
+        type(string_type), intent(inout) :: lhs, rhs
+        type(string_type) :: temp
+        temp = lhs ; lhs = rhs ; rhs = temp
+    end subroutine
+
+    #:for k1, t1 in BITSET_KINDS_TYPES
```
jvdp1, line 566: It could be in the same loop as for integers and reals.

src/stdlib_math.fypp
```diff
@@ -42,6 +42,22 @@ module stdlib_math
     #:endfor
     end interface clip
 
+    !> Swap the values of the lhs and rhs arguments
+    !> ([Specification](../page/specs/stdlib_math.html#swap))
```
jvdp1, line 46:
```suggestion
    !> ([Specification](../page/specs/stdlib_math.html#swap_subroutine))
```

jvdp1 (approved):
Thank you @jalvesz . LGTM. I left a few minor comments/suggestions. Feel free to ignore them.

src/stdlib_math.fypp
jalvesz: This is what @perazz pointed out here https://github.com/fortran-lang/stdlib/pull/869#issuecomment-2336693942 so for plain character there would be a loss of information. The string_type does enable avoiding this issue.

src/stdlib_math.fypp
jalvesz: It could, in that case I would need to move the `use stdlib_bitsets` to the top of the module. Should I? No preferences on my side.

src/stdlib_math.fypp
jalvesz: I'm not 100% sure, but I think that with the previous definition the `temp` variable should most likely be created in the stack instead of the heap, which should enable faster mem access. Should be profiled.

src/stdlib_math.fypp
jalvesz: done here https://github.com/fortran-lang/stdlib/pull/869/commits/7370020e002c4f75b050869bb316b1a6f26b1857

perazz:
I see that there are no further comments @jvdp1 @jalvesz, should we wait a couple more days and then merge?

jalvesz:
I've added a warning and extended the example specific to `character`s in https://github.com/fortran-lang/stdlib/pull/869/commits/111996497a40fef5b2a5c5d65872518b62a5112d

test/math/test_stdlib_math.fypp
```diff
@@ -246,6 +256,141 @@ contains
...
+        if (allocated(error)) return
+    end subroutine test_swap_c${k1}$
+    #:endfor
+
+    subroutine test_swap_str(error)
+        type(error_type), allocatable, intent(out) :: error
+        character(5) :: x(2), y(2)
+        
+        x = ['abcde','fghij'] 
+        y = ['fghij','abcde']
```
minhqdao, line 310: I think it would be great to have a test case where strings have different lengths.

jalvesz: done here https://github.com/fortran-lang/stdlib/pull/869/commits/7c5763c2b8fa4de5faa1898891366cb9551b2c0d

jalvesz:
I think this PR is ready to be merged if there are no further comments?

Issue #820, opened 2024-05-17
Adding matrix norms to the `stdlib_linalg` module.


loiseaujc:
### Motivation

I'm currently updating my package `LightKrylov` to make use as much as possible of the new linalg features offered in the `stdlib_linalg` module. Among the things I'll need eventually are functions to compute matrix norms. Is there any on-going work at the moment from @perazz, @jvdp1, or @jalvesz on this subject or could I make this my first task with `stdlib` ?

### Prior Art

- SciPy: `scipy.linalg.norm(A, ord=None, axis=None, keepdims=False, check_finite=True)`. It handles both standard vector norms as well as a variety of vector-induced and non-induced matrix norms.
- Julia: `LinearAlgebra.norm(x, p)` where `x` is an `n`-vector returns the $p$-norm of this vector.
- Julia: `LinearAlgebra.norm(A, p)` where `A` is an `m x n` matrix returns the "entry-wise" $p$-norm of `A`.
- Julia : `LinearAlgebra.opnorm(A, p)` where `A` is an `m x n` matrix returns the vector-induced $p$-norm of `A`.

### Additional Information

While most people might be accustomed to the SciPy standard, my personal preference would still go to the Julia principle of separating the true vector-induced norms (i.e. 1-norm, 2-norm, $\infty$-norm) from the entrywise ones (e.g. Frobenius) although I'm obviously open to discussion.

loiseaujc:
We probably could. Julia is more exhaustive than SciPy for this. Having `norm` and `opnorm` allows them to have both the element wise p-norm (for abitrary p) and the induced one (for p=1, 2 and infinity). SciPy is restricted to the standard operator induced norms + the Frobenius one. I think it mostly depends on how exhaustive we want to be (and how often the non-Frobenius element wise norms are actually being used which I think is not much).

The more I think about it, the more it feels like the Julia way may actually cover more use-cases than are actually used in practice. I still like the difference between `norm` and `opnorm` though just to recall the user that this are fundamentally different mathematical objects. Having the two interfaces might also make it more transparent when a 2D array represent an actual matrix or a collection of vectors instead of having to play with the `axis` keyword.

loiseaujc:
I took some time to think about all the different norms (both for vectors and matrices) I've ever needed in my scientific computing adventure over the past 15 years or so. Here is a non-exhaustive list.

#### Vector norms

- `norm(x, 1)`: 1-norm of a vector (i.e. sum of absolute values),
[...]

PR #871, opened 2024-09-13
linalg: vector norms

perazz:
Address #820 in joint effort with @jalvesz.
Compute several vector norms. Array $A$, real or complex, has general rank `n>=1`. 

### Proposed implementation

- `x = norm(a, order=2 [, dim=dim] [, err=err])`: (pure) function interface
- `call get_norm(a, nrm=x, order=2 [, dim=dim] [, err=err])`: pure subroutine interface

### Key facts
- The following implementations are provided: 

   - `1, '1'`: 1-norm, $\sum_i{ \left|a_i\right| }$
   -  `2, '2', 'Euclidean'`: 2-norm, $\sqrt{\sum_i{ a_i^2 }}$
   -   `>=3`: p-norm, $\left(  \sum_i{ \left|a_i\right| ^p }\right) ^{1/p}$ 
   -    `'inf', huge(0)`: $\infty$-norm, $\max_i{ \left|a_i\right| }$
   -    `'-inf', -huge(0)`: minimum norm, $\min_i{ \left|a_i\right| }$  
 
- `order` can either be an integer (`1, 2, ..., n, -huge(0), huge(0)`) or a character input (`'1', '2', '10', 'inf', '-inf', 'euclidean', ...`)
- The implementation currently only uses Fortran intrinsics that handle all `dim` cases by default: would be hard to unroll all them to enable calls to `BLAS` backends for all ranks 1:15. 

### Progress
- [x] base implementation
- [x] tests
- [x] documentation  
- [x] submodule 
- [x] examples
- [x] all `pure subroutine` interfaces, `pure function` interfaces if no error flag is requested.

### Prior art
- Numpy: `linalg.norm(x, ord=None, axis=None, keepdims=False)`
- Scipy: `norm(a, ord=None, axis=None, keepdims=False, check_finite=True)`


doc/specs/stdlib_linalg.md
```diff
@@ -1382,3 +1382,111 @@ If `err` is not present, exceptions trigger an `error stop`.
...
+
+Experimental
+
+### Description
+
+This function computes one of several vector norms of `real` or `complex` array \( A \), depending on 
+the value of the `order` input argument. \( A \) may be an array of any rank. 
+
+Result `x` returns a `real`, scalar norm value for the whole array; if `dim` is specified, `x` is a rank-1 
+array with the same shape as \(A \) and dimension `dim` collapsed, containing all norms evaluated along `dim`.
```
jvdp1, line 1397:
```suggestion
Result `x` returns a `real` scalar norm value for the whole array; if `dim` is specified, `x` is an array of rank n-1, 
where n equals the rank of ARRAY, and a shape similar to that of \( A \) with dimension `dim` dropped,
containing all norms evaluated along `dim`.
```

doc/specs/stdlib_linalg.md
```diff
@@ -1382,3 +1382,111 @@ If `err` is not present, exceptions trigger an `error stop`.
...
+
+`dim` (optional): Shall be a scalar `integer` value with a value in the range from `1` to `n`, where `n` is the rank of the array. It is an `intent(in)` argument.
+
+`err` (optional): Shall be a `type(linalg_state_type)` value. This is an `intent(out)` argument. If `err` is not present, the function is `pure`. 
+
+### Return value
+
+By default, the return value `x` is a scalar, and contains the norm as evaluated over all elements of the generic-rank array \( A \). 
+If the optional `dim` argument is present, `x` is a rank `n-1` array with the same shape as \( A \) except 
+for dimension `dim`, that is collapsed. Each element of `x` contains the 1D norm of the elements of \( A \),
```
jvdp1, line 1425:
```suggestion
for dimension `dim`, that is dropped. Each element of `x` contains the 1D norm of the elements of \( A \), 
```

doc/specs/stdlib_linalg.md
```diff
@@ -1382,3 +1382,111 @@ If `err` is not present, exceptions trigger an `error stop`.
...
+### Status
+
+Experimental
+
+### Description
+
+This `pure subroutine` interface computes one of several vector norms of `real` or `complex` array \( A \), depending on 
+the value of the `order` input argument. \( A \) may be an array of any rank. 
+
+Result `nrm` returns a `real`, scalar norm value for the whole array; if `dim` is specified, `nrm` is a rank-1
```
jvdp1, line 1449:
```suggestion
Result `nrm` returns a `real`, scalar norm value for the whole array; if `dim` is specified, `nrm` is a rank n-1 
```

doc/specs/stdlib_linalg.md
```diff
@@ -1382,3 +1382,111 @@ If `err` is not present, exceptions trigger an `error stop`.
...
+
+Experimental
+
+### Description
+
+This `pure subroutine` interface computes one of several vector norms of `real` or `complex` array \( A \), depending on 
+the value of the `order` input argument. \( A \) may be an array of any rank. 
+
+Result `nrm` returns a `real`, scalar norm value for the whole array; if `dim` is specified, `nrm` is a rank-1 
+array with the same shape as \(A \) and dimension `dim` collapsed, containing all norms evaluated along `dim`.
```
jvdp1, line 1450:
```suggestion
array with the same shape as \(A \) and dimension `dim` dropped, containing all norms evaluated along `dim`.
```

src/stdlib_linalg.fypp
```diff
@@ -1065,6 +1067,189 @@ module stdlib_linalg
...
+           type(linalg_state_type), intent(out), optional :: err   
+        end subroutine norm_${rank}$D_${ii}$_${ri}$
+        #:endfor
+        !> Array norms: ${rt}$
+        #:for rank in range(2, MAXRANK + 1)
+        pure module subroutine norm_${rank}$D_to_${rank-1}$D_${ii}$_${ri}$(a, nrm, order, dim, err)
+           !> Input matrix a[..]
+           ${rt}$, intent(in), target :: a${ranksuffix(rank)}$
+           !> Dimension to collapse by computing the norm w.r.t other dimensions
+           !  (dim must be defined before it is used for `nrm`)
```
jvdp1, line 1239: What do you mean?

src/stdlib_linalg_norms.fypp
```diff
@@ -0,0 +1,301 @@
...
+        
+        select case(norm_request)
+            case(NORM_ONE)
+                nrm = sum( abs(a) )
+            case(NORM_TWO)            
+                #:if rt.startswith('complex')
+                nrm = sqrt( real( sum( a * conjg(a) ), ${rk}$) )
+                #:else
+                nrm = sqrt( sum( a ** 2 ) )
+                #:endif
```
jvdp1, line 175: What is the difference with the intrinsic `norm2`? Could we use it instead of `sqrt(sum(a**2))`?

jvdp1 (commented):
Overall LGTM. Thank you @perazz and @jalvesz .
It could be worthwhile to add a note in the specs to highlight the differences (or not) with the intrinsic `norm2`.

src/stdlib_linalg.fypp
perazz: [aa734de](https://github.com/fortran-lang/stdlib/pull/871/commits/aa734de2707b808ccfcc2fc8a390c63543d37d5f)

src/stdlib_linalg_norms.fypp
perazz: [437b96e](https://github.com/fortran-lang/stdlib/pull/871/commits/437b96eb020516d25048b810833884f69689ac8d)

awvwgk:
I believe the BLAS API should allow also strided access to array elements, which would support `dim != 1`, the stride length would need to be computed based on the selected axis and the dimensions.

perazz:
The compiler will create a temporary array passing from `(:)` to the `(*)` BLAS API, because the latter implies `contiguous` data. So, it is possible to infer the stride of an `(:)` array like I did, but there is no guarantee that the compiler will not create a temporary array in between the calls. See [this example](https://godbolt.org/z/zr3MsfhMv). 

So, in the interest of safety, I suggest to avoid inferring strides and rather create the temporary ourselves, i.e. preprocessing it with a `reshape(a, order=sorted_dims)`

doc/specs/stdlib_linalg.md
```diff
@@ -1382,3 +1382,112 @@ If `err` is not present, exceptions trigger an `error stop`.
...
+### Status
+
+Experimental
+
+### Description
+
+This function computes one of several vector norms of `real` or `complex` array \( A \), depending on 
+the value of the `order` input argument. \( A \) may be an array of any rank. 
+
+Result `x` returns a `real` scalar norm value for the whole array; if `dim` is specified, `x` is an array of rank n-1,
```
jalvesz, line 1396: This paragraph is mentioned just below in the "return value" section. Maybe it can be dropped from the Description section.

doc/specs/stdlib_linalg.md
```diff
@@ -1382,3 +1382,112 @@ If `err` is not present, exceptions trigger an `error stop`.
...
+Raises `LINALG_VALUE_ERROR` if any of the arguments has an invalid size.
+If `err` is not present, exceptions trigger an `error stop`.
+
+### Example
+
+```fortran
+{!example/linalg/example_norm.f90!}
+```
+
+## `get_norm` - Computes the vector norm of a generic-rank array.
```
jalvesz, line 1439: Given that the subroutine contains the core implementation and the function calls it from within, I'm wondering if `get_norm` should be introduced before, and the description of `norm` should make reference to it to also avoid duplicating information and just describing the difference between the funcion interface and the subroutine interface?

src/stdlib_linalg_norms.fypp
```diff
@@ -0,0 +1,336 @@
...
+        integer(ilp), intent(out) :: norm_type
+        !> State return flag
+        type(linalg_state_type), intent(out) :: err
+        
+        select case (order)
+           case (1_ilp)
+               norm_type = NORM_ONE
+           case (2_ilp)
+               norm_type = NORM_TWO
+           case (3_ilp:huge(0_ilp)-1_ilp)
```
jalvesz, line 56:
```suggestion
           case (3_ilp:NORM_POW_LAST)
```

src/stdlib_linalg_norms.fypp
```diff
@@ -0,0 +1,336 @@
...
+               norm_type = order
+           case (huge(0_ilp):)
```
jalvesz, line 58:
```suggestion
           case (NORM_INF:)
```

src/stdlib_linalg_norms.fypp
```diff
@@ -0,0 +1,336 @@
...
+               norm_type = NORM_INF
+           case (:-huge(0_ilp))
```
jalvesz, line 60:
```suggestion
           case (:NORM_MINUSINF)
```

src/stdlib_linalg_norms.fypp
```diff
@@ -0,0 +1,336 @@
...
+                #:if rank==1
+                nrm = nrm2(sze,a,incx=1_ilp)
+                #:elif rt.startswith('complex')
+                nrm = sqrt( real( sum( a * conjg(a) ), ${rk}$) )                
+                #:else
+                nrm = norm2(a)
+                #:endif             
+            case(NORM_INF)
+                nrm = maxval( abs(a) )
+            case(-NORM_INF)
```
jalvesz, line 212:
```suggestion
            case(NORM_MINUSINF)
```

src/stdlib_linalg_norms.fypp
```diff
@@ -0,0 +1,336 @@
...
+                nrm = sum( abs(a) , dim = dim )                
+            case(NORM_TWO)                
+                #:if rt.startswith('complex')
+                nrm = sqrt( real( sum( a * conjg(a) , dim = dim ), ${rk}$) )
+                #:else
+                nrm = norm2( a , dim = dim )
+                #:endif
+            case(NORM_INF)
+                nrm = maxval( abs(a) , dim = dim )
+            case(-NORM_INF)
```
jalvesz, line 320:
```suggestion
            case(NORM_MINUSINF)
```

example/linalg/example_get_norm.f90
```diff
@@ -0,0 +1,51 @@
+! Vector norm: demonstrate usage of the function interface
+program example_get_norm
+  use stdlib_linalg, only: get_norm, linalg_state_type
+  implicit none
+  
+  real :: a(3,3),nrm,nrmd(3)
```
jalvesz, line 6:
```suggestion
  real :: a(3,3), nrm, nrmd(3)
```

example/linalg/example_get_norm.f90
```diff
@@ -0,0 +1,51 @@
...
+  print *, 'maxval(||a(i,:)||)  = ',nrmd  ! 3.00000000       4.00000000       5.00000000
+  
+  call get_norm(a, nrm, '-inf')
+  print *, 'minval(||a||)       = ',nrm   ! 0.00000000
+  
+  call get_norm(a, nrmd, '-inf', dim=1)
+  print *, 'minval(||a(:,i)||)  = ',nrmd  ! 1.00000000       0.00000000       3.00000000 
+  
+  ! Catch Error: 
+  ! [norm] returned Value Error: dimension 4 is out of rank for shape(a)= [3 3]
```
jalvesz, line 47:
```suggestion
  ! [norm] returned Value Error: dimension 4 is out of rank for shape(a)= [3, 3]
```

example/linalg/example_norm.f90
```diff
@@ -0,0 +1,40 @@
...
+  print *, 'Columns norms       = ',norm(a, 2, dim=1) ! 3.74165750       2.23606801       7.07106781 
+  
+  ! Infinity norms
+  print *, 'maxval(||a||)       = ',norm(a, 'inf')          ! 5.00000000
+  print *, 'maxval(||a(i,:)||)  = ',norm(a, 'inf', dim=2)   ! 3.00000000       4.00000000       5.00000000
+  print *, 'minval(||a||)       = ',norm(a, '-inf')         ! 0.00000000
+  print *, 'minval(||a(:,i)||)  = ',norm(a, '-inf', dim=1)  ! 1.00000000       0.00000000       3.00000000 
+  
+  ! Catch Error: 
+  ! [norm] returned Value Error: dimension 4 is out of rank for shape(a)= [3 3]
```
jalvesz, line 36:
```suggestion
  ! [norm] returned Value Error: dimension 4 is out of rank for shape(a)= [3, 3]
```

jalvesz (changes requested):

doc/specs/stdlib_linalg.md
perazz: [91882fa](https://github.com/fortran-lang/stdlib/pull/871/commits/91882fa9bf68871f2259033f01a26fff1389b8e4)

doc/specs/stdlib_linalg.md
perazz: [1a79128](https://github.com/fortran-lang/stdlib/pull/871/commits/1a79128d26fd55c103023d5a901fe59bc775b6d9)

perazz:
- [x] unify calls to norm evaluation function
- [x] $+\infty$ norm: replace `maxval(abs(.))` with calls to BLAS, add test 
- [x] ND and `dim`-med versions:  always call LAPACK/BLAS rather than Fortran intrinsics
- [x] address contiguity via `reshape(a, order=sorted_dims)`

jalvesz:
LGTM @perazz! On my end I have no further comments.

jvdp1 (approved):
Thank you @perazz . This sounds ready to be merged! Nice addition.

perazz:
Thank you @jvdp1, I'd wait a few more days and then merge in absence of comments, so we can pave the way for the matrix norms.

PR #873, opened 2024-09-13
API: make LAPACK-backed APIs since Release <=0.7.0 `stable`

perazz:
Due to lack of complaints, 
this is a proposal to stabilize the LAPACK-based high-level Linear Algebra APIs that have appeared in stdlib release 0.7.0 

(all but `Cholesky`, `norm`s and `qr` factorization). 


jvdp1 (approved):

perazz:
Thank you @jvdp1. Let's see if there are other comments. It would be valuable to have these older APIs stabilized before the next release.

awvwgk (approved):
Good idea.

perazz:
Merging.

Issue #201, opened 2020-05-30
File system operations


MarDiehl:
While some functionality for file system related operations exist in Fortran, some rather relevant operations are not standardized.
For example, figuring out whether a path is a directory:
https://stackoverflow.com/questions/9522933

A function like "isDirectory" can be based on the corresponding C-functionality. Would that be an appropriate solution? I would certainly require a bunch of #ifdefs in the C-side of the implementation

certik:
This is related to #14. I think we agreed that file system operations are in the scope. The naming convention should probably be `is_directory` or `is_dir`, and we should also look at how Python, Julia and Matlab name such functions. We are always trying to be consistent with other languages where it makes sense.

The main goal of stdlib is to figure out and document (in a spec) the API. The underlying implementation is secondary --- stdlib will provide a reference implementation and then compiler vendors are free to provide their own, different or more optimized implementation if they want. The only requirement is that it must run on all platforms, but probably calling into C if we have to would do it.

milancurcic:
#22 for POSIX systems is also related.

MarDiehl:
I have thought a little bit about a possible contribution from my side, and I propose to add general path related functionality. Since the other language I am frequently using is python, I got my inspiration from
1.  [os.path](https://docs.python.org/3/library/os.path.html)
1.  [pathlib](https://docs.python.org/3/library/pathlib.html)

In python, the object oriented approach of pathlib is often favored over os.path. However, unless I overlooked something, I believe that the os.path approach is better suited for Fortran because one can not chain function calls.

- Python: `p = path.absolute(),is_file()`
- Fortran: `p = path_is_file(path_absolute(Str))`

With Fortran variable length strings (allocatable), most operations are easily performed. However, there is one exception:
`os.path.commonpath(paths)` takes a list of strings, in Fortran this would be an array of strings/characters and all of them would need to have the same length. Alternatively, one can use an interface for a series of `fpp` generated functions with signatures of type `path_commonpath(path1,path2)`,  `path_commonpath(path1,path2,path3)`, ...

While most functionality will be implemented in pure Fortran, certain operations require file system functions from C.

There is one more thing I need to mention: If Windows support is needed, someone else has to provide it. I've never compiled a Fortran program on windows and the path operations differ significantly. `pathlib` has essentially two implementations and I assume we are in the same situation. If someone volunteers, I would prefer that we work in parallel on both implementation. My time budget is about 4h/week and I would hope to finish the whole implementation in 3 months.

[...]

PR #874, opened 2024-09-16
Add filesystem interaction

minhqdao:
This PR introduces filesystem interactions by pragmatically invoking the system shell. It is part of the broader strategy outlined in https://github.com/fortran-lang/stdlib/pull/865 (see https://github.com/fortran-lang/stdlib/pull/865#pullrequestreview-2259189900) to support file zipping and unzipping. These functionalities are essential for handling `.npz` files.


src/CMakeLists.txt
```diff
@@ -109,6 +109,7 @@ set(SRC
     stdlib_hashmaps.f90
     stdlib_hashmap_chaining.f90
     stdlib_hashmap_open.f90
+    stdlib_io_filesystem.f90
```
gnikit, line 112:
```suggestion
    stdlib_io_filesystem.F90
```

If you capitalise the extension compilers will turn on their respective preprocessors

test/io/test_filesystem.f90
```diff
@@ -0,0 +1,258 @@
...
+        if (stat /= 0) then
+            call test_failed(error, "Creating directory '"//temp_list_dir//"' failed."); return
+        end if
+
+        call run('touch '//temp_list_dir//'/'//filename1, iostat=stat)
+        if (stat /= 0) then
+            call test_failed(error, "Creating file 1 in directory '"//temp_list_dir//"' failed."); return
+        end if
+
+        call run('mkdir '//temp_list_dir//'/'//dir, iostat=stat)
```
zoziha, line 209: [There is an error in CI under Windows OS](https://github.com/fortran-lang/stdlib/actions/runs/10888590749/job/30213419329?pr=874#step:7:1004). This is because the shell command `mkdir` called may follow the `mkdir` command in `CMD.exe`, like `mkdir \a\b\c\d`. The path connector uses `\` instead of `/`. It may be possible to implement a routine to determine the current operating system in Stdlib later.

A temporary solution can be to make two judgments. If the `/` scheme fails, then try the `\` scheme.

src/stdlib_io_filesystem.f90
```diff
@@ -0,0 +1,112 @@
...
+        !> Error message.
+        character(len=:), allocatable, optional, intent(out) :: iomsg
+
+        integer :: unit, stat
+        character(len=256) :: line
+
+        stat = 0
+
+        if (.not. exists(temp_dir)) then
+            call run('mkdir '//temp_dir, stat)
```
zoziha, line 51: Note that under Windows OS, `mkdir temp` will create a `temp` folder under the current path `pwd`. Is this appropriate?

zoziha (approved):
Thanks for this work, @minhqdao. Looking forward to the Stdlib starting to integrate filesystem-related routines. This is the beginning of a new category. In addition to the two noted issues, the function interfaces introduced by this PR should be documented in the specifications later.

(see also [`fpm::filesystem`](https://github.com/fortran-lang/fpm/blob/da6d1bf89a241cd87d640b889ab1e570e8be5f35/src/fpm_filesystem.F90#L490))

test/io/test_filesystem.f90
jalvesz: A simple solution that I've been using and think would be useful to have in stdlib would be to add in the constants module something like:
```Fortran
#ifdef _WIN32
  character(len=1), parameter, public :: os_sep='\'
#else
  character(len=1), parameter, public :: os_sep='/'
#endif
```
Which then would enable the equivalent of python
```python
import os
print(os.sep)
```

arjenmarkus:
Windows has NO command "rm", I just confirmed that. You need "del" instead
and quite possibly "del /q" to make sure no questions aer asked.

To get a clean listing of all the files in a directory, use "dir /b".

A side note: many commands on Windows do accept the forward slash as well
as the backward slash to indicate directories. Although, you can get
strange surprises. Where "cd" accepts the forward slash, "dir" does not. Oh
well, best stick with the backslash ;).

Op do 19 sep 2024 om 08:22 schreef Federico Perini ***@***.***

test/io/test_filesystem.f90
zoziha: The C preprocessor **in Fortran** does not predefine the `_WIN32` macro definition. This may require CMake to determine the operating system and define `_WIN32`. There are indeed two solutions:
1. During compilation, solve this problem through the C preprocessor and CMake macro definitions.
2. During runtime, solve this problem by implementing a routine to determine the operating system.

jalvesz: The C preprocessor approach is already in use in stdlib_system, so it could be one solution. The runtime approach is also very good.

src/CMakeLists.txt
minhqdao: It worked, thanks 👍

minhqdao:
Changes made:

- Activated cpp by using capitalized `.F90` suffix.
- Check `is_windows` during runtime.
- Add platform-agnostic `path_separator` (`/` vs. `\`).
- Use `rmdir \s\q` on Windows to remove directories.
- Use `dir \b` on Windows.
- All tested.
- Add documentation.

@zoziha @jalvesz @arjenmarkus @perazz

src/stdlib_io_filesystem.f90
minhqdao: It is a very pragmatic solution, but I think it should be fine for now

src/stdlib_io_filesystem.f90
```diff
@@ -0,0 +1,112 @@
...
+            files = [files, string_type(line)]
+        end do
+        close(unit, status="delete")
+    end
+
+    !> Version: experimental
+    !>
+    !> Run a command in the shell.
+    !> [Specification](../page/specs/stdlib_io.html#run)
+    subroutine run(command, iostat, iomsg)
```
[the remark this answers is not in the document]
minhqdao: I think it slightly exceeds the scope of this PR but it can easily be done in a follow-up PR

test/io/test_filesystem.f90
minhqdao: I chose the runtime approach because, as mentioned by @zoziha, Windows-GFortran does not set the `_WIN32` macro. We had that problem in `fpm` already and I don't think the solution in `stdlib_system` works on Windows using `gfortran`.

src/stdlib_io_filesystem.F90
```diff
@@ -0,0 +1,181 @@
...
+
+    character(*), parameter :: temp_dir = 'temp'
+
+contains
+
+    !> Version: experimental
+    !>
+    !> Whether the operating system is Windows.
+    !> [Specification](../page/specs/stdlib_io.html#is_windows)
+    logical function is_windows()
```
zoziha, line 19: In [fpm](https://github.com/fortran-lang/fpm/blob/da6d1bf89a241cd87d640b889ab1e570e8be5f35/src/fpm_environment.f90#L66), the method of using local `save` variables seems very advisable to avoid repeatedly accessing environment variables when `is_windows` is frequently called.

jalvesz:
Hi I went through the PR again and would like to make a few comments. 

The first one is purely cosmetic: I think that all procedures should have `end function` or `end subroutine`, same for the module. While a standalone `end` works, it is not a common practice and it felt somehow unnatural. For the sake of coherence with the rest of the library, I would suggest to keep the full closing statement.

The second point: I would like to come back to the preprocessing... I know that it is a sad state of affairs that `gfortran` does not expose `gcc` defined macros. Yet, on one hand, I find terribly worrisome to add all those operations just to detirme the OS and then to get the proper path separator. I find it hard to justify all those runtime operations for something that can be a compile-time constant. So I would like to come back to proposing setting the path separator as a simple `character(1), parameter, public` constant which value is defined by the C preprocessor mechanism. Same for `is_windows` which can be a `logical, parameter, public` constant.

This does indeed impose: either relying on CMake, or documenting that one should pass the appropriate definition flags if using plain Makefiles or `fpm`. This build-time "burden" seems to me more reasonable compared to the runtime operations.

I've opened a discussion to see if `fpm` could also help improving this preprocessing situation here https://github.com/fortran-lang/fpm/issues/1084

While I find it "tolerable" for fpm having to rely on the runtime implementation as a tool for building, I'm expecting stdlib to be used for more computationally intensive tasks. Thus, it seems to me better to shift the issue to the building documentation and not the library.

jvdp1:
Thank you @minhqdao for this PR. In addition to all other comments, I think that the proposed features in this PR should have their own modules and specs , and not be related to `stdlib_io`. what about calling it `stdlib_filesystem` or even `stdlib_os`?
 See also #201 and [this initial project](https://github.com/MarDiehl/stdlib_os)

minhqdao:
@zoziha @jalvesz @jvdp1 

- Runtime checks to determine the OS are eliminated.

But please note for future apis that check `OS`: When building with `CMake`, we use [CMAKE_SYSTEM_NAME](https://gitlab.kitware.com/cmake/community/-/wikis/doc/tutorials/How-To-Write-Platform-Checks#platform-variables) and when building with `fpm`, we use [platform.system()](https://docs.python.org/3/library/platform.html#platform.system) (`python`). The results might not be the same. It seems to work for `OS == "Windows"`, though.

Further: 
- Added `end` annotations.
- Updated tests.
- Extracted `stdlib_filesystem`.

However, `is_windows` is not a filesystem feature but a `os` one. `exists` gives information on the filesystem. And `run` is ... neither (?). I don't want to have too many files as it might end up a bit messy (which is what happened in `fpm`), so maybe we should just generalize it under `stdlib_os`? Or `stdlib_fs_os`? What are your thoughts?

jalvesz:
Thanks @minhqdao for the updates. And sorry for the late replay!

I think that generalizing under `stdlib_os` module is a good idea!

Just one last request, could we keep the preprocessing as cpp instead of fypp? this would enable compiling the same fypp pre-pre-processed code on different OS. I personally do this systematically thanks to the WSL on windows.

With CMake nothing to do (revert change in current PR).

With the python script for fpm `fypp_deployment.py` add the `platform` module and change the function fpm_build:
```python
import os
import platform
...

def fpm_build(args,unknown):
    import subprocess
    #==========================================
    # check compilers
    FPM_FC  = os.environ['FPM_FC']  if "FPM_FC"  in os.environ else "gfortran"
    FPM_CC  = os.environ['FPM_CC']  if "FPM_CC"  in os.environ else "gcc"
    FPM_CXX = os.environ['FPM_CXX'] if "FPM_CXX" in os.environ else "gcc"
    #==========================================
    # Filter out flags
    preprocessor = { 'gfortran':'-cpp ' , 'ifort':'-fpp ' , 'ifx':'-fpp ' }
    flags = preprocessor[FPM_FC]
    for idx, arg in enumerate(unknown):
        if arg.startswith("--flag"):
            flags= flags + unknown[idx+1]
    flags = flags + "-D{}".format(platform.system().upper())
    #==========================================
    # build with fpm
    subprocess.run("fpm build"+
                   " --compiler "+FPM_FC+
                   " --c-compiler "+FPM_CC+
                   " --cxx-compiler "+FPM_CXX+
                   " --flag \"{}\"".format(flags), shell=True, check=True)
    return
```

In `stdlib_system.F90` we can change the preprocessing line for:
```Fortran
#if defined(WIN32) || defined(WIN64) || defined(WINDOWS)
```
(the first two are provided by CMake, the latter is what the python script would add to the flags)
This same idea can be used in the current module.

This enables building with fpm (throught the script) with
```
python .\config\fypp_deployment --build
```
I tested it on PowerShell / CMD / Ubuntu (WSL) and it works.

Issue #875, opened 2024-09-23
Compilation failed when using gfortran flag default-real-8 and stdlib


cyrilgandon:
### Description

I am trying to compile my program in 64 bits. I use gfortran and the option `default-real-8` to compile :

`-fdefault-real-8    Set the default real type to an 8 byte wide type. `

but got an error:

```
build\dependencies\stdlib\src\stdlib_specialfunctions_gamma.f90:1175:18:

 1175 |             res = l_gamma(n + 1, 1.0D0)
      |                  1
Error: There is no specific function for the generic 'l_gamma' at (1)
```

If I believe the error, the function containing this call is 

``` fortran
impure elemental function l_factorial_iint64(n) result(res)
    integer(int64), intent(in) :: n

    [...]
    
    res = l_gamma(n + 1, 1.0D0)
end function
```

So `n` being int(8) and `1.0D0` being real(8), I think it should find this overload:

```
impure elemental function l_gamma_iint64dp(z, x) result(res)
    integer(int64), intent(in) :: z
    real(dp), intent(in) :: x

    [...]

end function
```

But apparently not. Pretty new to Fortran and the `kind` stuff, I guess I am missing something!

Edit: I extract the code, and the error is more clear:

    Type mismatch in argument 'x' at (1); passed REAL(16) to REAL(8)

So I guess default-real-8 is promoting 1.0D0 from REAL(8) to REAL(16)

### Expected Behaviour

The stdlib should compile with a default real of size 8.

### Version of stdlib

ccdba91a24be1d70c6e211a76f71068a8bf261d5

### Platform and Architecture

Windows

### Additional Information

Also post a question on SO: https://stackoverflow.com/questions/79013476/compilation-failed-when-using-gfortran-flag-default-real-8-and-stdlib

PR #876, opened 2024-09-23
fix(factorial): result should be of the default kind, not double default kind

cyrilgandon:
Fixes https://github.com/fortran-lang/stdlib/issues/875

The parameter `1.0D0` is promoted to a real kind 16 when use in conjunction with the flag of gfortran `default-real-8`, and there is not such an overload for the `l_gamma` function.  
Using a double precision number seems a bug here, nowhere else in the stdlib we have this usage.  

Especially when the result is defined as a default real:

    real :: res

I don't see a reason to pass a double precision parameter to the `l_gamma` function. I propose to drop the double precision here.

Edit: here is a comment by OP

https://github.com/fortran-lang/stdlib/pull/625#issuecomment-1048341303

>By using higher precision during calculation, the desired precision digits are guaranteed to be correct. So single precision uses double precision for calculation, double precision uses quadruple precision for calculation. Extended precision calculation may need higher precision a little over quadruple, while quadruple precision calculation need much higher precision not available right now.

So I'm not sure anymore that I should drop that double precision.  
Maybe @Jim-215-Fisher can help here. What should happen if default real size is 64bit, should we still do calculation with 128bits real? I think not since most of architecture are not 128bits..

I'd say that double precision logic is included is the gamma functions, no need to have that at the level of the l_factorial.


perazz (approved):
LGTM, this looks like the right thing to do here. Thank you @cyrilgandon

cyrilgandon:
After reading attentively the code, the 2nd parameter of the `l_gamma` function is very important and use only for it's kind.  
So it is in fact a huge change to remove the doubling.

Here is some test on log(10!):

> Original code with doubling : 15.1044130325
> PR code removing doubling : 15.1044120789
> Expected                                : 15.1044125731

perazz:
@cyrilgandon, this is very interesting. It looks like the error here is close to `epsilon(0.0)`, so, perhaps it would be a good idea to shift all `real` variables to `real(dp)`? So, the internal accuracy would be higher, and anyways independent of the compiler's default real choice.

cyrilgandon:
@perazz Yes!! I think it's an easy fix and should answer the problem. For most people having the default real(4), it will not change the result.

perazz:
If I read the OP's comment well, a double precision version (integer(int64) input as far as I understand) may benefit from a quad-precision internal calculation. 
However, the `real` accuracy has not been templated here, so I believe it is good enough to promote it to a fixed 64-bit precision, like this PR is suggesting. Maybe, it would make sense to spread the same precision to all `real` variables, including the function result. That would not introduce a backward incompatible change, as a function result can be automatically converted (and the new version provides higher precision).

cyrilgandon:
Yes that makes sense, also rename zero_k2 to zero_dp, as the k2 paramter did not exist in this function

perazz (approved):
LGTM. @jvdp1 what do you think?

jvdp1 (approved):
LGTM. Thank you @cyrilgandon

jvdp1:
Thank you @cyrilgandon . I will merge this PR.

Issue #862, opened 2024-08-15
`loadtxt` test occasionally fails


minhqdao:
### Description

It's a little bit tricky to reproduce but the `loadtxt` test occasionally fails (maybe 10-20% of the time). Mostly, it happens in the CI with the combination ubuntu-latest/gcc/12/cmake:

```
159/324 Test #158: loadtxt ................................***Failed    0.01 sec
At line 171 of file /home/runner/work/stdlib/stdlib/build/src/stdlib_io.f90
Fortran runtime error: Bad value during floating point read

Error termination. Backtrace:
#0  0x7fafd0c23960 in ???
#1  0x7fafd0c244d9 in ???
#2  0x7fafd0e75aba in ???
#3  0x7fafd0e796b9 in ???
#4  0x7fafd0e7ae55 in ???
#5  0x7fafd0e77b7a in ???
#6  0x55e0c2cce272 in __stdlib_io_MOD_loadtxt_rsp
#7  0x55e0c2cc33c7 in MAIN__
#8  0x55e0c2cc345b in main
```

However, I've also seen it happening with `intel` and `intel-classic`:

```
157/324 Test #158: loadtxt ................................***Failed    0.01 sec
forrtl: severe (24): end-of-file during read, unit -129, file /home/runner/work/stdlib/stdlib/example/io/example.dat
Image              PC                Routine            Line        Source             
example_loadtxt    0000000000436426  Unknown               Unknown  Unknown
example_loadtxt    00000000004049DB  Unknown               Unknown  Unknown
example_loadtxt    000000000040423C  Unknown               Unknown  Unknown
example_loadtxt    00000000004041CD  Unknown               Unknown  Unknown
libc.so.6          00007F854EA29D90  Unknown               Unknown  Unknown
libc.so.6          00007F854EA29E40  __libc_start_main     Unknown  Unknown
example_loadtxt    00000000004040E5  Unknown               Unknown  Unknown
```

### Expected Behaviour

Test reliably not failing.

### Version of stdlib

master

### Platform and Architecture

CI: Mostly ubuntu-latest, gcc-12, cmake, but also with other combinations, e.g. `ifort`

chuckyvt:
Note the following thread on the Fortran discourse has some similarities to this issues.  

https://fortran-lang.discourse.group/t/why-am-i-getting-an-error-with-list-directed-read/8562

perazz:
[...]

PR #877, opened 2024-09-27
`loadtxt/savetxt`: do not require space after last entry

perazz:
Attempt to address #862. 

The current formats require an end-of-row space both in `savetxt` and `loadtxt`, i.e., 

```
format(*(i0,1x))
```

This PR proposes to not require a space after the last record, i.e., 
 
```
format(*(i0,:,1x))
```

This should hopefully improve the seldom crashing CI tests. 
@fortran-lang/stdlib @chuckyvt @minhqdao

### DIFF
--- a/src/stdlib_io.fypp
+++ b/src/stdlib_io.fypp
@@ -144,10 +144,15 @@ contains
       do i = 1, skiprows_
         read(s, *)
       end do
-
-      #:if 'real' in t1
+      
       ! Default to format used for savetxt if fmt not specified.
-      fmt_ = optval(fmt, "(*"//FMT_REAL_${k1}$(1:len(FMT_REAL_${k1}$)-1)//",1x))")
+      #:if 'real' in t1
+      fmt_ = optval(fmt, "(*"//FMT_REAL_${k1}$(1:len(FMT_REAL_${k1}$)-1)//",:,1x))")
+      #:elif 'complex' in t1
+      fmt_ = optval(fmt, "(*"//FMT_COMPLEX_${k1}$(1:len(FMT_COMPLEX_${k1}$)-1)//",:,1x))")
+      #:else
+      fmt_ = optval(fmt, "*")
+      #:endif      
 
       if ( fmt_ == '*' ) then
         ! Use list directed read if user has specified fmt='*'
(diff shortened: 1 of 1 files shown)


src/stdlib_io.fypp
```diff
@@ -120,7 +120,8 @@ contains
       !!     ...
       !!
       integer :: s
-      integer :: nrow, ncol, i, skiprows_, max_rows_
+      integer :: nrow, ncol, i, ios, skiprows_, max_rows_
+      character(len=128) :: iomsg,msgout
```
jvdp1, line 124:
```suggestion
      character(len=128) :: iomsg, msgout
```

src/stdlib_io.fypp
```diff
@@ -142,56 +143,51 @@ contains
       allocate(d(max_rows_, ncol))
 
       do i = 1, skiprows_
-        read(s, *)
+        read(s, *, iostat=ios, iomsg=iomsg)
+        
+        if (ios/=0) then 
+           write(msgout,1) trim(iomsg),i,trim(filename) 
+           call error_stop(msg=trim(msgout))
```
jvdp1, line 150: Is it still wise to use `call error stop(msg)` instead of `error stop msg`? If I remember well, `call error_stop` was introduced because some compilers did not support `error stop` yet. However, I don't think it is still the case now.

src/stdlib_io.fypp
```diff
@@ -142,56 +143,51 @@ contains
...
+          
+          if (ios/=0) then 
+             write(msgout,1) trim(iomsg),i,trim(filename) 
+             call error_stop(msg=trim(msgout))
+          end if          
+          
         enddo
       else
-        ! Otherwise pass default user specified fmt string.
+        ! Otherwise pass default or user specified fmt string.
```
jvdp1, line 176:
```suggestion
        ! Otherwise pass default or user specified fmt string.
```

src/stdlib_io.fypp
```diff
@@ -218,20 +214,31 @@ contains
       !!```
       !!
 
-      integer :: s, i
+      integer :: s, i, ios
+      character(len=128) :: iomsg,msgout
```
jvdp1, line 218:
```suggestion
      character(len=128) :: iomsg, msgout
```

jvdp1 (approved):
Thank you @perazz . Pending minor comments, LGTM.
Note: I thought that `write(*,'(*(i0,1x)')` would not result in a line with a space at its end. Is this compiler-specific?

src/stdlib_io.fypp
perazz: Yes! I used it for consistency. I agree that `error stop` is a better solution. Here [15de15d](https://github.com/fortran-lang/stdlib/pull/877/commits/15de15d951aa988229dcf15e4b7c767d0ab74b75) is the update.

perazz:
> Is this compiler-specific?

I'm never sure about these string formats, so I went test it [here](https://godbolt.org/z/vWTq1rYTo). It seems like all most used compilers do add a space at the end. After this PR we will have an error message, so if those tests fail again, we will hopefully know why (I've never had any errors testing on my machine locally).

PR #879, opened 2024-10-02
Fix warnings [-Wunused-xxx] from compilation

cyrilgandon:
Cleaning duty, warnings are generates with

`fpm build --verbose  --flag '-Wunused-variable -Wunused-dummy-argument -Wunused-parameter'
`
> ### Progress
> * [x]  42 unused-variable
> * [x]  15 unused-parameter
> * [x]  12/67 unused-dummy-argument
> * [ ]  55/67 unused-dummy-argument

Leaving 55 unused-dummy-argument warnings unresolvable, see conversation about those.

### DIFF
--- a/src/stdlib_linalg_least_squares.fypp
+++ b/src/stdlib_linalg_least_squares.fypp
@@ -203,12 +203,18 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
          !! Local variables
          type(linalg_state_type) :: err0
          integer(ilp) :: m,n,lda,ldb,nrhs,ldx,nrhsx,info,mnmin,mnmax,arank,lrwork,liwork,lcwork
-         integer(ilp) :: nrs,nis,ncs,nsvd
+         integer(ilp) :: nrs,nis,nsvd
+         #:if rt.startswith('complex')         
+         integer(ilp) :: ncs
+         #:endif
          integer(ilp), pointer :: iwork(:)
          logical(lk) :: copy_a,large_enough_x
          real(${rk}$) :: acond,rcond
          real(${rk}$), pointer :: rwork(:),singular(:)
-         ${rt}$, pointer :: xmat(:,:),amat(:,:),cwork(:)
+         ${rt}$, pointer :: xmat(:,:),amat(:,:)
+         #:if rt.startswith('complex')         
+         ${rt}$, pointer :: cwork(:)
+         #:endif  
 
          ! Problem sizes
          m     = size(a,1,kind=ilp)
--- a/example/hashmaps/example_hashmaps_get_other_data.f90
+++ b/example/hashmaps/example_hashmaps_get_other_data.f90
@@ -1,6 +1,6 @@
 program example_get_other_data
   use stdlib_kinds, only: int8, int64
-  use stdlib_hashmaps, only: chaining_hashmap_type, int_index
+  use stdlib_hashmaps, only: chaining_hashmap_type
   use stdlib_hashmap_wrappers, only: fnv_1_hasher, key_type, other_type, set, get
   implicit none
   logical                     :: conflict
--- a/example/hashmaps/example_hashmaps_remove.f90
+++ b/example/hashmaps/example_hashmaps_remove.f90
@@ -1,6 +1,6 @@
 program example_remove
   use stdlib_kinds, only: int8, int64
-  use stdlib_hashmaps, only: open_hashmap_type, int_index
+  use stdlib_hashmaps, only: open_hashmap_type
   use stdlib_hashmap_wrappers, only: fnv_1_hasher, &
                                      fnv_1a_hasher, key_type, other_type, set
   implicit none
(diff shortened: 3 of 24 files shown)


perazz (approved):
LGTM, thank you @cyrilgandon.

awvwgk (approved):
Thanks for sharing

src/stdlib_linalg_least_squares.fypp
```diff
@@ -203,12 +203,18 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+         #:if rt.startswith('complex')         
+         integer(ilp) :: ncs
+         #:endif
          integer(ilp), pointer :: iwork(:)
          logical(lk) :: copy_a,large_enough_x
          real(${rk}$) :: acond,rcond
          real(${rk}$), pointer :: rwork(:),singular(:)
-         ${rt}$, pointer :: xmat(:,:),amat(:,:),cwork(:)
+         ${rt}$, pointer :: xmat(:,:),amat(:,:)
+         #:if rt.startswith('complex')
```
jvdp1, line 215: Is this procedure `startwith` supported by all versions of Python?

jvdp1 (commented):
Thank you @cyrilgandon for this PR. I have only one minor comment regarding Python procedures. Recently, a procedure of Python >3.9 (?) was introduced in the Fortran `stdlib`, limiting its compilation on systems supporting most recent versions of Python. It would be nice if this issue can be avoided, hence my question below.

src/stdlib_linalg_least_squares.fypp
perazz: @jvdp1 I think @cyrilgandon is just following the same style that I've used through the BLAS/LAPACK routines for those cases that have differences between `real` and `complex` versions. As far as I can tell, string's `startwith` was supported in Python 2 already. Another way to accomplish the same result would be to use as used elsewhere:

https://github.com/fortran-lang/stdlib/blob/de9b1f8f6cb5482c897f93e06062243b4c635e01/src/stdlib_io.fypp#L138

jvdp1: Thank you for the explanation @perazz . As you can see, my knowledge in Python is quite limited ;) If this approach (using `startwith`) is already used and works, I am fine with it!

jvdp1 (approved):
Based on @perazz 's answer, this PR can be merged IMO. Thank you @cyrilgandon for cleaning the code.

cyrilgandon:
I still have some fix to do on unused-dummy-argument. This is a bit more complicated because I have to really understand what the code does for some cases

src/stdlib_linalg_lapack_aux.fypp
```diff
@@ -321,7 +321,7 @@ module stdlib_linalg_lapack_aux
      end function stdlib_ilauplo
 
 
-     pure integer(ilp) function stdlib_iparmq( ispec, name, opts, n, ilo, ihi, lwork )
```
perazz, line 324: @cyrilgandon please be careful changing the BLAS/LAPACK backend API. I would believe we don't want it modified, so it is always compatible with the library's standard interfaces

jvdp1: I agree with @perazz : API of BLAS/LAPACK should not be modified (and I would extend that to all procedures of `stdlib`). Such API changes should have their own PR.

src/stdlib_str2num.fypp
```diff
@@ -95,6 +95,7 @@ module stdlib_str2num
         integer(int8) :: p !! position within the number
         integer(int8)  :: stat !! error status
         !----------------------------------------------
+        if (.false.) v = mold ! removes unused-dummy-argument warning
```
jvdp1, line 98: What does the compiler with this line? Does it remove it? What about the efficiency of adding this line of the whole procedure?

Romendakil: I would refrain from such changes. In OO programming,  variables defined in abstract routines of the base type can easily be unused in some inherited types. Adding some lines just to suppress the warning is artificial and error-prone. I would rather advocate an option in fpm to build without the argument -Wunused-XXX

src/stdlib_linalg_lapack_aux.fypp
cyrilgandon: Oh yeah, ,thank you for pointing that, for a minute I believe iparmq was an internal function, my bad.
BLAS and LAPACK generates most of the unused-dummy warnings, do you have a preferred way to deal with those?

perazz: There is not much we can do about it as far as I can tell. There may be ways some those warnings could be suppressed (i.e., force their initialization, at least if they are `intent(out)/intent(inout)`, but maybe a better strategy would be to flag the BLAS/LAPACK backend sources with `-Wnounused-...` as pointed out by @Romendakil

src/stdlib_str2num.fypp
jalvesz: Fully second leaving the unused dummy arguments as they are. In this simple case the compiler will remove the line, but it is indeed error prone and artificial. Unless one is 100% sure that the procedures signature can be safely changed, then the best solution in this case is to don't do anything.

perazz:
It seems like all comments have been addressed. Let's wait one or two days for further comments, and then merge.

PR #880, opened 2024-10-04
Fix missing argument to linalg_invert_inplace_ call in linalg_invert_split

cyrilgandon:
During the fix of `unused-dummy-argument` in https://github.com/fortran-lang/stdlib/pull/879, I see this case where the pivot argument is not used, but it seems it should be definitely be passed to the linalg_invert_inplace function.

I made a different PR since it is no longer a refactoring but a real fix.


perazz (approved):

jalvesz:
Thanks for this fix @cyrilgandon! LGTM

awvwgk (approved):
Good catch

jvdp1:
Thank you @cyrilgandon. With 3 approvals, I will merge this PR.

PR #882, opened 2024-10-19
fix fypp_deployment build and set maxrank default to 4 as in cmake

jalvesz:
Small fix for the build option in the fypp_deployment script and change default maxrank. The previous command line construc worked in Linux but was not working on windows.


jvdp1 (approved):
Thank you @jalvesz

perazz (approved):
I also see no issues with it. Thanks @jalvesz

PR #886, opened 2024-11-10
promote ascii functions to elemental

jalvesz:
Promote ascii functions from pure to elemental for potentially better optimization and ease use for array expressions manipulation.

### DIFF
--- a/doc/specs/stdlib_ascii.md
+++ b/doc/specs/stdlib_ascii.md
@@ -38,7 +38,7 @@ Converts input character variable to all lowercase.
 
 #### Class
 
-Pure function.
+Elemental function.
 
 #### Argument
 
@@ -70,7 +70,7 @@ Converts input character variable to all uppercase.
 
 #### Class
 
-Pure function.
+Elemental function.
 
 #### Argument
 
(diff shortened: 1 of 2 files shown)


perazz (approved):
I believe I remember that some older versions of gfortran had issues with `elemental` functions and `character `. However, that seems not an issue here as all tests pass with all supported gfortran versions. LGTM, thank you @jalvesz

jalvesz:
Do you remember if it was a compile-time problem or a run-time problem? maybe it would be better to add array-like tests also ?

jvdp1 (approved):
Thank you @perazz . I guess that the specs should be updated too.
Re: the history, I don't remember any specific reasons why they were not defined as `elemental`. These procedures were some of the first to be implemented in `stdlib` and the structure was not well defined yet (see e.g., #32 ).

jalvesz:
On my side I think it could be merged. The fails in the macOS CI are due to an update in github, which can be fixed in another PR.

```
The macOS-12 environment is deprecated, consider switching to macOS-13, macOS-14 (macos-latest) or macOS-15. For more details, see https://github.com/actions/runner-images/issues/10721
```

perazz:
Merging now.

PR #887, opened 2024-11-12
adds cursor control ANSI escape codes

### LINKED ISSUE #229: ANSI Colors support

wassup05:
The PR adds cursor control ANSI escape codes as a separate module `stdlib_ansi_cursor`

This was brought up scarcely in #229 but was not really discussed.

## Functions added are
1. `move_to(x,y)`
2. `move_to_column(x)`
3. `move_<direction>` where `<direction>` is any one of up, down, left, right
4. A couple of constants for clearning the terminal, clearning a line and it's different variants

## Motivation
There is already a module `stdlib_ansi` for ANSI escape codes of foreground, background colors and the different styles... If a person finds the use for such module it is very probable that he will also at some point want to control the cursor position too, to let's say, implement a custom loading animation or for elegant management of terminal output

### Side note
I chose to make it a different module because of the difference in approach between the two, the functions in this module talk in terms of strings whereas `stdlib_ansi` has a whole type surrounding it and also because you won't need to add the cursor_ to all the functions to make it more readable... But I don't really see a problem in merging the two too...

Edit: This is my first attempt at contributing here and I am not sure about who I am supposed to ask for feedback.. Could you guide me regarding this @perazz @jalvesz (I figured as you were the most active right now, hence pinging you would be appropriate)


src/stdlib_ansi_cursor.f90
```diff
@@ -0,0 +1,107 @@
...
+    !> erases from the cursor till the end of line
+    character(len=*), parameter :: clear_till_line_end = esc//"[0K"
+    !> erases from the cursor till the beginning of the line
+    character(len=*), parameter :: clear_till_line_start = esc//"[1K"
+    !> erases the entire line
+    character(len=*), parameter :: clear_entire_line = esc//"[2K"
+
+contains
+    !> moves the cursor to `(line, column)`
+    !> returns an empty string if any of them is negative
```
jvdp1, line 23: For each procedure, a `version` and a link to the specs should be provided.

src/stdlib_ansi_cursor.f90
```diff
@@ -0,0 +1,107 @@
...
+    pure function move_to(line, col) result(str)
```
jvdp1, line 24: Could some of these functions be `elemental`?

test/terminal/test_ansi_cursor.f90
```diff
@@ -0,0 +1,90 @@
...
+        type(error_type), allocatable, intent(out) :: error
+        character(len=:), allocatable :: str
+
+        str = move_to(-10, 20)
+        call check(error, str, "")
+        if (allocated(error)) return
+        str = move_to(10, 20)
+        call check(error, iachar(str(1:1)), 27)
+        if (allocated(error)) return
+        call check(error, str(2:), "[10;20H")
```
jvdp1, line 31: Could you provide clear error messages, please?

jvdp1 (commented):
Thank you @wassup05 for this PR. It sounds like a good start. I left a few comments.

src/stdlib_ansi_cursor.f90
```diff
@@ -0,0 +1,107 @@
+module stdlib_ansi_cursor
```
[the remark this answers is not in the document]
wassup05: Thank you for reviewing!

It could be, but I thought separating it would be better since `stdlib_ansi` revolves around it's own derived type whereas this doesn't... I'm open to either way though

src/stdlib_ansi_cursor.f90
```diff
@@ -0,0 +1,107 @@
...
+    use stdlib_strings, only: to_string
+    implicit none
+
+    character(len=*), parameter :: esc = achar(27)
+    !> moves the cursor to home => `(0,0)`
+    character(len=*), parameter :: home = esc//"[H"
```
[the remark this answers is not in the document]
wassup05: Yes, It's more of a terminal thing than an OS thing.. But almost all of the modern terminals support these ansi codes

src/stdlib_ansi_cursor.f90
wassup05: They could be made so, But I don't think it's worth it since I don't see anyone using it like that

test/terminal/test_ansi_cursor.f90
```diff
@@ -0,0 +1,120 @@
...
+        str = move_up(-15)
+        call check(error, str, "")
+        if (allocated(error)) then
+            print *, "ERROR: move_up fails with negative values"
+            return
+        end if
+        str = move_up(15)
+        call check(error, iachar(str(1:1)), 27)
+        if (allocated(error)) then
+            print *, "ERROR: move_up doesn't add ESC character at the beggining"
```
perazz, line 58:
```suggestion
            print *, "ERROR: move_up doesn't add ESC character at the beginning"
```

perazz:
Nice work @wassup05, thank you. I'm no expert on ANSI terminal handling so I will add some of my thoughts here. 
- The `stdlib_ansi` module has a derived type for ANSI escape codes: 

https://github.com/fortran-lang/stdlib/blob/7511064e45476cc2deaa5af195fdb70cb73214d1/src/stdlib_ansi.f90#L23-L31

If the terminal handling codes provided here match the same pattern, I believe it would be better to uniform their handling with that of the module, and make them `type(ansi_code)` as well.
- Testing and examples are hard with console handling. However, I think it would be very useful to have examples that show some console interaction with these functions (for example, draw a box on the terminal, or something like that).

wassup05:
Thanks for replying, @perazz !

When I started working on this, I thought of ways to integrate the two but couldn't really find any such way... As these are more action oriented whereas those of ansi_type were more of like changing the properties of the text being printed

Regarding examples... I added an example in the recent commit which basically draws a blue box on the screen using a subroutine `draw_box`, whose width, height and the position (line, column) were it is drawn can be changed.. I added the link to it in the spec as well

Please let me know what you think about it

PR #888, opened 2024-11-14
linalg: introduce 64-bit integer size BLAS/LAPACK implementation (`ilp64`)

perazz:
I'm introducing 64-bit integer support throughout the BLAS/LAPACK implementation to address #881 and introduce support for the 64-bit MKL library. 

With this update: 
- Users will always seamlessly use either 32- or 64-bit `integer` sizes
- Can link against 64-bit external libraries such as MKL
- If linking against a 64-bit external library, the internal 32-bit version is provided, and vice versa 

## Key facts 
- [x] Introduce `ilp64` integer kind 
- [x] Template all procedures in our BLAS/LAPACK implementation to support either `ilp`, `ilp64`, or both
- [x] Extend all generic BLAS/LAPACK interfaces to both integer types
- [x] Allow linking to external `ilp64` libraries (such as MKL) via `STDLIB_EXTERNAL_BLAS_I64` and `STDLIB_EXTERNAL_LAPACK_I64`

The code shows large differences because all interfaces are now templated. These are the key lines: 
https://github.com/fortran-lang/stdlib/blob/4b5a6e62c49221efa02403d9cd645023888d4931/include/common.fypp#L77-L81

- `ilp` and `ilp64` are now the two supported integer kinds
- 64-bit subroutine names start with `stdlib_I64_*`, while 32-bit versions are still `stdlib_*`

If we support this implementation, next I will start templating the high-level APIs. 


include/common.fypp
```diff
@@ -74,6 +74,12 @@
 $:"s" if cmplx=="c" else "d" if cmplx=="z" else "x" if cmplx=="y" else "q" if cmplx=="w" else "ERROR"
 #:enddef
 
+#! BLAS/LAPACK/Linear Algebra Integer Kinds
+#:set LINALG_INT_KINDS       = ["ilp","ilp64"]
```
jalvesz, line 78: Using fypp it would be possible then to initialize the 3 lists with only the 32bit support and then add a conditional something like
```
#:set LINALG_INT_KINDS       = ["ilp"]
#:set LINALG_INT_SUFFIX      = [""]
#:if WITH_LINALG_ILP64
#:set LINALG_INT_KINDS       = LINALG_INT_KINDS+["ilp64"]
#:set LINALG_INT_SUFFIX      = LINALG_INT_SUFFIX+["_I64"]
#:endif
#:set LINALG_INT_TYPES       = ["integer({})".format(k) for k in LINALG_INT_KINDS]
#:set LINALG_INT_KINDS_TYPES = list(zip(LINALG_INT_KINDS, LINALG_INT_TYPES, LINALG_INT_SUFFIX))
```
Like this, it would be easy to parametrize which version to preprocess in case one doesn't require the ilp64 version. This can be added then as a parameter in the fypp_deployment.py script:
```python
parser.add_argument("--with_linalg_ilp64",action='store_true', help="Include the WITH_LINALG_ILP64 preprocessing keyword for BLAS/LAPACK")
```

jalvesz (changes requested):

include/common.fypp
perazz: Amazing, thanks @jalvesz. See the deployment config edits in [34b7a81](https://github.com/fortran-lang/stdlib/pull/888/commits/34b7a81ad657789f6beaeff8af657dd181eae7b1).
I've used `WITH_ILP64` instead of `WITH_LINALG_ILP64` because it's shorter and seems closer to the other flags `WITH_QP` etc. What do you think? Also `WITH_LINALG_ILP64` is OK imho.

jalvesz: I like also the shorter keyword! Perfect!

I'm not sure regarding the default, I'm guessing it should be important to weight: binary size, compilation time and users demand for ilp64 support out of the box. No dogma on my side.

awvwgk:
> If linking against a 64-bit external library, the internal 32-bit version is provided, and vice versa

Not sure whether I got this point correctly. If we link against the ilp64 version the stdlib built-in blas version is provided or automatically converted from ilp to ilp64? Either choice might bring surprises to the user in the former case performance depends on the choice of integer kind and in the other case conversation might require additional memory.

perazz:
> If we link against the ilp64 version the stdlib built-in blas version is provided or automatically converted from ilp to ilp64? 

The idea is that if stdlib is built `WITH_ILP64`, both 32- and 64-bit versions are provided. If the user has an external 64-bit library, the 32-bit procedures are provided by the internal implementation; and vice versa. As far as I can tell, it is not possible to have both 32/64 external libraries, due to name conflict (e.g. two external subroutines with different interfaces but both named `dgemm` )

include/common.fypp
perazz: Default `ILP64` build is currently `False`.

jalvesz (changes requested):
LGTM @perazz! would it be possible to add a comment in the README regarding the use of the macro in order to parametrize the build process?

perazz:
@jalvesz and all: the build script allows choosing what versions to include which is great. For the pre-processed fpm versions, What do you think about creating a new `stdlib-fpm-ilp64` branch that includes both 32- and 64-bit versions by default?

perazz:
- [x] added parameter flags in `stdlib_linalg_constants` the user can check whether an external library is being used
- [x] linear algebra options documented in README 
- [x] `stdlib-fpm-ilp64` branch added, receives deployment of the `ilp64`-enabled version. (thanks @jalvesz, was very easy to adapt the deployment script) 

PS: I believe the actual deployment with commit to the fpm branches is only triggered after the PR is merged

jalvesz (approved):

perazz:
Thank you @jalvesz and all for the reviews. To expedite the introduction of #908, I invite anyone that has further comments on this PR to please do so.

jvdp1 (approved):
LGTM. Thank you @perazz

config/fypp_deployment.py
```diff
@@ -140,8 +148,9 @@ def fpm_build(args,unknown):
     parser.add_argument("--maxrank",type=int, default=4, help="Set the maximum allowed rank for arrays")
     parser.add_argument("--with_qp",action='store_true', help="Include WITH_QP in the command")
     parser.add_argument("--with_xdp",action='store_true', help="Include WITH_XDP in the command")
+    parser.add_argument("--with_ilp64",action='store_true', help="Include WITH_ILP64 to build 64-bit BLAS/LAPACK")
```
ivan-pi, line 151:
```suggestion
    parser.add_argument("--with_ilp64",action='store_true', help="Include WITH_ILP64 to build 64-bit integer BLAS/LAPACK")
```

jalvesz:
> Thank you @jalvesz and all for the reviews. To expedite the introduction of #908, I invite anyone that has further comments on this PR to please do so.

LGTM @perazz! As soon as you merge this, I'll merge with my branch to open a PR on the build speed issue.

perazz:
Thank you all for the reviews! Let us wait another couple of days, and then merge if there are no further comments.

jalvesz:
Thanks @perazz for this nice PR, I'll merge it in order to take over with #908

PR #889, opened 2024-11-14
Fix Gaussian distribution formula in documentation

moriglia:
There was a missing $\pi$ in the normalization term in the documentation of the pdf_normal function.


perazz (approved):
Good catch, thank you @moriglia

moriglia:
Thank you for your effort in maintaining the stdlib @perazz

moriglia:
Sure, no rush, the documentation is already clear, regardless of the 
missing yummy $\pi$.

Issue #890, opened 2024-11-16
Compilation error in stdlib_io.f90


MilanSkocic:
### Description

I've got an error when compiling the stdlib in the io module either with gfortran 10 or gfortran 14 (std=f2008).

The error says:
```
  153 |            error stop trim(msgout)
      |                           1
Error: STOP code at (1) must be a scalar default CHARACTER or INTEGER constant expression
```
I think that the compilation error might be related to the change in the commit [6439a2c](https://github.com/fortran-lang/stdlib/commit/6439a2cc2c6ed825e8d71071502c9aa9400a99c4) when `call error_strop` was replaced with `error stop`.

Does anyone have an idea how fix it? 

### Expected Behaviour

No error.

### Version of stdlib

stdlib=* (in fpm.toml)

### Platform and Architecture

Windows, Linux

perazz:
Thank you for reporting @MilanSkocic, I did not realize that some compiler options would cause issues with `error stop`

MilanSkocic:
@perazz I just wanna let you know that I also tested with gfortran std=f2018 and it failed too. The issue might be only related to gfortran but I haven't tested other compilers.

PR #894, opened 2024-11-21
Revert "`call error_stop` -> `error stop`"

perazz:
Address #890. 

Some compilers, if run with `-std=f2008`, do not accept `error stop + character` .

@jvdp1 @MilanSkocic


jvdp1 (approved):
LGTM. Thank you @perazz

perazz:
This is a simple PR, I would say let's wait one or two days, and then merge if OK.

Issue #901, opened 2024-11-30
linalg: generalize return type for `eye`


perazz:
### Motivation

It was [recently suggested](https://fortran-lang.discourse.group/t/fortran-stdlib-sufficient-for-scientific-programming/8905/3) that the `eye` (#481) function is misleading. I would like to suggest that the `eye` function is extended by allowing a general return type. This could be accomplished by introducing an optional `mold` keyword: 

```
A = eye(m, n [, mold])
```

that would tell the `eye` interface which return type should have `A`:

```
pure function eye_${rk}$(m, n, mold)
   integer, intent(in) :: m
   integer, intent(in), optional :: n
   ${rt}$, optional, intent(in) :: mold
   ${rt}$, dimension(:,:), allocatable :: eye
end function eye_${rk}$
```

### Prior Art

`mold` is a Fortran Standard keyword that specifies that the return type of a polymorphic allocation should be taken from another variable: 
```
allocate(a, mold=b)
```

Similarly, here we may want to specify that the return type of the `eye` matrix should be taken from the `mold` variable.

PR #902, opened 2024-12-02
linalg `eye`: allow generalized return type and kind

perazz:
It was recently suggested (#901) that it may be confusing that `eye` always returns an `integer(int8)` result. 
This PR proposes to extend the current identity matrix implementation via an optional `mold` keyword. 
If present, `mold` is used to give a type and kind to the return matrix.

### Proposed interface

- `A = eye(dim1 [, dim2] [, mold])`

### Key facts
- Default behavior is unchanged (implementation is backward compatible)
- [Fortran Discourse](https://fortran-lang.discourse.group/t/fortran-stdlib-sufficient-for-scientific-programming/8905/3) thread 

### Prior art
`mold` already exists, and with the same meaning, both in the Fortran standard (`allocate(a, mold=b)`), and here in stdlib:
- [codata](https://stdlib.fortran-lang.org/page/specs/stdlib_constants.html#codata) constants
- [string to number](https://stdlib.fortran-lang.org/page/specs/stdlib_str2num.html#to-num-conversion-of-strings-to-numbers) conversion


doc/specs/stdlib_linalg.md
```diff
@@ -239,33 +239,34 @@ Pure function.
...
 ### Arguments
 
-`dim1`: Shall be a scalar of default type `integer`.
-This is an `intent(in)` argument. 
-
-`dim2`: Shall be a scalar of default type `integer`.
-This is an `intent(in)` and `optional` argument. 
+- `dim1`: A scalar of type `integer`. This is an `intent(in)` argument and specifies the number of rows.
+- `dim2`: A scalar of type `integer`. This is an optional `intent(in)` argument specifying the number of columns. If not provided, the matrix is square (`dim1 = dim2`).
+- `mold`: A scalar of any supported `integer`, `real`, or `complex` type. This is an optional `intent(in)` argument. If provided, the returned identity matrix will have the same type and kind as `mold`. If not provided, the matrix will be of type `integer(int8)` by default.
```
jvdp1, line 252: Based on the Fortran Discourse thread, is it still wise to return `integer(int8)` by default (instead of `integer`)? (the API is still experimental and so could still be changed).

jvdp1 (approved):
Thank you @perazz . This is a nice extension. I just wonder if we should keep the default output as `integer(int8)` or as `integer`?

doc/specs/stdlib_linalg.md
perazz: Thanks a lot @jvdp1 for the review. I left default behavior unchanged to not introduce API changes. However, my personal preference would be that the default return type should be `real(real32)` or `real(real64)`. The reason is that we've all made this Fortran mistake of integer division far too many times :) So I think that in the interest of safety for the least experienced Fortran programmers, it would be safe(r) to use a `real` return value.

doc/specs/stdlib_linalg.md
```diff
@@ -239,33 +239,34 @@ Pure function.
...
+- By default, the return value is of type `integer(int8)`, which is recommended for storage efficiency.
+- If the `mold` argument is provided, the return value will match the type and kind of `mold`, allowing for arbitrary `integer`, `real`, or `complex` return types.
 
 #### Warning
 
-Since the result of `eye` is of `integer(int8)` type, one should be careful about using it in arithmetic expressions. For example:
+When using the default `integer(int8)` type, be cautious when performing arithmetic operations, as integer division may occur. For example:
+
 ```fortran
-!> Be careful
```
perazz, line 265: Integer division warning is mentioned in the function description here @jvdp1, so maybe it would be better design to avoid it? Memory is cheap nowadays, it should not be a huge issue imho.

doc/specs/stdlib_linalg.md
jvdp1: I agree with you: I also think it would be better to return a `real` identity matrix. The `mold` argument will let the user to return another type and kind if desired.

doc/specs/stdlib_linalg.md
perazz: [57e3f89](https://github.com/fortran-lang/stdlib/pull/902/commits/57e3f89d367174c908b78bb8dd744e9fc336de12)

perazz:
@jvdp1 let me know if you have further comments since I switched the default return to `real(real64)`. Otherwise, I believe this can be merged soon.

jalvesz:
Thanks @perazz for this improvement! I think it is indeed almost ready. One suggestion, feel free to neglect it:

Quite often it is required to initialize arrays with zero or one, why not add these parameters to one of the constants modules? taking for instance the one added in the stdlib_sparse_constants:
```
#:for k1, t1, s1 in (R_KINDS_TYPES)
    ${t1}$, parameter :: zero_${s1}$ = 0._${k1}$
    ${t1}$, parameter :: one_${s1}$ = 1._${k1}$
#:endfor
#:for k1, t1, s1 in (C_KINDS_TYPES)
    ${t1}$, parameter :: zero_${s1}$ = (0._${k1}$,0._${k1}$)
    ${t1}$, parameter :: one_${s1}$ = (1._${k1}$,1._${k1}$)
#:endfor
```
Make them accesible elsewhere, in that case, I would propose the following small modification to the `eye` template:
```
...
allocate(result(dim1, dim2_), source = zero_${s1}$)
        
do i = 1, min(dim1, dim2_)
      result(i, i) = one_${s1}$
end do
...
```

perazz:
Thank you @jalvesz, with two approvals, I will merge this one. 
Regarding your suggestion: I think it's a good idea and we should think of a good way to do it (perhaps, inside the `stdlib_linalg_constants` module?)

jalvesz:
@jvdp1 I think that if consistently used, it would avoid the small kind conversion errors under assignment.

Agree, let's postpone that for another PR.

jvdp1:
> @jvdp1 I think that if consistently used, it would avoid the small kind conversion errors under assignment.

OK. I see your point and fully agree with it!

perazz:
@jvdp1 @jalvesz note that we already have [many such constants](https://github.com/fortran-lang/stdlib/blob/35e7146bf77d63c33f633273ba315bcf9522e07e/src/stdlib_linalg_blas_d.fypp#L50-L77) (including precision-related) in the BLAS modules. 
- For the user API, maybe all we need is a consistent way to export them? One way this was done in the `codata` PR was to use a `mold` parameter and functions
- For development, maybe the best approach would be to template them directly in `fypp`? this works well:

```
#! Return the unitary numeric constant for a given type and kind
#!
#! Args:
#!   type (string): an intrinsic type (complex, real, integer)
#!
#! Returns:
#!   Parameter value string
#!
#! E.g.,
#!   one('complex(dp)')  -> (1.0_dp, 0.0_dp)
#!   one('real(sp)') -> 1.0_sp
#!   one('integer') -> 1
#!
#:def one(type)
  #:assert 'integer' in type or 'real' in type or 'complex' in type
  #:if '(' in type and ')' in type
    #:set kind="_" + type.split('(')[1].split(')')[0]
  #:else
    #:set kind=""
  #:endif
  #:if 'integer' in type
    ${"1"+kind}$
  #:elif 'real' in type
    ${"1.0"+kind}$
  #:elif 'complex' in type
    ${"(1.0"+kind+",0.0"+kind+')'}$
  #:endif
#:enddef
```

perazz:
I will merge this PR now, and open an "idea" issue to discuss and start the constants implementation, thank you @jvdp1 @jalvesz for the reviews.

Issue #782, opened 2024-04-02
Add `library` configuration to `stdlib-fpm`


perazz:
### Motivation

Users may want to build stdlib using `fpm`, and then use it as a static library in other projects. 
To do so, I suggest to add a `[library]` configuration in the `fpm.toml` manifest. 

### Prior Art

See also: https://github.com/fortran-lang/stdlib/issues/640
https://github.com/fortran-lang/stdlib/pull/781

PR #903, opened 2024-12-09
Add [install] section to fpm.toml

moriglia:
This commit allows for installation of the `stdlib`, instructing `fpm` to install (if the prefix is not modified):
-  `libstdlib.a` file --> `~/.local/lib/
- `.[s]mod` files --> `~/.local/include/`

Example:
`fpm install --compiler=gfortran-11 --profile release`

Resolves fortran-lang/stdlib#782

### DIFF
--- a/fpm.toml
+++ b/fpm.toml
@@ -5,6 +5,11 @@ author = "stdlib contributors"
 maintainer = "@fortran-lang/stdlib"
 copyright = "2019-2021 stdlib contributors"
 
+
+[install]
+library = true
+
+
 [dev-dependencies]
 test-drive.git = "https://github.com/fortran-lang/test-drive"
 test-drive.tag = "v0.4.0"


perazz (approved):
Thank you @moriglia, LGTM. Definitely a good idea, since this is a library. I have no idea why that was not there in the first place.

fpm.toml
```diff
@@ -5,6 +5,11 @@ author = "stdlib contributors"
 maintainer = "@fortran-lang/stdlib"
 copyright = "2019-2021 stdlib contributors"
 
+
+[install]
+library = true
+
+
```
jvdp1, line 12:
```suggestion
[install]
library = true

```

jvdp1 (commented):
Thank you @moriglia . It makes sense to me. It might be good to mention this new behaviour in the `README`.

jvdp1:
cc: @jalvesz

jalvesz:
LGTM @moriglia thanks for this contribution!

moriglia:
Thank you all for your positive feedback!

> It might be good to mention this new behaviour in the `README`.

Sure, I think I should open a separate PR though, since the `README` is present only on the `master` branch. Unless you want it to appear also on the `stdlib-fpm` branch. In which branch do you think the `README` edit would be more appropriate?

jalvesz:
The `stdlib-fpm` branch is a subproduct of the automatic github CI. All contributions should be made through the master branch. You can add the info either around here https://github.com/fortran-lang/stdlib/blob/master/README.md#build-with-fortran-langfpm or here https://github.com/fortran-lang/stdlib/blob/master/README.md#using-stdlib-with-fpm. Also, all `fpm` actions are possible with the master branch following the fypp preprocessing as discribed in the first link. So if one uses the master branch, installation through fpm is possible.

perazz:
@jalvesz is right, thank you for noticing it. These changes should be moved to the `main` branch fpm.toml

PR #904, opened 2024-12-09
filesystem: error handling, `delete_file`, `is_directory` and other

perazz:
this PR follows up (without overlapping) on the excellent work ongoing at #874. 
Also CC: #14 and https://github.com/j3-fortran/fortran_proposals/issues/100
An initial `stdlib_io_filesystem` functionality is proposed: because such minimal functionality is intertwined, this PR proposes to introduce the following: 

- extend `state_type` error handling, that was well accepted for linear algebra, to the whole library (without API changes: just make `state_type->linalg_state_type` polymorphism)
- Runtime OS query (saved to static storage for performance)
- Basic file I/O and sub-processing functions: `delete_file`, `is_directory`, `run`, `getfile`

## `stdlib_io_filesystem`
- `is_directory(path)`: `logical` function
- `call delete_file(path [, err])`: optional return `state`, otherwise exceptions trigger an `error stop` 

## `stdlib_system`
- `call run(cmd [, exit_state] [, command_state] [, stdout] [, stderr])`: with optional stdout/stderr outputs loaded onto `string_type` variables
- function `os_type()` returns a cached OS type flag, `OS_NAME` its name
- function `null_device()` returns a cross-platform string path to suppress terminal output

## `stdlib_io`
- `type(string_type) function getfile(fileName [,err] [,delete])` loads a whole ASCII file into a `string_type` variable all at once, with option to return a state flag, or to request file deletion (`delete=.true.`) after loading. This function is necessary within this PR for loading screen output. 

### Comments

- Usage of external C code has been avoided, but I'm happy to change implementation details if requested. In `fpm`, there is a policy that the package must be fully compiled as a single-source Fortran file, so C usage is not well integrated, but I don't know what the stdlib policy is in this respect.
- should `delete` functionality be unique for either files or directories? otherwise, I will work out a `rmdir` or `delete_dir` alternative for them.

cc: @fortran-lang/stdlib @jvdp1 @jalvesz @zoziha @minhqdao


perazz:
The contents of this PR have now be fully integrated as split PRs #938, #939, #942, #945, #946, #966, so I will close this one.

PR #906, opened 2024-12-10
Add `[install]` section to `fpm.toml`

moriglia:
Moved PR from `stdlib-fpm` to `master` branch.

Initial discussion here: https://github.com/fortran-lang/stdlib/pull/903


moriglia:
Another thing I noticed is that `fpm` will use the project name in the library archive, i.e. `stdlib`.
In contrast, in [this section](https://github.com/moriglia/stdlib?tab=readme-ov-file#using-stdlib-with-a-regular-makefile) of the `README`, the suggested flag and name is `fortran_stdlib`.

Should I also change the project name to `fortran_stdlib` to have `fpm` produce `libfortran_stdlib.a` instead of `libstdlib.a`?

fpm.toml
```diff
@@ -5,6 +5,9 @@ author = "stdlib contributors"
 maintainer = "@fortran-lang/stdlib"
 copyright = "2019-2021 stdlib contributors"
```
perazz, line 6:
```suggestion
copyright = "2019-2024 stdlib contributors"
```

perazz (approved):
LGTM, thank you @moriglia. 

Regarding the fpm library name, I believe it should remain `stdlib`, because that is the name that is typically used as a `fpm` dependency (including as a "metapackage").

moriglia:
Thank you! @perazz 
Somehow it felt wrong to link against a `fortran_stdlib` to me, also :)

README.md
```diff
@@ -219,7 +219,29 @@ python config/fypp_deployment.py --help
...
+
+
+#### Installing with fpm
+
+Either option you chose for building the `stdlib`, you can install it with:
+```sh
+fpm install --profile release
+```
+The command above will install the following files:
+- `libstdlib.a` into `~/.local/lib/`
```
jalvesz, line 231: Maybe add : into `~/.local/lib/` (Unix) or `C:\Users\<username>\AppData\Roaming\local` (Windows)

jalvesz (changes requested):

README.md
moriglia: Added a commit below, I was wondering whether also the next line needs to be modified for Windows (although I don't know where windows would store the `.[s]mod` files)

jalvesz: Same pattern, what changes is that longer root, so it would be:
`C:\Users\<username>\AppData\Roaming\local\lib` and `C:\Users\<username>\AppData\Roaming\local\include`

moriglia:
Ok, thanks, got it! I'll fix and amend my latest commit as soon as I'm back to my laptop.

jalvesz (approved):

jalvesz:
I think this PR can be merged. Let's wait a couple more days and if there are no remarks I'll merge it.

PR #912, opened 2024-12-26
sparse linalg: simplify SELLC spmv kernel

jalvesz:
@ivan-pi realized that the initial proposal for the SELLC spmv could be improved by defaulting the 0-valued-existing indexes of the column array to an actual index in the column space of the matrix. This change avoids the need for a conditional within the kernel.

This PR proposes to change the default index to 1.


src/stdlib_sparse_spmv.fypp
```diff
@@ -529,7 +529,7 @@ contains
             integer :: j, k
             do j = 1, n
```
perazz, line 530: would it be suitable to replace the inner loops with an array operations?
```
y(col(:,j)) = y(col(:,j)) + alpha_ * a(:,j) * x(:)
```

src/stdlib_sparse_spmv.fypp
```diff
@@ -580,7 +580,7 @@ contains
...
                 do k = 1, r
```
perazz, line 582: Same here, could the `k` loop be collapsed onto an array operation?

src/stdlib_sparse_spmv.fypp
perazz, line 1: Couple comments that I can't inline: 

- https://github.com/fortran-lang/stdlib/blob/3463c4390ee5fde5d0d29d7710eb3667926e84be/src/stdlib_sparse_spmv.fypp#L579, https://github.com/fortran-lang/stdlib/blob/3463c4390ee5fde5d0d29d7710eb3667926e84be/src/stdlib_sparse_spmv.fypp#L566, https://github.com/fortran-lang/stdlib/blob/3463c4390ee5fde5d0d29d7710eb3667926e84be/src/stdlib_sparse_spmv.fypp#L554, https://github.com/fortran-lang/stdlib/blob/3463c4390ee5fde5d0d29d7710eb3667926e84be/src/stdlib_sparse_spmv.fypp#L541, https://github.com/fortran-lang/stdlib/blob/3463c4390ee5fde5d0d29d7710eb3667926e84be/src/stdlib_sparse_spmv.fypp#L528, https://github.com/fortran-lang/stdlib/blob/3463c4390ee5fde5d0d29d7710eb3667926e84be/src/stdlib_sparse_spmv.fypp#L516

Could we employ assumed shape `(:)` in these cases? I believe the performance penalty is negligible (especially with array operations involved), compilers may still optimize these calls, and that stdlib encourages safer coding standards.

perazz (approved):
Nice improvement, thank you @jalvesz. LGTM.

src/stdlib_sparse_spmv.fypp
jalvesz: Yes, this can be done without issues and without evident penalties. done here https://github.com/fortran-lang/stdlib/pull/912/commits/1d25a61d30d6a333ca7695d1788bb30213b8ef39

src/stdlib_sparse_spmv.fypp
jalvesz: unfortunatelly no, it can happen that `col(:,j)` will request a jump-back to a "previous" index. When such access is on the lhs of the assigment with max optimization it will happen that the second request to a previous index will not happen and the result will be incorrect.
I had alreay seen this before and tested it again with the unit test, indeed if I try to use array expression instead of a plain do-loop the test fails because of that.

perazz (approved):

jvdp1 (approved):
LGTM. Thank you @jalvesz for the PR, and well seen from @ivan-pi

Issue #908, opened 2024-12-12
Improve build speed for stdlib


jalvesz:
Ever since adding BLAS/LAPACK to stdlib the build times have exploded.

I have observed in my local machine (Windows + gfortran + CMake + Ninja with 16 threads) that the build takes **199 seconds**. Using this tool https://ui.perfetto.dev/ as suggested by @ivan-pi it is possible to load the .ninja_log file and observe where are the bottle necks:
![Image](https://github.com/user-attachments/assets/1466a75f-123d-4781-bfdc-7a86d426055d)

In the figure we observe the build chocking up at the lapack per kind files.

I started a major refactoring of stdlib's BLAS/LAPACK here with help from @ivan-pi and @perazz : https://github.com/jalvesz/stdlib/tree/split_lapack using a python script to rewrite the blas/lapack-per-kind files into submodule files per groups.

Doing that I managed to get down the build time to **38 seconds** (x5 faster):

![Image](https://github.com/user-attachments/assets/23c3a0f3-2e8e-4095-a2a8-079fc14023f3)

There are still some spots in which the build stalls but it is quite evident that by splitting into several submodule files enables better parallel build.

Before opening a PR, I would like to gather some broad opinions on this refactoring.

Some observations/open questions:
1. In order to achieve this splitting, the `refactor_blaslapack_subm.py` script contains two dictionaries `blas_groups` and `lapack_groups`. They define the procedures to be included in each submodule file. This splitting started by following the lapack reference manual, then I had to arbitrarily split further. I'm sure that with a proper dependency graph it could be possible to achieve a better grouping. If anyone feels like proposing a mechanism to find such grouping that would be very nice. Here a couple of references I've found interesting about such topic: https://paulcavallaro.com/blog/optimizing-function-placement-with-call-chain-clustering/ & https://medium.com/@avidaneran/using-a-graph-representation-to-analyze-python-dependencies-a57cd681fa09

2. In the draft branch, the per-kind files were moved to a `legacy` folder. When opening the PR I think these should not be there as there is no point in having so many duplicates of the sources. Nonetheless, the splitting script can be improved and it is based on those files. Should these files be saved in a separate repo for reference?

3. The submodules approach is proving very neat to organize and increase build performance. Now, this implies a very large header module file in order to have all interfaces for lapack functions. A better work on (1) could probably lead to finding clearer macro dependencies which could enable splitting also this header module.

[...]

Issue #897, opened 2024-11-26
implicit none(type,external) may not support by some fortran compiler


hgrhgy:
### Description

```log
stdlib_linalg_constants.F90 (5) : error #5136: Illegal syntax for IMPLICIT NONE statement.
**implicit none(type,external)**
```

Some compiler may not support this syntax.
Intel fortran 2017 is used. 

### Expected Behaviour

MAYBE fixed it, or other suggestions to deal with this issue.

### Version of stdlib

v0.7.0

### Platform and Architecture

linux

perazz:
I will mark this as closed, it was addressed by #913.

Issue #540, opened 2021-09-27
Should all implementations be placed in submodules?


ivan-pi:
Quoting @milan:

> General question for everybody, do we have a guide on when to put a function implementation in its own submodule and when to leave it in the module? I see that we have `clip` in the module but the others have their own submodules.

On the long term I think the public modules should contain only the interfaces, and all of the implementation should be placed in sub-modules. This would facilitate compiler vendors providing their own optimized implementations in the (distant?) future. It would also make it easier to swap between debug and release binaries by re-linking. I think the current reasoning was simply when launching a new module with a small function, it was just easier to prepare a single module.

A second reason in favor of separating interface and implementation *might* be to use a stdlib library compiled by another Fortran compiler. In this case you would only need to compile the interface (see https://github.com/fortran-lang/stdlib/pull/530#discussion_r711945701 and the discussion that follows). As an example, Intel MKL ships libraries that can be used with both Intel and gfortran compilers.

_Originally posted by @ivan-pi in https://github.com/fortran-lang/stdlib/issues/539#issuecomment-927824562_

milancurcic:
It makes sense to me, I agree. 

Another advantage to submodules in their own source files is that the implementations that are independent from one another could be compiled in parallel.

ivan-pi:
A few issues semi-related to submodules:
* NAG 7.0 is required to support submodules, see #108
* CMake version 3.16.3 is required to support submodules, see #108
* In #307 I advocate the separation of interface and implementation using submodules as a way to provide non-Fortran implementations of stdlib functionality using existing C or C++ libraries (like Boost)

Besides avoiding compilation cascades during development, another benefit of submodules is the reduction in memory usage during compilation (see https://github.com/fortran-lang/stdlib/pull/283).

[...]

PR #913, opened 2024-12-31
linalg: refactor blas/lapack modules

jalvesz:
Major refactoring of the blas/lapack modules into submodules for improved build speed and potentially better organization of procedures. This PR follows #908 

* The reference files and refactoring script are currently within the `legacy` folder which should be removed before merging.
* Two folders have been added: `blas` and `lapack` to organize all required files within.
* The splitting is managed by the `legacy\refactor_blaslapack_subm.py` script which contains 3 dictionaries: `blas_groups`, `lapack_groups`, `lapack_subgroups`. The first two list the procedures in groups following lapacks documentation with some extra arbitrary splitting in order to improve parallel build. Given that the use of submodules requires a header module to declare the procedures signature, putting all lapack procedures into a single header implies a stall in the build process while building this single header. The dictionary `lapack_subgroups` proposes to split into `modules` which define macro-groups for the `submodules`.


src/lapack/stdlib_lapack_solve_chol.fypp
```diff
@@ -0,0 +1,4320 @@
+#:include "common.fypp" 
+submodule(stdlib_lapack_solve) stdlib_lapack_solve_chol
+  implicit none(type,external)
```
perazz, line 3:
```suggestion
  implicit none
```

src/lapack/stdlib_lapack_solve_ldl.fypp
```diff
@@ -0,0 +1,4865 @@
+#:include "common.fypp" 
+submodule(stdlib_lapack_solve) stdlib_lapack_solve_ldl
+  implicit none(type,external)
```
perazz, line 3:
```suggestion
  implicit none
```

src/lapack/stdlib_lapack_solve_ldl_comp3.fypp
```diff
@@ -0,0 +1,9625 @@
+#:include "common.fypp" 
+submodule(stdlib_lapack_solve) stdlib_lapack_solve_ldl_comp3
+  implicit none(type,external)
```
perazz, line 3:
```suggestion
  implicit none
```

src/lapack/stdlib_lapack_solve_ldl_comp4.fypp
```diff
@@ -0,0 +1,12443 @@
+#:include "common.fypp" 
+submodule(stdlib_lapack_solve) stdlib_lapack_solve_ldl_comp4
+  implicit none(type,external)
```
perazz, line 3:
```suggestion
  implicit none
```

perazz (commented):
Thank you @jalvesz, it is very very good already. I only have a few comments. To summarize: 
- `implicit none(type, external)` -> `implicit none`, to satisfy #897 
- please remove the `legacy/` folder altogether: it would cause confusion to have some code duplication. (thanks for leaving it there for comparison)
- please do not `use stdlib_blas_constants_*` in the interfaces, as that increases parsing complexity, but no constants are used there. 
- Something in the module/submodule structure can be slightly streamlined.

perazz (commented):
Thank you @jalvesz for the review. I think it is almost ready. 
I noticed the vast change in naming convention, and imho: 
- `_subm_` should not be used: it does not conform with the submodule names everywhere else in the library. Also, we should try to [stay away](https://fortran-lang.discourse.group/t/catastrophic-compile-error-on-ifx-when-defining-a-pointer-to-a-long-function-name/8625) from names that become too long
- For the same reason, a `submodule` would better not be used if there's only one (the `auxiliary` case). I suggest to not sacrifice clarity and simplicity now for the sake of a potential and currently unplanned future expansion :)

jalvesz:
I see your points @perazz, the `_subm_` was an idea, I'm fully open to a more structured and agreed upon naming convention. I also agree that too long path names are an issue. For instance building with Ninja under windows has a limit of 129 characters for the files path, so one has to be careful with that.

That being said, I do think that module/submodule splitting by:
* Modules should contain only interfaces (or private implementations)
* submodules the implementation for public procedures

Should be the preferred structuring for stdlib. There was a discussion here https://github.com/fortran-lang/stdlib/issues/540 around this topic.

perazz:
@jalvesz as you point out there is a prior specific discussion, I think it is a separate topic, that will find agreement for the whole library, which should not be part of the current PR which addresses the BLAS/LAPACK backends. So my suggestion would be to restore the names you proposed in the initial commits, and then you can propose the naming convention change as part of a separate PR where it can be properly addressed and discussed (separation of concerns).

jalvesz:
I have reverted the naming, I agree that it is better to follow that in a second round for a different PR. Current one is already big enough.

perazz (approved):
LGTM @jalvesz, thank you, afaict all comments I raised have been addressed. It is a very good PR imho.

perazz:
@jvdp1 do you have comments, and think we could merge this PR soon? I would think so, since the API has not changed. The overall build time is faster.

jvdp1 (approved):
Thank you @jalvesz and @perazz . I don't have additional comments. As all APIs seem to remain the same, then it should be ready to be merged!

jalvesz:
Thank you @perazz, @ivan-pi and @jvdp1! With 2 approvals I'll merge this PR. This will ease some CI issues :). I kept the old files in case we need to review anything down the road.

PR #911, opened 2024-12-26
system: subprocessing interface

perazz:
With this PR I want to introduce for discussion a comprehensive API for managing external processes in Fortran. 
This is a fully cross-platform implementation for synchronous (similar to `execute_command_line`) or asynchronous (fully detached process) processes, and mimics functionality from the Python `subprocess` module. 

### Previous discussion
- #22 
- #738 
- #266 

### Proposed API

Here’s the simplified list of interfaces with their types and example usage:  
- [x] **`type(process_type)`**: A derived type representing the external process state containing the state handler as well as `stdin`, `stdout`, `stderr` as strings.  
- [x] **`run`** a synchronous (blocking) process: 
`process = run(cmd [, stdin=string] [, want_stdout=.false.] [, want_stderr=.false.])`  
- [x] **`runasync`** an asynchronous (non-blocking) process: 
`process = runasync(cmd [, stdin=string] [, want_stdout=.false.] [, want_stderr=.false.])`  
- [x] **`update`** query process state and update the state variable: 
`call update(process)`  
- [x] **`is_running`** check if a process is still running and update the handler: 
`status = is_running(process)`  
- [x] **`is_completed`** check if a process has finished and update the handler: 
`status = is_completed(process)`  
- [x] **`wait`** for completion (option for max waiting time in seconds): 
`call wait(process [, max_wait_time=10.0])`  
- [x] **`elapsed`** time since process start, or total process lifetime: 
`duration = elapsed(process)`  
- [x] **`kill`** a running external process: 
`call kill(process, success)`  
- [x] **`sleep`** implementation is moved to C so that a slightly better wrapper based on `nanosecond` can be used 

This provides a concise overview of the interfaces for quick reference.

### Key facts
- Traditional Fortran `interface`s are used rather than type-bound procedures because that seems more similar to the other stdlib interfaces, but I'm happy to introduce them as well and further discuss the API. 
- `stderr`, `stdout`, `stdin` are stored via temporary files, but that is completely internal to the implementation. So, we can later implement faster memory-only communication via pipes without changing the API.
- A cross-platform C interface routine was necessary. Because C macros are more informative, all platform-dependent operation is kept on the C side only.  
- Error handling is currently made via integer exit codes. If/when the `type(state_type)` general error handler is approved, we can improve this and use it throughout the subprocessing API too.

### Prior art
- Python [subprocess](https://docs.python.org/3/library/subprocess.html) module

Issue #738, opened 2023-09-15
'Sleep' not working when compiled in Win32 (x86)


davidpfister:
### Description

The Sleep function does not even compile on Windows using ifort (ia-32 2021.5.0). One gets 
fatal error LNK1120: 1 unresolved external
error LNK2019: unresolved external symbol _Sleep

It works fine when building in x64 though.

### Expected Behaviour

The code should compile for x86 and x64 platforms.

### Version of stdlib

v0.3.0

### Platform and Architecture

Windows

### Additional Information

After playing around I came up with a solution that works for me. 
In x86 the sleep function on Windows should be called ```Sleep@4```. 

I changed the code as follows 
```fortran
#ifdef _WIN32
#ifdef _WIN64
        subroutine  winsleep(dwMilliseconds) bind (C, name='Sleep')
            import :: DWORD
            integer(DWORD) :: dwMilliseconds
        end subroutine
#else
        subroutine  winsleep(dwMilliseconds) bind (C, name='Sleep@4')
            import :: DWORD
            integer(DWORD) :: dwMilliseconds
        end subroutine
#endif
#else
        integer(c_int) function usleep(usec) bind (C)
            import c_int
            integer(c_int), value, intent(in) :: usec
        end function
#endif

subroutine sleep(millisec)
        integer, intent(in) :: millisec
        integer(c_int) :: ierr

#ifdef _WIN32
        call winsleep(%val(transfer(millisec, DWORD)))
#else
        ierr = usleep(int(millisec * 1000, c_int))
        if (ierr/=0) error stop 'problem with usleep() system call'
#endif
    end subroutine
```

and the chole thing works like a charm

jvdp1:
Thank you for reporting the bug.
Should this solution be for Intel ifort on Windows ia-32? Or is it a more general solution for all compilers on Windows ia-32? (My experience with Window is close to 0:()

davidpfister:
@jvdp1 Good question. It might be an ifort specificity on Windows. I do not have access to other compilers at the moment. I may be able to give it a try with gfortran and keep you posted. As far as I know the pragma _WIN32/_WIN64 are only standard for ifort (maybe ifx) on Windows, so the proposed code would only work with ifort anyway.

davidpfister:
After playing around with different function names I figured out that one should call `_sleep`. By doing so, there is no different between x86 and x64
```fortran
#ifdef _WIN32
[...]

PR #914, opened 2025-01-01
Remove mingw32

chuckyvt:
This PR removes MINGW32 from the Windows CI checks.  This version of MINGW receives limited support, and I don't believe it makes sense to use for validation of future development.  This change was originally suggested in this [PR](https://github.com/fortran-lang/stdlib/pull/854).

### DIFF
--- a/.github/workflows/ci_windows.yml
+++ b/.github/workflows/ci_windows.yml
@@ -13,8 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         include: [
-          { msystem: MINGW64, arch: x86_64 },
-          { msystem: MINGW32, arch: i686   }
+          { msystem: MINGW64, arch: x86_64 }
         ]
     defaults:
       run:
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ Name | Version | Platform | Architecture
 GCC Fortran | 10, 11, 12, 13 | Ubuntu 22.04.2 LTS | x86_64
 GCC Fortran | 10, 11, 12, 13 | macOS 12.6.3 (21G419) | x86_64
 GCC Fortran (MSYS) | 13 | Windows Server 2022 (10.0.20348 Build 1547) | x86_64
-GCC Fortran (MinGW) | 13 | Windows Server 2022 (10.0.20348 Build 1547) | x86_64, i686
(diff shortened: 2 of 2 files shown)


perazz:
Thank you @chuckyvt for this PR. I'm generally in favor of dropping support for ancient platforms. However, I find that: 
- 32-bit windows support seems still relevant for stdlib (#738)
- It was useful for a robust cross-platform implementation (#911), as `DWORD` has different sizes.
- I haven't seen issues with 32-bit Windows builds so far, is there any practical reason for this platform to be discarded?

So, maybe this can be deferred until issues related to 32-bit Windows may arise?

chuckyvt:
Will try to address your comments/concerns.  

Have issues been seen with MINGW32:   Yes, there have been.   Work on PR [#843](https://github.com/fortran-lang/stdlib/pull/843) was paused due to CI failures with MINGW32.   The failure traced back to a memory access error when reallocating an unlimited polymorphic variable.   The failure is only seen with Mingw32.  I tried every code refactor I could think of, but nothing seemed to address the issue,

Value of Win32 support:   The company I work for is highly dependent on Win32 IFORT due to the amount of legacy binary libraries in use and will continue to use it for years to come.  I suspect we aren't the only ones stuck in that boat.  However, this legacy support is specific to Intel IFORT.  (Note the Sleep issue mentioned is specific to IFORT).  I doubt there are many, if any, uses of Stdlib in a Mingw32 project.   As a side note, a future PR I would like to work on is to include Intel IFX in the Windows CI, and as part of that work could consider including 32-bit IFORT.  However, with IFORT no longer being actively devloped, I'm not sure if that is right thing to do or not.

Value in cross platform validation:   I think the work you're doing on subprocess type implementation is great.  However, at first glance, I don't see anything there that is specific to Win32.   The `_WIN32` variable is common to all versions of Windows as best I know, so the existing 64-bit Windows GCC check should capture it. (Windows Intel IFX needs to be included in Windows CI checks as well as mentioned above and is future work)

perazz (approved):
LGTM given the motivations.

jalvesz (approved):

perazz:
With two approvals, I suggest to wait another couple of days, and then merge if there are no further comments.

jalvesz:
I'll merge this PR as there have been no more comments! Thanks @chuckyvt

PR #919, opened 2025-01-17
Add character array interface to loadtxt

chuckyvt:
This PR is to add functionality to the loadtxt function to load a text file into a 1-D allocatable character array.   The len of the array will be set based on the longest line found in the file, and the size of the array will equal the number of lines in the file.  Added a 'skip_blank_lines' option, which is usueful to match list directed read capability, since that typically would skip blank lines.   

The PR functionality and code is inspired by this [post](https://fortran-lang.discourse.group/t/allocation-of-array-for-reading-strings-from-a-text-file/5986/2)   and thread though it was heavily modified from Beliavsky's initial code.

### DIFF
--- a/src/stdlib_io.fypp
+++ b/src/stdlib_io.fypp
@@ -66,6 +66,7 @@ module stdlib_io
     #:for k1, t1 in KINDS_TYPES
       module procedure loadtxt_${t1[0]}$${k1}$
     #:endfor
+    module procedure :: loadtxt_char
   end interface loadtxt
 
   interface savetxt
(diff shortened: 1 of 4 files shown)


perazz:
@chuckyvt `loadtxt` is meant for reading 2D tabular data, so I'm not sure that an interface with the same name would fit a generic text file read. 

Could we instead extend the (`getfile`)[https://github.com/fortran-lang/stdlib/blob/14f58b6f97a30d6a7c6469a2fb83d6c895ba054a/src/stdlib_io.fypp#L547-L637] in #904 to also return individual lines?

chuckyvt:
My thought was "loadtxt" is a generic name and shouldn't necessarily be limited to numeric arrays.  Using "loadtxt" for numeric arrays and "getfile" for character and string arrays isn't obvious at first glance.  Compare that to extending loadtxt to interface with numeric, character and string arrays, which I think is an easier approach for someone new to Fortran.

All that said, I'm open to putting this functionality under getfile, however that procedure being a function instead of a subroutine complicates it a bit.   We would need to add an argument to getfile to define the type to return.   Alternatively, we could convert getfile to a subroutine.

jvdp1:
I think that overloading `loadtxt` is a good idea.  However, the shape of the output might be different. A new set of subroutines (called `getfile` or another name) could be also an option. And if it has the same properties as `loadtxt`, then `loadtxt` could be depreceted.

src/stdlib_io.fypp
```diff
@@ -192,6 +193,119 @@ contains
...
+                end_idx = len(text)
+
+                do while (idx <= len(text))
+                    !! Find line end
+                    ! Look for either CR or LR
+                    ascii_idx = iachar(text(idx:idx))
+                    
+                    if (ascii_idx == 13) then
+                        ! Found CR return.  Check for LR
+                        if (iachar(text(idx+1:idx+1)) == 10) then
```
perazz, line 290: 2 questions: 
1) since `idx<=len(text)`, should we check that this does not go out of bounds?
2) just a style comment: would it be better to refer to the line feed using the intrinsic `new_line`? i.e.: 

```suggestion
                        if (text(idx+1:idx+1) == new_line('a')) then
```

perazz (commented):
Thank you for the PR @chuckyvt, I've added some comments in light of the recent `get_line` and `get_file` addition.

chuckyvt:
My thought is this this functionality makes more sense in the updated get_file subroutine?

perazz:
I think it’s a good idea @chuckyvt. Here is a couple of thoughts for discussion: 
- if the numpy equivalent `loadtxt` can also return strings, I think this PR could be moved forward 
- the `skip_lines` flag would well be useful also in `get_file`
- because `get_file` is a subroutine, to add the line versions we may just add to procedures to the interface, that return either an allocatable character array, or a `stringlist_type` type?

PR #926, opened 2025-01-29
Proposal to upgrade `number_of_columns`

### LINKED ISSUE #924: `loadtxt` CI failure
### Description

While testing stdlib I noticed `loadtxt` [still fails](https://github.com/perazz/stdlib/actions/runs/13026476737/job/36336362585) sometimes in the CI. 

Thanks to the recently added error messages, we now see: 

```
 loadtxt: error <Bad real number in item 1 of list input> reading 6 values from line 1 of example.dat.
ERROR STOP 
```

example.dat however only has 2 columns (not 6), but it has 6 total values: 

https://github.com/fortran-lang/stdlib/blob/6ec8cd0ea5988ccccb805af709d161389eee5a83/example/io/example.dat#L1-L3

 so the issue must lay in the `number_of_columns` function. 

### Expected Behaviour

No test should fail

### Version of stdlib

0.7.0

### Platform and Architecture

CI - ubuntu + gcc

### Additional Information

_No response_

perazz:
Using `getline` instead of non-advancing io, we may be able to fix #924 

cc: @jvdp1 @jalvesz

### DIFF
--- a/src/stdlib_io.fypp
+++ b/src/stdlib_io.fypp
@@ -252,6 +252,7 @@ contains
 
     integer :: ios, skiprows_, i
     character :: c
+    character(len=:), allocatable :: line
     logical :: lastblank
 
     skiprows_ = optval(skiprows, 0)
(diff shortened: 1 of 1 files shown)


src/stdlib_io.fypp
```diff
@@ -261,12 +262,16 @@ contains
     do i = 1, skiprows_
       read(s, *)
     end do
-
+
```
jvdp1, line 265:
```suggestion
```

jvdp1 (approved):
LGTM. Thank you @perazz

perazz:
Thank you @jvdp1 - forgot to mention: this is clearly a compiler issue: sometimes, it skips the end-of-line character and keeps reading in all values as if they were from the first line imho.

jalvesz (approved):
Nice catch @perazz

perazz:
Thank you @jvdp1 @jalvesz, I'll merge this with two approvals.

PR #929, opened 2025-01-30
Fix generalized `eig` rwork size

### LINKED ISSUE #927: "double free or corruption (out)" when requesting eigenvectors in complex generalized eigenvalue problem
### Description

Calling the eig subroutine for a generalized eigenvalue problem with complex A and B matrices leads to a memory fault when right or left eigenvectors are requested. The error seems to only occur for this specific combination, if I change A and B to real, or remove the optional left and right eigenvector arguments there are no problems.

Here is a reproducer:
```
program test
  use stdlib_linalg, only: eig
  implicit none

  double complex, dimension(3,3) :: A_Z,S_Z
  double precision, dimension(3,3) :: A_D,S_D
  double complex, dimension(3,3) :: vecs_r
  double complex,dimension(3) :: eigs
  integer :: i

  ! Set matrix
  A_Z = reshape( [ [1, 6, 3], &
                 [9, 2, 1], &
                 [8, 3, 4] ], [3,3] )

  S_Z = 0
  do i= 1,3
    S_Z(i,i) = i
  end do

  A_D = real(A_Z)
  S_D = real(S_Z)

  call eig(A_D,S_D,eigs,right=vecs_r) !Works
  call eig(A_Z,S_Z,eigs,right=vecs_r) !Fails
  write(6,*) eigs

end program test
```
and here is the error message and backtrace:
```
double free or corruption (out)

Program received signal SIGABRT: Process abort signal.

Backtrace for this error:
#0  0x7f0809780d11 in ???
#1  0x7f080977fee5 in ???
#2  0x7f08093d108f in ???
	at /build/glibc-LcI20x/glibc-2.31/signal/../sysdeps/unix/sysv/linux/x86_64/sigaction.c:0
#3  0x7f08093d100b in __GI_raise
	at ../sysdeps/unix/sysv/linux/raise.c:51
#4  0x7f08093b0858 in __GI_abort
	at /build/glibc-LcI20x/glibc-2.31/stdlib/abort.c:79
#5  0x7f080941b26d in __libc_message
	at ../sysdeps/posix/libc_fatal.c:155
#6  0x7f08094232fb in malloc_printerr
	at /build/glibc-LcI20x/glibc-2.31/malloc/malloc.c:5347
#7  0x7f0809424f9f in _int_free
	at /build/glibc-LcI20x/glibc-2.31/malloc/malloc.c:4314
#8  0x556798caef27 in ???
#9  0x556798c98f6c in test
	at /home/edvinolofsson/stdlib_gen_eig_bug/test.f90:25
#10  0x556798c99092 in main
	at /home/edvinolofsson/stdlib_gen_eig_bug/test.f90:2
Aborted (core dumped)
```

### Expected Behaviour

The call to eig should return the eigenvalues and eigenvectors also for complex A and B.

### Version of stdlib

8916fc52fb8df18e79f5877c0bedf986c4ca55e5

### Platform and Architecture

Ubuntu 20.04, x86 , gfortran 10.5.0  (but can reproduce it also on compiler explorer with latest gfortran)

### Additional Information

compiler explorer link: [https://godbolt.org/z/bYW6v6jso](https://godbolt.org/z/bYW6v6jso)

perazz:
Fix #927. 

`rwork` size is different in `*GGEV` (`8*n`) vs. in `*GEEV` (`2*n`), I had overlooked that. 

cc: @jalvesz @jvdp1 
Thank you

### DIFF
--- a/test/linalg/test_linalg_eigenvalues.fypp
+++ b/test/linalg/test_linalg_eigenvalues.fypp
@@ -320,9 +320,8 @@ module test_linalg_eigenvalues
     subroutine test_issue_927_${ci}$(error)
       type(error_type), allocatable, intent(out) :: error 
 
-      ${ct}$, dimension(3,3) :: A_Z,S_Z,vecs_r
-      ${ct}$,dimension(3) :: eigs
-      real(${ck}$), dimension(3,3) :: A_D,S_D
+      ${ct}$ :: A_Z(3,3),S_Z(3,3),vecs_r(3,3),eigs(3)
+      real(${ck}$) :: A_D(3,3),S_D(3,3)
       type(linalg_state_type) :: state
       integer :: i
 
--- a/test/linalg/test_linalg_eigenvalues.fypp
+++ b/test/linalg/test_linalg_eigenvalues.fypp
@@ -3,7 +3,7 @@
 module test_linalg_eigenvalues
     use stdlib_linalg_constants
     use stdlib_linalg_state
-    use stdlib_linalg, only: eig, eigh, eigvals, eigvalsh, diag
+    use stdlib_linalg, only: eig, eigh, eigvals, eigvalsh, diag, eye
     use testdrive, only: error_type, check, new_unittest, unittest_type    
 
     implicit none (type,external)
@@ -33,7 +33,8 @@ module test_linalg_eigenvalues
         #:for ck,ct,ci in CMPLX_KINDS_TYPES
         #:if ck!="xdp"    
         tests = [tests,new_unittest("test_eig_complex_${ci}$",test_eig_complex_${ci}$), &
-                       new_unittest("test_eig_generalized_complex_${ci}$",test_eigvals_generalized_complex_${ci}$)]                
+                       new_unittest("test_eig_generalized_complex_${ci}$",test_eigvals_generalized_complex_${ci}$), &
+                       new_unittest("test_eig_issue_927_${ci}$",test_issue_927_${ci}$)]                
         #:endif
         #: endfor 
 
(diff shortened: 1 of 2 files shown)


jvdp1 (approved):
nice catch! Thank you @perazz

test/linalg/test_linalg_eigenvalues.fypp
```diff
@@ -324,10 +316,41 @@ module test_linalg_eigenvalues
         
     end subroutine test_eigvals_generalized_complex_${ci}$
 
-    #:endif
+    ! Generalized eigenvalues should not crash
+    subroutine test_issue_927_${ci}$(error)
+      type(error_type), allocatable, intent(out) :: error 
+
+      ${ct}$, dimension(3,3) :: A_Z,S_Z,vecs_r
```
jalvesz, line 323: Following the style guide https://github.com/fortran-lang/stdlib/blob/master/STYLE_GUIDE.md#attributes I would recommend avoiding the use of the `dimension` attribute and prefere the shorter array notation.

jalvesz (changes requested):

test/linalg/test_linalg_eigenvalues.fypp
perazz: Good idea @jalvesz, you may want to look further into the whole library as from a quick search, there are several occurrences

jalvesz (approved):

jalvesz (approved):

perazz:
Thanks for the quick review @jvdp1 @jalvesz! I will merge this now as it is better to close the bug earlier rather than later.

PR #936, opened 2025-02-11
strings: `join`, `to_c_char`

perazz:
This PR is spun off #911 for clarity.  

#### `join`  
```fortran
result = join(strings [, separator])  
```  
- Joins an array of strings into a single string, using the specified separator 
- Separator is optional, may have length>1 (default: 1 space). 
- accepts `character(*)` or `type(string_type)` arrays.

This simplifies string concatenation for command construction and formatted output.  

#### `to_c_char`  
```fortran
result = to_c_char(value)  
```  
Converts a Fortran string to a null-terminated C character array (`c_char`), enabling interoperability with C libraries that require C-style strings. 
- accepts `character(*)` or `type(string_type)` variables.
- name choice: to be similar to `to_string`. maybe we should consider `to_c_char`? or `c_char` to mimic the internal `char` interface?


doc/specs/stdlib_strings.md
```diff
@@ -459,6 +459,43 @@ The result is of the same type as `string`.
...
+
+#### Class
+
+Pure function
+
+#### Argument
+
+- `strings`: Array of strings (either `type(string_type)` or `character(len=*)`).
+  This argument is `intent(in)`. It is an array of strings that will be concatenated together.
+- `separator`: Character scalar (optional).
```
perazz, line 487:
```suggestion
- `separator`: `character(len=*)` scalar (optional).
```

doc/specs/stdlib_strings.md
```diff
@@ -498,3 +535,38 @@ The result is an `allocatable` length `character` scalar with up to `128` cached
 ```fortran
 {!example/strings/example_to_string.f90!}
 ```
+
+### `to_c_string`
```
jvdp1, line 540: As the output is a C character, the  name `to_c_char` might be more appropriate.

doc/specs/stdlib_strings.md
```diff
@@ -498,3 +535,38 @@ The result is an `allocatable` length `character` scalar with up to `128` cached
...
+Experimental
+
+#### Class
+
+Pure function.
+
+#### Argument
+
+- `value`: Shall be a `character(len=*)` string or a `type(string_type)` variable. It is an `intent(in)` argument.  
+  The Fortran string that will be converted to a C character array.
```
jvdp1, line 562:
```suggestion
  This Fortran variable will be converted to a C character array.
```

example/strings/example_join.f90
```diff
@@ -0,0 +1,21 @@
...
+
+  ! Custom separator
+  line = join(words, "_")
+  print *, "'" // line // "'"  !! 'Hello_World_Fortran'
+
+  ! Custom 2-character separator
+  line = join(words, ", ")
+  print *, "'" // line // "'"  !! 'Hello, World, Fortran'
+
+  stop 0
```
jvdp1, line 20:
```suggestion
```

Why is there a `stop 0`?

example/strings/example_to_c_string.f90
```diff
@@ -0,0 +1,22 @@
...
+  character(*), parameter :: hello = "Hello, World!"
+  
+  ! Convert character array
+  cstr = to_c_string(hello)
+  
+  ! Convert string type  
+  cstr2 = to_c_string(string_type(hello))
+    
+  if (size(cstr)==size(cstr2) .and. all(cstr==cstr2)) then 
+     stop 0
```
jvdp1, line 17:
```suggestion
     exit
```
Is this needed?

jalvesz:
>maybe we should consider to_c_char? or c_char to mimic the internal char interface?

I think `to_c_char` is a very good option and its clear about the intent of the procedure. `c_char` being the kind designator it would be confusing.

jvdp1 (approved):
LGTM! Thank you @perazz

perazz:
That was fast, thank you @jvdp1 @jalvesz!

I've renamed to `to_c_char`. As Jose points out, `c_char` would collapse with the integer kind, that can easily lead to errors

doc/specs/stdlib_strings.md
perazz: [23e535b](https://github.com/fortran-lang/stdlib/pull/936/commits/23e535bc8abea89d2b193961af5030d8a1ec7ddf)

example/strings/example_to_c_string.f90
perazz: [89df63a](https://github.com/fortran-lang/stdlib/pull/936/commits/89df63a8d2b9e57543dff790e64a70633b5b5f10)

example/strings/example_join.f90
perazz: I agree, it is not necessary.

jalvesz (changes requested):
I saw in several places that you use `cmd` to refer to the output of the string manipulation. Would it be possible to use one of `res`, `out` or `str` ? I think `cmd` signals an explicit use of commands, but these implementations are much more generic than that.

doc/specs/stdlib_strings.md
```diff
@@ -459,6 +459,43 @@ The result is of the same type as `string`.
...
+
+#### Description
+
+Joins an array of strings into a single string. This function concatenates the strings from the input array, 
+inserting a separator between each string (default: space). A user-defined separator may be provided, The resulting string is returned.
+
+
+#### Syntax
+
+`cmd = ` [[stdlib_strings(module):join(interface)]] ` (strings, separator)`
```
jalvesz, line 473:
```suggestion
`joined = ` [[stdlib_strings(module):join(interface)]] ` (strings, separator)`
```

jalvesz (approved):

jalvesz:
LGTM @perazz! Great addition, no further comments on my side :)

perazz:
Thank you for the reviews @jvdp1 @jalvesz, let's wait another day or two and then merge, if there are no further comments.

PR #938, opened 2025-02-17
error: implement general `state_type` handler

perazz:
This PR was spun off #904 for clarity. 

Because the adoption of a `linalg_state_type` was successful and flexible enough, it is proposed to extend the linear algebra error handling type by introducing a new `state_type` base class as part of the `stdlib_error` module. 

This would provide a unified framework for managing error states across all different modules of `stdlib`, not only linear algebra but also I/O, filesystem and other modules.  

The `state_type` class has the same contents as `linalg_state_type`: integer error flag, fixed-size character storage for messages and error locations, and type-bound procedures for error checking and reporting. The existing `linalg_state_type` now extends `state_type`, ensuring consistency - and interoperability.  

The linear algebra modules still require `type(linalg_state_type)` outputs. We could make them `class(state_type), optional, intent(out)`, so they would seamlessly handle both derived types. However, in Fortran, polymorphic arguments cannot appear in `pure` procedures, so another alternative would be to template them with `fypp`.


jvdp1 (approved):
LGTM! Thank you @perazz

perazz:
Thank you @jvdp1 for the review. If you agree, I suggest to wait a bit longer for additional comments, and then merge if none, so we can make more progress on ohters.

perazz:
Thank you @jvdp1 @jalvesz, merging now.

PR #942, opened 2025-03-03
system: OS type query

perazz:

Add OS detection routines:  
- `get_runtime_os` – Determines the OS at runtime using environment checks.  
- `OS_TYPE` – Cached version of `get_runtime_os` for efficiency.  
- `OS_NAME` – Return the OS name as a string.    

Test programs require that stdlib is never run on an unknown OS. 

cc: @jvdp1 @jalvesz

### DIFF
--- a/src/stdlib_system.F90
+++ b/src/stdlib_system.F90
@@ -5,6 +5,70 @@ implicit none
 private
 public :: sleep
 
+!! version: experimental
+!!
+!! Cached OS type retrieval with negligible runtime overhead.
+!! ([Specification](../page/specs/stdlib_system.html#os_type-cached-os-type-retrieval))
+!!
+!! ### Summary
+!! Provides a cached value for the runtime OS type. 
+!!
+!! ### Description
+!! 
+!! This function caches the result of `get_runtime_os` after the first invocation. 
+!! Subsequent calls return the cached value, ensuring minimal overhead.
+!!
+public :: OS_TYPE
+
+!! version: experimental
+!!
+!! Determine the current operating system (OS) type at runtime.
+!! ([Specification](../page/specs/stdlib_system.html#get_runtime_os-determine-the-os-type-at-runtime))
+!!
+!! ### Summary
+!! This function inspects the runtime environment to identify the OS type. 
+!!
+!! ### Description
+!! 
+!! The function evaluates environment variables (`OSTYPE` or `OS`) and filesystem attributes
+!! to identify the OS. It distinguishes between several common operating systems:
+!! - Linux
+!! - macOS
+!! - Windows
+!! - Cygwin
+!! - Solaris
+!! - FreeBSD
+!! - OpenBSD
+!!
+!! Returns a constant representing the OS type or `OS_UNKNOWN` if the OS cannot be determined.
+!!                    
+public :: get_runtime_os
+
+!> Version: experimental
+!>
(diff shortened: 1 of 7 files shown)


src/stdlib_system.F90
```diff
@@ -430,4 +497,130 @@ end function process_get_ID
...
+
+    if (rc == 0 .and. length > 0) then
+        ! Linux
+        if (index(val, 'linux') > 0) then
+            os = OS_LINUX
+            return
+        end if
+
+        ! macOS
+        if (index(val, 'darwin') > 0) then
```
jalvesz, line 529: I wonder, shouldn't these conditionals be if/else if with a single return at the end?

src/stdlib_system.F90
```diff
@@ -5,6 +5,70 @@ module stdlib_system
...
+!!
+!! Returns a constant representing the OS type or `OS_UNKNOWN` if the OS cannot be determined.
+!!                    
+public :: get_runtime_os
+
+!> Version: experimental
+!>
+!> Integer constants representing known operating system (OS) types
+!> ([Specification](../page/specs/stdlib_system.html))
+integer, parameter, public :: &
```
jalvesz, line 51: These parameters could be defined with the `enum` facility:
```Fortran
enum, bind(c)
    enumerator :: OS_UNKNOWN !> Represents an unknown operating system
    enumerator :: OS_LINUX   !> Represents a Linux operating system
    enumerator :: OS_MACOS   !> Represents a macOS operating system
    enumerator :: OS_WINDOWS !> Represents a Windows operating system
    enumerator :: OS_CYGWIN  !> Represents a Cygwin environment
    enumerator :: OS_SOLARIS !> Represents a Solaris operating system
    enumerator :: OS_FREEBSD !> Represents a FreeBSD operating system
    enumerator :: OS_OPENBSD !> Represents an OpenBSD operating system
end enum
```

perazz: I don't disagree with this idea, but looking at the rest of the code, `enum`s are never employed, while the integer `parameter` approach is used in several places. So I would think that the approach used in the PR is more in line with the rest of the code, although this is not specified in the style guide.

src/stdlib_system.F90
perazz: [fbaacc9](https://github.com/fortran-lang/stdlib/pull/942/commits/fbaacc92e07db97cb1dac4d67b4ff0c8f3575fde)

src/stdlib_system.F90
jvdp1: If I remember correctly, there were some issues with `enum` with some compilers at the start of the project. I don't disagree with using `enum`.

jalvesz: I'm not saying we should take this way, I just find it "elegant", if you will, because the numbering is automatic. We introduced `enum` here https://github.com/fortran-lang/stdlib/blob/69eaa2056aca32c3f7e0119c1bb4b208d99b48cd/src/stdlib_sparse_constants.fypp#L10 without problems, I have not experienced issues on my own projects either.

No dogmas here really.

jalvesz (approved):

jvdp1 (approved):
LGTM. Thank you @perazz

perazz:
Thank you @jvdp1 @jalvesz, let's merge this PR soon, so we can continue discussing more system-related functions.

Issue #898, opened 2024-11-27
Compiling with warning maybe-uninitialized show a potential bug in gamma functions


cyrilgandon:
### Description

In https://github.com/fortran-lang/stdlib/blob/68524b3280a25a4a36995c431cac59551f0c4fb7/src/stdlib_specialfunctions_gamma.fypp#L580 the value `ss` is declared:

and then use here: https://github.com/fortran-lang/stdlib/blob/68524b3280a25a4a36995c431cac59551f0c4fb7/src/stdlib_specialfunctions_gamma.fypp#L673

The `ss` variable is never initalized, and generates 2 warnings with 

`fpm build --verbose  --flag '-Wmaybe-uninitialized'
`

> 
> .\.\src\temp\stdlib_specialfunctions_gamma.f90:1424:24:
> 
>  1424 |             m = nint(ss)
>       |                        ^
> Warning: 'ss' may be used uninitialized [-Wmaybe-uninitialized]
> .\.\src\temp\stdlib_specialfunctions_gamma.f90:1331:52:
> 
>  1331 |         real(qp) :: res, p_lim, a, b, g, c, d, y, ss
>       |                                                    ^
> note: 'ss' was declared here
> .\.\src\temp\stdlib_specialfunctions_gamma.f90:1290:24:
> 
>  1290 |             m = nint(ss)
>       |                        ^
> Warning: 'ss' may be used uninitialized [-Wmaybe-uninitialized]
> .\.\src\temp\stdlib_specialfunctions_gamma.f90:1197:52:
> 
>  1197 |         real(dp) :: res, p_lim, a, b, g, c, d, y, ss
>       |                                                    ^
> note: 'ss' was declared here
> [ 12%] stdlib_specialfunctions_gamma.  done.

### Expected Behaviour

Either the value to be initialized, of the chunk of code deleted.

### Version of stdlib

68524b3280a25a4a36995c431cac59551f0c4fb7

### Platform and Architecture

Windows

perazz:
Thanks @cyrilgandon, from a quick look, it really looks like a bug. 
The mentioned reference can be found [at this link](https://hal.science/hal-01329669v2/document)

PR #943, opened 2025-03-06
fix: gamma functions

wassup05:
This PR tries to address #898 

The problem was in the interface `gpx` (meant for internal use only) which had two procedures for `x, p are real` and `x is real, p is an integer`, `p being > 0` in all cases

Everything was fine with the second procedure but in the first one it is required that ([Fast and accurate evaluation of a
generalized incomplete gamma function](hal.science/hal-01329669v2/document)) `when x < 0, p be an integer` which wasn't actually programmed in the procedure itself but was being taken care of as and when requiring to call `gpx` in such a case. like so https://github.com/fortran-lang/stdlib/blob/master/src/stdlib_specialfunctions_gamma.fypp#L1078-L1099

And hence this was not found earlier as the chunk of code was in the case when `x<0 and p<p_lim(x)` (check Algorithm 3) which is now unreachable, hence removed

### DIFF
--- a/src/stdlib_specialfunctions_gamma.fypp
+++ b/src/stdlib_specialfunctions_gamma.fypp
@@ -4,6 +4,7 @@
 #:set CI_KINDS_TYPES = INT_KINDS_TYPES + C_KINDS_TYPES
 module stdlib_specialfunctions_gamma
     use iso_fortran_env, only : qp => real128
+    use ieee_arithmetic, only: ieee_value, ieee_quiet_nan
     use stdlib_kinds, only :  sp, dp, int8, int16, int32, int64
     use stdlib_error, only : error_stop
 
@@ -670,6 +671,9 @@ contains
                 if(abs(y - one) < tol_${k2}$) exit
 
             end do
+
+        else
+            g = ieee_value(1._${k1}$, ieee_quiet_nan)
 
         end if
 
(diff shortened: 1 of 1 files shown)


src/stdlib_specialfunctions_gamma.fypp
```diff
@@ -668,30 +671,6 @@ contains
 
             end do
```
perazz, line 673: Although this should never happen due to the previous `error_stop`, I believe it's good to have this `if` block consider all possible branches, to suppress potential compiler warnings. 

```suggestion
        else                           
            g = ieee_value(1._${k1}$, ieee_quiet_nan)
```

perazz (commented):
LGTM with the above suggestion @wassup05. Thank you for the fix.

As an additional request, it would be great to see gamma functions a little better documented in the docs files, so the lazy user doesn't have to search into the source code to understand what ranges and types they work with.

jvdp1 (approved):
LGTM. Thank you.

perazz (approved):

jalvesz (approved):
LGTM @wassup05

perazz:
With 3 approvals, I will merge this PR. Thank you @wassup05 for the PR, and @jalvesz @jvdp1 for the reviews.

PR #946, opened 2025-03-07
system: `is_directory`

perazz:

- It provides a `logical` function to evaluate whether a given path is a directory. 
- On Windows, it natively works both with `/` and `\` separators. 
- Relies on C functions, to avoid Fortran pre-processor macros and path swapping, that would be slow. 

Some questions to discuss that come to my mind: 
- Should this be part of `stdlib_system`, `stdlib_io`, or maybe a `submodule(stdlib_system) stdlib_system_filesystem`? or else?
- Should we begin standardizing the C-side names somehow? They use the same global namespace, so we should start being careful imho. 

cc: @jvdp1 @jalvesz

### DIFF
--- a/src/stdlib_system.F90
+++ b/src/stdlib_system.F90
@@ -93,8 +93,8 @@ public :: is_windows
 !!
 !!### Description
 !! 
-!! This function checks if a given file system path is a directory. It is cross-platform and avoids reliance 
-!! on external C libraries by utilizing system calls. It supports common operating systems such as Linux, macOS, 
+!! This function checks if a given file system path is a directory. It is cross-platform and utilizes
+!! native system calls. It supports common operating systems such as Linux, macOS, 
 !! Windows, and various UNIX-like environments. On unsupported operating systems, the function will return `.false.`.
 !!
 public :: is_directory
--- a/src/stdlib_system.F90
+++ b/src/stdlib_system.F90
@@ -650,7 +650,7 @@ logical function is_directory(path)
 
     end interface        
     
-    is_directory = logical(stdlib_is_directory(to_c_char(path)))
+    is_directory = logical(stdlib_is_directory(to_c_char(trim(path))))
     
 end function is_directory
 
--- a/src/stdlib_system.F90
+++ b/src/stdlib_system.F90
@@ -1,6 +1,7 @@
 module stdlib_system
 use, intrinsic :: iso_c_binding, only : c_int, c_long, c_null_ptr, c_int64_t
-use stdlib_kinds, only: int64, dp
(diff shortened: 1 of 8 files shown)


perazz:
Funny enough, `loadtxt` is still failing sometimes: 
```
169/368 Test #170: loadtxt ................................***Failed    0.01 sec
 loadtxt: error <End of file> reading 0 values from line 2 of example.dat.
ERROR STOP 
```

src/stdlib_system.F90
```diff
@@ -82,6 +83,22 @@ module stdlib_system
...
+!! Tests if a given path matches an existing directory.
+!! ([Specification](../page/specs/stdlib_io.html#is_directory-test-if-a-path-is-a-directory))
+!!
+!!### Summary
+!! Function to evaluate whether a specified path corresponds to an existing directory.
+!!
+!!### Description
+!! 
+!! This function checks if a given file system path is a directory. It is cross-platform and avoids reliance 
+!! on external C libraries by utilizing system calls. It supports common operating systems such as Linux, macOS,
```
jvdp1, line 97: Is this correct, as stdlib_system_subprocess relies on some C libraries?

jvdp1 (approved):
LGTM. Thank you  @perazz

src/stdlib_system.F90
```diff
@@ -82,6 +83,22 @@ module stdlib_system
...
+!! ([Specification](../page/specs/stdlib_io.html#is_directory-test-if-a-path-is-a-directory))
+!!
+!!### Summary
+!! Function to evaluate whether a specified path corresponds to an existing directory.
+!!
+!!### Description
+!! 
+!! This function checks if a given file system path is a directory. It is cross-platform and avoids reliance 
+!! on external C libraries by utilizing system calls. It supports common operating systems such as Linux, macOS, 
+!! Windows, and various UNIX-like environments. On unsupported operating systems, the function will return `.false.`.
```
perazz, line 98:
```suggestion
!! This function checks if a given file system path is a directory. It is cross-platform and utilizes
!! native system calls. It supports common operating systems such as Linux, macOS, 
!! Windows, and various UNIX-like environments. On unsupported operating systems, the function will return `.false.`.
```

src/stdlib_system.F90
perazz: This was not very clear!

jalvesz (approved):
LGTM @perazz! Thank for this nice PR!!

perazz:
Thank you @jvdp1 @jalvesz, let's wait a bit more, and then we can merge imho.

perazz:
Merging.

PR #949, opened 2025-03-11
small fix for `loadtxt`

wassup05:
Using the previous format directive would have caused a runtime error if there was an error skipping the required rows as there were only 3 format arguments given (instead of 4)

This PR fixes that and provides a useful error message with the required arguments


perazz (approved):
LGTM, thank you @wassup05.

jalvesz (approved):

perazz:
With two approvals, I will merge this one, thank you @wassup05.

Issue #931, opened 2025-02-06
Extend the intrinsic `matmul` to handle more than just two matrices.


loiseaujc:
### Motivation

The Fortran intrinisc `matmul` can handle only two matrices, i.e. `result = matmul(A, B)`. While working on problems involving multiple low-rank matrices, I've been faced recently with expressions like `result = A * B * C`, or `result = A * B * C * D`, where `A`, `B`, `C` and `D` tend to be tall or wide matrices. Depending on how you evaluate these matrix-matrix multiplications, you can either have a pretty fast code or a terribly slow one.

Except for myself, would it be of interest to anyone to extend the intrinsic `matmul` to handle expressions involving multiple matrices (say between 2 and 5, I don't think there are situations where you need much more than that). The corresponding interfaces would simply be

```fortran
result = matmul(A, B) ! Intrinsic matmul
result = matmul(A, B, C)
result = matmul(A, B, C, D)
result = matmul(A, B, C, D, E)
```
Given the dimensions of each array, the optimal order in which to perform the operations can be determined by solving the corresponding optimization problem using dynamic programming. Some details can be found [here](https://en.wikipedia.org/wiki/Matrix_chain_multiplication).

### Prior Art

- `numpy` exposes the `numpy.linalg.multi_dot(arrays, *, out=None)` function which chains calls to `numpy.dot` using the optimal parenthesization of the matrices.

- I'm pretty sure `julia` has a similar feature but a quick search didn't bring actual references to it

As far as I know, Fortran does not allow for arbitrary long lists of dummy arguments (except for intrisic `max` and `min` and a couple of others I suppose). In practice, I don't think the extended `matmul` interface would need to handle more than say 5-ish arrays though.

### Additional Information

Pinging @perazz, @jvdp1 and @jalvesz.

PR #951, opened 2025-03-13
feat: extend intrinsic `matmul`

wassup05:
This PR attempts to address #931 

interface `stdlib_matmul` is created and extended to handle 3-5 matrices. (works for `integer`, `real`, `complex`)

**API**
```fortran
A = stdlib_matmul(B, C, D, E, F)
A = stdlib_matmul(B, C, D, E)
A = stdlib_matmul(B, C, D)
```

The Algorithm for optimal parenthesization is as given in "Introduction to Algorithms" by Cormen, ed4, ch14, section-2.
numpy's [`linalg.multidot`](https://numpy.org/doc/2.2/reference/generated/numpy.linalg.multi_dot.html) also uses the same algorithm.

Although as @jalvesz had suggested I should have used `gemm` for the multiplication of component matrices this uses `matmul` for now, I can make this if deemed appropriate once the major implementation has been given a green signal.

I have added a very basic example to play around with, and I will be adding the detailed docs, specs and tests once everybody approves of the major implementation.

I am not really happy with some parts of the code like computing the `size` of all the matrices individually, If anyone has any suggestions regarding that or any cleaner way of implementing some other stuff (perhaps some `fypp` magic) please do let me know

**Notes**
- I think another interesting enhancement would be, if the first array is 1-D, then treat it as a row vector. If the last error is 1-D treat it as a column vector just as numpy does it.

### DIFF
--- a/src/stdlib_intrinsics_matmul.fypp
+++ b/src/stdlib_intrinsics_matmul.fypp
@@ -8,7 +8,7 @@ submodule (stdlib_intrinsics) stdlib_intrinsics_matmul
 
 contains
 
-    ! Algorithm for the optimal bracketization of matrices
+    ! Algorithm for the optimal parenthesization of matrices
     ! Reference: Cormen, "Introduction to Algorithms", 4ed, ch-14, section-2
     ! Internal use only!
     pure function matmul_chain_order(n, p) result(s)
--- a/src/stdlib_intrinsics_matmul.fypp
+++ b/src/stdlib_intrinsics_matmul.fypp
@@ -1,0 +1,119 @@
+#:include "common.fypp"
+#:set I_KINDS_TYPES = list(zip(INT_KINDS, INT_TYPES, INT_KINDS))
+#:set R_KINDS_TYPES = list(zip(REAL_KINDS, REAL_TYPES, REAL_SUFFIX))
+#:set C_KINDS_TYPES = list(zip(CMPLX_KINDS, CMPLX_TYPES, CMPLX_SUFFIX))
+
+submodule (stdlib_intrinsics) stdlib_intrinsics_matmul
+    implicit none
+
+contains
+
+    ! Algorithm for the optimal bracketization of matrices
+    ! Reference: Cormen, "Introduction to Algorithms", 4ed, ch-14, section-2
+    ! Internal use only!
+    pure function matmul_chain_order(n, p) result(s)
+        integer, intent(in) :: n, p(:)
+        integer :: s(1:n-1, 2:n), m(1:n, 1:n), l, i, j, k, q
+        m(:,:) = 0
+        s(:,:) = 0
+
+        do l = 2, n
+            do i = 1, n - l + 1
+                j = i + l - 1
+                m(i,j) = huge(1)
+                
+                do k = i, j - 1
+                    q = m(i,k) + m(k+1,j) + p(i)*p(k+1)*p(j+1)
+
+                    if (q < m(i, j)) then
+                        m(i,j) = q
+                        s(i,j) = k
+                    end if
+                end do
+            end do
+        end do
+    end function matmul_chain_order
+
+#:for k, t, s in I_KINDS_TYPES + R_KINDS_TYPES + C_KINDS_TYPES
+
+    pure module function stdlib_matmul_${s}$_3 (a, b, c) result(d)
+        ${t}$, intent(in) :: a(:,:), b(:,:), c(:,:)
+        ${t}$, allocatable :: d(:,:)
+        integer :: sa(2), sb(2), sc(2), cost1, cost2
+        sa = shape(a)
+        sb = shape(b)
+        sc = shape(c)
+
+        if ((sa(2) /= sb(1)) .or. (sb(2) /= sc(1))) then
+            error stop "stdlib_matmul: Incompatible array shapes"
+        end if
+
+        ! computes the cost (number of scalar multiplications required)
+        ! cost(A, B) = shape(A)(1) * shape(A)(2) * shape(B)(2)
+        cost1 = sa(1) * sa(2) * sb(2) + sa(1) * sb(2) * sc(2) ! ((AB)C)
+        cost2 = sb(1) * sb(2) * sc(2) + sa(1) * sa(2) * sc(2) ! (A(BC))
+
+        if (cost1 < cost2) then
+            d = matmul(matmul(a, b), c)
+        else
+            d = matmul(a, matmul(b, c))
+        end if
+    end function stdlib_matmul_${s}$_3
+
+    pure module function stdlib_matmul_${s}$_4 (a, b, c, d) result(e)
+        ${t}$, intent(in) :: a(:,:), b(:,:), c(:,:), d(:,:)
+        ${t}$, allocatable :: e(:,:)
+        integer :: p(5), i
+        integer :: s(3,2:4)
+
+        p(1) = size(a, 1)
+        p(2) = size(b, 1)
+        p(3) = size(c, 1)
+        p(4) = size(d, 1)
+        p(5) = size(d, 2)
+
+        s = matmul_chain_order(4, p)
+
+        select case (s(1,4))
+            case (1)
+                e = matmul(a, stdlib_matmul(b, c, d))
+            case (2)
+                e = matmul(matmul(a, b), matmul(c, d))
+            case (3)
+                e = matmul(stdlib_matmul(a, b ,c), d)
+            case default
+                error stop "stdlib_matmul: unexpected error unexpected s(i,j)"
+        end select
+    end function stdlib_matmul_${s}$_4
+
+    pure module function stdlib_matmul_${s}$_5 (a, b, c, d, e) result(f)
+        ${t}$, intent(in) :: a(:,:), b(:,:), c(:,:), d(:,:), e(:,:)
+        ${t}$, allocatable :: f(:,:)
+        integer :: p(6), i
+        integer :: s(4,2:5)
+
+        p(1) = size(a, 1)
+        p(2) = size(b, 1)
+        p(3) = size(c, 1)
+        p(4) = size(d, 1)
+        p(5) = size(e, 1)
+        p(6) = size(e, 2)
+
+        s = matmul_chain_order(5, p)
+
+        select case (s(1,5))
+            case (1)
+                f = matmul(a, stdlib_matmul(b, c, d, e))
+            case (2)
+                f = matmul(matmul(a, b), stdlib_matmul(c, d, e))
+            case (3)
+                f = matmul(stdlib_matmul(a, b ,c), matmul(d, e))
+            case (4)
+                f = matmul(stdlib_matmul(a, b, c, d), e)
+            case default
+                error stop "stdlib_matmul: unexpected error unexpected s(i,j)"
+        end select
+    end function stdlib_matmul_${s}$_5
+
+#:endfor
+end submodule stdlib_intrinsics_matmul
--- a/example/intrinsics/CMakeLists.txt
+++ b/example/intrinsics/CMakeLists.txt
@@ -1,2 +1,3 @@
 ADD_EXAMPLE(sum)
 ADD_EXAMPLE(dot_product)
+ADD_EXAMPLE(matmul)
--- a/example/intrinsics/example_matmul.f90
+++ b/example/intrinsics/example_matmul.f90
@@ -1,0 +1,7 @@
+program example_matmul
+    use stdlib_intrinsics, only: stdlib_matmul
+    complex :: a(2,2)
+    a = reshape([(0, 0), (0, -1), (0, 1), (0, 0)], [2, 2]) ! pauli y-matrix
+
+    print *, stdlib_matmul(a, a, a, a, a) ! should be sigma_y
+end program example_matmul
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -19,6 +19,7 @@ set(fppFiles
     stdlib_hash_64bit_spookyv2.fypp
     stdlib_intrinsics_dot_product.fypp
     stdlib_intrinsics_sum.fypp
+    stdlib_intrinsics_matmul.fypp
     stdlib_intrinsics.fypp
     stdlib_io.fypp
     stdlib_io_npy.fypp
@@ -32,14 +33,14 @@ set(fppFiles
     stdlib_linalg_kronecker.fypp
     stdlib_linalg_cross_product.fypp
     stdlib_linalg_eigenvalues.fypp
-    stdlib_linalg_solve.fypp    
+    stdlib_linalg_solve.fypp
     stdlib_linalg_determinant.fypp
     stdlib_linalg_qr.fypp
     stdlib_linalg_inverse.fypp
     stdlib_linalg_pinv.fypp
     stdlib_linalg_norms.fypp
     stdlib_linalg_state.fypp
-    stdlib_linalg_svd.fypp 
+    stdlib_linalg_svd.fypp
     stdlib_linalg_cholesky.fypp
     stdlib_linalg_schur.fypp
     stdlib_optval.fypp
--- a/src/stdlib_intrinsics.fypp
+++ b/src/stdlib_intrinsics.fypp
@@ -157,9 +157,10 @@ module stdlib_intrinsics
         !!### Description
         !!
         !! matrix multiply more than two matrices with a single function call
-        !! the multiplication with the optimal bracketization is done automatically
+        !! the multiplication with the optimal parenthesization for efficiency of computation is done automatically
         !! Supported data types are `real`, `integer` and `complex`.
         !!
+        !! Note: The matrices must be of compatible shapes to be multiplied
         #:for k, t, s in I_KINDS_TYPES + R_KINDS_TYPES + C_KINDS_TYPES
             pure module function stdlib_matmul_${s}$_3 (a, b, c) result(d)
                 ${t}$, intent(in) :: a(:,:), b(:,:), c(:,:)
--- a/src/stdlib_intrinsics.fypp
+++ b/src/stdlib_intrinsics.fypp
@@ -146,6 +146,38 @@ module stdlib_intrinsics
         #:endfor
     end interface
     public :: kahan_kernel
+
+    interface stdlib_matmul
+        !! version: experimental
+        !!
+        !!### Summary
+        !! compute the matrix multiplication of more than two matrices with a single function call.
+        !! ([Specification](../page/specs/stdlib_intrinsics.html#stdlib_matmul))
+        !!
+        !!### Description
+        !!
+        !! matrix multiply more than two matrices with a single function call
+        !! the multiplication with the optimal bracketization is done automatically
+        !! Supported data types are `real`, `integer` and `complex`.
+        !!
+        #:for k, t, s in I_KINDS_TYPES + R_KINDS_TYPES + C_KINDS_TYPES
+            pure module function stdlib_matmul_${s}$_3 (a, b, c) result(d)
+                ${t}$, intent(in) :: a(:,:), b(:,:), c(:,:)
+                ${t}$, allocatable :: d(:,:)
+            end function stdlib_matmul_${s}$_3
+
+            pure module function stdlib_matmul_${s}$_4 (a, b, c, d) result(e)
+                ${t}$, intent(in) :: a(:,:), b(:,:), c(:,:), d(:,:)
+                ${t}$, allocatable :: e(:,:)
+            end function stdlib_matmul_${s}$_4
+
+            pure module function stdlib_matmul_${s}$_5 (a, b, c, d, e) result(f)
+                ${t}$, intent(in) :: a(:,:), b(:,:), c(:,:), d(:,:), e(:,:)
+                ${t}$, allocatable :: f(:,:)
+            end function stdlib_matmul_${s}$_5
+        #:endfor
+    end interface stdlib_matmul
+    public :: stdlib_matmul
     
 contains
 


src/stdlib_intrinsics_matmul.fypp
```diff
@@ -0,0 +1,133 @@
...
+
+        select case (s(1,5))
+            case (1)
+                f = matmul(a, stdlib_matmul(b, c, d, e))
+            case (2)
+                f = matmul(matmul(a, b), stdlib_matmul(c, d, e))
+            case (3)
+                f = matmul(stdlib_matmul(a, b ,c), matmul(d, e))
+            case (4)
+                f = matmul(stdlib_matmul(a, b, c, d), e)
```
wassup05, line 126: I just noticed this part is quite inefficient, because it calls to`stdlib_matmul` with 3, 4 arguments and that in turn computes the whole `s(i,j)` again which is a costly operation (`O(n^3)`)... Something cleaner is required here instead of the alternate 3 level nesting of select statements.

perazz: Thank you for this PR @wassup05. Wherever it is possible, I believe it would be useful to call the `gemm` backend for lapack. This would allow better performance when one can link against an external optimize BLAS/LAPACK library.

wassup05: Thank you @perazz for the review, I tried doing that and it seems like `gemm` is not defined for integer type arrays... Is that so? Or am I missing something? And also I noticed that trying to integrate with `gemm` introduces a lot of verboseness, Is that how it is?

jalvesz:
Here a few ideas for consideration:

1. Regarding the use of gemm or plain matmult: both gfortran and intel (ifort/ifx) enable replacing the intrinsic matmul by gemm using flags: worth reading https://stackoverflow.com/questions/31494176/will-fortrans-matmul-make-use-of-mkl-if-i-include-the-library and https://community.intel.com/t5/Intel-Fortran-Compiler/qopt-matmul-with-mkl-sequential/td-p/1003110 si it might not be that harmful after all to implement using matmul.
2. Regarding the signature of the key functions, you could consider:
```Fortran
pure function matmul_chain_order(p) result(s)
    integer, intent(in) :: p(:)
    integer :: s(1:size(p)-1, 2:size(p))
    integer :: m(1:size(p), 1:size(p))
    integer :: l, i, j, k, q, n
    n = size(p)
    ...
end function matmul_chain_order

pure module function stdlib_matmul (m1, m2, m3, m4, m5) result(e)
    real, intent(in) :: m1(:,:), m2(:,:)
    real, intent(in), optional :: m3(:,:), m4(:,:), m5(:,:) !> from the 3rd matrix they can be all optional
    real, allocatable :: e(:,:)
    integer :: p(5), i, num_present
    integer :: s(3,2:4)

    p(1) = size(m1, 1)
    p(2) = size(m2, 1)
    num_present = 2
    if(present(m3)) then
        p(3) = size(m3, 1)
        num_present = num_present + 1
    end if
    if(present(m4)) then
        p(4) = size(m4, 1)
        num_present = num_present + 1
    end if
    if(present(m5)) then
        p(5) = size(m5, 2)
        num_present = num_present + 1
    end if

    s = matmul_chain_order(p(1:num_present))

    ... 
end function stdlib_matmul
```
For the procedure computing the orders you only need the `p` array, the dimension can be obtained. One might argue that this induces extra time. I think it will most likely be transparent.

For the main procedure, you could consider using optional arguments from the 3rd matrix onwards and then manage the rest within the same procedure without recursion.

Just some food-for-thought.

wassup05:
Thank you very much @jalvesz for your detailed remarks, they are quite helpful.
I have refactored the algorithm to only accept `p`, as `n` can very well be calculated from that.

Regarding the signature I am a bit confused, because if we don't use recursion we would have to handle 14 cases (4th catalan number) if the number of matrices is 5, which would become quite messy quickly. For 3 matrices it is much efficient to compare the two ordering costs and dispatch an ordering appropriately (numpy claims that it is 15 more efficient than calculating a `s` matrix for it), 4 matrices case is also somewhat fine considering there are only 5 cases, But I am not sure about how to handle the 5 matrix one without handling all the cases explicitly.. (Or maybe we can just limit this to 4 matrices?)

And yes I will add some non-trivial examples for sure.

Looking forward to hearing your thoughts.

jalvesz:
Indeed! another idea:

For 3 and 4 matrices make them internal procedures including passing as extra argument the ordering slice corresponding to those matrices. For the one public procedure with 2+3optionals, you compute just once the ordering and call the internal versions for 3 and 4 depending on the case.

wassup05:
Thank you @jalvesz, I have implemented them accordingly and added some better examples although I am not sure of how I may test the correctness of multiplication with large matrices or compare the time taken by native `matmul` vs this..

jalvesz:
In terms of correctness I would say that the simplest approach would be to write down a couple of analytical cases for which the exact solution is fixed, also a couple of cases with random matrices for which the optimal ordering is different from the trivial one and compare: stdlib_matlul vs trivial sequencial calls to intrinsic matmal. For integer arrays the error should be zero, for real/complex arrays the error tolerance should be somewhere around `100*epsilon(0._wp)`. The test could be done using the Frobenius norm `error = mnorm(A-A_ref)`, where `A` would result from the current proposal and `A_ref` from the intrinsic matmul using the naive sequence.

In terms of performance, It might not be necessary to add that to the PR but, a separate small program showcasing two scenarios might be useful:
1. Many multiple calls for `stdlib_matmul` vs naive matmul for small matrices
2. Few multiple calls for `stdlib_matmul` vs naive matmul for medium/large matrices

@loiseaujc maybe you have some use cases worth testing?

loiseaujc:
Oh boy ! I had not seen this PR. Super happy with it. I'll take some time this afternoon to look at the code and give you some feedback. As for the test cases, anything involving multiplications with a low-rank matrix already factored would do. As far as I'm concerned, these are the typical cases I encounter where such an interface would prove very useful syntactically. I'll see as I get to the lab if I have a sufficiently simple yet somewhat realistic set of data we could use for testing the performances.

src/stdlib_intrinsics_matmul.fypp
perazz: Yes, I am not sure if such API would have wide usage for `integer` kinds. So as far as I'm concerned, it could deal with real and complex matrices only. Regarding the `gemm` backend, yes I believe it is important because it is a staple of the whole linear algebra library: that it provides a clean, high-level API, that the user can build with external high-performance BLAS/LAPACK libraryes.

wassup05:
I have replaced all the `matmul`'s by calls to `gemm`... The code has become a bit complicated and verbose... If I have missed anything or there was a shorter way of doing the same, do let me know please.
And also now the interface handles only `real` and `complex` values.

@jalvesz @perazz

src/stdlib_intrinsics_matmul.fypp
```diff
@@ -0,0 +1,226 @@
...
+        ord = s(start, start + 2)
+        allocate(r(p(start), p(start + 3)))
+
+        if (ord == start) then
+            ! m1*(m2*m3)
+            m = p(start + 1)
+            n = p(start + 3)
+            k = p(start + 2)
+            allocate(temp(m,n))
+            call gemm('N', 'N', m, n, k, one_${s}$, m2, m, m3, k, zero_${s}$, temp, m)
```
perazz, line 57: Very good progress @wassup05, thank you! Imho this PR is almost ready to be merged. As you suggest, it would be good to have a nice wrapper for `gemm`. It has been discussed before. I would like to suggest that all calls to `gemm` are also wrapped into a `stdlib_matmul` function - now with two matrices only. This would give stdlib fully functional matmul functionality. 

Here I suggest two possible APIs, and I will ask @jalvesz @jvdp1 @loiseaujc to discuss that together: 

The first would be similar to `gemm`
```Fortran
! API Similar to gemm
${t1}$ function stdlib_matmul(A, Astate, B, Bstate) result(C)
  ${t1}$, intent(in) :: A(:,:), B(:,:)
  character, intent(in), optional :: Astate, Bstate
```

and could use the matrix state definitions already in use for the sparse operations
https://github.com/fortran-lang/stdlib/blob/5c64ee65fcaf01f710bd1caf587700117fe1fc2a/src/stdlib_sparse_constants.fypp#L16-L18

The second would be more ambitious and essentially zero-overhead, it would wrap the operation in a derived type: (to be templated of course)
```Fortran
type :: transposed_rdp
   real(dp), pointer :: A(:,:) => null()
   character(1) :: Astate = 'N'
end type transposed_rdp

interface transposed
   module procedure transposed_new_rdp
   ...
end interface transposed    

type(transposed_rdp) function transposed_new_rdp(A) result(AT)
    real(dp), intent(inout), target :: A(:,:)
    AT%A => A
    AT%Astate = 'T'
end function
```

Then we could define a templated base interface
```
! Work with normal matrices
${t1}$ function stdlib_matmul(A, B) result(C)
  ${t1}$, intent(in) :: A(:,:), B(:,:)

! Work with transposed/hermitian swaps
${rt}$ function stdlib_matmul(A, B) result(C)
  templated_${rn}$, intent(in) :: A, B
```  

So the user writing code would have it clear: 

```
C = strlib_matmul(A, B)
C = stdlib_matmul(transposed(A), B)
C = stlib_matmul(A, hermitian(C))
```

without it triggering any actual data movement.

perazz (commented):
Very good progress @wassup05.

src/stdlib_intrinsics_matmul.fypp
loiseaujc:
```
C = stdlib_matmul(transposed(A), B)
C = stlib_matmul(A, hermitian(C))
```

How would that work in the written code? Would `A` and `B` have to be declared as `type(matrix_state_type)`?
If so, I'm a bit afraid most people would prefer the more "natural" `real, dimension(m, n) :: A, B` and play around either directly with the intrinsic `transpose` or the classical `gemm` dummy arguments.

src/stdlib_intrinsics_matmul.fypp
```diff
@@ -0,0 +1,226 @@
...
+        integer :: p(6), num_present, m, n, k
+        integer, allocatable :: s(:,:)
+
+        p(1) = size(m1, 1)
+        p(2) = size(m2, 1)
+        p(3) = size(m2, 2)
+
+        num_present = 2
+        if (present(m3)) then
+            p(3) = size(m3, 1)
```
loiseaujc, line 138: Before overwritting `p(3)`, I think adding a check returning an informative error if the number of columns of `m2` and the number of rows of `m3` are inconsistent might be a good idea. Sure enough, the inconsistency will be catched later on anyway (by `gemm` I guess), but by then users might not directly realize which of their matrices ain't quite right. The same comment would apply for `p(4)` and `p(5)`. I think, in general, "dumb-proofing" things is a good idea.

src/stdlib_intrinsics_matmul.fypp
jalvesz: >How would that work in the written code? Would A and B have to be declared as type(matrix_state_type)?

I also had the same reaction at first, after thinking about it longer I saw that it is actually a pretty clever solution, the user would declare the matrices as regular dense matrices. It is the internal interface that would make the distinction. This would imply though implementing internally several versions to account for the combinations (dense,dense) / (dense,type) / (type,dense). This looks interesting but I wonder if it should be pursued at this stage. The first proposal by @perazz looks easier and totally valid but I would propose it with a slight modification:
```Fortran
${t1}$ function stdlib_matmul(A, B, op_a, op_b) result(C)
  ${t1}$, intent(in) :: A(:,:), B(:,:)
  character, intent(in), optional :: op_a, op_b
```
to let all optional arguments of a procedure at the end of the signature.

perazz: Yes @jalvesz @loiseaujc, I don't know how to write it better, but it is outlined in the previous post. 
There would be an interface for `matmul(A,B)` (resolved at compile time) with 4 options for each kind: 
1. both A, B are simple matrices (2d arrays)
2. A is a `matrix_state_`, B is a 2D array
3. B is a `matrix_state_`, A is a 2d array
4. Both are `matrix_state_`*

Only function 4. is actually implemented, and is a `gemm` wrapper: the other implementations just wrap against it with `fypp`.

src/stdlib_intrinsics_matmul.fypp
perazz: > informative error

yes, using a `linalg_state_type` variable for error handling is recommended imho.

perazz: The approach we used for other `linalg` routines is to have an interface: 
- version with no output error state: -> it is `pure`, triggers `error stop` on issues
- version with output error state: -> not `pure`, returns `intent(out)` error state. 

This way, the function is resolved at compile time. However, you have to write one more interface

wassup05: I have implemented error handling as pointed out by both of you, let me know what you think of it

wassup05:
I thought CI failing on windows was just pure chance, but I don't think so now.. most probably has something to do with random numbers... I am not sure what exactly tho

src/stdlib_intrinsics_matmul.fypp
wassup05: I like the first proposal quite a lot too! the 2nd proposal seems convenient too with the `.t.`, `.h.` operators but maybe the initialization of arrays as `type(matrix_state_*)` would put some people off..? Maybe we should go ahead with the 1st proposal for now?

src/stdlib_intrinsics_matmul.fypp
```diff
@@ -0,0 +1,281 @@
...
+            call linalg_error_handling(err0, err)
+            allocate(res(0, 0))
+            return
+        end if
+
+        num_present = 2
+        if (present(m3)) then
+
+            if (size(m3, 1) /= p(3)) then
+                err0 = linalg_state_type(this, LINALG_VALUE_ERROR, 'matrices m2, m3 not of compatible sizes')
```
perazz, line 154:
```suggestion
                err0 = linalg_state_type(this, LINALG_VALUE_ERROR, 'matrices m2=',shape(m2),', m3=',shape(m3),'have incompatible sizes')
```

src/stdlib_intrinsics_matmul.fypp
```diff
@@ -0,0 +1,281 @@
...
+        integer, allocatable :: s(:,:)
+
+        type(linalg_state_type) :: err0
+
+        p(1) = size(m1, 1)
+        p(2) = size(m2, 1)
+        p(3) = size(m2, 2)
+
+        if (size(m1, 2) /= p(2)) then
+            err0 = linalg_state_type(this, LINALG_VALUE_ERROR, 'matrices m1, m2 not of compatible sizes')
```
perazz, line 144:
```suggestion
            err0 = linalg_state_type(this, LINALG_VALUE_ERROR, 'matrices m1=',shape(m1),', m2=',shape(m2),'have incompatible sizes')
```

src/stdlib_intrinsics_matmul.fypp
```diff
@@ -0,0 +1,281 @@
...
+            end if
+
+            p(3) = size(m3, 1)
+            p(4) = size(m3, 2)
+            num_present = num_present + 1
+        end if
+        if (present(m4)) then
+
+            if (size(m4, 1) /= p(4)) then
+                err0 = linalg_state_type(this, LINALG_VALUE_ERROR, 'matrices m3, m4 not of compatible sizes')
```
perazz, line 167:
```suggestion
                err0 = linalg_state_type(this, LINALG_VALUE_ERROR, 'matrices m3=',shape(m3),', m4=',shape(m4),' have incompatible sizes')
```

src/stdlib_intrinsics_matmul.fypp
```diff
@@ -0,0 +1,281 @@
...
+            end if
+
+            p(4) = size(m4, 1)
+            p(5) = size(m4, 2)
+            num_present = num_present + 1
+        end if
+        if (present(m5)) then
+
+            if (size(m5, 1) /= p(5)) then
+                err0 = linalg_state_type(this, LINALG_VALUE_ERROR, 'matrices m4, m5 not of compatible sizes')
```
perazz, line 180:
```suggestion
                err0 = linalg_state_type(this, LINALG_VALUE_ERROR, 'matrices m4=',shape(m4),', m5=',shape(m5),' have incompatible sizes')
```

src/stdlib_intrinsics_matmul.fypp
```diff
@@ -0,0 +1,281 @@
...
+                call gemm('N', 'N', m, n, k, one_${s}$, temp, m, temp1, k, zero_${s}$, res, m)
+            case (4)
+                ! (m1*m2*m3*m4)*m5
+                temp = matmul_chain_mult_${s}$_4(m1, m2, m3, m4, 1, s, p)
+                m = p(1)
+                n = p(6)
+                k = p(5)
+                call gemm('N', 'N', m, n, k, one_${s}$, temp, m, m5, k, zero_${s}$, res, m)
+            case default
+                error stop "stdlib_matmul: internal error: unexpected s(i,j)"
```
perazz, line 258:
```suggestion
                err0 = linalg_state_type(this,LINALG_INTERNAL_ERROR,"internal error: unexpected s(i,j)")
                call linalg_error_handling(err0,err)
```

src/stdlib_intrinsics_matmul.fypp
```diff
@@ -0,0 +1,281 @@
+#:include "common.fypp"
+#:set I_KINDS_TYPES = list(zip(INT_KINDS, INT_TYPES, INT_KINDS))
+#:set R_KINDS_TYPES = list(zip(REAL_KINDS, REAL_TYPES, REAL_SUFFIX))
+#:set C_KINDS_TYPES = list(zip(CMPLX_KINDS, CMPLX_TYPES, CMPLX_SUFFIX))
+
+submodule (stdlib_intrinsics) stdlib_intrinsics_matmul
+    use stdlib_linalg_blas, only: gemm
+    use stdlib_linalg_state, only: linalg_state_type, linalg_error_handling, LINALG_VALUE_ERROR
```
perazz, line 8:
```suggestion
    use stdlib_linalg_state, only: linalg_state_type, linalg_error_handling, LINALG_VALUE_ERROR, LINALG_INTERNAL_ERROR
```

perazz (commented):
@wassup05 the state type introduction LGTM. I have polished the error messages a bit. Are you still planning to include the `gemm` wrapper as part of this PR? Otherwise, we could defer that to a separate PR imho.

jalvesz:
@perazz I think the following error:
```
 1827 |                 err0 = linalg_state_type(this, LINALG_VALUE_ERROR, 'matrices m4=',shape(m4),', m5=',shape(m5),' have incompatible sizes')
      |                                                                                                                1
Error: Unterminated character constant beginning at (1)
```
is related with the line limit by default of 132 characters. This line has 146 characters. Maybe if you split it with a line continuation it should pass.

src/stdlib_intrinsics_matmul.fypp
```diff
@@ -0,0 +1,282 @@
...
+        integer, allocatable :: s(:,:)
+
+        type(linalg_state_type) :: err0
+
+        p(1) = size(m1, 1)
+        p(2) = size(m2, 1)
+        p(3) = size(m2, 2)
+
+        if (size(m1, 2) /= p(2)) then
+            err0 = linalg_state_type(this, LINALG_VALUE_ERROR, 'matrices m1=',shape(m1),', m2=',shape(m2),'have incompatible sizes')
```
perazz, line 144:
```suggestion
            err0 = linalg_state_type(this, LINALG_VALUE_ERROR, 'matrices m1=',shape(m1),&
                   ', m2=',shape(m2),'have incompatible sizes')
```

src/stdlib_intrinsics_matmul.fypp
```diff
@@ -0,0 +1,282 @@
...
+            call linalg_error_handling(err0, err)
+            allocate(res(0, 0))
+            return
+        end if
+
+        num_present = 2
+        if (present(m3)) then
+
+            if (size(m3, 1) /= p(3)) then
+                err0 = linalg_state_type(this, LINALG_VALUE_ERROR, 'matrices m2=',shape(m2),', m3=',shape(m3),'have incompatible sizes')
```
perazz, line 154:
```suggestion
                err0 = linalg_state_type(this, LINALG_VALUE_ERROR, 'matrices m2=',shape(m2), &
                       ', m3=',shape(m3),'have incompatible sizes')
```

src/stdlib_intrinsics_matmul.fypp
```diff
@@ -0,0 +1,282 @@
...
+            end if
+
+            p(3) = size(m3, 1)
+            p(4) = size(m3, 2)
+            num_present = num_present + 1
+        end if
+        if (present(m4)) then
+
+            if (size(m4, 1) /= p(4)) then
+                err0 = linalg_state_type(this, LINALG_VALUE_ERROR, 'matrices m3=',shape(m3),', m4=',shape(m4),' have incompatible sizes')
```
perazz, line 167:
```suggestion
                err0 = linalg_state_type(this, LINALG_VALUE_ERROR, 'matrices m3=',shape(m3), &
                       ', m4=',shape(m4),' have incompatible sizes')
```

src/stdlib_intrinsics_matmul.fypp
```diff
@@ -0,0 +1,282 @@
...
+            end if
+
+            p(4) = size(m4, 1)
+            p(5) = size(m4, 2)
+            num_present = num_present + 1
+        end if
+        if (present(m5)) then
+
+            if (size(m5, 1) /= p(5)) then
+                err0 = linalg_state_type(this, LINALG_VALUE_ERROR, 'matrices m4=',shape(m4),', m5=',shape(m5),' have incompatible sizes')
```
perazz, line 180:
```suggestion
                err0 = linalg_state_type(this, LINALG_VALUE_ERROR, 'matrices m4=',shape(m4), &
                       ', m5=',shape(m5),' have incompatible sizes')
```

wassup05:
yes @perazz defering that to another PR seems a better option to me.

So, some final touches to this PR (I think) would be for me to add
- Some tests
- an example
- specs

perazz:
@wassup05 please also note that the example program seems to be failing on Windows in the CI.

wassup05:
@jalvesz @perazz I have added the specs, example and some tests. I did not add any 5 argument test cases because as you can see the error tolerance should be increased quite a lot since the error is increasing as args are increasing... For 3 args it needed `epsilon*300` and for 4 `epsilon*1500` and even with that some tests have failed in the `CI`

wassup05:
I tried that @jalvesz, I think the tolerance went even higher than the individual checking did, for 3 args it took around `*800` and for 4 args it is `*2500` and ongoing... Maybe I should reduce the `size` of the matrix

test/intrinsics/test_intrinsics.fypp
```diff
@@ -249,6 +251,45 @@ subroutine test_dot_product(error)
...
+    block
+        ${t}$ :: x(10,20), y(20,30), z(30,10), r(10,10), r1(10,10)
+        call random_number(x)
+        call random_number(y)
+        call random_number(z)
+
+        r = stdlib_matmul(x, y, z) ! the optimal ordering would be (x(yz))
+        r1 = matmul(matmul(x, y), z) ! the opposite order to induce a difference
+
+        call check(error, all(abs(r-r1) <= epsilon(0._${k}$) * 300), "real, ${k}$, 3 args: error too large")
```
loiseaujc: I'm probably late for this party but isn't `epsilon(0.0_wp)` like the smallest number that can be represented in the working precision `wp`? If it is indeed the case, then  I wouldn't be surprised to see the tests fail.

I guess, a better indicator would be to consider the matrix 2-norm (i.e. the largest singular value of $A - \tilde{A}$ where $A$ is the matrix you'd compute with a chain of `matmul` and $\tilde{A}$ the one from your implementation).

jalvesz: >but isn't epsilon(0.0_wp) like the smallest number that can be represented in the working precision wp

That would be `tiny(0.0)` which is indeed a very tiny number. `epsilon()` represents the smallest significant difference between two consecutive real numbers in the given precision.

perazz:
In absence of further comments and activity, this can be merged soon imho.

jalvesz (approved):

loiseaujc:
Should we merge it now or is there anything else you guys would like to improve?

Issue #948, opened 2025-03-08
Suggest that `loadtxt` supports reading CSV files of pure floating-point type, that is, add a argument `delimiter` to `savetxt` and `loadtxt` interfaces.


zoziha:
### Motivation

Numpy's `loadtxt` supports reading CSV files with pure floating-point values, through `data = np.loadtxt('example.csv', delimiter=',')`. In Fortran's default data reading rules, data separated by commas or spaces is very natural. The `loadtxt` in stdlib supporting the argument `delimiter=','` will be a practical function. This is when I use the `loadtxt` in stdlib to read CSV files and find that only the first column of CSV data can be read.

In this case, `savetxt` also needs to support the `delimiter` argument.

### Prior Art

- [`numpy.loadtxt`](https://numpy.org/doc/stable/reference/generated/numpy.loadtxt.html);
- [`numpy.savetxt`](https://numpy.org/doc/stable/reference/generated/numpy.savetxt.html).

Issue #532, opened 2021-09-19
[Help] Maybe a bug that will affect `stdlib_io:open`: open files with `access=stream` in gfortran.


zoziha:
### Description

The following example can get our expected results on `ifort`, **but in the case of gfortran with `access=stream`, for the characters in the output example (`A10>A2`), unexpected encoding will occur!** 
This may affect the use of the `open` function in `stdlib_io` with gfortran, which uses `access=stream`.

#### Demo: Characters output

```fortran
program main

    open (2, file="2.txt", access="stream", form="formatted")
    write(2, "(2A10)") "12","34"
    close(2)

    open (3, file="3.txt", access="sequential", form="formatted")
    write(3, "(2A10)") "12","34"
    close(3)

end program main
```

#### Run results on my laptop (windows10, msys2 ucrt64-gfortran 10.3)
**2.txt**
```sh
        12343.txts        343.txtseq

```
**3.txt**
```sh
        12        34

```

### Expected Behaviour

In the above example, I think `access` should output consistent results in both `sequential` and `stream` cases. `ifort` is like this, gfortran unexpectedly garbled when `access=stream`.

### Version of stdlib

8bfcdf9fdf8d3d897eee76512ba5083ecab29f96

### Platform and Architecture

Windows 10, msys2 ucrt64-gfortran 10.3

### Additional Information

I think this should be a bug. If so, do we need to report to gfortran?

awvwgk:
See https://gcc.gnu.org/bugs/ for details on reporting bugs to GCC. You can create an account at https://gcc.gnu.org/bugzilla/ to file the actual bugreport or reach out to the developers on the mailing list: https://gcc.gnu.org/mailman/listinfo/fortran.

PR #958, opened 2025-03-20
Support csv file reading and writing in loadtxt and savetxt.

zoziha:
### Description

This PR will add a `delimiter` argument to `loadtxt` and `savetxt` functions for issue #948 to support potential CSV data file support.

Additionally, it corrects the title errors of `stdlib_math` and `stdlib_sparse` docs.


src/stdlib_io.fypp
```diff
@@ -236,17 +240,27 @@ contains
       !! call savetxt("log.txt", data)
       !!```
       !!
-
+
```
jvdp1, line 243:
```suggestion

```

src/stdlib_io.fypp
```diff
@@ -236,17 +240,27 @@ contains
...
       !!
-
+      
       integer :: s, i, ios
+      character(len=1) :: delimiter_
+      character(len=3) :: delim_str
+      character(len=:), allocatable :: fmt_
       character(len=1024) :: iomsg, msgout
+
+      delimiter_ = optval(delimiter, " ")
```
jvdp1, line 250: To avoid any potential issues in the future, I suggest to introduce the parameter `delimiter_default`
```fortran
character(len=1), parameter :: delimiter_default = " "
```

```suggestion
      delimiter_ = optval(delimiter, delimiter_default)
```

src/stdlib_io.fypp
```diff
@@ -266,19 +280,22 @@ contains
...
 
     integer :: ios, skiprows_, i
     character :: c
     character(len=:), allocatable :: line
-    logical :: lastblank
+    character(len=1) :: delimiter_
+    logical :: last_delim
 
     skiprows_ = optval(skiprows, 0)
+    delimiter_ = optval(delimiter, " ")
```
jvdp1, line 298:
```suggestion
    delimiter_ = optval(delimiter, delimiter_default)
```

src/stdlib_io.fypp
```diff
@@ -291,12 +308,23 @@ contains
...
     if (ios/=0 .or. .not.allocated(line)) return
 
-    lastblank = .true.
-    do i = 1,len(line)
-      c = line(i:i)
-      if (lastblank .and. .not. is_blank(c)) number_of_columns = number_of_columns + 1
-      lastblank = is_blank(c)
-    end do
+    last_delim = .true.
+    if (delimiter_ == " ") then
```
jvdp1, line 312:
```suggestion
    if (delimiter_ == delimiter_default) then
```

src/stdlib_io.fypp
```diff
@@ -147,6 +149,7 @@ contains
 
       skiprows_ = max(optval(skiprows, 0), 0)
       max_rows_ = optval(max_rows, -1)
+      delimiter_ = optval(delimiter, " ")
```
jvdp1, line 152:
```suggestion
      delimiter_ = optval(delimiter, delimiter_default)
```

jvdp1 (approved):
LGTM. Thank you @zoziha

src/stdlib_io.fypp
zoziha: @jvdp1 , thanks for your review. The `delimiter_default` constant has been added according to the feedback.

src/stdlib_io.fypp
```diff
@@ -141,12 +143,14 @@ contains
       !!     11 12 13
       !!     ...
       !!
+      character(len=1), parameter :: delimiter_default = " "
```
perazz, line 146: To remove duplication, the `delimiter_default` parameter could be moved to the module declarations section.

src/stdlib_io.fypp
```diff
@@ -157,9 +161,9 @@ contains
 
       ! determine number of columns
       ncol = 0
-      if ( skiprows_ < nrow ) ncol = number_of_columns(s, skiprows=skiprows_)
+      if ( skiprows_ < nrow ) ncol = number_of_columns(s, skiprows=skiprows_, delimiter=delimiter_)
       #:if 'complex' in t1
-      ncol = ncol / 2
+      if (is_blank(delimiter_)) ncol = ncol / 2
```
perazz, line 166: Why does the complex case only need to half the number of columns in case of a space delimiter? Let's assume delimiter = `;`, what will happen then? It may be a good idea to add a test case with a different delimiter i.e. `;` or `-`.

perazz (commented):
Nice work @zoziha, I also think that this is about ready to merge. I only have a couple comments.

src/stdlib_io.fypp
jalvesz: Second this

src/stdlib_io.fypp
zoziha: > Why does the complex case only need to half the number of columns in case of a space delimiter

This is to be compatible with the logic of the previous loadtxt function. Since a complex number has a real part and an imaginary part, it has been deliberately divided by 2. `nrow` and `ncol` are used for the allocation of the allocatable array.

> Let's assume delimiter = `;`, what will happen then? It may be a good idea to add a test case with a different delimiter i.e. `;` or `-`.

Thanks for your reminder. I added similar tests and indeed found an oversight. Now it has been resolved.

PS. `-` may not be a good delimiter for separating data and is only tested in `test_int32`.

zoziha:
@jvdp1, @perazz Thanks for your review.

There is a bug that made me dig deep into the reason and it took a long time. I once mentioned in #532 that the combination of `form="formatted"` and `access="stream"` in the `stdlib_io.open` function will lead to a bug in `gfortran`. But this time I also encountered this problem in `ifort`. You can refer to the [CI failure](https://github.com/fortran-lang/stdlib/actions/runs/13999185134/job/39201532261?pr=958). For safety, I think `form="formatted"` should better be combined with `access="sequential"`.

Therefore, I have changed [this internal detail to the `open` function](https://github.com/fortran-lang/stdlib/pull/958/commits/b80448ce0001b9a5b496d48ca6851c302ca82aad).

perazz (approved):
LGTM, thank you @zoziha.

jalvesz:
Given that this PR enables reading .csv files by default I wonder if the separator symbol should also be set as a function of the file extension ?

Something like:
```
if( .not.present(delimiter)) then
   i = scan(filename,'.',back=.true.)
   if( filename(i+1:i+4) .eq. 'csv' ) delimiter_ = ','
end if
```

zoziha:
@jalvesz Thanks for your proposal. Although your proposal has improved the ease of use of loadtxt and savetxt for reading and writing CSV files, I think explicit rules are better than implicit rules. Explicitly passing the delimiter `','` will be easier to make these interfaces keep simple.

jalvesz (approved):

jalvesz:
Thanks for this nice improvement @zoziha, with 3 approvals I'll merge this PR

Issue #940, opened 2025-02-24
Implement a `unique` function returning only the unique values in a vector.


loiseaujc:
### Motivation

Recently, I've run into the problem of extracting unique values in a vector (of any integer, real or complex type or possibly even character). Consider for instance the following vector `x = [1, 2, 3, 3, 4]`. What I'd need is a function taking `x` as input and returning the vector `y = [1, 2, 3, 4]` as output. The interface for a real-valued vector could be as simple as

```fortran
pure function unique(x, sorted) result(y)
     real(dp), intent(in) :: x(:)
     !! Array whose unique values need to be extracted.
     logical(lk), optional, intent(in) :: sorted
     !! Whether the output vector needs to be sorted or not (default .false. ?)
     real(dp), allocatable :: y(:)
     !! Vector containing only the unique values from x.
end function
```

The output vector could be sorted or not, depending on the user's choice. I know that there are no Fortran intrinsic functions for that purpose, but I ain't sure something like that is already available in `stdlib`. If I'm wrong, could anyone point me to the correct function?

### Prior Art

- In Matlab, there is the `unique` function whose description is available [here](https://www.mathworks.com/help/matlab/ref/double.unique.html).
- Python has the `set` function taking as input a `list` and returning only the unique elements of this list.
- Numpy has `np.unique` whose description is available [here](https://numpy.org/doc/stable/reference/generated/numpy.unique.html).
- @jacobwilliams provides an integer-based implementation on his blog ([here](https://numpy.org/doc/stable/reference/generated/numpy.unique.html)).

### Additional Information

Both [Matlab](https://www.mathworks.com/help/matlab/ref/double.unique.html) and [Numpy](https://numpy.org/doc/stable/reference/generated/numpy.unique.html)'s implementations cover a relatively large set of cases (1D-array, multidimensional arrays, different types, etc) and return values (the unique elements, the corresponding indices, indices to the reconstruct the original array from this unique set, etc).

I don't know if absolutely all these cases need to be covered (at least as a starting point). I would probably recommend to start with the simplest ones (i.e. only input vectors and output vector with the unique elements) as these are probably the most common situations where a `unique` function might be needed. That would include `integer`, `real`, `complex` and `character` 1D-arrays.

loiseaujc:
I'm not sure either into which module this utility function should be included. Maybe `stdlib_sorting`?

loiseaujc:
Oh sure! I completely overlooked this issue.

[...]

PR #965, opened 2025-03-25
Implement a unique function returning only the unique values in a vector. #940

demoncoder-crypto:
Please let me know if this is the right approach


perazz:
Linked to #940

src/stdlib_sorting_unique_impl.fypp
```diff
@@ -0,0 +1,72 @@
...
+        
+        want_sorted = optval(sorted, .false.)
+        
+        n = size(array)
+        if (n == 0) then
+            allocate(unique_values(0))
+            return
+        endif
+        
+        ! Create a temporary copy that may be sorted
```
loiseaujc, line 39: Just a matter of style here, but you could be more concise with:

```fortran
 temp_array = array ; if (want_sorted) call sort(temp_array)
```

src/stdlib_sorting_unique_impl.fypp
```diff
@@ -0,0 +1,72 @@
...
+            call sort(temp_array)
+        else
+            temp_array = array
+        endif
+        
+        ! Find unique elements using a mask
+        ! Start with first element always marked as unique
+        mask(1) = .true.
+        
+        ! Compare each element with previous to mark duplicates
```
loiseaujc, line 51: The left-hand sign assignment could be in any order. You could use

```fortran
do concurrent (i=2:n)
    mask(i) = temp_array(i) /= temp_array(i-1)
enddo
```

I believe. If multithreading is being used, my understanding is that it'll enable the compiler to better optimize and speed-up the calculation (notably for very large arrays).

loiseaujc: Another possibility using array syntax is simply `mask(:n-1) = temp_array(:n-1) == temp_array(2:)`. Not sure what would be the fastest here.

src/stdlib_sorting_unique_impl.fypp
```diff
@@ -0,0 +1,72 @@
...
+        do i = 2, n
+            mask(i) = temp_array(i) /= temp_array(i-1)
+        end do
+        
+        ! Count unique elements and allocate result array
+        unique_count = count(mask)
+        allocate(unique_values(unique_count))
+        
+        ! Extract unique elements to result array
+        j = 0
```
loiseaujc, line 61: Your `do` loop can be replaced by a one-liner: `unique_values = pack(temp_array, mask)`.

src/stdlib_sorting_unique_impl.fypp
```diff
@@ -0,0 +1,72 @@
...
+        
+        ${t2}$ :: temp_array(size(array))
+        logical :: mask(size(array))
+        integer :: i, j, n, unique_count
+        logical :: want_sorted
+        
+        want_sorted = optval(sorted, .false.)
+        
+        n = size(array)
+        if (n == 0) then
```
loiseaujc, line 34: You can also treat the case `n == 1` here. If `n == 1`, there is only a single value to return.

jalvesz:
Thanks @demoncoder-crypto for this PR. Before reviewing it, could you please add a short summary of the key points regarding the What(s), Why(s) and/or How(s) of the PR in your first comment? It is very helpful for the reviewing process to have summary of what is being proposed. You can edit your first comment following the three dots to the right.

doc/specs/stdlib_sorting_unique.md
```diff
@@ -0,0 +1,176 @@
...
+```fortran
+! Get unique values from an integer array
+integer :: x(5) = [1, 2, 3, 3, 4]
+integer, allocatable :: y(:)
+y = unique(x)    ! y will be [1, 2, 3, 4]
+
+! Get sorted unique values from a real array
+real :: a(8) = [3.1, 2.5, 7.2, 3.1, 2.5, 8.0, 7.2, 9.5]
+real, allocatable :: b(:)
+b = unique(a, sorted=.true.)  ! b will be [2.5, 3.1, 7.2, 8.0, 9.5]
```
loiseaujc, line 38: To me, having the syntax

```
b = unique(a, sorted=.true.)
```

would imply that on entry `a` is already sorted such that some internal logic can be skipped. Clearly, `a` is not sorted in this example and so I find the syntax a bit counter-intuitive. Maybe you need change this internally.

demoncoder-crypto:
@jalvesz for sure going ahead i will put summary of everything to make sure to reduce the load for reviews

demoncoder-crypto:
Yes I am currently working on this issue. I will update you shortly on this. I am understanding every comment made here so it might take some time but I am really grateful for all the responsiveness and guidance. Thanks means a lot

perazz:
> 3\. You should not allocate an array to size 0, this is an error.

I agree on the excellent reviews, please keep going! The only objection is on this point: function results with the `allocatable` attribute MUST always return an allocated value, per [Fortran standard 15.6.2.2](https://j3-fortran.org/doc/year/24/24-007.pdf) "If the function result is not a pointer, its value shall be defined by the function.". It is necessary, (though not sufficient), for an allocatable function result (any object) to be allocated to be defined. [see also](https://stackoverflow.com/questions/48155025/fortran-function-returning-unallocated-array-causes-segmentation-fault) where I also learned this the hard way.

So when empty, we should return an empy array i.e. `allocate(array(0))`

src/stdlib_sorting_unique.fypp
```diff
@@ -0,0 +1,133 @@
...
+!!
+!!   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+!!   OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+!!   MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+!!   IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+!!   CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+!!   TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+!!   SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+module stdlib_sorting_unique
```
jvdp1, line 42: Would this be a submodule of `stdlib_sort` more appropriate, as it is suggested to access it through `stdlib_sorting` in the examples?

src/stdlib_sorting_unique_impl.fypp
```diff
@@ -0,0 +1,72 @@
...
+        mask(1) = .true.
+        
+        ! Compare each element with previous to mark duplicates
+        do i = 2, n
+            mask(i) = temp_array(i) /= temp_array(i-1)
+        end do
+        
+        ! Count unique elements and allocate result array
+        unique_count = count(mask)
+        allocate(unique_values(unique_count))
```
jvdp1, line 58: These two lines could be combined into a single one.

src/stdlib_sorting_unique_impl.fypp
jvdp1: Actually, the lines 56-67 could be replaced as follows:
```fortran
        allocate(unique_values, source = pack(temp_array, mask)
```

or

```fortran
        unique_values = pack(temp_array, mask)
```

jvdp1 (commented):
Thank you @demoncoder-crypto for this PR. It will be quite useful to many users IMO.
Here are some suggestions regarding the API, code and its integration in the `stdlib` structure

jalvesz:
@jvdp1 maybe you could check out the modified version proposed here https://github.com/fortran-lang/stdlib/pull/965#issuecomment-2763345081 regarding the internal implementation.

jvdp1:
> @jvdp1 maybe you could check out the modified version proposed here [#965 (comment)](https://github.com/fortran-lang/stdlib/pull/965#issuecomment-2763345081) regarding the internal implementation.

Sorry, I started my review early today and missed your comment. I am puzzled with the aim of `sorted` (is it for the input (as in the code) or for the output (as in the specs)?). Otherwise, your code LGTM. It probably avoids a temporary array generated by `pack`. It could include the case `n == 1` in addition to `n==0`.
Also, ` allocate(temp(n), source = array)` could be written as `allocate(temp, source = array)`.

jalvesz:
>I am puzzled with the aim of sorted (is it for the input (as in the code) or for the output

I think it only makes sense for the input as in "is the input sorted or not?" Because in any case the output will be sorted. This flag allows controlling whether to apply the sorting or not to the temporary array which only makes sense to skip if one knows the input is already ordered but that contains duplicates.

perazz:
The build fails due to a circular dependency. As @jvdp1 suggested: 
> Would this be a submodule of `stdlib_sort` more appropriate, as it is suggested to access it through `stdlib_sorting` in the examples?

there should be a `submodule(stdlib_sorting) stdlib_sorting_unique` that only contains the implementation. To achieve that, you could remove the `_impl` submodule, put its contents into `stdlib_sorting_unique`, then put the interface directly in module `stdlib_sorting`

loiseaujc:
Rather than seeing if it works only whence you commit, the easiest for you would be to actually build and run the tests locally. Assuming you have `cmake` installed, it is as simple as

```bash
cmake -B build # sets up the build process.
cmake --build build # Actually compile stdlib
cmake --build build --target test # Compile and runs the tests
```

Doing so, you'll be able to test everything locally and only push to github once you have a working implementation.

PR #966, opened 2025-03-26
system: `delete_file`

perazz:

The `delete_file` subroutine removes a file from the filesystem.
The function ensures the file exists and is not a directory before attempting deletion.
If deletion fails due to permissions, being a directory, or other issues, an error is raised.

Some questions to discuss:

- Naming: Should this be `delete`, `delete_file`, or another name?
- Should it work on files only (current version), or should it delete directories natively (without asking), or should a logical argument be added to control this? The directory case is more complicated because it should handle either empty directories, or directories with contents, recursively.
- Should this belong in `stdlib_system` or `stdlib_io`, or elsewhere?

cc: @jvdp1 @jalvesz


jalvesz:
Hi @perazz thanks for this PR!

Regarding the naming, I think it will depend indeed on the functionality. If deleting folders is also included then it could be `delete` otherwise `delete_file` is better to remain explicit.

Another possibility would be to leave this procedure for deleting only single files and then add another procedure for deleting a whole folder tree

I think that it makes sense to leave it in the `system` module as it is an OS related feature.

perazz:
I agree with your opinion: all OSes have a similar differentiation between 3 options basically: 
- delete files only (maybe with globbing)
- delete directories (empty only)
- delete entire directory trees (recursive)

doc/specs/stdlib_system.md
```diff
@@ -492,3 +492,43 @@ None.
...
+
+### Arguments
+
+`path`: Shall be a character string containing the path to the file to be deleted. It is an `intent(in)` argument.
+
+`err` (optional): Shall be a `type(state_type)` variable for error handling. If provided, errors are returned as a state object. If not provided, the program stops execution on error.
+
+### Behavior
+
+- Checks if the file exists. If not, an error is raised.
```
jvdp1, line 522: Should it not be a warning, instead of an error, if the file doesn't exist?

jvdp1 (approved):
Thank you @perazz . LGTM. I only have a minor question regarding its behaviour. I wonder if it should raise a warning, instead of an error, if the file does not exist.

doc/specs/stdlib_system.md
perazz: Great idea, I totally agree: [16d3f28](https://github.com/fortran-lang/stdlib/pull/966/commits/16d3f28278e61aa7aa8d401ddf0f7512025508cf)

jalvesz (approved):
LGTM @perazz Thanks for this nice addition!

PR #968, opened 2025-03-27
Exponential of a matrix

loiseaujc:
Following a comment by Meromorphic on the discourse [here](https://fortran-lang.discourse.group/t/some-reflections-about-stdlib-where-it-stands-and-what-it-could-become/9376/53?u=loiseaujc), this PR implements the matrix exponential function `expm`. It does so in a new `stdlib_linalg_matrix_functions` submodule which might eventually accomodate later on implementations of other matrix functions (e.g. `logm`, `sqrtm`, `cosm`, etc).

### Proposed interface

- `E = expm(A [, order, err])` where `A` is the input `n x n` matrix, `order` (optional) is the order of the Pade approximation, and `err` of type `linalg_state_type` for error handling.

### Key facts

The implementation makes use of a standard "scaling and squaring" approach to compute the Pade approximation with a fixed order. It is adapted from the implementation by John Burkardt.

### Progress

- [x] Base implementation.
- [x] Tests
- [x] in-code Documentation
- [x] Specifications
- [x] Example

### Possible improvements

The implementation is fully working and validated. Computational performances and/or robustness could potentially be improved eventually by considering:

- Automatic estimation of the best `order` for the Pade approximation based on the data. This is used for instance in `scipy.linalg.expm`.
- Balancing is not used currently, but it might improve the robustness.
- ~At the end of the algorithm, the squaring step is implemented using a simple loop involving `matmul`. Additional performances can be gained by implementing a fast `matrix_power` function (see `np.matrix_power` for instance).~ (irrelevant for this particular task since the scaling factor is chosen as a power of 2, might still be a useful function in the grand scheme of things)

In practice, it has however never failed me so far.

cc @perazz, @jvdp1, @jalvesz


loiseaujc:
I think it is pretty much ready for review. Note that the optimizations I've mentioned (i.e. variable order for the Pade approximation, balancing, etc) wouldn't change the signature of the function. Either we take it as is and gradually improve it over time, or we improve it right away (but I won't have much time to do right now). As you like.

test/linalg/test_linalg_expm.fypp
```diff
@@ -0,0 +1,90 @@
...
+        
+        #:for rk,rt,ri in RC_KINDS_TYPES
+        tests = [tests, new_unittest("expm_${ri}$",test_expm_${ri}$)]
+        #:endfor
+
+    end subroutine test_expm_computation
+
+    !> Matrix exponential with analytic expression.
+    #:for rk,rt,ri in RC_KINDS_TYPES
+    subroutine test_expm_${ri}$(error)
```
jvdp1, line 30: Tests that trigger the errors should be provided.

src/stdlib_linalg_matrix_functions.fypp
```diff
@@ -0,0 +1,130 @@
...
+            call linalg_error_handling(err0, err)
+        end block
+
+        ! This loop should eventually be replaced by a fast matrix_power function.
+        do k = 1, s
+            E = matmul(E, E)
+        enddo
+        return
+    contains
+        elemental subroutine handle_gesv_info(info,lda,n,nrhs,err)
```
jvdp1, line 106: Should this procedure not provided by `stdlib_lapack` (or similar)? Otherwise, developers and users might replicate this type of procedures quite often (cc @perazz )

src/stdlib_linalg.fypp
```diff
@@ -1678,6 +1679,53 @@ module stdlib_linalg
       #:endfor            
   end interface mnorm
 
+  !> Matrix exponential: function interface
+  interface expm
+    !! version : experimental
+    !!
+    !! Computes the exponential of a matrix using a rational Pade approximation.
+    !! ([Specification](../page/specs/stdlib_linalg.html#expm))
```
jvdp1, line 1687: is this link correct?

src/stdlib_linalg_matrix_functions.fypp
```diff
@@ -0,0 +1,130 @@
...
+        block
+            integer(ilp) :: ipiv(n), info
+            call gesv(n, n, Q, n, ipiv, E, n, info) ! E = inv(Q) @ E
+            call handle_gesv_info(info, n, n, n, err0)
+            call linalg_error_handling(err0, err)
+        end block
+
+        ! This loop should eventually be replaced by a fast matrix_power function.
+        do k = 1, s
+            E = matmul(E, E)
```
jvdp1, line 102: As `gesv` is used above, `matmul` could be replaced by `gemm`

jvdp1 (commented):
Thank you @loiseaujc ! LGTM. I have some minor suggestions and some additional ones that might lead to some discussions.

src/stdlib_linalg.fypp
loiseaujc: It does work when I build the doc locally.

test/linalg/test_linalg_expm.fypp
loiseaujc: Done.

src/stdlib_linalg_matrix_functions.fypp
loiseaujc: Done.

src/stdlib_linalg_matrix_functions.fypp
perazz: Yes, I agree that code duplication should be avoided. Looking at the current lapack module structure, there are so many modules and submodules already. So I think that we should try to not create a new one only for the error message handling? Perhaps we should use `stdlib_lapack_base` or `stdlib_linalg_lapack_aux`? The latter is not a large module and it would seem a better fit to me.

loiseaujc:
Seems like I have issues with `mys

src/stdlib_linalg_matrix_functions.fypp
```diff
@@ -0,0 +1,141 @@
...
+                                    needs to be positive, order=', order_)
+            call linalg_error_handling(err0, err)
+            return
+        endif
+
+        ! Compute the L-infinity norm.
+        a_norm = mnorm(A, "inf")
+
+        ! Determine scaling factor for the matrix.
+        ee = int(log(a_norm) / log(2.0_${rk}$)) + 1
```
jalvesz: maybe define the `log(2)` as a compile time constant parameter within the procedure.

src/stdlib_linalg_matrix_functions.fypp
```diff
@@ -0,0 +1,141 @@
...
+
+        ! Scale the input matrix & initialize polynomial.
+        A2 = A/2.0_${rk}$**s ; X = A2
+
+        ! First step of the Pade approximation.
+        c = 0.5_${rk}$
+        allocate (E, source=A2) ; allocate (Q, source=A2)
+        do concurrent(i=1:n, j=1:n)
+            E(i, j) =  c*E(i, j) ; if (i == j) E(i, j) = 1.0_${rk}$ + E(i, j) ! E = I + c*A2
+            Q(i, j) = -c*Q(i, j) ; if (i == j) Q(i, j) = 1.0_${rk}$ + Q(i, j) ! Q = I - c*A2
```
jalvesz: Given that you either assign one value or the other depending on the index, a compact and efficient way would be to replace using `merge`
```
E(i, j) =  merge( 1.0_${rk}$ + E(i, j) , c*E(i, j) , i == j ) ! E = I + c*A2
Q(i, j) =  merge( 1.0_${rk}$ + Q(i, j) , -c*Q(i, j) , i == j ) ! Q = I - c*A2
```

test/linalg/test_linalg_expm.fypp
```diff
@@ -0,0 +1,126 @@
...
+    use stdlib_linalg_state, only: linalg_state_type, linalg_error_handling, LINALG_ERROR, &
+         LINALG_INTERNAL_ERROR, LINALG_VALUE_ERROR
+
+    implicit none (type,external)
+    
+    public :: test_expm_computation
+
+    contains
+
+    !> schur decomposition tests
```
jalvesz: I think this is a copy-paste typo: "Matrix exponential tests" ?

src/stdlib_linalg_matrix_functions.fypp
```diff
@@ -0,0 +1,141 @@
...
+    use stdlib_linalg_blas, only: gemm
+    use stdlib_linalg_lapack, only: gesv
+    use stdlib_linalg_state, only: linalg_state_type, linalg_error_handling, LINALG_ERROR, &
+         LINALG_INTERNAL_ERROR, LINALG_VALUE_ERROR
+    implicit none
+
+contains
+
+    #:for rk,rt,ri in RC_KINDS_TYPES 
+    module function stdlib_expm_${ri}$(A, order, err) result(E)
```
jalvesz: For the base implementation I would have done it using a `subroutine` that pre-supposes the matrix is already allocated in memory, e.g.: `real(<>), intent(inout) :: E(:,:)`. The `function` interface could be simply built on top of it to have automatic (costly) allocations.

test/linalg/test_linalg_expm.fypp
```diff
@@ -0,0 +1,126 @@
...
+    subroutine test_expm_computation(tests)
+        !> Collection of tests
+        type(unittest_type), allocatable, intent(out) :: tests(:)
+        
+        allocate(tests(0))
+        
+        #:for rk,rt,ri in RC_KINDS_TYPES
+        tests = [tests, new_unittest("expm_${ri}$",test_expm_${ri}$)]
```
jalvesz: Maybe the current fails with msys CI is related to the changes in #1008

jalvesz (changes requested):

src/stdlib_linalg_matrix_functions.fypp
loiseaujc: I feel like there are many algorithms that may use it (although I don't have any particular in mind right now). Should we define it in `stdlib_constants` ?

src/stdlib_linalg_matrix_functions.fypp
loiseaujc: Oh nice trick ! Will update the code shorlty.

test/linalg/test_linalg_expm.fypp
loiseaujc: Yes indeed. Will update shortly.

src/stdlib_linalg_matrix_functions.fypp
loiseaujc: Fair enough.

src/stdlib_linalg_matrix_functions.fypp
jalvesz: Makes sense, no strong opinions.

src/stdlib_linalg_matrix_functions.fypp
loiseaujc: I mimicked what had been for the Cholesky factorization. The lowest-level driver is a subroutine doing in-place computation. I've also added the `matrix_exp` interface if anyone wants to do a subroutine call rather than a function one.

test/linalg/test_linalg_expm.fypp
loiseaujc: I merged the latest additions from `master` into my branch but it does not seem to solve the problem. Really not sure what's going on since all other tests pass with flying colors. Will investigate further.

loiseaujc: Looking a bit online, it seems that the exit code `0xC0000374` returned when the test fails indicates a heap corruption problem.

jalvesz: Managed to reproduce the fails with GNU 15. Using the `add_test` proposed by @perazz in #1008 solves the fails:
<details>

<summary>(click to expand) test_linalg_expm.fypp</summary>

```fortran
#:include "common.fypp"
! Test Schur decomposition
module test_linalg_expm
    use testdrive, only: error_type, check, new_unittest, unittest_type
    use stdlib_constants
    use stdlib_linalg_constants
    use stdlib_linalg, only: expm, eye, norm, matrix_exp
    use stdlib_linalg_state, only: linalg_state_type, linalg_error_handling, LINALG_ERROR, &
         LINALG_INTERNAL_ERROR, LINALG_VALUE_ERROR

    implicit none (type,external)
    
    public :: test_expm_computation

    contains

    ! gcc-15 bugfix utility
    subroutine add_test(tests,new_test)
        type(unittest_type), allocatable, intent(inout) :: tests(:)    
        type(unittest_type), intent(in) :: new_test

        integer :: n
        type(unittest_type), allocatable :: new_tests(:)

        if (allocated(tests)) then 
            n = size(tests)
        else
            n = 0
        end if

        allocate(new_tests(n+1))
        if (n>0) new_tests(1:n) = tests(1:n)
                 new_tests(1+n) = new_test
        call move_alloc(from=new_tests,to=tests)        

    end subroutine add_test

    !> Exponent of matrix tests
    subroutine test_expm_computation(tests)
        !> Collection of tests
        type(unittest_type), allocatable, intent(out) :: tests(:)
        
        call add_test(tests,new_unittest("expm",test_expm))
        call add_test(tests,new_unittest("error_handling_expm",test_error_handling_expm))

    end subroutine test_expm_computation

    !> Matrix exponential with analytic expression.
    subroutine test_expm(error)
        ! Problem dimension.
        integer(ilp), parameter :: n = 5, m = 6
        ! Test matrix.
        integer(ilp) :: i, j

        #:for rk,rt,ri in RC_KINDS_TYPES
        block
            ${rt}$ :: A(n, n), E(n, n), Eref(n, n)
            real(${rk}$) :: err
            
            ! Initialize matrix.
            A = zero_${rk}$
            do i = 1, n-1
                A(i, i+1) = m*one_${rk}$
            enddo

            ! Reference with analytical exponential
            Eref = eye(n, mold=one_${rk}$)
            do i = 1, n-1
                do j = 1, n-i
                    Eref(i, i+j) = Eref(i, i+j-1)*m/j
                enddo
            enddo

            ! Compute matrix exponential.
            E = expm(A)

            ! Check result.
            err = norm(Eref - E, "inf")
            print *, err , (n**2)*epsilon(1.0_${rk}$)
            call check(error, err < (n**2)*epsilon(1.0_${rk}$), "Analytical matrix exponential.")
            if (allocated(error)) return
        end block     
        #:endfor  
    end subroutine test_expm

    !> Test error handler.
    subroutine test_error_handling_expm(error)
        ! Problem dimension.
        integer(ilp), parameter :: n = 5, m = 6
        ! Test matrix.
        
        type(linalg_state_type) :: err
        integer(ilp) :: i

        #:for rk,rt,ri in RC_KINDS_TYPES
        block
            ${rt}$ :: A(n, n), E(n, n)
            ! Initialize matrix.
            A = zero_${rk}$
            do i = 1, n-1
                A(i, i+1) = m*one_${rk}$
            enddo

            ! Compute matrix exponential.
            call matrix_exp(A, E, order=-1, err=err)
            ! Check result.
            call check(error, err%error(), "Negative Pade order")
            if (allocated(error)) return

            call matrix_exp(A, order=-1, err=err)
            ! Check result.
            call check(error, err%error(), "Negative Pade order")
            if (allocated(error)) return

            ! Compute matrix exponential.
            call matrix_exp(A, E(:n, :n-1), err=err)
            ! Check result.
            call check(error, err%error(), "Invalid matrix size")
            if (allocated(error)) return

            ! Compute matrix exponential.
            call matrix_exp(A(:n, :n-1), err=err)
            ! Check result.
            call check(error, err%error(), "Invalid matrix size")
            if (allocated(error)) return
        end block
        #:endfor
    end subroutine test_error_handling_expm

end module test_linalg_expm

program test_expm
    use, intrinsic :: iso_fortran_env, only : error_unit
    use testdrive, only : run_testsuite, new_testsuite, testsuite_type
    use test_linalg_expm, only : test_expm_computation
    implicit none
    integer :: stat, is
    type(testsuite_type), allocatable :: testsuites(:)
    character(len=*), parameter :: fmt = '("#", *(1x, a))'

    stat = 0

    testsuites = [ &
        new_testsuite("linalg_expm", test_expm_computation) &
        ]

    do is = 1, size(testsuites)
        write(error_unit, fmt) "Testing:", testsuites(is)%name
        call run_testsuite(testsuites(is)%collect, error_unit, stat)
    end do

    if (stat > 0) then
        write(error_unit, '(i0, 1x, a)') stat, "test(s) failed!"
        error stop
    end if
end program test_expm
```

</details>

loiseaujc: Works like a charm ! Thanks

jalvesz (approved):

src/stdlib_linalg_matrix_functions.fypp
```diff
@@ -43,10 +43,10 @@ contains
         lda = size(A, 1, kind=ilp) ; n = size(A, 2, kind=ilp)
         lde = size(E, 1, kind=ilp) ; ne = size(E, 2, kind=ilp)
           
-        if (lda<1 .or. n<1 .or. lda<n .or. lde<n .or. ne<n) then     
+        if (lda<1 .or. n<1 .or. lda/=n .or. lde/=n .or. ne/=n) then     
             err0 = linalg_state_type(this,LINALG_VALUE_ERROR, &
-                                     'invalid matrix sizes: A=',[lda,n], &
-                                                          ' E=',[lde,ne])
+                                     'invalid matrix sizes: A must be square (lda=', lda, ', n=', n, '), &
```
jalvesz: missing the last '
```
'invalid matrix sizes: A must be square (lda=', lda, ', n=', n, ')', &
```

jalvesz (changes requested):

jalvesz (approved):

PR #974, opened 2025-04-03
hashmap-init-update

chuckyvt:
This PR is to update the hashmap initialization routine to be more user friendly.   It does two things: 

- Changes hasher to be an optional argument instead of required, with the default being fnv_1_hasher if no hasher is specified. 
    This perhaps was the original intent of the code, as the abstract base hashmap class sets the hasher to fnv_1_hasher, but then the downstream init functions currently require the hasher to be specified again.  I believe the choice of hasher function should be abstracted away for the majority of users.   

- Includes a logical flag `initialized` to track initialized state, and if a map_entry call is made prior to initializing, it will call the init routine using default arguments.    
   I believe this is a relatively simple ease of use feature, that will make the hashmaps more in line with implicit initialization commonly used with other languages (but init can still be explicitly called if needed).


perazz (approved):
Nice improvement @chuckyvt, looks good imho.

test/hashmaps/test_chaining_maps.f90
```diff
@@ -48,7 +48,7 @@ program test_chaining_maps
 
     test_8_bits(:) = transfer( rand_object, 0_int8, test_size )
 
-    call map % init( fnv_1_hasher, slots_bits=10 )
+    ! Test implicit initalization by leaving skipping init call for first test.
```
jvdp1, line 51:
```suggestion
    ! Test implicit initalization by skipping init call for first test.
```

test/hashmaps/test_open_maps.f90
```diff
@@ -49,7 +49,7 @@ program test_open_maps
 
     test_8_bits(:) = transfer( rand_object, 0_int8, test_size )
 
-    call map % init( fnv_1_hasher, slots_bits=10 )
+    ! Test implicit initalization by leaving skipping init call for first test.
```
jvdp1, line 52:
```suggestion
    ! Test implicit initalization by skipping init call for first test.
```

jvdp1 (commented):
Thank you @chuckyvt ! LGTM! Here are some minor comments

chuckyvt:
Added the suggested changes.  Also realized I forgot to update init so that it updates the initialization flag to true and added that.  This PR is complete as far as I know.    The last update shows a CI failure that is not related to hashmaps.

jalvesz (approved):
LGTM @chuckyvt I think this PR is ready to be merged

jvdp1 (approved):
LGTM. Thank you @chuckyvt

perazz:
With three approvals, I will merge. Thank you @chuckyvt @jvdp1 @jalvesz.

Issue #973, opened 2025-04-02
Make more functions in `stdlib_ascii.f90` elemental


Beliavsky:
The following functions in `stdlib_ascii.f90` can be made `elemental`:

```
is_alpha
is_alphanum
is_ascii
is_control
is_digit
is_octal_digit
is_hex_digit
is_punctuation
is_graphical
is_printable
is_lower
is_upper
is_white
is_blank
```

The following code where pure functions were designated elemental compiles with gfortran and ifx.

``` fortran
module m
implicit none
contains
    elemental logical function is_alpha(c)
        character(len=1), intent(in) :: c !! The character to test.
        is_alpha = (c >= 'A' .and. c <= 'Z') .or. (c >= 'a' .and. c <= 'z')
    end function

    !> Checks whether `c` is a letter or a number (0 .. 9, a .. z, A .. Z).
    elemental logical function is_alphanum(c)
        character(len=1), intent(in) :: c !! The character to test.
        is_alphanum = (c >= '0' .and. c <= '9') .or. (c >= 'a' .and. c <= 'z') &
            .or. (c >= 'A' .and. c <= 'Z')
    end function

    !> Checks whether or not `c` is in the ASCII character set -
    !> i.e. in the range 0 .. 0x7F.
    elemental logical function is_ascii(c)
        character(len=1), intent(in) :: c !! The character to test.
        is_ascii = iachar(c) <= int(z'7F')
    end function

    !> Checks whether `c` is a control character.
    elemental logical function is_control(c)
        character(len=1), intent(in) :: c !! The character to test.
        integer :: ic
        ic = iachar(c)
        is_control = ic < int(z'20') .or. ic == int(z'7F')
    end function

    !> Checks whether `c` is a digit (0 .. 9).
    elemental logical function is_digit(c)
        character(len=1), intent(in) :: c !! The character to test.
        is_digit = ('0' <= c) .and. (c <= '9')
    end function

    !> Checks whether `c` is a digit in base 8 (0 .. 7).
    elemental logical function is_octal_digit(c)
        character(len=1), intent(in) :: c !! The character to test.
        is_octal_digit = (c >= '0') .and. (c <= '7');
    end function

    !> Checks whether `c` is a digit in base 16 (0 .. 9, A .. F, a .. f).
    elemental logical function is_hex_digit(c)
        character(len=1), intent(in) :: c !! The character to test.
[...]

PR #977, opened 2025-04-10
promote other ascii functions to elemental

wassup05:
Attempts to address #973 

There was a `subroutine` which would not compile due to this change as procedure pointers cannot point to elemental procedures (according to `gfortran`) and as that `subroutine` was not used anywhere in the tests, is now removed.

Maybe we should add some array test cases to ensure it behaves correctly?

### DIFF
--- a/test/ascii/test_ascii.f90
+++ b/test/ascii/test_ascii.f90
@@ -721,66 +721,6 @@ contains
         end do
     end subroutine
 
-    !
-    !   This test reproduces the true/false table found at
-    !   https://en.cppreference.com/w/cpp/string/byte
-    !
-    subroutine test_ascii_table
-        integer :: i, j
-        logical :: table(15,12)
-
-        abstract interface
-            pure logical function validation_func_interface(c)
-                character(len=1), intent(in) :: c
-            end function
-        end interface
-
-        type :: proc_pointer_array
-            procedure(validation_func_interface), pointer, nopass :: pcf
-        end type proc_pointer_array
-
-        type(proc_pointer_array) :: pcfs(12)
-
-        pcfs(1)%pcf => is_control
-        pcfs(2)%pcf => is_printable
-        pcfs(3)%pcf => is_white
-        pcfs(4)%pcf => is_blank
-        pcfs(5)%pcf => is_graphical
-        pcfs(6)%pcf => is_punctuation
-        pcfs(7)%pcf => is_alphanum
-        pcfs(8)%pcf => is_alpha
-        pcfs(9)%pcf => is_upper
-        pcfs(10)%pcf => is_lower
-        pcfs(11)%pcf => is_digit
-        pcfs(12)%pcf => is_hex_digit
-
-        ! loop through functions
-        do i = 1, 12
-            table(1,i)  = all([(pcfs(i)%pcf(achar(j)),j=0,8)])      ! control codes
-            table(2,i)  = pcfs(i)%pcf(achar(9))                     ! tab
-            table(3,i)  = all([(pcfs(i)%pcf(achar(j)),j=10,13)])    ! whitespaces
-            table(4,i)  = all([(pcfs(i)%pcf(achar(j)),j=14,31)])    ! control codes
-            table(5,i)  = pcfs(i)%pcf(achar(32))                    ! space
-            table(6,i)  = all([(pcfs(i)%pcf(achar(j)),j=33,47)])    ! !"#$%&'()*+,-./
-            table(7,i)  = all([(pcfs(i)%pcf(achar(j)),j=48,57)])    ! 0123456789
-            table(8,i)  = all([(pcfs(i)%pcf(achar(j)),j=58,64)])    ! :;<=>?@
-            table(9,i)  = all([(pcfs(i)%pcf(achar(j)),j=65,70)])    ! ABCDEF
-            table(10,i) = all([(pcfs(i)%pcf(achar(j)),j=71,90)])    ! GHIJKLMNOPQRSTUVWXYZ
-            table(11,i) = all([(pcfs(i)%pcf(achar(j)),j=91,96)])    ! [\]^_`
-            table(12,i) = all([(pcfs(i)%pcf(achar(j)),j=97,102)])   ! abcdef
-            table(13,i) = all([(pcfs(i)%pcf(achar(j)),j=103,122)])  ! ghijklmnopqrstuvwxyz
-            table(14,i) = all([(pcfs(i)%pcf(achar(j)),j=123,126)])  ! {|}~
-            table(15,i) = pcfs(i)%pcf(achar(127))                   ! backspace character
-        end do
-
-        ! output table for verification
-        write(*,'(5X,12(I4))') (i,i=1,12)
-        do j = 1, 15
-            write(*,'(I3,2X,12(L4),2X,I3)') j, (table(j,i),i=1,12), count(table(j,:))
-        end do
-        write(*,'(5X,12(I4))') (count(table(:,i)),i=1,12)
-    end subroutine test_ascii_table
-
     subroutine test_to_lower_string(error)
         !> Error handling
         type(error_type), allocatable, intent(out) :: error
--- a/src/stdlib_ascii.fypp
+++ b/src/stdlib_ascii.fypp
@@ -107,13 +107,13 @@ module stdlib_ascii
 contains
 
     !> Checks whether `c` is an ASCII letter (A .. Z, a .. z).
-    pure logical function is_alpha(c)
+    elemental logical function is_alpha(c)
         character(len=1), intent(in) :: c !! The character to test.
         is_alpha = (c >= 'A' .and. c <= 'Z') .or. (c >= 'a' .and. c <= 'z')
     end function
 
     !> Checks whether `c` is a letter or a number (0 .. 9, a .. z, A .. Z).
-    pure logical function is_alphanum(c)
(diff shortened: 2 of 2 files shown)


perazz (commented):
Thank you for this improvement @wassup05. 
I have restored the former table function. 

A good test case would be to reproduce the results of the table as reported at https://en.cppreference.com/w/cpp/string/byte

the comparison table should look like

```
logical, parameter :: ascii_class_table(15,12) = transpose(reshape([ &
    ! iscntrl  isprint  isspace  isblank  isgraph  ispunct  isalnum  isalpha  isupper  islower  isdigit  isxdigit
    .true.,   .false., .false., .false., .false., .false., .false., .false., .false., .false., .false., .false., & ! 0–8
    .true.,   .false., .true.,  .true.,  .false., .false., .false., .false., .false., .false., .false., .false., & ! 9
    .true.,   .false., .true.,  .false., .false., .false., .false., .false., .false., .false., .false., .false., & ! 10–13
    .true.,   .false., .false., .false., .false., .false., .false., .false., .false., .false., .false., .false., & ! 14–31
    .false.,  .true.,  .true.,  .true.,  .false., .false., .false., .false., .false., .false., .false., .false., & ! 32 (space)
    .false.,  .true.,  .false., .false., .true.,  .true.,  .false., .false., .false., .false., .false., .false., & ! 33–47
    .false.,  .true.,  .false., .false., .true.,  .false., .true.,  .false., .false., .false., .true.,  .true.,  & ! 48–57
    .false.,  .true.,  .false., .false., .true.,  .true.,  .false., .false., .false., .false., .false., .false., & ! 58–64
    .false.,  .true.,  .false., .false., .true.,  .false., .true.,  .true.,  .true.,  .false., .false., .true.,  & ! 65–70
    .false.,  .true.,  .false., .false., .true.,  .false., .true.,  .true.,  .true.,  .false., .false., .false., & ! 71–90
    .false.,  .true.,  .false., .false., .true.,  .true.,  .false., .false., .false., .false., .false., .false., & ! 91–96
    .false.,  .true.,  .false., .false., .true.,  .false., .true.,  .true.,  .false., .true.,  .false., .true.,  & ! 97–102
    .false.,  .true.,  .false., .false., .true.,  .false., .true.,  .true.,  .false., .true.,  .false., .false., & ! 103–122
    .false.,  .true.,  .false., .false., .true.,  .true.,  .false., .false., .false., .false., .false., .false., & ! 123–126
    .true.,   .false., .false., .false., .false., .false., .false., .false., .false., .false., .false., .false.  & ! 127
], shape=[12,15]))
```

wassup05:
Thank you very much for the refactor @perazz 

I have added the test incorporating the procedure, do let me know what you think about it.

test/ascii/test_ascii.f90
```diff
@@ -779,6 +755,60 @@ pure logical function validation_func_interface(c)
...
+        .false.,  .true.,  .false., .false., .true.,  .false., .true.,  .true.,  .false., .true.,  .false., .true.,  & ! 97–102
+        .false.,  .true.,  .false., .false., .true.,  .false., .true.,  .true.,  .false., .true.,  .false., .false., & ! 103–122
+        .false.,  .true.,  .false., .false., .true.,  .true.,  .false., .false., .false., .false., .false., .false., & ! 123–126
+        .true.,   .false., .false., .false., .false., .false., .false., .false., .false., .false., .false., .false.  & ! 127
+        ], shape=[12,15]))
+
+        call ascii_table(arr)
+        call check(error, all(arr .eqv. ascii_class_table), "ascii table was not accurately generated")
+
+        if (allocated(error)) return
```
wassup05, line 811: Noticed this is not required actually...

perazz, line 811:
```suggestion
```

perazz (approved):
LGTM @wassup05. I think this PR is ready to merge imho.

jalvesz:
Another way of performing the test and validating the `elemantal` attribute to check simultaneously all elements could be:

```Fortran
subroutine test_ascii_table
    integer :: i, j
    logical :: table(15,12)

    type :: tests
        character(1), allocatable :: chars(:)
    end type
    type(tests) :: my_tests(15)

    my_tests(1)%chars  = [(achar(j),j=0,8)]    ! control codes
    my_tests(2)%chars  = [(achar(j),j=9,9)]    ! tab
    my_tests(3)%chars  = [(achar(j),j=10,13)]  ! whitespaces
    my_tests(4)%chars  = [(achar(j),j=14,31)]  ! control codes
    my_tests(5)%chars  = [(achar(j),j=32,32)]  ! space
    my_tests(6)%chars  = [(achar(j),j=33,47)]  ! !"#$%&'()*+,-./
    my_tests(7)%chars  = [(achar(j),j=48,57)]  ! 0123456789
    my_tests(8)%chars  = [(achar(j),j=58,64)]  ! :;<=>?@
    my_tests(9)%chars  = [(achar(j),j=65,70)]  ! ABCDEF
    my_tests(10)%chars = [(achar(j),j=71,90)]  ! GHIJKLMNOPQRSTUVWXYZ
    my_tests(11)%chars = [(achar(j),j=91,96)]  ! [\]^_`
    my_tests(12)%chars = [(achar(j),j=97,102)] ! abcdef
    my_tests(13)%chars = [(achar(j),j=103,122)]! ghijklmnopqrstuvwxyz
    my_tests(14)%chars = [(achar(j),j=123,126)]! {|}~
    my_tests(15)%chars = [(achar(j),j=127,127)]! backspace character

    ! loop through functions
    do i = 1, 15
        table(i,1)  = all(is_control(my_tests(i)%chars)) 
        table(i,2)  = all(is_printable(my_tests(i)%chars))
        table(i,3)  = all(is_white(my_tests(i)%chars))
        table(i,4)  = all(is_blank(my_tests(i)%chars))
        table(i,5)  = all(is_graphical(my_tests(i)%chars)) 
        table(i,6)  = all(is_punctuation(my_tests(i)%chars))
        table(i,7)  = all(is_alphanum(my_tests(i)%chars))
        table(i,8)  = all(is_alpha(my_tests(i)%chars))
        table(i,9)  = all(is_upper(my_tests(i)%chars))  
        table(i,10) = all(is_lower(my_tests(i)%chars)) 
        table(i,11) = all(is_digit(my_tests(i)%chars)) 
        table(i,12) = all(is_hex_digit(my_tests(i)%chars)) 
    end do

    ! output table for verification
    write(*,'(5X,12(I4))') (i,i=1,12)
    do j = 1, 15
        write(*,'(I3,2X,12(L4),2X,I3)') j, (table(j,i),i=1,12), count(table(j,:))
    end do
    write(*,'(5X,12(I4))') (count(table(:,i)),i=1,12)
end subroutine test_ascii_table
```
No need for procedure pointers either.

wassup05:
Thank you for your reviews everyone.

It seems these procedures are not documented at all. So there is nothing to change really. But maybe I should add them now? since they are complete with functionality now.

@jalvesz ohh yes that is indeed a way of doing two things at once and seems a good way to me. Maybe some other reviewers could also pitch in and let me know what they think about this way (since I wasn't the one who added the table procedure)?

perazz:
> But maybe I should add them now

Yes, please add them to https://github.com/fortran-lang/stdlib/blob/master/doc/specs/stdlib_ascii.md , thank you.

jalvesz:
@wassup05 I was testing a methodology for collaboration and pushed directly to your PR. Hope it is fine ( ? ) in the last commit you'll find a mix of the test that @perazz proposed before and the propal from the comment. Feel free to change it.

jalvesz (approved):

wassup05:
I don't mind at all @jalvesz, thank you very much for the refactor.

`ascii` module was missing docs of the constants and some procedures (that are now elemental) it provides, I added the constants to the docs and will be adding the procedures shortly.

jalvesz:
@wassup05 your last commit for the specs look good to me. Do you consider it done? If yes, I would say lets wait a couple of days in case there are some comments and then merge.

wassup05:
Yes @jalvesz I have added all the missing specs of the module.

perazz (changes requested):
Sorry for the late reply - The changes introduced in 4bc022d9244cbf5438b602b69113be5097341461 look a bit confusing to me - there is no need to store integer ascii codes in a derived type imho, it’s not immediately clear those are ascii codes. Would you mind going back to the previous version? LGTM otherwise.

jalvesz:
>there is no need to store integer ascii codes in a derived type imho, it’s not immediately clear those are ascii codes.

The intention is not storing ascii codes in a DT, simply having a list of allocatable characters such that each one is seen as an array. Thus exactly the same test is performed. The only difference being that the `elemental` attribute is being put to test as it is applied on the whole array at once.

perazz:
Hello @wassup05 @jalvesz, when you are ready to revert commit https://github.com/fortran-lang/stdlib/commit/4bc022d9244cbf5438b602b69113be5097341461, then I believe this PR can be merged.

jalvesz:
@perazz for the moment I haven't had the time to think about a different way of doing the test. As we discussed, the issue I had with the previous proposal is that the test did not enable doing the following:
```Fortran
... = all( is_<fun_name>( array(:) ) )
```
Which the current test does.

This is important because it is the only actual valuable difference between an `elemental` and a `pure` procedure (whole array evaluations). So, say that a compiler for some reason has a bug in the treatment of the `elemental` attribute, the current test would fail. The previous test would apply the same logic as if the procedure had no `elemental` attribute on it.

PR #980, opened 2025-04-20
stdlib_linalg: declare the `diag` function as pure

zoziha:
This is a minor update that sets the `diag` function as pure.

### DIFF
--- a/doc/specs/stdlib_linalg.md
+++ b/doc/specs/stdlib_linalg.md
@@ -187,6 +187,10 @@ to the original.
 
 Stable
 
+### Class
+
+Pure function.
+
 ### Description
 
 Create a diagonal array or extract the diagonal elements of an array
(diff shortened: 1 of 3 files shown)


jvdp1 (commented):
Thank you @zoziha . I am surprised that `diag` was not `pure` yet, as it is the case for `eye`.
The API of `diag` is set to `stable`. However, changing it to `pure` is backward-compatible, and is ok IMO.

Could you also update the specs, please?

zoziha:
> Could you also update the specs, please?

@jvdp1 Thanks for your review. I will mark the `class` of the `diag` function as a pure function in `stdlib_linalg.md`. Does _updating the specifications_ mean this? Or are there other updates needed?

jalvesz (approved):
LGTM @zoziha

jvdp1 (approved):
LGTM! Thank yoju @zoziha

zoziha:
@jvdp1 @jalvesz Thank you for your review. With two approvals, this minor update will be merged now.

PR #981, opened 2025-04-20
stdlib_io: add `print_array` function to print arrays to output units

### LINKED ISSUE #40: Pretty printing of matrices (and multidimensional arrays)

zoziha:
### Description

Add `print_array` subroutine, which is used to print two-dimensional arrays of integer and floating-point types. These two types are more commonly used in Fortran. It supports arguments such as delimiter, output unit, output format, and whether to print briefly.

In the non-brief output mode, the print_array function can produce an output result style similar to that of `savetxt` in stdlib_io. The print_array function is mainly convenient for users to output to the screen (standard output) and is likely to be mostly used for code debugging.

Prior Art:

+ [disp](https://ww2.mathworks.cn/help/matlab/ref/[disp](https://ww2.mathworks.cn/help/matlab/ref/disp.html?lang=en).html?lang=en) in Matlab;
+ The default print of Python and Julia;
+ [brief_print](https://arma.sourceforge.net/docs.html#brief_print) in Armadillo.

This PR will replace #520, and is related to #40, but is not sufficient to close #40.

### Remaining to be discussed

+ Is it necessary to support delimiters of any length, not just those with a length of 1? In addition to the `print_array`, it also includes the `savetxt` routine in stdlib_io.
+ Currently, print_array is designed to assist users in program debugging and does not provide file write error reports like savetxt does.


doc/specs/stdlib_io.md
```diff
@@ -305,3 +305,38 @@ Exceptions trigger an `error stop` unless the optional `err` argument is provide
 {!example/io/example_get_file.f90!}
 ```
 
+## `print_array` - Print an array to an output unit
```
jvdp1, line 308: Minor comment: the name could be discussed with the community. Personally, I would prefer `disp`.

src/stdlib_io_print_array.fypp
```diff
@@ -0,0 +1,88 @@
...
+    implicit none
+
+contains
+
+    #:for k1, t1 in KINDS_TYPES
+    module subroutine print_array_${t1[0]}$${k1}$(array, unit, fmt, delimiter, brief)
+        ${t1}$, intent(in) :: array(:, :)
+        integer, intent(in), optional :: unit
+        character(len=*), intent(in), optional :: fmt
+        character(len=1), intent(in), optional :: delimiter
```
jvdp1, line 15: It could be 
```suggestion
        character(len=*), intent(in), optional :: delimiter
```
and then declare `delimiter_` `allocatable`

src/stdlib_io_print_array.fypp
```diff
@@ -0,0 +1,88 @@
...
+        logical, intent(in), optional :: brief
+
+        integer :: i, j, unit_, shape_(2)
+        character(len=:), allocatable :: fmt_
+        character(len=1) :: delimiter_
```
jvdp1, line 20:
```suggestion
        character(len=:), allocatable :: delimiter_
```

src/stdlib_io_print_array.fypp
```diff
@@ -0,0 +1,88 @@
...
+        character(len=3) :: delim_str
```
jvdp1, line 21:
```suggestion
        character(len=:), allocatable :: delim_str
```

jvdp1 (commented):
Thank you @zoziha . Here are some minor comments.

Mahmood-Sinan (changes requested):
Thank you @zoziha. One suggestion:
Can each entry of complex type array be printed as (a,b) form instead of using a delimiter between the real and imaginary parts. For example, for an array 
`complex, dimension(2, 2) :: a = reshape([cmplx(1,2), cmplx(3,4), cmplx(5,6), cmplx(7,8)], [2,2])`
can the output be displayed as:
```
(1.0,2.0) | (5.0,6.0)
(3.0,4.0) | (7.0,8.0)
```
instead of 
```
    1.0|    2.0|    5.0|    6.0
    3.0|    4.0|    7.0|    8.0
```

where delimiter is `|`.

PR #985, opened 2025-04-22
`linalg`: slightly increase test tolerance

perazz:
this PR attempts to fix some CI failures by relaxing the `pinv` (pseudoinverse) test tolerance.
`pinv` works with singular matrices and the accuracy of the inverse in the test may be affected by the random number choice.


jvdp1 (approved):
LGTM. Thank you @perazz

perazz:
Thank you for the review @jvdp1: because this is a short PR, I will go ahead and merge.

PR #991, opened 2025-05-08
modified stats exponential distribution procedures to use  `loc` and `scale` 

sebastian-mutz:
There’s some inconsistency in passed arguments for normal and exponential distributions. Consider the pdfs:
`result = pdf_normal(x, loc, scale)`
`result = pdf_exp(x, lambda)`

While the normal distribution procedure uses `loc` and `scale` (reminiscent of scipy) as opposed to mu and sigma, the exponential distribution procedure uses `lambda` rather than `scale` (also often referred to as beta, which is 1/lambda) and is missing `loc`.

As discussed on [fortran-lang discourse](https://fortran-lang.discourse.group/t/api-design-and-consistency-for-stdlib-stats-distribution-modules/9614), I suggest introducing more consistency in the stats distribution procedures API. I have made some modifications to `stdlib_stats_distribution_exponential`  (and associated examples and tests) to:
1) use `scale` instead of `lambda`, and
2) allow passing an additional location parameter `loc`.

With the suggested modifications in the PR, rvs, pdf and cdf procedures can now be called as such:
`result = rvs_exp(x, loc, scale)`,
`result = pdf_exp(x, loc, scale)`, and
`result = cdf_exp(x, loc, scale)`.


jalvesz:
Thanks for this PR @sebastian-mutz! code wise it looks good to me. Since an API change is being proposed this is a breaking change that might impact some users. I would suggest just to link back on discourse with the summary of the proposal here to alert possible users.

perazz (commented):
Thank you for this PR @sebastian-mutz. LGTM, I only have a couple comments. 

- Regarding the API, I would be in favor of _adding_ the new one as an interface rather than _replacing_ the old. This would be in favor of backward compatibility and user flexibility. The former one could be removed in a next version, if deemed obsolete. 

- Would you mind to also update the [specs](https://stdlib.fortran-lang.org/page/specs/stdlib_stats_distribution_exponential.html#cdf_exp-exponential-cumulative-distribution-function) file for the exponential? it is at 

https://github.com/fortran-lang/stdlib/blob/master/doc/specs/stdlib_stats_distribution_exponential.md

sebastian-mutz:
Thanks, Federico.

> Regarding the API, I would be in favor of _adding_ the new one as an interface rather than _replacing_ the old. This would be in favor of backward compatibility and user flexibility. .

Keeping this flexibility makes sense. Since the different parameterisation requires (slightly) different calculations, it may not be as simple as adding the new ones as an interface (or perhaps I misunderstand your suggestion?). However, I can explore adding the former procedures back in and creating a generic interface for both parameterisation schemes to achieve the same goal.

> Would you mind to also update the specs file for the exponential? 

I missed this. Thanks. I'll update it.

test/stats/test_distribution_exponential.fypp
```diff
@@ -72,46 +72,46 @@ contains
...
-                             0.137541023585612635452927558314210417_${k1}$,    &
-                             0.134921508232253721063879462841820585_${k1}$,    &
-                              1.33766060689229752493171569464417802_${k1}$,    &
-                             0.111148487576340881943792737729381770_${k1}$,    &
-                             0.533951653963536868966836361020492979_${k1}$,    &
-                              1.96897428558727671799033487332053483_${k1}$,    &
-                             0.371111977992924465160247867364281152_${k1}$,    &
-                             0.811918715695663687862785688290993341_${k1}$,    &
-                             0.404637854946697868759504975362991277_${k1}$]
+                            [ 1.37178108290154243675829093263018876_${k1}$,    &
```
jvdp1, line 82: Ideally, these values should not be modified. Only the calls should be modified and return the same values as the previous versions.

src/stdlib_stats_distribution_exponential.fypp
```diff
@@ -294,19 +294,19 @@ contains
...
 
-        if (lambda <= 0._${k1}$) then
+        if (scale <= 0._${k1}$) then
             res = ieee_value(1._${k1}$, ieee_quiet_nan)
-        else if (x < 0._${k1}$) then
+        else if (x < loc) then
             res = 0._${k1}$
         else
-            res = 1.0_${k1}$ - exp(- x * lambda)
+            res = 1.0_sp - exp(- (x - loc) / scale)
```
jvdp1, line 309:
```suggestion
            res = 1.0_sp - exp(- (x - loc) / scale)
```
```suggestion
            res = 1.0_${k1}$ - exp(- (x - loc) / scale)
```

test/stats/test_distribution_exponential.fypp
```diff
@@ -120,16 +120,18 @@ contains
...
         scale = (0.7_${k1}$, 1.3_${k1}$)
         #:endif
 
         do i = 1, k
-            res(i) = expon_rvs(scale)      ! 1 dummy
+            res(i) = expon_rvs(loc, scale)      ! 1 dummy
         end do
-        res(6:10) = expon_rvs(scale, k)    ! 2 dummies
+        res(6:10) = expon_rvs(loc, scale, k)    ! 2 dummies
         call check(all(abs(res - ans) < ${k1}$tol),                            &
```
jvdp1, line 135: Could you provide a new API (and new tests) and keep this one (at least until a next release)?

The API could be something like:
```fortran
function expon_rvs(scale) result(res)
  ...
  res = expon_rrvs(loc = 0, scale = 1/scale,...)
end function
```

jvdp1 (commented):
Here are some comments. 
* I suggest to keep the older API, if possible and if everybody agrees on it.
* The specs should be updated.
* The tests (at least the true values) should not be modified (except if they are wrong). If new true values are needed, new tests should be added.

test/stats/test_distribution_exponential.fypp
sebastian-mutz: Thanks for the review and comments, @jvdp1. If I understand correctly, you suggest renaming the updated procedures and creating a wrapper to keep the old API in tact for now. So in your example, scale in `expon_rvs(scale)` would be actually `lambda`.

doc/specs/stdlib_stats_distribution_exponential.md
```diff
@@ -84,11 +107,20 @@ Elemental function
...
 All arguments must have the same type.
 
 ### Return value
 
-The result is a scalar or an array, with a shape conformable to the arguments, and the same type as the input arguments. If `lambda` is non-positive, the result is `NaN`.
+If `lambda` is passed, the result is a scalar or an array, with a shape conformable to the arguments, and the same type as the input arguments. If `lambda` is non-positive, the result is `NaN`.
+
+
+If `loc` and `scale` are passed, the result is a scalar or an array, with a shape conformable to the arguments, and the same type as the input arguments. If `scale` is non-positive, the result is `NaN`.
+
```
perazz, line 123:
```suggestion
The result is a scalar or an array, with a shape conformable to the arguments, and the same type as the input arguments. If non-positive `lambda` or `scale`, the result is `NaN`.

```

doc/specs/stdlib_stats_distribution_exponential.md
```diff
@@ -112,10 +144,16 @@ For a complex variable  \(z=(x + y i)\) with independent real \(x\) and imaginar
 
 $$F(x+\mathit{i}y)=F(x)F(y)=\begin{cases} (1 - e^{-\lambda_{x} x})(1 - e^{-\lambda_{y} y}) &x\geqslant 0, \;\; y\geqslant 0 \\\\ 0 & \text{otherwise} \end{cases}$$
 
+Instead of of the inverse scale parameter `lambda`, it is possible to pass `loc` and `scale`, where \(scale = \frac{1}{\lambda}\) and `loc` specifies the value by which the distribution is shifted.
```
perazz, line 147:
```suggestion
Alternative to the inverse scale parameter `lambda`, it is possible to pass `loc` and `scale`, where \(scale = \frac{1}{\lambda}\) and `loc` specifies the value by which the distribution is shifted.
```

doc/specs/stdlib_stats_distribution_exponential.md
```diff
@@ -127,11 +165,19 @@ Elemental function
...
+
 All arguments must have the same type.
 
 ### Return value
 
-The result is a scalar or an array, with a shape conformable to the arguments, and the same type as the input arguments. If `lambda` is non-positive, the result is `NaN`.
+If `lamba` is passed, the result is a scalar or an array, with a shape conformable to the arguments, and the same type as the input arguments. If `lambda` is non-positive, the result is `NaN`.
+
+
+If `loc` and `scale` are passed, the result is a scalar or an array, with a shape conformable to the arguments, and the same type as the input arguments. If `scale` is non-positive, the result is `NaN`.
```
perazz, line 180:
```suggestion
The result is a scalar or an array, with a shape conformable to the arguments, and the same type as the input arguments. With non-positive `lambda` or `scale`, the result is `NaN`.
```

perazz (approved):
LGTM with the attached edits, thank you @sebastian-mutz! Let us wait for @jvdp1's review as well.

doc/specs/stdlib_stats_distribution_exponential.md
sebastian-mutz: Thanks, @perazz. That's a more economic way to document this.  I'll wait for further comments from @jvdp1 and tackle all together.

sebastian-mutz:
Thanks for commiting the specs update already, @perazz.

perazz:
Yes, let's wait another couple of days for additional comments, and if not, this is ready to merge imho.

doc/specs/stdlib_stats_distribution_exponential.md
```diff
@@ -31,6 +36,10 @@ The algorithm used for generating exponential random variates is fundamentally l
 
 `result = ` [[stdlib_stats_distribution_exponential(module):rvs_exp(interface)]] `([lambda] [[, array_size]])`
```
jvdp1, line 37: I suggest to remove this one from the specs or to mention that it is deprecated.

doc/specs/stdlib_stats_distribution_exponential.md
```diff
@@ -69,10 +86,16 @@ For a complex variable \(z=(x + y i)\) with independent real \(x\) and imaginary
 
 $$f(x+\mathit{i}y)=f(x)f(y)=\begin{cases} \lambda_{x} \lambda_{y} e^{-(\lambda_{x} x + \lambda_{y} y)} &x\geqslant 0, y\geqslant 0 \\\\ 0 &\text{otherwise}\end{cases}$$
 
+Instead of of the inverse scale parameter `lambda`, it is possible to pass `loc` and `scale`, where \(scale = \frac{1}{\lambda}\) and `loc` specifies the value by which the distribution is shifted.
+
 ### Syntax
 
 `result = ` [[stdlib_stats_distribution_exponential(module):pdf_exp(interface)]] `(x, lambda)`
```
jvdp1, line 93: Same here: remove or mention it is deprecated.

example/stats_distribution_exponential/example_exponential_cdf.f90
```diff
@@ -4,60 +4,80 @@ program example_exponential_cdf
...
+  complex :: cloc, cscale
   integer :: seed_put, seed_get, i
 
   seed_put = 1234567
   call random_seed(seed_put, seed_get)
 
-  ! standard exponential cumulative distribution at x=1.0
+  ! standard exponential cumulative distribution at x=1.0 with lambda=1.0
   print *, exp_cdf(1.0, 1.0)
   ! 0.632120550
```
jvdp1, line 17: In the example, I suggest to remove the old API.

jvdp1 (approved):
LGTM. Thank you @sebastian-mutz 

I wonder if the old API should still be mentioned in the specs. The wrappers are importatn for backward compatibility,but I think that they should not be mentioned in the specs to avoid that people use them. Or a note mentioning that they are deprecated should be at least added in the  specs. @perazz what do you think about that?

perazz:
In fine either way @jvdp1: my personal preference would be to keep offering both versions, I don’t see a problem having a specialized option (`lambda` only)  that complements the shared API. But if that will be gradually deprecated, then I agree with you and it should not be in the top level specs anymore

sebastian-mutz:
Thanks, @jvdp1 and @perazz.

> I don’t see a problem having a specialized option (`lambda` only) that complements the shared API

The only problem I can think of is if any arguments are made optional at some point (so the module procedures could become indistinguishable). However, at this point, it'd be another API change requiring more attention and changes in documentation anyway.

sebastian-mutz:
Updated md file and examples to prioritise new API by mentioning it at the top as a "soft" encouragment to use it for now.

jvdp1:
> Updated md file and examples to prioritise new API by mentioning it at the top as a "soft" encouragment to use it for now.

Thank you, @sebastian-mutz . It sounds like a good compromise. Let wait a couple of days before merging this PR.

perazz:
Thank you @sebastian-mutz @jvdp1 @jalvesz, I will merge, as suggested.

PR #996, opened 2025-05-29
fix: collab diagram layout fix

jalvesz:
I noticed that the diagram included to explain the collab script didn't show properly in the documentation html. Adding the text quotes enables a proper visualization of the diagram for both the original markdown and the html.

Current layout in the docs:
image

After fix:
image


jvdp1 (approved):
good catch. Thank you @jalvesz

jalvesz:
Thanks @jvdp1 ! I'll merge this quick fix

PR #997, opened 2025-05-29
fix: make ansi examples from documentation executable

jalvesz:
I was reviewing the documentation of the ansi module and realized the examples were not runnable. Here I propose to add them to the list of executable examples.


jvdp1 (approved):
LGTM. Thank you @jalvesz

PR #939, opened 2025-02-21
io: `getfile`

perazz:
This PR is spun off #904 for discussion with #919. 

The present option provides a `getfile` function that has a similar functionality as the existing `getline` from the `stdlib_io` module: 

## `stdlib_io`
- `type(string_type) function getfile(fileName [,err] [,delete])` loads a whole ASCII file into a `string_type` variable all at once, with option to return a state flag, or to request file deletion (`delete=.true.`) after loading. 

- The file is loaded all at once for maximum computational efficiency. 
- The new `state_type` error handler is optionally used to provide clear error information. Its optional as commonly done in linear algebra routines: if provided, an error message may be returned; if not provided, exceptions trigger an `error stop`

cc: @jvdp1 @chuckyvt @jalvesz @fortran-lang/stdlib

PR #999, opened 2025-06-07
`stdlib_system`: essential path functionality

wassup05:
A few path related functions for ease of future functionality have been added.

- `joinpath`: joins the given paths according to the platform's `path-separator`.
- `operator(/)`: as was suggested in the Fortran discourse [here](https://fortran-lang.discourse.group/t/gsoc-25-stdlib-filesystem/9798/4?u=suprit05) an operator is also provided for the same functionality
- `splitpath`: splits the path following the last `path-separator` and returns the `head` and `tail`.
- `basename`: just returns the `tail` of `splitpath`
- `dirname`: just returns the `head` of `splitpath`

The `pathsep` `parameter` contains either `/` or `\` depending on the platform and is a compile-time constant now, so is the `parameter`, `ISWIN`

Do let me know your thoughts.


wassup05:
I will add that with the other reviews @sebastian-mutz, I don't want to trigger the ci/cd workflow just for a few comments. Also on a second thought maybe instead of the `if (ISWIN) ...` thing I could separate the two examples...

doc/specs/stdlib_system.md
```diff
@@ -532,3 +532,180 @@ The file is removed from the filesystem if the operation is successful. If the o
...
+
+`p`: A character string containing the path to be split. `intent(in)`
+`head`: The first part of the path. `allocatable, intent(out)`
+`tail`: The rest part of the path. `allocatable, intent(out)`
+
+### Behavior
+
+- If `p` is empty, `head` is set to `.` and `tail` is empty
+- If `p` consists entirely of path-separators. `head` is set to the path-separator and `tail` is empty
+- `head` ends in a path-separator iff and only if `p` appears to be a root directory or child of one
```
jalvesz: Iff > if

example/system/example_path_1.f90
jalvesz, line 1: I would advise to split the example per function such that each function in the specs gets it's own code snippet.

wassup05: https://github.com/fortran-lang/stdlib/pull/999/commits/c561108ce8c5d8f82955858057f4b38feed19a84

example/system/example_path_1_windows.f90
```diff
@@ -0,0 +1,18 @@
...
+program example_join_windows
+    use stdlib_system, only: joinpath, operator(/)
+    character(len=:), allocatable :: p1, p2, p3
+    character(len=20) :: parr(4)
+
+    p1 = 'C:'/'Users'/'User1'/'Desktop'
+    p2 = joinpath('C:\Users\User1', 'Desktop')
+
+    parr = [character(len=20) :: 'C:', 'Users', 'User1', 'Desktop']
+    p2 = joinpath(parr)
```
wassup05: should be `p3`, will change it.

jalvesz:
The issue with Ninja detecting a "circular dependency" (there was non) was related to the stdlib_system.F90 file not being visible in a correct manner in the CMakeLists.txt file.

The group:
```cmake
# Preprocessed files to contain preprocessor directives -> .F90
set(cppFiles
```
corresponds to `.fypp` files for which the suffix should be changed to `.F90`. Not for already `.F90` files.

example/system/example_path_1.f90
jalvesz: further split https://github.com/fortran-lang/stdlib/pull/999/commits/272eb675b3df42b12812379dbdc908d0ed795fa7

CMakeLists.txt
```diff
@@ -31,6 +31,12 @@ if(CMAKE_Fortran_COMPILER_ID STREQUAL GNU AND CMAKE_Fortran_COMPILER_VERSION VER
   message(FATAL_ERROR "GCC Version 9 or newer required")
 endif()
 
+# Convert CMAKE_SYSTEM_NAME to uppercase
```
perazz: It is not currently possible to pass system-specific pre-processor flags to an `fpm` build of `stdlib`. So, I suggest CMake-determined macros should be removed.

config/fypp_deployment.py
```diff
@@ -115,6 +116,7 @@ def fpm_build(args,unknown):
     for idx, arg in enumerate(unknown):
         if arg.startswith("--flag"):
             flags= flags + unknown[idx+1]
+    flags = flags + "-D{}".format(platform.system().upper())
```
perazz: System-dependent pre-processor flags cannot be passed to the `stdlib-fpm-*` branches.

doc/specs/stdlib_system.md
```diff
@@ -532,3 +532,175 @@ The file is removed from the filesystem if the operation is successful. If the o
 ```fortran
 {!example/system/example_delete_file.f90!}
 ```
+
+## `joinpath` - Joins the provided paths according to the OS
```
perazz: Quoting https://github.com/fortran-lang/stdlib/pull/939#pullrequestreview-2638507159, I agree it would be a good idea to keep the naming convention as homogeneous as possible. In this case it would be `join_path` with snake case.

src/stdlib_system.F90
```diff
@@ -83,7 +83,22 @@ module stdlib_system
 public :: kill
 public :: elapsed
 public :: is_windows
-     
+
+!! Public path related functions and interfaces
```
perazz, line 88: Because of the said limitations, I would discourage from defining these `parameter`s with pre-processor flags. I would suggest the same approach for [OS_TYPE](https://github.com/fortran-lang/stdlib/blob/ca867f8134ac32352f92d92c57d70b23c759e8d7/src/stdlib_system.F90#L743) be also adopted here. This is not going to have any impact on computational performance.

src/stdlib_system.F90
```diff
@@ -83,7 +83,22 @@ module stdlib_system
...
+#ifdef WINDOWS
+    character(len=1), parameter, public :: pathsep = '\'
+    logical, parameter, public :: ISWIN = .true.
+#else
+    character(len=1), parameter, public :: pathsep = '/'
+    logical, parameter, public :: ISWIN = .false.
```
perazz: `ISWIN` looks like a duplicate definition for `OS_TYPE()==OS_WINDOWS`, so I suggest to remove it altogether and use what `stdlib` already offers at this stage. `pathsep` or rather `path_sep` would become a function: how about just `sep` such as in Matlab? Or, something familiar to other libraries would be fine imho.

src/stdlib_system.F90
```diff
@@ -83,7 +83,22 @@ module stdlib_system
...
+#endif
+
+public :: joinpath
+public :: operator(/)
+public :: splitpath
+public :: basename
+public :: dirname
```
perazz: Could you please add short FORD comment blocks for theese APIs? As an example: https://github.com/fortran-lang/stdlib/blob/ca867f8134ac32352f92d92c57d70b23c759e8d7/src/stdlib_system.F90#L102-L116

test/system/test_path.f90
```diff
@@ -0,0 +1,145 @@
...
+
+        if (ISWIN) then
+            path = joinpath('C:\Users', 'Alice')
+            call checkpath(error, 'joinpath', 'C:\Users\Alice', path)
+            if (allocated(error)) return
+
+            paths = [character(20) :: 'C:','Users','Bob','Pictures','2025']
+            path = joinpath(paths)
+
+            call checkpath(error, 'joinpath', 'C:\Users\Bob\Pictures\2025', path)
```
perazz: One of the usual issues with MS Windows is paths with spaces. What will happen if we try to join `"C:\Users\John Doe"` (note the quotes) + `Pictures` + `2025`?

src/stdlib_system.F90
```diff
@@ -550,6 +565,87 @@ end function process_get_ID
...
+    !! A binary operator to join the paths provided according to the OS-specific path-separator
+    !! ([Specification](../page/specs/stdlib_system.html#operator(/)))
+    !!
+    module pure function join_op(p1, p2) result(path)
+        character(:), allocatable :: path
+        character(*), intent(in) :: p1, p2
+    end function join_op
+end interface operator(/)
+
+interface splitpath
```
perazz: I suggest to rename this to `split_path` for stdlib style guide.

src/stdlib_system.F90
```diff
@@ -550,6 +565,87 @@ end function process_get_ID
...
+    !! If the path only consists of separators, `head` is set to the separator and tail is empty
+    !! If the path is a root directory, `head` is set to that directory and tail is empty
+    !! `head` ends with a path-separator iff the path appears to be a root directory or a child of the root directory
+    module subroutine splitpath(p, head, tail)
+        character(*), intent(in) :: p
+        character(:), allocatable, intent(out) :: head, tail
+    end subroutine splitpath
+end interface splitpath
+
+interface basename
```
perazz: I suggest to rename this to `base_name` for consistency with the style guide.

src/stdlib_system.F90
```diff
@@ -550,6 +565,87 @@ end function process_get_ID
...
+    !!
+    !!### Description
+    !! The value returned is the `tail` of the interface `splitpath`
+    module function basename(p) result(base)
+        character(:), allocatable :: base
+        character(*), intent(in) :: p
+    end function basename
+end interface basename
+
+interface dirname
```
perazz: I suggest to rename this to `dir_name` for consistency with the style guide.

perazz (changes requested):
Nice start @wassup05, thank you for this PR. It's a bit long so I tried to give some initial comments first hand. Let's first discuss and iterate around these comments, and then we can start planning about more path functionality, thank you.

src/stdlib_system.F90
jalvesz: If the documentation is added here, it should be placed following the `interface` declaration or within the `function` declaration if there is no additional interface, and not before the `public` statement as it is done within this module currently. I actually have issues with FORD locally to produce the documentation because of that but haven't mention it yet to avoid mixing issues within the same PR.

perazz: Yes, it is a good idea to address (these and perhaps more of) FORD issues in a separate PR. For the current PR, I think it's ok to stay consistent with the current syntax.

wassup05: I have added the docs following the `interface` blocks of these interfaces... Isn't that enough?

test/system/test_path.f90
wassup05: Its the same behaviour, it should return `"C:\Users\John Doe\Pictures\2025"` which is what `ntpath.join` of `Python` returns as well...

perazz: Sounds good - could you please add one simple test for it?

src/stdlib_system.F90
wassup05: I think `is_windows()` would also work fine here?

wassup05: I read the code for the `get_runtime_os` and I think I get what you mean...

CMakeLists.txt
```diff
@@ -31,6 +31,12 @@ if(CMAKE_Fortran_COMPILER_ID STREQUAL GNU AND CMAKE_Fortran_COMPILER_VERSION VER
...
+string(TOUPPER "${CMAKE_SYSTEM_NAME}" SYSTEM_NAME_UPPER)
+
+# Pass the uppercase system name as a macro
+add_compile_options(-D${SYSTEM_NAME_UPPER})
+
```
perazz: These build flags are not used anymore, could you please remove them?

config/fypp_deployment.py
perazz: This build flag is not used anymore, could you please remove it?

test/system/test_path.f90
```diff
@@ -0,0 +1,149 @@
...
+            call checkpath(error, 'join_path', 'C:\Users\Alice', path)
+            if (allocated(error)) return
+
+            paths = [character(20) :: 'C:','Users','Bob','Pictures','2025']
+            path = join_path(paths)
+
+            call checkpath(error, 'join_path', 'C:\Users\Bob\Pictures\2025', path)
+            if (allocated(error)) return
+
+            path = join_path('C:\Users\John Doe', 'Pictures\2025') ! path with spaces
```
perazz: Could we please check the address I was referring to?
```suggestion
            path = join_path('"C:\Users\John Doe"', 'Pictures\2025') ! path with spaces
```

perazz (changes requested):

test/system/test_path.f90
wassup05: I don't think I get the additional quotes.. `'C:\Users\John Doe'` is a valid path if the quotes are included when doing `cd` etc your suggestion would result in `'"C:\Users\John Doe"\Pictures\2025'` (which is what `Python` returns as well) But I don't think it is a valid path or of any practical significance

config/fypp_deployment.py
wassup05: I'm sorry I forgot, I'll remove them

test/system/test_path.f90
perazz: What I meant is that on Windows, passing paths with spaces to the environment requires an analysis of the characters and appropriate quoting of the global path (and possibly of its chunks, as I was suggesting here). So I suggest we assume that `join_path` only wants clean paths as input (no quotes, or already de-quoted). Later, imho we want a `quote_path` or similar function that is capable of evaluating the path and appropriately creates quotes so that it can be passed to the shell

perazz (commented):
Thank you for implementing these updates @wassup05.

doc/specs/stdlib_system.md
```diff
@@ -532,3 +532,175 @@ The file is removed from the filesystem if the operation is successful. If the o
...
+`res = ` [[stdlib_system(module):join_path(interface)]] ` (p1, p2)`
+
+`res = ` [[stdlib_system(module):join_path(interface)]] ` (p)`
+
+### Class
+Pure function
+
+### Arguments
+
+`p1, p2`: Shall be a character string. It is an `intent(in)` argument.
```
perazz: Because stdlib offers a string type, it would be nice if these string manipulation functions were `interfaces` working with either character strings (as currently proposed), or the internal `string_type` type.

doc/specs/stdlib_system.md
```diff
@@ -532,3 +532,175 @@ The file is removed from the filesystem if the operation is successful. If the o
...
+`call `[[stdlib_system(module):split_path(interface)]]`(p, head, tail)`
+
+### Class
+Subroutine
+
+### Arguments
+
+`p`: A character string containing the path to be split. `intent(in)`
+`head`: The first part of the path. `allocatable, intent(out)`
+`tail`: The rest part of the path. `allocatable, intent(out)`
```
perazz:
```suggestion
`p`: A character string containing the path to be split. It is an `intent(in)` argument.
`head`: The first part of the path. It is an `allocatable, intent(out)` argument.
`tail`: The rest part of the path. It is an `allocatable, intent(out)` argument.
```

doc/specs/stdlib_system.md
```diff
@@ -532,3 +532,175 @@ The file is removed from the filesystem if the operation is successful. If the o
...
+
+### Behavior
+
+- If `p` is empty, `head` is set to `.` and `tail` is empty
```
perazz:
```suggestion
- If `p` is empty, `head` is set to `.` and `tail` is left empty.
```

doc/specs/stdlib_system.md
```diff
@@ -532,3 +532,175 @@ The file is removed from the filesystem if the operation is successful. If the o
...
+- If `p` consists entirely of path-separators. `head` is set to the path-separator and `tail` is empty
+- `head` ends in a path-separator if and only if `p` appears to be a root directory or child of one
```
perazz:
```suggestion
- If `p` consists entirely of path-separators, `head` is set to the path-separator and `tail` is left empty.
- `head` ends with a path-separator if and only if `p` appears to be a root directory or child of one.
```

doc/specs/stdlib_system.md
```diff
@@ -532,3 +532,175 @@ The file is removed from the filesystem if the operation is successful. If the o
...
+### Syntax
+
+`res = ` [[stdlib_system(module):base_name(interface)]]`(p)`
+
+### Class
+Function
+
+### Arguments
+
+`p`: the path, a character string, `intent(in)`
```
perazz:
```suggestion
`p`: the path, a character string. It is an `intent(in)` argument.
```

doc/specs/stdlib_system.md
```diff
@@ -532,3 +532,175 @@ The file is removed from the filesystem if the operation is successful. If the o
...
+### Syntax
+
+`res = ` [[stdlib_system(module):dir_name(interface)]]`(p)`
+
+### Class
+Function
+
+### Arguments
+
+`p`: the path, a character string, `intent(in)`
```
perazz:
```suggestion
`p`: the path, a character string. It is an `intent(in)` argument.
```

doc/specs/stdlib_system.md
```diff
@@ -532,3 +532,175 @@ The file is removed from the filesystem if the operation is successful. If the o
...
+
+### Behavior
+
+- The `head` of `stdlib_system(module):split_path(interface)` is exactly what is returned. Same Behavior.
```
perazz: please add a link to `split_path` in FORD syntax.

perazz (commented):
Thank you for this PR @wassup05, I have left a few final comments.

doc/specs/stdlib_system.md
wassup05: I made these functions interfaces keeping this possibility in mind... But what is considered a good way of doing this? Writing procedures for `string_type` and then, when calling them with `character` variables just assign these to the new `string_type` variables and call the `string_type` version of the interface itself or code duplication, keeping the code separate for these two although they are in essence the same?

wassup05: Ah, I got it!

perazz: Yes: the easiest option is to keep the `character` implementation and then return the result to a string via `move`: https://github.com/fortran-lang/stdlib/blob/d8a90aa878afb4cc315d81384d9077285a6ebd2d/src/stdlib_string_type.fypp#L220-L227

wassup05: I have added the relevant changes... But it doesn't use `move`, it uses `assignment(=)` and `char`, Let me know what you think of it.

perazz: It works @wassup05, however, there are two copies: `string->char` for the input arguments, and then `char -> string` on the assignment. Using `move` would at least avoid incurring the last copy. Just do:
```
character(:), allocatable :: join_char
join_char = join_path(char(a),char(b))
call move(from=join_char,to=...result variable...)
```

perazz: Because the internal `string_type` representation is just an allocatable character variable, I would also support (to be discussed with the community as a separate PR) the implementation of a tiny zero-copy "view" function for the string type, [see here](https://godbolt.org/z/dvaK7Tc1E): 

```fortran
    !> Zero-copy view of the string as a character array
    pure function view(string) result(char_array)
        type(string_type), intent(in), target :: string
        character(:), pointer :: char_array
        if (allocated(string%raw)) then
            char_array => string%raw
        else
            nullify(char_array)
        end if
    end function view
```

perazz:
In absence of further comments, I will merge this one, thank you.

PR #1000, opened 2025-06-08
fix: to_num_from_stream pointer position correction

jalvesz:
This PR fixes the computation of the position in a character stream for the `to_num_from_stream` function to properly advance when the character is a long chain. Otherwise, negative numbers can be obtained from `len(s, lind=int8)` if the string is in effect larger than 127.

### DIFF
--- a/src/stdlib_str2num.fypp
+++ b/src/stdlib_str2num.fypp
@@ -109,7 +109,7 @@ module stdlib_str2num
         integer(int8) :: err
         !----------------------------------------------
         call to_num_base(s,v,p,err)
-        p = min( p , len(s, kind = int8) )
+        p = min( p , len(s) )
         s => s(p:)
         if(present(stat)) stat = err
     end function


src/stdlib_str2num.fypp
```diff
@@ -109,7 +109,7 @@ module stdlib_str2num
         integer(int8) :: err
         !----------------------------------------------
         call to_num_base(s,v,p,err)
-        p = min( p , len(s, kind = int8) )
+        p = min( p , len(s) )
```
jvdp1, line 112: As `p` is declared as `integer(int8)`, similar issues might still happen.  Maybe also declare `p` as `integer :: p`? Or do you think it is not needed?

jvdp1 (approved):
Thank you @jalvesz . LGTM. See my suggestion below.

src/stdlib_str2num.fypp
jalvesz: It is not needed as the kind is converted as an output of `min`. It could happen if an individual number is written on more than 127 characters. But even a quad number would fit.

PR #1006, opened 2025-06-17
feat: filesystem error handling

wassup05:
A derived type `fs_error` containg an integer `code` and a fixed-length string is added for concise error handling

This is kind of a stripped down version of `state_type` already present in the `stdlib`, syntax is also kept very similar... except the introduction of `code` to store integers returned by C functions such as `GetLastError` (Windows API), global variables like `errno` (Posix platforms)

Do let me know your thoughts


wassup05:
> My understanding is that the current implementation would store platform-dependent error code, that I would refrain from doing. Instead, IMHO platform wrappers would be useful to standardize error messages

As an example, on my main system (linux) in `errno.h` there are 135 errors defined such as `EACCES`, `ENOTDIR` etc and these are NOT the same across all platforms, meaning the macro defining `EACCES` could be defined using a different integer... It is simply not feasible to cover all these cases separately, That's why I feel giving out the code of the error to the user (as an `integer` preferably) would be very beneficial as in the user has a way at least to figure out what is happening, using his OS specifications etc instead of probably a vague `Unkown error occured` in case the error is too uncommon.

ofcourse user friendly messages are required and for that you might want to look at #1011 where I have given out both the code and the message as a string, directly binding to the CRT function [`strerr`](https://man7.org/linux/man-pages/man3/strerror.3.html) which is available on both Windows and Linux, which converts the `errno` to a `char*`

And of course having useful developer extensions as you pointed out would be desirable.

PR #957, opened 2025-03-20
Tridagonal matrices and associated spmv kernel

loiseaujc:
This PR is the first of a series to port what I had been doing in [`SpecialMatrices`](https://github.com/loiseaujc/SpecialMatrices/tree/main) to `stdlib`. This first one provides the base type and spmv kernel for `Tridiagonal` matrices. At the moment, it is included in `stdlib_sparse_kinds` and `stdlib_sparse_spmv` files out of simplicity. It could also be (yet) another module, I don't have a strong opinion about that.

### Proposed interface

- `A = Tridiagonal(dl, dv, du)` where `dl`, `dv` and `du` are rank-1 arrays describing the tridiagonal elements.
- `call spmv(A, x, y [, alpha] [, beta] [, op])` for the matrix-vector product.

### Key facts

- The `spmv` kernel uses `lagtm` LAPACK backend under the hood.
- Tridiagonal matrices of all the kinds provided by `stdlib` can be created.
- The kind generality of the `spmv` kernel has the same limitations as `stdlib_linalg_lapack`.

### Progress

[x] Base implementation
[x] Tests
[ ] Documentation
[ ] Examples

cc @jvdp1, @jalvesz, @perazz

Issue #10, opened 2019-12-16
Proposal for linalg


certik:
`eig`, `eigh`, `inv`, `solve`, `det`, `svd`, ... . A possible implementation is here: https://github.com/certik/fortran-utils/blob/b43bd24cd421509a5bc6d3b9c3eeae8ce856ed88/src/linalg.f90.

In particular, let's start with `eig`, for an eigenvalue and eigenvectors of a general (non-symmetric) matrix. Both NumPy and Matlab have a very similar interface called `eig`, that I propose we use:

* NumPy: https://docs.scipy.org/doc/numpy/reference/generated/numpy.linalg.eig.html
* Matlab: https://www.mathworks.com/help/matlab/ref/eig.html

Julia seem to have more of an "object oriented" interface called `eigen`: https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/index.html#LinearAlgebra.eigen, which uses some Julia language features to emulate the Matlab style `vals, vecs = eigen([1.0 0.0 0.0; 0.0 3.0 0.0; 0.0 0.0 18.0])`.

certik:
Let's discuss the API for `eig`. Fortran cannot return more than one result in a function, and `eig` must return both eigenvalues and eigenvectors. So the only solution is a subroutine. The first argument should be the input matrix, to be consistent with NumPy and Matlab. The second argument can be an optional `B` matrix for the generalized eigenvector problem. The next argument can be eigenvalues and the last argument eigenvectors (NumPy style) or they can be switched (Matlab style). For efficiency reasons the eigenvalues should be just a vector (NumPy style) not a diagonal matrix (Matlab style).

That reasoning leads to the following API:
```fortran
  interface eig
     module procedure seig
     module procedure seig_generalized
     module procedure deig
     module procedure deig_generalized
     module procedure zeig
     module procedure zeig_generalized
  end interface eig
```
Here `s` means single precision, `d` double, `z` complex (double). Here is an example of a double precision interface:
```fortran
    subroutine deig(A, lam, c)
    ! Solves a standard eigenproblem A*c = lam*c
    real(dp), intent(in) :: A(:, :)  ! matrix A
    complex(dp), intent(out) :: lam(:)  ! eigenvalues: A c = lam c
    complex(dp), intent(out) :: c(:, :)  ! eigenvectors: A c = lam c; c(i,j) = ith component of jth vec.
    ...
    end subroutine

    subroutine deig_generalized(A, B, lam, c)
[...]

PR #1007, opened 2025-06-27
fix: CI issue with test-drive and gnu 15

jalvesz:
Hacking away the CI for gnu 15: https://github.com/fortran-lang/stdlib/pull/957#issuecomment-3013754809

### DIFF
--- a/.github/workflows/ci_windows.yml
+++ b/.github/workflows/ci_windows.yml
@@ -40,7 +40,7 @@ jobs:
         -Wdev
         -B build
         -DCMAKE_BUILD_TYPE=Debug
-        -DCMAKE_Fortran_FLAGS_DEBUG="-Wall -Wextra -Wimplicit-interface -fPIC -g -fcheck=all -fbacktrace"
+        -DCMAKE_Fortran_FLAGS_DEBUG="-Wall -Wextra -Wno-external-argument-mismatch -Wimplicit-interface -fPIC -g -fcheck=all -fbacktrace"
         -DCMAKE_MAXIMUM_RANK:String=4
         -DCMAKE_INSTALL_PREFIX=$PWD/_dist
         -DFIND_BLAS:STRING=FALSE


jalvesz:
@Romendakil @perazz  I tried here to address the issue with gcc 15. It builds and runs but now there are several tests failing. Any advise?
cc @loiseaujc

jalvesz:
I'm tempted to say that this version of gnu 15.1.0 is bugged ... I had it and reproduced these issues, I just downgraded locally to gnu 14.3.0 and all tests pass. If someone knows how to bring these issues to the attention of gcc developers it would be great.

For `stdlib` CI maybe we should consider downgrading to gcc 14.3.0

Romendakil:
> I'm tempted to say that this version of gnu 15.1.0 is bugged ... I had it and reproduced these issues, I just downgraded locally to gnu 14.3.0 and all tests pass. If someone knows how to bring these issues to the attention of gcc developers it would be great.
> 
> For `stdlib` CI maybe we should consider downgrading to gcc 14.3.0

As I said, this problem is fixed. Of course, this single version does have this problem. Actually, I think the official 15.0 release doesn't.

perazz:
@jalvesz some debugging shows that another (!) gcc-15 bug is being hit when self-growing an array using the array initializer:

```
       tests = [tests,new_unittest("$eye_det_rsp",test_rsp_eye_determinant)]
        tests = [tests,new_unittest("$eye_det_multiple_rsp",test_rsp_eye_multiple)]
        tests = [tests,new_unittest("$eye_det_rdp",test_rdp_eye_determinant)]
      ...
``` 

as also addressed in fpm here https://github.com/krystophny/fpm/blame/928fae3160039d42c0ed07736bb8512d30028113/src/metapackage/fpm_meta_mpi.f90#L770

(something like `a = [a, b]` will crash)

jalvesz:
Thanks @perazz for finding this out! This looks like a serious regression bug on the gfortran's end.

All of the tests are written using this syntax. And it is not only `stdlib` that is concerned...

Should we change all of the tests to avoid recursive allocation? Doesn't look reasonable.

I think this problem should be brought to gfortran developers attention and in the meantime time we downgrade the gcc version in stdlib's CI.

Romendakil:
> @jalvesz some debugging shows that another (!) gcc-15 bug is being hit when self-growing an array using the array initializer:
> 
> ```
>        tests = [tests,new_unittest("$eye_det_rsp",test_rsp_eye_determinant)]
>         tests = [tests,new_unittest("$eye_det_multiple_rsp",test_rsp_eye_multiple)]
>         tests = [tests,new_unittest("$eye_det_rdp",test_rdp_eye_determinant)]
>       ...
> ```
> 
> ```
> (lldb) bt
> * thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
>   * frame #0: 0x000000018f741388 libsystem_kernel.dylib`__pthread_kill + 8
>     frame #1: 0x000000018f77a88c libsystem_pthread.dylib`pthread_kill + 296
>     frame #2: 0x000000018f683c60 libsystem_c.dylib`abort + 124
>     frame #3: 0x000000018f588174 libsystem_malloc.dylib`malloc_vreport + 892
>     frame #4: 0x000000018f58bc90 libsystem_malloc.dylib`malloc_report + 64
>     frame #5: 0x000000018f59021c libsystem_malloc.dylib`___BUG_IN_CLIENT_OF_LIBMALLOC_POINTER_BEING_FREED_WAS_NOT_ALLOCATED + 32
>     frame #6: 0x00000001000010a8 test_linalg_determinant`__test_linalg_determinant_MOD_test_matrix_determinant at test_linalg_determinant.f90:25:83
>     frame #7: 0x000000010015bc08 test_linalg_determinant`__testdrive_MOD_run_testsuite at testdrive.F90:328:27
>     frame #8: 0x00000001000072f4 test_linalg_determinant`MAIN__ at test_linalg_determinant.f90:533:68
>     frame #9: 0x000000010015f4c0 test_linalg_determinant`main at test_linalg_determinant.f90:518:40
>     frame #10: 0x000000018f3dab98 dyld`start + 6076
> ```
> 
> as also addressed in fpm here https://github.com/krystophny/fpm/blame/928fae3160039d42c0ed07736bb8512d30028113/src/metapackage/fpm_meta_mpi.f90#L770
> 
> (something like `a = [a, b]` will crash)

This seems to be known as well to the team and a bugfix is under testing right now: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=120711
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=120656

perazz:
@Romendakil great to see the team is working on it. However, people will stumble upon these issues on systems that ship gcc 15.1.0, and we don't want stdlib users to deal with compiler stuff. So @jalvesz I would advise to have workarounds in place across the library and I endorse this PR.

Romendakil:
> @Romendakil great to see the team is working on it. However, people will stumble upon these issues on systems that ship gcc 15.1.0, and we don't want stdlib users to deal with compiler stuff. So @jalvesz I would advise to have workarounds in place across the library and I endorse this PR.

Yes, but note that this is not in the official 15.1 release from gcc. So it is an inconsistent version that you get from somewhere

perazz:
@Romendakil yep, from the CI it would look like MSYS2 ships that version. We are not testing against gcc-15 on other systems yet, though.

PR #1008, opened 2025-06-28
GCC 15.1 fixes

perazz:
Introduces workarounds for known gcc-15 issues: 
- tests: replace array constructors `tests = [tests, new_unittest(...)]` with a tiny helper function `add_test`
- strings: do not use array constructors `[string_t(), string_t(),...]`
- build: add `-Wno-external-argument-mismatch` (to be replaced by PR #1007 when merged). 

cc: @jvdp1 @zoziha @jalvesz

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,13 @@ if(CMAKE_Fortran_COMPILER_ID STREQUAL GNU AND CMAKE_Fortran_COMPILER_VERSION VER
   message(FATAL_ERROR "GCC Version 9 or newer required")
 endif()
 
+# --- silence gfortran-15 argument-mismatch warnings
+if(CMAKE_Fortran_COMPILER_ID STREQUAL GNU
+   AND CMAKE_Fortran_COMPILER_VERSION VERSION_GREATER_EQUAL 15.0
+   AND CMAKE_Fortran_COMPILER_VERSION VERSION_LESS 16.0)
+    add_compile_options("$<$<COMPILE_LANGUAGE:Fortran>:-Wno-external-argument-mismatch>")
+endif()
+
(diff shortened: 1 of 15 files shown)


jalvesz:
>build: add -Wno-external-argument-mismatch (to be replaced by PR https://github.com/fortran-lang/stdlib/pull/1007 when merged).

@perazz the PR with only this modification is ready. Would you preferer to merge that one and then bring it here or simply takeover those changes directly in this PR? we can simply discard the other PR.

zoziha (approved):
Thank you for this fix, @perazz. I think this patch addresses the current bug in GCC15.1.0.

I have another way to deal with `a = [a, b]` that occurred during the merging of the `unittest_type` variables in `testdrive` tests, which may be helpful for developers who write new tests for the `stdlib` in the future:

```Fortran

!> Cholesky factorization tests

subroutine test_cholesky_factorization(tests)
    !> Collection of tests
    type(unittest_type), allocatable, intent(out) :: tests(:)
    type(unittest_type) :: empty(0)  ! gcc-15 bugfix 

    tests = [ empty &
    #:for rk,rt,ri in RC_KINDS_TYPES
    , new_unittest("least_cholesky_${ri}$",test_cholesky_${ri}$) &
    #:endfor
    ]

end subroutine test_cholesky_factorization

```

perazz:
> Would you preferer to merge that one and then bring it here or simply takeover those changes directly in this PR? we can simply discard the other PR.

I'm fine either way @jalvesz - setting `-Wno-external-argument-mismatch` via CMake seems more future-proof for when gcc-15 will be part of the unix and Mac CIs, so you may want to port that to #1007 or let's just merge this one.

jalvesz (approved):

PR #1012, opened 2025-07-02
fix: index handling in `chomp`

wassup05:
To prevent any indices from becoming non-positive

Cause of CI failures in #999

### DIFF
--- a/src/stdlib_strings.fypp
+++ b/src/stdlib_strings.fypp
@@ -303,8 +303,12 @@ contains
         last = len(string)
         nsub = len(substring)
         if (nsub > 0 .and. nsub <= last) then
-            do while(last >= nsub .and. string(last-nsub+1:last) == substring)
-                last = last - nsub
+            do while(last >= nsub)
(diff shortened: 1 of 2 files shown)


src/stdlib_strings.fypp
```diff
@@ -302,11 +302,17 @@ contains
...
                 last = last - nsub
             end do
         end if
+
+        if (last <= 0) then
+            chomped_string = ''
+            return
+        end if
+
         chomped_string = string(1:last)
```
perazz:
```suggestion
        else
            chomped_string = string(1:last)
        end if        
```

jalvesz:
Thanks @wassup05 for this PR, seems like this was the only procedure from the interface that is not being tested. Could you add 1 test to verify its behavior in `test\string\test_string_strip_chomp.f90` ?

perazz (commented):
Thank you for this PR @wassup05. I agree on @jalvesz's comment to add a simple test case before this can be merged.

jalvesz (approved):

PR #1015, opened 2025-07-12
helper functions for Filesystem error handling

wassup05:
Functions added are:
- `fs_error`: just sets the `flag` of `state_type` to `STDLIB_FS_ERROR` and passes the remaining 20 arguments to it
- `fs_error_code`: sets the `flag` as above and also prefixes the `code` (`integer` argument) accordingly.

I tried using an `interface` block to handle both of these together, but that introduces ambiguity for the compiler.

Could make `code` an `optional` argument and depending on if it's present or not pass 18 or 20 arguments to `state_type` but I feel that the 2 unused arguments in that case would not be correct.

Inspired from @perazz in https://github.com/fortran-lang/stdlib/pull/1006#issuecomment-3034714220


wassup05:
Oh no, It was just because it was an error and I felt it was better to emphasize that, but I'll change it

jalvesz (approved):
Thanks @wassup05 LGTM

doc/specs/stdlib_system.md
```diff
@@ -418,6 +418,83 @@ Returns one of the `integer` `OS_*` parameters representing the OS type, from th
...
+
+`err = fs_error([a1,a2,a3,a4...... a20])`
+
+### Class
+Pure Function
+
+### Arguments
+
+`a1,a2,a3.....a20`(optional) : They are of type `class(*), dimension(..), optional, intent(in)`. 
+An arbitrary list of `integer`, `real`, `complex`, or `character` variables. Numeric variables may be provided as either scalars or rank-1 (array) inputs.
```
perazz:
```suggestion
An arbitrary list of `integer`, `real`, `complex`, `character` or `string_type` variables. Numeric variables may be provided as either scalars or rank-1 (array) inputs.
```

example/system/example_fs_error.f90
```diff
@@ -0,0 +1,23 @@
...
+    err0 = fs_error("Could not create directory", "`temp.dir`", "- Already exists")
+
+    if (err0%state == STDLIB_FS_ERROR) then
+        ! Error encountered: Filesystem Error: Could not create directory `temp.dir` - Already exists
+        print *, err0%print() 
+    end if
+
+    err1 = fs_error_code(1, "Could not create directory", "`temp.dir`", "- Already exists")
+
+    if (err1%state == STDLIB_FS_ERROR) then
```
perazz: The lowercase (function) vs. uppercase (parameter) syntax clashes imho @jalvesz. Becuase the function wraps a parameter, I think it would be more consistent have capital letters. However, not a blocker.

perazz (commented):
Thanks for this PR @wassup05. I added a few comments.

perazz (approved):
LGTM, thank you @wassup05.

Issue #1016, opened 2025-07-17
`stdlib_dot_product` and `stdlib_dot_product_kahan` return erroneous results for complex vectors


loiseaujc:
### Description

The implementations of the improved `dot_product` kernels work like a charm for real vectors but return erroneous results for complex ones. It seems like the complex versions are missing a complex-conjugate operation. See the following code snippet for reproducibility.

```fortran
program main
   use stdlib_linalg_constants
   use stdlib_intrinsics
   implicit none

   integer(ilp), parameter :: n = 128
   complex(dp), dimension(n) :: x, y
   real(dp), dimension(n, 2) :: z

   call random_number(z); x%re = z(:, 1); x%im = z(:, 2)
   call random_number(z); y%re = z(:, 1); y%im = z(:, 2)

   print *, dot_product(x, y)
   print *, stdlib_dot_product(x, y)
   print *, stdlib_dot_product_kahan(x, y)
end program main
```

### Expected Behaviour

Results between `stdlib_dot_product`, `stdlib_dot_product_kahan` and the Fortran intrinsic `dot_product` should be identical (up to a multiple of the machine precision).

### Version of stdlib

latest

### Platform and Architecture

Ubuntu 22.04

### Additional Information

Ping @perazz, @jalvesz

jalvesz:
humm indeed ... here is a CE snippet to test it https://godbolt.org/z/Mza5P7bae ... So it would seem that a final  `conj( )` is missing though it looks strange as `X*Y = sum_{i=1,N}( x_i * conj(y_j) )` by definition, right?

loiseaujc:
It is actually the opposite: <x, y> = x.H @ y

bjbraams:
If I read it correctly the test program for this Kahan dot product and the Kahan sum only checks that the Kahan implementation gives a result consistent with standard floating point arithmetic, not that it gives a result of the desired high accuracy. (I am looking at test/intrinsics/test_intrinsics.fypp and under the tests for the real sum and dot product I notice the statement `tolerance = epsilon(1._${k}$)*100`.) In some cases it might be enough to guard against gross errors, as these tests would do, but Kahan summation is vulnerable to subtle errors too. Or maybe my reading is wrong.

jalvesz:
@bjbraams sorry, I didn't understand your question or proposal. Would you mind reformulating?

[...]

PR #1017, opened 2025-07-17
fix: complex dot_product formulation

jalvesz:
addresses #1016 

cc: @loiseaujc


loiseaujc (approved):
Looks good to me.

test/intrinsics/test_intrinsics.fypp
```diff
@@ -246,6 +246,27 @@ subroutine test_dot_product(error)
         call check(error, all(err(:)<tolerance) , "complex dot_product is not accurate" )
         if (allocated(error)) return
     end block
+
+    block ! test for https://github.com/fortran-lang/stdlib/issues/1016
+        ${t}$ :: x(128), y(128)
+        real(${k}$) :: z(128,2)
+        real(${k}$), parameter :: tolerance = epsilon(1._${k}$)*100000
```
perazz, line 253: For a 32-bit precision dot product, this error is ~0.01, which seems a bit large? I believe using `sqrt(epsilon(0.0_${k}$))` would be a more established error threshold in numerical analysis. LGTM otherwise.

PR #1018, opened 2025-07-20
Make collaboration script executable

wassup05:


perazz (approved):

banana-bred:
Thanks for letting me know, consider it done. I did not think this would affect the base stdlib repo at all, but I see all the mentions now 😅😅 definitely not intended.

PR #1019, opened 2025-07-20
Unlimited polymorphic hashmap key type

chuckyvt:
I ran across [this](https://community.intel.com/t5/Intel-Fortran-Compiler/unlimited-polymorphic-parameter-and-transfer-intrinsic/td-p/1062112) thread on Intel Fortran forums, which based on their interpretation of the standard, the transfer intrinsic should accept an unlimited polymorphic argument.  Was excited to see this, as this could potentially simplify the hashmap code, and allow the routine to accept any type as key value.   This PR incorporates that work.   

However, based on the CI runs, while Intel compilers support it, GFortran does not and has a compilation failure.   Based on the need for the stdlib to work with GFortran, I assume this PR is dead in the water for the time being, but I thought it still worth submitting for reference.   

Based on this [bugreport](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=98534) it appears there has been some effort on intrinsic support of unlimited polymorhpic variables, if anyone would like to file a follow up bug report based on this.


jalvesz:
@chuckyvt I think it is better if you keep this PR in **draft** state until a solution is found that enables compiling with gfortran as well.

Indeed, stdlib should not be locked by a single compiler, this of course implies that only features that are supported by at least a few versions of intel and gnu compilers will be allowed. It might hold back on certain aspects but also guarantees a certain level of robustness. You could explore options with C preprocessor macros.

PR #1025, opened 2025-08-01
Hashmap test update

chuckyvt:
As part of development work on the stdlib hashmaps, I have run into CI failures on the stdlib test that I'm not sure if it's due to code updates, or because of format of hashmap test program.  For example, the way the current implantation uses transfer to generate character values, I'm not sure it's always generating valid character values.  Also fypp makes it harder to develop and update the test routine as the hashmaps evolve.   

This PR is an attempt to address those shortcomings.  The intent is to improve test coverage of hashmap functions, improve test quality (ensure valid data is being used for key and values), and simplify the code and remove fypp preprocessor format.


chuckyvt:
This PR also makes the abstract hashmap_type public so that can be used for interface definition that works for both open and chaining types.   I thought it was worth adding a simple example showing the use of that.

example/hashmaps/example_hashmaps_abstract_type.f90
```diff
@@ -0,0 +1,53 @@
+
+! For procedure interfaces, consider using abstract hashmap_type for interface definition. 
+! This allows the procedure to be used for both chaining and open hashmap types.
+    
+program example_abstract_type
+    use stdlib_kinds, only: int8, int64
```
jvdp1:
```suggestion
```

Both are not used in this example

example/hashmaps/example_hashmaps_abstract_type.f90
```diff
@@ -0,0 +1,53 @@
+
```
jvdp1:
```suggestion
```

Nice suggestion to make `hashmap_type` `public`

Could you update the specs, too, please?

jvdp1 (commented):
Thank you @chuckyvt for this PR. Overall it looks good to me. I have only minor suggestions

chuckyvt:
Thank you for the review.  Suggested updates have been applied.  The specs already listed the abstract hashmap_type as public, so I believe this change in essentially updated the code to the original intent.  I added an added line and a reference to the abstract type example.

perazz (commented):
What was the shortcoming of keeping `fypp` for this update? While I totally agree that working with templated code is hard (for the linear algebra part, many fypp functions had to be create), however I believe it ultimately makes everything more maintainable. So, it may be possible to keep the former fypp implementation wherever possible and perhaps only add non-templated code outside of the templated part?

PR #1026, opened 2025-08-02
feat: `exists` to return the type of a path

wassup05:
User facing stuff added are
- `integer` function `exists (path [, err])` to check if a file exists and return it's type.
- `integer` parameters, `type_unknown`, `type_regular_file`, `type_directory`, `type_symlink` which are the return values of the `exists` function.

To accomplish this, on unix [`lstat`](https://pubs.opengroup.org/onlinepubs/009695399/functions/lstat.html) is used and on windows [`GetFileAttributesA`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfileattributesa) is used.

1. If any errors, `type_unknown` is returned
2. the code and the string describing the code are provided through `FS_ERROR_CODE` (using `strerror` on unix and [`FormatMessageA`](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-formatmessagea) on windows).
3. and to accomplish this platform based error handling `c_getstrerr` now takes a `logical` `winapi` argument to switch between these two functions (`strerror` and `FormatMessageA`).


src/stdlib_system.F90
```diff
@@ -1134,6 +1177,31 @@ pure function FS_ERROR(a1,a2,a3,a4,a5,a6,a7,a8,a9,a10,a11,&
...
+
+    integer(kind=c_int) :: stat
+
+    fs_type = stdlib_exists(to_c_char(trim(path)), stat)
+
+    if (stat /= 0) then
+        err0 = FS_ERROR_CODE(stat, c_get_strerror())
+        call err0%handle(err)
+    end if
+end function exists
```
sebastian-mutz: I strongly recommend adding a few comments here (see `delete_file` in the same module for reference). While it's clear what's happening when reviewing the specific PR and being familiar with `stdlib_system`, you want to make sure it remains easily understandable and maintainable by the wider community.

doc/specs/stdlib_system.md
```diff
@@ -646,6 +646,51 @@ Subroutine
...
+This function performs a system call (syscall) to the operating system, to retrieve the metadata
+corresponding to the path, and identifies the type of path it is. 
+It can distinguish among the following path types
+
+- Regular File
+- Directory
+- Symbolic Link
+
+Returns a constant representing the path type or `type_unknown` if it cannot be determined.
+If there has been an error, It is handled using `state_type`.
```
sebastian-mutz: Suggestion for clarity:

```suggestion
This function makes a system call (syscall) to retrieve metadata for the specified path and determines its type.
It can distinguish between the following path types:

- Regular File
- Directory
- Symbolic Link

It returns a constant representing the detected path type, or `type_unknown` if the type cannot be determined. Any encountered errors are handled using `state_type`.
```

doc/specs/stdlib_system.md
```diff
@@ -646,6 +646,51 @@ Subroutine
...
+
+### Arguments
+
+`path`: Shall be a character string containing the path. It is an `intent(in)` argument.
+
+`err`(optional): Shall be of type `state_type`, and is used for error handling. It is an `optional, intent(out)` argument.
+
+### Return values
+
+`err` is an optional state return flag. On error if not requested, an `FS_ERROR` will trigger an error stop.
```
sebastian-mutz:
```suggestion
`err` is an optional state return flag. If it is not provided and an error occurs, an `FS_ERROR` will trigger an error stop.
```

example/system/example_exists.f90
```diff
@@ -0,0 +1,26 @@
...
+    end if
+
+    select case (t)
+    case (type_unknown); print *, "Unknown type!"
+    case (type_regular_file); print *, "Regular File!"
+    case (type_directory); print *, "Directory!"
+    case (type_symlink); print *, "Symbolic Link!"
+    end select
+end program example_exists
+
```
sebastian-mutz: Please also elaborate a little with comments, as was done for most of the system module examples. Examples is where many go first to understand a procedures, so let's make it easier for them :)

sebastian-mutz (changes requested):
Thanks for this PR, @wassup05!
Functionality seems fine; would strongly recommend commenting the code a little more here and there to make it easier to maintain down the line by people other than those intimately familiar with the system modules.

example/system/example_exists.f90
wassup05: https://github.com/fortran-lang/stdlib/pull/1026/commits/306bc8626a3af59a1810f70e7141cdb66ec0d68a

doc/specs/stdlib_system.md
wassup05: I have changed that to make it consistent with the other procedure's docs here - https://github.com/fortran-lang/stdlib/pull/1026/commits/306bc8626a3af59a1810f70e7141cdb66ec0d68a

doc/specs/stdlib_system.md
wassup05: https://github.com/fortran-lang/stdlib/pull/1026/commits/306bc8626a3af59a1810f70e7141cdb66ec0d68a

src/stdlib_system.F90
wassup05: I have added 2-3 comments, honestly they seem kind of forced to me because all this procedure does is declare an interface and call it, but let me know if it's fine

doc/specs/stdlib_system.md
```diff
@@ -646,6 +646,50 @@ Subroutine
...
+
+### Arguments
+
+`path`: Shall be a character string containing the path. It is an `intent(in)` argument.
+
+`err`(optional): Shall be of type `state_type`, and is used for error handling. It is an `optional, intent(out)` argument.
+
+### Return values
+
+`err` is an optional state return flag. If not requested and an error occurs, an `FS_ERROR` will trigger an error stop.
```
zoziha: Should the description of the return value `res` be included here?

src/stdlib_system.F90
```diff
@@ -204,6 +204,43 @@ module stdlib_system
...
+!> ([Specification](../page/specs/stdlib_system.html))
+integer, parameter, public :: &
+    !> Represents an unknown path type
+    type_unknown      = 0, &
+    !> Represents a regular file
+    type_regular_file = 1, &
+    !> Represents a directory
+    type_directory    = 2, &
+    !> Represents a symbolic link
+    type_symlink      = 3
```
perazz: `type` correctly flags filesystem items, but it may be confused with a Fortran type imho. Some ideas to rename these parameters are: `FS_TYPE_*` or `FILESYSTEM_TYPE_*` or `FS_ITEM_`

src/stdlib_system.c
```diff
@@ -44,3 +72,50 @@ int stdlib_remove_directory(const char* path){
...
+
+// Wrapper to the platform's `stat`(status of path) call.
+// Uses `lstat` on unix, `GetFileAttributesA` on windows.
+// Returns the `type` of the path, and sets the `stat`(if any errors).
+int stdlib_exists(const char* path, int* stat){
+    // All the valid types
+    const int type_unknown = 0;
+    const int type_regular_file = 1;
+    const int type_directory = 2;
+    const int type_symlink = 3;
```
perazz: See above: let's please keep the naming consistent between C and Fortran versions.

src/stdlib_system.c
```diff
@@ -44,3 +72,50 @@ int stdlib_remove_directory(const char* path){
...
+        // `lstat` failed
+        *stat = errno;
+        return type_unknown;
+    }
+
+    switch (buf.st_mode & S_IFMT) {
+        case S_IFREG: type = type_regular_file; break;
+        case S_IFDIR: type = type_directory;    break;
+        case S_IFLNK: type = type_symlink;      break;
+        default:      type = type_unknown;      break;
```
perazz: Should we differentiate between a _missing_ and an _invalid_ object? What do other libraries do in this respect?

perazz (commented):
Thank you for this PR @wassup05, I think the implementation is almost ready. The biggest comment I have is about harmonization with the rest of the library. 

We already have `is_directory`, so I think we should introduce similar functionality to `is_link` and `is_file`. Same thing for `delete_file` and `remove_directory` , we probably need `unlink` (or `remove_link`) and `create_link`?

src/stdlib_system.c
wassup05: I don't think it is possible to do that, because the respective syscalls may error out, then we don't truly know the status of the path, whether it is _missing_ or _invalid_, also windows only returns `INVALID_FILE_ATTRIBUTES`, not differentiating between some runtime error or the path not existing, that part is done by the `GetLastError` and I think we could also keep it that way for simplicity.

doc/specs/stdlib_system.md
wassup05: Oh yes, thank you, have been added here - https://github.com/fortran-lang/stdlib/pull/1026/commits/a5a340ef2d4629bddfb90bcef4a868eba333831a

src/stdlib_system.F90
wassup05: I agree, I have changed them here - https://github.com/fortran-lang/stdlib/pull/1026/commits/a5a340ef2d4629bddfb90bcef4a868eba333831a

src/stdlib_system.c
wassup05: https://github.com/fortran-lang/stdlib/pull/1026/commits/a5a340ef2d4629bddfb90bcef4a868eba333831a

src/stdlib_system.c
perazz: Ok so for now, I agree to just leave it `unknown` and assume we cannot differentiate between an actually missing entity and a malformed path yet.

perazz:
This is the way that feels most natural to me: 

- regular file -> `is_file = .true.`, `is_directory = .false.`, `is_symlink=.false.`, `exists = type_file`
- regular dir -> `is_file = .false.`, `is_directory = .true.`, `is_symlink=.false.`, `exists = type_dir`
- link to file -> `is_file = .true.`, `is_directory = .false.`, `is_symlink=.true.`, `exists = type_link`
- link to dir -> `is_file = .false.`, `is_directory = .true.`, `is_symlink=.true.`, `exists = type_link`

But of course I suggest further ideas about this be discussed.

wassup05:
That seems reasonable @perazz, I have added `is_symlink` and `is_regular_file`.
Let me know if it looks good and I will add the specs and examples

wassup05:
Another point to discuss I feel is, should these `is_` functions provide optional error handling capabilities using the usual `state_type` instead of just being `logical`?

sebastian-mutz:
Can you highlight some specific advantages you see in using `state_type`/what error handling capabilities you think would be beneficial to these?

doc/specs/stdlib_system.md
```diff
@@ -535,6 +575,46 @@ The function returns a `logical` value:
...
+### Status
+
+Experimental
+
+### Description
+
+This function checks if a specified file system path is a symbolic link to either a file or a directory.
+Use [[stdlib_system(module):is_file(function)]] and [[stdlib_system(module):is_directory(function)]] functions
+to check further if the link is to a file or a directory respectively.
+It is designed to work across multiple platforms. On Windows, paths with both forward `/` and backward `\` slashes are accepted.
```
perazz, line 589: Since we are advertising it, would it be reaonable to add a test that checks `./test_file` on all systems, and also `.\test_file` on Windows

perazz (approved):
LGTM with the attached comment @wassup05, thank you for this PR. 
Regarding the `logical` functions, I agree that (separation of concerns) they should not return a state flag, but just match the underlying logic such is done already. 

Think about condition "value>0"
-> true if value = 1.0, 1234.5, +Inf
-> false if value = 0.0, -123. NaN, -Inf

So it seems consistent to me that the edge cases are already considered by the boolean condition.

doc/specs/stdlib_system.md
wassup05: I don't think I understand what you mean, do you mean a sort of utility test function to make the tests simpler and extensive by checking both the kinds of paths for all the tests?

perazz: I believe it would be useful if the tests tested `filename` but also `./filename` (on all OSes) and `.\filename` (on Windows only)

wassup05: I don't believe that's necessary @perazz, because we haven't written any custom logic for that ourselves, it is how the underlying C functions that do the `syscalls` behave, so adding tests for such would be like testing the C standard library or the Windows API and not necessarily the test of our own application or library.

perazz:
Request to add a test for `.\file` on Windows seems stale; however, in absence of further comments I suggest this PR is merged soon anyways.

PR #1027, opened 2025-08-11
feat: path functions `is_abs` and `abs_path`

wassup05:
User facing functions added are:
- `is_abs(path)`: returns a `logical` indicating if the path is absolute.
- `abs_path(path [, err])`: returns the absolutized version of the path.

`is_abs` does pure string manipulation checking if
- on POSIX systems, `path` starts with `/`
- on Windows systems it is either an `UNC` path or starts with a drive letter like `C:\`

`abs_path` performs a syscall to get the current working directory and then `joins` it to the `path`, returning the `joined` path

## Prior Art
[`os.path.abspath`](https://docs.python.org/3/library/os.path.html#os.path.abspath) (Python)
[`os.path.isabs`](https://docs.python.org/3/library/os.path.html#os.path.isabs) (Python)


doc/specs/stdlib_system.md
```diff
@@ -978,3 +978,79 @@ A character string or `type(string_type)`.
...
+Function
+
+### Arguments
+
+`p`: the path, a character string or `type(string_type)`. It is an `intent(in)` argument.
+
+
+### Return values
+
+A `logical` indicating if the the path is absolute.
```
sebastian-mutz: While it's an intuitive return value, it's best to be explicit somewhere about what value is returned in what case. Either here or in the description. "Returns `.true.` if the path is absolute."

doc/specs/stdlib_system.md
```diff
@@ -978,3 +978,79 @@ A character string or `type(string_type)`.
...
+
+`p`: the path, a character string or `type(string_type)`. It is an `intent(in)` argument.
+
+`err`(optional): Shall be of type `state_type`, and is used for error handling. It is an `optional, intent(out)` argument.
+
+### Return values
+
+`res`: the absolutized version of the path, It is of type `character(:), allocatable`.
+
+`err`: It is an optional state return flag. On error if not requested, an `FS_ERROR` will trigger an error stop.
```
sebastian-mutz: Use consistent capitalisation after colon. I think most of stdlib uses lower case.

sebastian-mutz (commented):
Thanks, @wassup05! 
I'm wondering if `is_abs_path` may be better to use. While "path" is the first thing many will think of from context, it is entirely implied. Even in the use statement `stdlib_system, only: is_abs`, it's not visible (c.f. Python's `os.path.isabs`).

doc/specs/stdlib_system.md
wassup05: https://github.com/fortran-lang/stdlib/pull/1027/commits/62e1efceff044effd4be7f3cb2d1151f018395e6

doc/specs/stdlib_system.md
wassup05: https://github.com/fortran-lang/stdlib/pull/1027/commits/62e1efceff044effd4be7f3cb2d1151f018395e6

sebastian-mutz (approved):
LGTM with the recent change, @wassup05. Many thanks.

src/system/stdlib_system_path.f90
```diff
@@ -167,4 +167,58 @@ module function dir_name_string(p) result(dir)
...
+        ! get the current working directory
+        call get_cwd(cwd, err0)
+
+        if (err0%error()) then
+            abs_p = ''
+            call err0%handle(err)
+        end if
+
+        ! join the cwd and path
+        abs_p = cwd / p
```
perazz, line 217: what happens if `p` is already an absolute path on input?

src/system/stdlib_system_path.f90
```diff
@@ -167,4 +167,58 @@ module function dir_name_string(p) result(dir)
 
         call split_path(p, dir, temp)
     end function dir_name_string
+
+    module logical function is_abs_path_char(p)
```
perazz, line 171: the `is_abs_path` function only tests the _form_ of a path specification, not the fact that it must actually exist in the filesystem. So I wonder if giving these functions the same name as `is_file`, `is_directory` etc., that instead test the *actual* filesystem, wouldn't be confusing for some. Also, do other library implementations only check the form, or do they require the actual path to exist?

doc/specs/stdlib_system.md
```diff
@@ -978,3 +978,82 @@ A character string or `type(string_type)`.
...
+
+## `is_abs_path` - Checks if the path is absolute
+
+### Status
+
+Experimental
+
+### Description
+
+This function checks if the path is absolute (i.e not relative).
```
jvdp1:
```suggestion
This function checks if the path is absolute (i.e. not relative).
```

doc/specs/stdlib_system.md
```diff
@@ -978,3 +978,82 @@ A character string or `type(string_type)`.
...
+
+### Arguments
+
+`p`: the path, a character string or `type(string_type)`. It is an `intent(in)` argument.
+
+`err`(optional): Shall be of type `state_type`, and is used for error handling. It is an `optional, intent(out)` argument.
+
+### Return values
+
+`res`: the absolutized version of the path, It is of type `character(:), allocatable`.
```
jvdp1:
```suggestion
`res`: the absolutized version of the path. It is of type `character(:), allocatable`.
```

example/system/example_path_abs.f90
```diff
@@ -0,0 +1,29 @@
...
+    absolute_path = abs_path(path, err)
+
+    if (err%error()) then
+        ! there was an error! print it
+        print *, "error converting to absolute path: " // err%print()
+    else
+        print *, "absolute path => " // absolute_path
+    end if
+end program example_path_abs
+
```
jvdp1:
```suggestion
```

src/stdlib_system_path.f90
```diff
@@ -167,4 +167,58 @@ module function dir_name_string(p) result(dir)
...
+        ! get the current working directory
+        call get_cwd(cwd, err0)
+
+        if (err0%error()) then
+            abs_p = ''
+            call err0%handle(err)
+        end if
+
+        ! join the cwd and path
+        abs_p = cwd / p
```
jvdp1: I think that using `is_abs_path` as a check before this row could be a solution.

src/system/stdlib_system_path.f90
wassup05: Fixed in https://github.com/fortran-lang/stdlib/pull/1027/commits/85745a461dc05b29a83ed3524c32878968021a3a, it just returns the path unchanged.

src/system/stdlib_system_path.f90
wassup05: That's a valid concern I feel but that is how most path related functions work.. even in other libraries like [python](https://docs.python.org/3/library/os.path.html), [go](https://pkg.go.dev/path/filepath@go1.26.1) etc and considering all these functions have *path* in their name should make the required distinction I believe.

perazz (approved):

PR #1029, opened 2025-08-23
feat: `get_file_size` - gets the size of a file in bytes

wassup05:
User facing function added is `get_file_size(path [, err])`

It returns the file size in bytes. It returns an error if the `path` is a directory or a symlink to a directory (because for the size of directories you will need to recurse into the directory and add all its component sizes to get its actual size)

It uses [`stat`](https://man7.org/linux/man-pages/man2/stat.2.html) and [`_stati64`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/stat-functions?view=msvc-170) on POSIX and windows respectively. It returns an integer of `kind` `int64` to handle large enough file sizes.

### DIFF
--- a/test/system/test_filesystem.f90
+++ b/test/system/test_filesystem.f90
@@ -370,7 +370,7 @@ contains
     subroutine test_get_file_size_file(error)
         type(error_type), allocatable, intent(out) :: error
         character(len=256) :: filename
-        integer :: ios, iunit
+        integer :: ios, iunit, iocmd
         character(len=512) :: msg
         character(len=20) :: text
 
@@ -379,8 +379,8 @@ contains
 
         filename = "test_file.txt"
 
-        ! Create a file
-        open(newunit=iunit, file=filename, status="replace", action='write', iostat=ios, iomsg=msg)
+        ! Create a file and open it in `stream` access
+        open(newunit=iunit, file=filename, status="replace", action='write', access='stream', iostat=ios, iomsg=msg)
         call check(error, ios == 0, "Cannot create test file: " // trim(msg))
         if (allocated(error)) return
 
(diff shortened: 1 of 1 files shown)


doc/specs/stdlib_system.md
```diff
@@ -720,6 +720,46 @@ Subroutine
 
 ---
 
+## `get_file_size` - Gets the size (in bytes) of a file.
```
perazz, line 723: We have not used `get_*` patterns anywhere else so, this function may perhaps just be `file_size`? The fact that it is a function already means we're `get_`ting stuff.

src/stdlib_system.F90
```diff
@@ -1112,6 +1127,39 @@ end function stdlib_set_cwd
...
+        return
+    end if
+
+    code = stdlib_get_file_size(to_c_char(trim(path)), c_size)
+
+    if (code /= 0) then
+        err0 = FS_ERROR_CODE(code, c_get_strerror())
+        call err0%handle(err)
+        return
+    end if
```
perazz, line 1158: This functionality can be achieved by Fortran intrinsics only. Perhaps that would be a good way to avoid cross-platform CI issues and avoid the C backend calls?

```suggestion
    type(state_type) :: err0
    integer :: code
    character(len=256) :: iomsg
    integer(c_int64_t) :: c_size

    if (is_directory(path)) then
        err0 = FS_ERROR("Is a directory!")
        call err0%handle(err)
        return
    end if

    inquire(file=path, size=c_size, iostat=code, iomsg=iomsg)

    if (code /= 0) then
        err0 = FS_ERROR_CODE(code, iomsg)
        call err0%handle(err)
        return
    end if
```

Issue #1031, opened 2025-09-17
Proposal to divide the stdlib CMake project in multiple libraries


eduardz1:
### Motivation

Currently, the stdlib exports a single CMake target, this means that to use even a single feature one has to compile the entire standard library. This is particularly problematic for cases where compiler support is lacking such as https://github.com/fortran-lang/stdlib/issues/107.

In the software I am currently working on, https://ffaucher.gitlab.io/hawen-website/ (note that the public source code is not up-to-date), I would consider proposing the usage of the stdlib but compiler limitations currently make it non-feasible.

The solution would be to export a library for each module, so I believe it could be solved with something like

```cmake
# src/CMakeLists.txt

add_library(stdlib_io ${IO_SOURCES})
add_library(stdlib_array ${ARRAY_SOURCES})
# ...

add_library(${PROJECT_NAME} INTERFACE)
target_link_libraries(${PROJECT_NAME} INTERFACE 
    stdlib_io 
    stdlib_array
    # ...
)

install(TARGETS 
    ${PROJECT_NAME}
    stdlib_io
    stdlib_array
    # ...
    EXPORT ${PROJECT_NAME}-targets
    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
    PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
)
```

So that in our application we could, for example, only link `fortran_stdlib::stdlib_io` (whilst still keeping the `fortran_stdlib::fortran_stdlib` target available) and therefore limit the compatibility issues across different compilers (and potential points of failure in general).

This is not super straight forward to implement as dependencies will have to be managed individually (for example I see that `stdlib_ansi` will depend on `stdlib_kinds` and `stdlib_string_type` and will also make the `CMakeLists.txt` file under `src` quite verbose.

I would be willing to try to implement this if I find agreement in the proposal

PR #1033, opened 2025-09-22
Modularize BLAS and LAPACK modules in CMake to make it easier to use in external projects

eduardz1:
This PR is the start of https://github.com/fortran-lang/stdlib/issues/1031, it currently only separates the BLAS and LAPACK. This means that, when including the stdlib in a CMake project through FetchContent, we can link only the `fortran_stdlib::blas` or `fortran_stdlib::lapack` targets. This kind of approach might become a bit difficult to manage in the future so there should be some careful discussion about which library could be exported as separate targets. I have started with BLAS and LAPACK because writing interfaces for the two is something that happens way too often. I have also created aliased targets with the namespace to make it a bit nicer to work with when included.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -36,6 +36,9 @@ set(fppFiles
     stdlib_linalg_determinant.fypp
     stdlib_linalg_qr.fypp
     stdlib_linalg_inverse.fypp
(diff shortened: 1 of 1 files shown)


perazz (approved):
LGTM, thank you @eduardz1 for this improvement.

eduardz1:
Thanks!

PR #994, opened 2025-05-17
feat: [linalg] add iterative solvers (CG and PCCG)

jalvesz:
This **daft** PR introduces the base for iterative solvers.

Two methods are proposed:
* Conjugate Gradient
* Preconditioned Conjugate Gradient

Each method is made public with two public interface flavors:
* `solve_<method>_kernel`: All arguments are mandatory (no optionals/no internal allocations), it contains the methods steps. The linear system (and preconditioner) is defined through a public DT `linop` which enables to extend two key procedures: `apply` equivalent to a matrix-vector product and `inner` equivalent to the dot_product. This is the interface recommended to extend the method when dealing with a matrix type not available in `stdlib` or when working in distributed-memory frameworks for which the matvec, dot_product and factorization steps need to be adapted to account for parallel synchronization.
* `solve_<method>`: Flexible method with optional arguments, the linear system can be defined with dense or `CSR_<>_type` matrices. For the PCCG, the following preconditioners are available: `none` (identity), `jacobi` (1/diagonal), `SSOR`, `LDLt`.

- [ ] tests
- [ ] examples
- [ ] documentation

PR #1034, opened 2025-09-28
feat : [linalg] add bi-conjugate gradient stabilized method

jalvesz:
This PR co-developed with CoPilot introduces the Bi-Conjugate Gradient stabilized method. It is inspired from SciPy's implementation https://github.com/scipy/scipy/blob/v1.16.2/scipy/sparse/linalg/_isolve/iterative.py#L158-L304 but following the internal framework introduced  in #994 

Key interfaces :
* `call stdlib_solve_bicgstab_kernel (A, M, b, x, rtol, atol, maxiter, workspace)`
* `call stdlib_solve_bicgstab(A, b, x [, di, rtol, atol, maxiter, restart, precond, M, workspace])`

- [x] Implementation
- [x] Documentation
- [x] unit testing (symmetric and non-symmetric)
- [x] examples ( 2 : dense and CSR )


example/linalg/example_solve_bicgstab_wilkinson.f90
```diff
@@ -0,0 +1,52 @@
...
+    type(COO_dp_type) :: COO
+    type(CSR_dp_type) :: A
+    type(stdlib_solver_workspace_dp_type) :: workspace
+    real(dp) :: b(n), x(n), norm_sq0
+    integer :: i
+
+    ! Construct the Wilkinson's matrix in COO format
+    ! https://en.wikipedia.org/wiki/Wilkinson_matrix
+    call COO%malloc(n, n, n + 2*(n-1))
+    COO%data(1:n) = [( dble(abs(i)), i=-(n-1)/2, (n-1)/2)]
```
jvdp1:
```suggestion
    COO%data(1:n) = [( real(abs(i), kind=dp), i=-(n-1)/2, (n-1)/2)]
```

doc/specs/stdlib_linalg_iterative_solvers.md
```diff
@@ -248,4 +248,104 @@ Subroutine
...
+
+#### Argument(s)
+
+`A`: `dense` or `CSR_<kind>_type` matrix defining the linear system. This argument is `intent(in)`.
+
+`b`: 1-D array of `real(<kind>)` defining the loading conditions of the linear system. This argument is `intent(in)`.
+
+`x`: 1-D array of `real(<kind>)` which serves as the input initial guess and the output solution. This argument is `intent(inout)`.
+
+`di` (optional): 1-D mask array of type `logical(int8)` defining the degrees of freedom subject to dirichlet boundary conditions. The actual boundary conditions values should be stored in the `b` load array. This argument is `intent(in)`.
```
jvdp1:
```suggestion
`di` (optional): 1-D mask array of type `logical(int8)` defining the degrees of freedom subject to dirichlet boundary conditions. The actual boundary condition values should be stored in the `b` load array. This argument is `intent(in)`.
```

example/linalg/example_solve_bicgstab_wilkinson.f90
```diff
@@ -0,0 +1,52 @@
...
+    COO%data(1:n) = [( real(abs(i), kind=dp), i=-(n-1)/2, (n-1)/2)]
+    COO%data(n+1:) = 1.0_dp
+    do i = 1, n
+        COO%index(1:2, i) = [i,i]
+    end do
+    do i = 1, n-1
+        COO%index(1:2, n+i) = [i,i+1]
+        COO%index(1:2, n+n-1+i) = [i+1,i]
+    end do
+    call coo2ordered(COO,sort_data=.true.)
```
jvdp1, line 27: Is this needed?

jvdp1 (approved):
LGTM! Thank you @jalvesz

example/linalg/example_solve_bicgstab_wilkinson.f90
jalvesz: Yes, one should not create the CSR matrix from an un-ordored COO.

jvdp1: I didn't play enough with this part. However, if it is the case, I would suggest to include it `coo2csr` (maybe with an additional flag to check/trigger the sorting?)
Anyway, it is not related to this PR, and should not stop this PR.

jalvesz: Indeed, the `coo2csr` implementation assumes that the coo matrix is ordered and unique. We could enhance it by triggering calling the ordering step from within with a boolean flag.

PR #1036, opened 2025-10-05
CMake: isssue with an "add_library" added in #1033

jvdp1:
Following the changes in #1033, the following Cmake directives are broken:
```cmake
...
  if("${method}" STREQUAL "fetch")
    message(STATUS "Retrieving ${_lib} from ${_url}")
    include(FetchContent)
    FetchContent_Declare(
      "${_lib}"
      GIT_REPOSITORY "${_url}"
      GIT_TAG "HEAD"
      )
    FetchContent_MakeAvailable("${_lib}")

    add_library("${_lib}::${_lib}" INTERFACE IMPORTED)
    target_link_libraries("${_lib}::${_lib}" INTERFACE "${_lib}")

    # We need the module directory in the subproject before we finish the configure stage
    FetchContent_GetProperties("${_lib}" SOURCE_DIR "${_pkg}_SOURCE_DIR")
    FetchContent_GetProperties("${_lib}" BINARY_DIR "${_pkg}_BINARY_DIR")
    if(NOT EXISTS "${${_pkg}_BINARY_DIR}/mod_files")
      make_directory("${${_pkg}_BINARY_DIR}/mod_files")
    endif()

    break()
  endif()

...
```

Commenting out the directive `add_library(${PROJECT_NAME}::${target_name} ALIAS ${target_name})` introduced in #1033 solved my issue. However, I am not an experxt in CMake. So, is this required?

### DIFF
--- a/cmake/stdlib.cmake
+++ b/cmake/stdlib.cmake
@@ -69,7 +69,7 @@ function(configure_stdlib_target target_name regular_sources_var fypp_files_var 
     list(APPEND all_sources ${${regular_sources_var}})
 
     add_library(${target_name} ${all_sources})
-    add_library(${PROJECT_NAME}::${target_name} ALIAS ${target_name})
+    #add_library(${PROJECT_NAME}::${target_name} ALIAS ${target_name})
 
     set_target_properties(
       ${target_name}


jvdp1:
@perazz @eduardz1

eduardz1:
> Following the changes in #1033, the following Cmake directives are broken:
> 
> ```cmake
> ...
>   if("${method}" STREQUAL "fetch")
>     message(STATUS "Retrieving ${_lib} from ${_url}")
>     include(FetchContent)
>     FetchContent_Declare(
>       "${_lib}"
>       GIT_REPOSITORY "${_url}"
>       GIT_TAG "HEAD"
>       )
>     FetchContent_MakeAvailable("${_lib}")
> 
>     add_library("${_lib}::${_lib}" INTERFACE IMPORTED)
>     target_link_libraries("${_lib}::${_lib}" INTERFACE "${_lib}")
> 
>     # We need the module directory in the subproject before we finish the configure stage
>     FetchContent_GetProperties("${_lib}" SOURCE_DIR "${_pkg}_SOURCE_DIR")
>     FetchContent_GetProperties("${_lib}" BINARY_DIR "${_pkg}_BINARY_DIR")
>     if(NOT EXISTS "${${_pkg}_BINARY_DIR}/mod_files")
>       make_directory("${${_pkg}_BINARY_DIR}/mod_files")
>     endif()
> 
>     break()
>   endif()
> 
> ...
> ```
> 
> Commenting out the directive `add_library(${PROJECT_NAME}::${target_name} ALIAS ${target_name})` introduced in #1033 solved my issue. However, I am not an experxt in CMake. So, is this required?

I won't be able to test it in the next two weeks but you should be able to remove the two lines "add_library" and "target_link_libraries", the error is due to the target existing already. I see that the code you wrote is probably part of some function or some automated installation. In case you still want to keep these two lines for some repositories then change it to something like

```cmake
...
if (NOT TARGET "${_lib}::{_lib}")
     add_library("${_lib}::${_lib}" INTERFACE IMPORTED)
     target_link_libraries("${_lib}::${_lib}" INTERFACE "${_lib}")
endif()
...
```

jvdp1:
> I won't be able to test it in the next two weeks 

Thank you for your quick answer. It is not urgent on my side

> but you should be able to remove the two lines "add_library" and "target_link_libraries", the error is due to the target existing already.

Only the second `add_library` was added in #1033, as far as I can see. So, I wonder why it was added there (there might be a reason that I am not aware of).

> I see that the code you wrote is probably part of some function or some automated installation. 

Indeed, it is based on the `stdlib`  [`Findtest-drive.cmake`](https://github.com/fortran-lang/stdlib/blob/master/config/cmake/Findtest-drive.cmake)

>In case you still want to keep these two lines for some repositories then change it to something like
> 
> ```cmake
> ...
> if (NOT TARGET "${_lib}::{_lib}")
>      add_library("${_lib}::${_lib}" INTERFACE IMPORTED)
>      target_link_libraries("${_lib}::${_lib}" INTERFACE "${_lib}")
> endif()
> ...
> ```

It is also a possibility. But I will have to adapt many projects :( Of course, it is possible. But in that case, should the `stdlib` [`README `](https://github.com/fortran-lang/stdlib/blob/80a8164c417db5caa02722418698181a9c62f076/README.md?plain=1#L267) be adapted?

cmake/stdlib.cmake
```diff
@@ -69,7 +69,7 @@ function(configure_stdlib_target target_name regular_sources_var fypp_files_var
     list(APPEND all_sources ${${regular_sources_var}})
 
     add_library(${target_name} ${all_sources})
-    add_library(${PROJECT_NAME}::${target_name} ALIAS ${target_name})
+    #add_library(${PROJECT_NAME}::${target_name} ALIAS ${target_name})
```
perazz: we can just remove this line. 
```suggestion
```

jvdp1: It is what I proposed initially, mainly to get the same behaviour as previously, but @eduardz1 didn't seem to agree, and I didn't take the time to understand his explanations.

perazz: please take it as a style comment @jvdp1 - I noticed it was commented out already, so I just cleaned it up - no need to force deletion on my end.

jvdp1: Sorry, @perazz , I was not against your proposition, as it will make it cleaner. 
Currently I am wondering what the best approach will be, that is keeping the `add_library` here and modifying the Find*cmake, or to get back to the previous situation.

PR #1038, opened 2025-10-07
Matrix exponential

loiseaujc:
Following a comment by Meromorphic on the discourse [here](https://fortran-lang.discourse.group/t/some-reflections-about-stdlib-where-it-stands-and-what-it-could-become/9376/53?u=loiseaujc), this PR implements the matrix exponential function `expm`. It does so in a new `stdlib_linalg_matrix_functions` submodule which might eventually accomodate later on implementations of other matrix functions (e.g. `logm`, `sqrtm`, `cosm`, etc).

### Proposed interfaces

- `E = expm(A [, order, err])` 
- `call matrix_exp(A [, order, err])` (in-place)
- `call matrix_exp(A, E [, order, err])` (out-of-place)
 
where `A` is the input `n x n` matrix, `order` (optional) is the order of the Pade approximation, and `err` of type `linalg_state_type` for error handling.

### Key facts

The implementation makes use of a standard "scaling and squaring" approach to compute the Pade approximation with a fixed order. It is adapted from the implementation by John Burkardt.

### Progress

- [x] Base implementation.
- [x] Tests
- [x] in-code Documentation
- [x] Specifications
- [x] Example

### Possible improvements

The implementation is fully working and validated. Computational performances and/or robustness could potentially be improved eventually by considering:

- Automatic estimation of the best `order` for the Pade approximation based on the data. This is used for instance in `scipy.linalg.expm`.
- Balancing is not used currently, but it might improve the robustness.

In practice, it has however never failed me so far.

cc @perazz, @jvdp1, @jalvesz 

PS : This is actually a re-opening of #968 which was automatically closed when I deleted my own fork of `stdlib` (don't ask me why, I can't remember).


example/linalg/example_expm.f90
```diff
@@ -0,0 +1,7 @@
+program example_expm
+   use stdlib_linalg, only: expm
+   implicit none
+   real :: A(3, 3), E(3, 3)
+   A = reshape([1, 2, 3, 4, 5, 6, 7, 8, 9], [3, 3])
+   E = expm(A)
```
jvdp1, line 7: I suggest to print the output to stdout

src/stdlib_linalg.fypp
```diff
@@ -28,6 +28,7 @@ module stdlib_linalg
   public :: eigh
   public :: eigvals
   public :: eigvalsh
+  public :: expm, matrix_exp
```
jvdp1, line 31: `matrix_exp` is not mentioned in the specs.

src/stdlib_linalg.fypp
```diff
@@ -1678,6 +1679,107 @@ module stdlib_linalg
...
+    !! ([Specification](../page/specs/stdlib_linalg.html#expm))
+    !!
+    !! ### Description
+    !!
+    !! This interface provides methods for computing the exponential of a matrix
+    !! represented as a standard Fortran rank-2 array. Supported data types include
+    !! `real` and `complex`.
+    !!
+    !! By default, the order of the Pade approximation is set to 10. It can be changed
+    !! via the `order` argument which must be non-negative.
```
jvdp1: Should it be (I am not sure about that): 
```suggestion
    !! via the `order` argument that must be non-negative.
```
?

src/stdlib_linalg.fypp
```diff
@@ -1678,6 +1679,107 @@ module stdlib_linalg
...
+    !!  ! Default Pade approximation of the matrix exponential.
+    !!  E = expm(A)
+    !!
+    !!  ! Pade approximation with specified order.
+    !!  E = expm(A, order=12)
+    !! ```
+    !!
+    #:for rk,rt,ri in RC_KINDS_TYPES
+    module function stdlib_linalg_${ri}$_expm_fun(A, order) result(E)
+        !> Input matrix a(n, n).
```
jvdp1:
```suggestion
        !> Input matrix a(:, :).
```

src/stdlib_linalg.fypp
```diff
@@ -1678,6 +1679,107 @@ module stdlib_linalg
...
+    !!
+    !!  ! Default Pade approximation of the matrix exponential.
+    !!  E = expm(A)
+    !!
+    !!  ! Pade approximation with specified order.
+    !!  E = expm(A, order=12)
+    !! ```
+    !!
+    #:for rk,rt,ri in RC_KINDS_TYPES
+    module function stdlib_linalg_${ri}$_expm_fun(A, order) result(E)
```
jvdp1, line 1716: This API does not correspond to the one described in the specs (the `err` is not part of it)

src/stdlib_linalg_matrix_functions.fypp
```diff
@@ -0,0 +1,159 @@
...
+
+            block
+                integer(ilp) :: ipiv(n), info
+                call gesv(n, n, Q, n, ipiv, A, n, info) ! E = inv(Q) @ E
+                call handle_gesv_info(this, info, n, n, n, err0)
+            end block
+
+            ! Matrix squaring.
+            block
+                ${rt}$, allocatable :: E_tmp(:, :)
```
jvdp1:
```suggestion
                ${rt}$ :: E_tmp(n, n)
```
Otherwise , the declaration can be moved to the top of the procedure to avoid a block structure.

src/stdlib_linalg_matrix_functions.fypp
jalvesz, line 1: you could consider the following macros:
```python
#:set R_KINDS_TYPES = list(zip(REAL_KINDS, REAL_TYPES, REAL_SUFFIX))
#:set C_KINDS_TYPES = list(zip(CMPLX_KINDS, CMPLX_TYPES, CMPLX_SUFFIX))
```
That way, the third value of the list will give the proper suffix naming for complex or reals, which would then enable to write this in a single line:
```fortran
#:if rt.startswith('complex')
call gemm("N", "N", n, n, n, one_c${rk}$, A2, n, X_tmp, n, zero_c${rk}$, X, n)
#:else
call gemm("N", "N", n, n, n, one_${rk}$, A2, n, X_tmp, n, zero_${rk}$, X, n)
#:endif
```
now:
```fortran
#:for k, t, s in RC_KINDS_TYPES
...
call gemm("N", "N", n, n, n, one_${s}$, A2, n, X_tmp, n, zero_${s}$, X, n)
```

src/stdlib_linalg_matrix_functions.fypp
```diff
@@ -0,0 +1,159 @@
...
+        ! Check E sizes
+        lda = size(A, 1, kind=ilp) ; n = size(A, 2, kind=ilp)
+        lde = size(E, 1, kind=ilp) ; ne = size(E, 2, kind=ilp)
+          
+        if (lda<1 .or. n<1 .or. lda/=n .or. lde/=n .or. ne/=n) then     
+            err0 = linalg_state_type(this,LINALG_VALUE_ERROR, &
+                                     'invalid matrix sizes: A must be square (lda=', lda, ', n=', n, ')', &
+                                     ' E must be square (lde=', lde, ', ne=', ne, ')')
+        else
+            E(:n, :n) = A(:n, :n) ; call stdlib_linalg_${ri}$_expm_inplace(E, order, err0)
```
jalvesz: I would advise that all  `call`s are placed in a separate line for future debugging reasons. Easier to set a breakpoint and then enter the procedure.

src/stdlib_linalg_matrix_functions.fypp
jalvesz: One key point here would be: whether the memory of this temp variable should live on the heap or the stack, which would then have consequences on performance and size limits.

src/stdlib_linalg_matrix_functions.fypp
```diff
@@ -0,0 +1,159 @@
...
+    use stdlib_linalg_state, only: linalg_state_type, linalg_error_handling, LINALG_ERROR, &
+         LINALG_INTERNAL_ERROR, LINALG_VALUE_ERROR
+    implicit none
+
+    character(len=*), parameter :: this = "matrix_exponential"
+
+contains
+
+    #:for rk,rt,ri in RC_KINDS_TYPES 
+    module function stdlib_linalg_${ri}$_expm_fun(A, order) result(E)
```
jalvesz: pure style, I'm an advocate of having the kind suffix at the end of the procedure name signature. Kind of like the "hh:mm:ss" principle (the thing that changes the most last).

jalvesz (changes requested):

src/stdlib_linalg_matrix_functions.fypp
loiseaujc: Agreed. I went for the `allocatable` to be on the safe side. In practice though, you would never compute the matrix exponential if the matrix is too large, just like `LU` or other matrix factorizations. I don't think having it on the stack would be a major in practical applications.

src/stdlib_linalg_matrix_functions.fypp
loiseaujc: Oh cool. Will do the changes !

src/stdlib_linalg_matrix_functions.fypp
loiseaujc: I don't have a strong preference there. I went with whatever @perazz used for the Cholesky:

```
 pure module function stdlib_linalg_${ri}$_cholesky_fun(a,lower,other_zeroed) result(c)
```

Either ways are fine with me.

example/linalg/example_expm.f90
loiseaujc: Done.

src/stdlib_linalg.fypp
loiseaujc: Done.

src/stdlib_linalg.fypp
loiseaujc: Added matrix_exp to the specs.

src/stdlib_linalg.fypp
loiseaujc: Corrected.

src/stdlib_linalg_matrix_functions.fypp
loiseaujc: Done.

src/stdlib_linalg_matrix_functions.fypp
loiseaujc: I have now gotten rid of `E_tmp` as I could re-used another array already allocated.
The gist of the discussion now reduces to should we replace the line

```
        ${rt}$, allocatable     :: A2(:, :), Q(:, :), X(:, :), X_tmp(:, :)
```

with

```
        ${rt}$ :: A2(n, n), Q(n, n), X(n, n), X_tmp(n, n)
```

?

loiseaujc: I've put everything on the stack in my latest commit. I can always revert this choice if heap appears as a safer/better choice.

src/stdlib_linalg_matrix_functions.fypp
loiseaujc: Done.

Issue #1035, opened 2025-10-01
Stablish rule for submodule procedures declaration compatible with older cmake versions


jalvesz:
### Motivation

I have had problems with this issue https://gitlab.kitware.com/cmake/cmake/-/issues/18427#note_983426 when compiling stdlib under systems with cmake 3.17.5 (which for several reasons can't be upgraded) and then having build errors with spurious modules.

I had to patch locally the procedures by replacing for instance `module pure function` by `pure module function` etc etc.

I would like to propose that module procedures are to be declared as:
`<attribute> <attribute> module <function/subroutine> <name>` 
instead of 
`module <attribute> <attribute> <function/subroutine> <name>`

This would ensure an easy retro compatibility.

Affected modules :
stdlib_system_subprocess
stdlib_system
stdlib_linalg_diag
stdlib_linalg

PR #1041, opened 2025-10-14
submodule procedures declaration made compatible with older cmake version

BhoomishGupta:
Made change for issue no. #1035 (Stablish rule for submodule procedures declaration compatible with older cmake versions)

### DIFF
--- a/STYLE_GUIDE.md
+++ b/STYLE_GUIDE.md
@@ -62,8 +62,8 @@ focus on the semantics of the proposed changes rather than style and formatting.
   When defining many arrays of the same dimension, `dimension` can be used as an exception if it makes the code less verbose.
 * If the `optional` attribute is used to declare a dummy argument, it should follow the `intent` attribute.
 * For module procedures, it is recommended to declare attributes before the module keyword for better retro compatibility (Projects using CMake versions lower than CMake 3.25.0 are concerned see [Spurious modules](https://gitlab.kitware.com/cmake/cmake/-/issues/18427#note_983426)).
-Prefer the following pattern:
-```
+  Prefer the following pattern:
+  ```
   <attribute> <attribute> module <function/subroutine> <name>
   ```
   instead of:


jalvesz:
Thanks @BhoomishGupta for opening this PR.

@jvdp1 I saw that in the cmake files there is `cmake_minimum_required(VERSION 3.14.0)` but if indeed the available version is more recent, then this kind of issues goes unnoticed. When you propose to have a CI/CD job for this, do you mean to say to force installation of a specific old cmake version under one runner to ensure that minimum is actually functional?

@BhoomishGupta beyond just changing the code, it would be useful for this modification to be documented in the style_guide.md as a suggestion.

STYLE_GUIDE.md
```diff
@@ -61,6 +61,15 @@ focus on the semantics of the proposed changes rather than style and formatting.
   ```  
   When defining many arrays of the same dimension, `dimension` can be used as an exception if it makes the code less verbose.
 * If the `optional` attribute is used to declare a dummy argument, it should follow the `intent` attribute.
+* For module procedures, declare attributes before the `module` keyword for better retro compatibility:
```
jalvesz: * For module procedures, it is recommended to declare attributes before the `module` keyword for better retro compatibility (Projects using CMake versions lower than CMake 3.25.0 are concerned see [Spurious modules](https://gitlab.kitware.com/cmake/cmake/-/issues/18427#note_983426)). 
Prefere the following pattern:  
  ```
  <attribute> <attribute> module <function/subroutine> <name>
  ```  
  instead of:  
  ```
  module <attribute> <attribute> <function/subroutine> <name>
  ```

BhoomishGupta:
@jalvesz that is incredibly helpful, thank you. I didn't realize I needed to wrap it that way, so I appreciate you teaching me something new!

jvdp1 (approved):
LGTM! Thank you @BhoomishGupta for this PR. As @jalvesz approved it too, I will merge it.

PR #1042, opened 2025-10-17
Migration from macOS 13 to macOS 14

jvdp1:
* As macOS 13 is [deprecated](https://github.com/actions/runner-images), here is a migration to macOS 14.
* Support of GCC 14
* Update of the section "compilers" in the README


jalvesz (approved):
LGTM @jvdp1, thanks for this upgrade

PR #1043, opened 2025-10-17
Addition of Cmake 3.14 in Github CI

jvdp1:
Suppport of CMake 3.14 in Github CI

Suggested in #1041


jvdp1:
@jalvesz  @perazz I tried to support CMake 3.10, but it seems it is not supported anymore with linux: https://pypi.org/project/cmake/3.10.3/

I couldn't install it with macOS-14 either.
Any ideas?

Issue #930, opened 2025-02-06
Rank-revealing QR decomposition


loiseaujc:
### Motivation

The `stdlib_linalg` module includes the `qr` decomposition. As far as I know, the `stdlib_linalg_lapack` module exposes interfaces for `xGEQP3` which is the lapack routine for computing the QR with pivot which can be use as a rank-revealing factorization. I don't think it'll be a lot of work to extend the `qr` interface to allow for switching between the standard `qr` and the pivoting `qr`. The updated call to `qr` could look something like

```fortran
call qr(A, Q, R, pivots)
```

where `pivots` is an `integer(ilp), optional, intent(out)` dummy argument (possibly `allocatable`). If `pivots` is not passed, the driver falls back to the regular `qr` factorization.

### Prior Art

- Currently, `stdlib_linalg` exposes `qr(A, Q, R, overrite_a, storage, err)` which computes only the regular QR factorization.
 
- `scipy` exposes `scipy.linalg.qr(a, overwrite_a=False, lwork=None, mode="full", pivoting=False, check_finite=True)` for the QR factorization of `A`. Using `pivoting=True` returns `Q, R, p` where `p` is the vector of pivots.

- The `LinearAlgebra` module in `Julia` exposes `LinearAlgebra.qr(A, pivot=NoPivot(); blocksize) -> F` where one can switch between regular QR factorization (`pivot=NoPivot()`) and pivoted QR (`pivot=ColumnNorm()`).

### Additional Information

Pinging @perazz, @jvdp1 and @jalvesz.

PR #1045, opened 2025-10-20
Pivoting QR decomposition

loiseaujc:
Following #930, this PR extends the `qr` interface to enable the computation of the QR factorization with column pivoting. It is based on the `xGEQP3` driver from `lapack`.

### Proposed interfaces

- `call qr(a, q, r, pivots [, overwrite_a, storage, err])`

where `a` is the matrix to be factorized, `q` the orthonormal basis for `colspan(a)`, `r` the upper triangular matrix and `pivots` an integer array with indices of the pivoted columns.

### Progress

- [x] Interface
- [x] Base implementation
- [x] Tests
- [x] In-code documentation
- [x] Specifications
- [x] Example

Ping: @perazz, @jvdp1, @jalvesz


loiseaujc:
I think this PR is almost ready to be reviewed. I have the specs left to write but this should pretty quick as it mainly is a small modification of the `qr` specs. I'll try to do that by the end of the week. It seems like there is a small issue with the `unbuntu-22.04`/`cmake`/`inter-classic 2021.10` setup though. Not sure what it is and I don't have the intel compilers on my laptop so I can't really dig into it at the moment.

doc/specs/stdlib_linalg.md
```diff
@@ -953,7 +953,7 @@ the full problem is solved. On reduced matrices (`shape(Q)==[m,k]`, `shape(R)==[
 
 ### Syntax
 
-`call ` [[stdlib_linalg(module):qr(interface)]] `(a, q, r, [, storage] [, overwrite_a] [, err])`
+`call ` [[stdlib_linalg(module):qr(interface)]] `(a, q, r [, pivots] [, overwrite_a] [, storage] [, err])`
```
perazz, line 956:
```suggestion
`call ` [[stdlib_linalg(module):qr(interface)]] `(a, q, r [, storage] [, overwrite_a] [, pivots] [, err])`
```

Excellent work @loiseaujc -
Just a preliminary comment - I will have more time this weekend. I would try to not change the previous API. Adding `pivots` after `overwrite_a` instead of swapping it still allows running (no keywords)
```
call qr(a, q, r, pivots)
```
because `storage` is `kind(A)`, so, there is no ambiguity.

loiseaujc: I made the change in the docs because the actual interface in `stdlib_linalg.fypp` reads

```
pure module subroutine stdlib_linalg_${ri}$_qr(a, q, r, overwrite_a, storage, err) 
```

I am not sure why there would be ambiguity though. `pivots` is an integer array. The `call qr(a, q, r, pivots)` is exactly the one used in the example.

```fortran
program example_pivoting_qr
   use stdlib_linalg, only: qr
   implicit none
   real :: A(104, 32), Q(104, 32), R(32, 32)
   integer :: pivots(32)

   ! Create a random matrix
   call random_number(A)

   ! Compute its QR factorization (reduced)
   call qr(A, Q, R, pivots)

   ! Test factorization: Q*R = A
   print *, maxval(abs(matmul(Q, R) - A(:, pivots)))

end program example_pivoting_qr
```

doc/specs/stdlib_linalg.md
```diff
@@ -963,15 +963,17 @@ the full problem is solved. On reduced matrices (`shape(Q)==[m,k]`, `shape(R)==[
 
 `r`: Shall be a rank-2 array of the same kind as `a`, containing the upper triangular matrix `r`. It is an `intent(out)` argument. It should have a shape equal to either `[m,n]` or `[k,n]`, whether the full or the reduced problem is sought for.
 
-`storage` (optional): Shall be a rank-1 array of the same type and kind as `a`, providing working storage for the solver. Its minimum size can be determined with a call to [[stdlib_linalg(module):qr_space(interface)]]. It is an `intent(out)` argument.
+`pivots` (optional): Shall be an `integer` array of size `n`. If provided, QR factorization with column-pivoting is being computed.
```
jvdp1:
```suggestion
`pivots` (optional): Shall be an `integer` array of size `n`. If provided, QR factorization with column-pivoting is being computed. It is an `intent(out)` argument.
```

doc/specs/stdlib_linalg.md
```diff
@@ -995,20 +999,24 @@ This subroutine computes the internal working space requirements for the QR fact
...
-`call ` [[stdlib_linalg(module):qr_space(interface)]] `(a, lwork, [, err])`
+`call ` [[stdlib_linalg(module):qr_space(interface)]] `(a, lwork, [, pivoting] [, err])`
 
 ### Arguments
 
 `a`: Shall be a rank-2 `real` or `complex` array containing the coefficient matrix. It is an `intent(in)` argument.
 
 `lwork`: Shall be an `integer` scalar, that returns the minimum array size required for the working storage in [[stdlib_linalg(module):qr(interface)]] to factorize `a`.
 
+`pivoting` (optional): Shall a `logical` flag (default: `.false.`). If `.true.`, on exit `lwork` is the optimal workspace size for the QR factorization with column pivoting. If `.false.`, `lwork` is the optimal workspace size for the standard QR factorization.
```
jvdp1:
```suggestion
`pivoting` (optional): Shall a `logical` flag (default: `.false.`). If `.true.`, on exit `lwork` is the optimal workspace size for the QR factorization with column pivoting. If `.false.`, `lwork` is the optimal workspace size for the standard QR factorization. It is an `intent(in)` argument.
```

src/stdlib_linalg_lapack.fypp
```diff
@@ -3230,6 +3230,77 @@ module stdlib_linalg_lapack
...
+            pure subroutine zgeqp3(m, n, a, lda, jpvt, tau, work, lwork, rwork, info)
+                import sp, dp, qp, ${ik}$, lk
+                implicit none
+                integer(${ik}$), intent(in) :: m, n, lda, lwork
+                integer(${ik}$), intent(out) :: info
+                integer(${ik}$), intent(inout) :: jpvt(*)
+                complex(dp), intent(inout) :: a(lda, *)
+                complex(dp), intent(out) :: tau(*), work(*)
+                real(dp), intent(out) :: rwork(*)
+            end subroutine zgeqp3
```
jvdp1: It seems that all of these interfaces could be defined using a `fypp` pre-processing approach.

test/linalg/test_linalg_pivoting_qr.fypp
```diff
@@ -0,0 +1,326 @@
...
+
+    !> QR factorization of a random matrix
+    #:for rk,rt,ri in RC_KINDS_TYPES
+    subroutine test_pivoting_qr_random_${ri}$(error)
+        type(error_type), allocatable, intent(out) :: error
+
+        !-------------------------------
+        !-----     Tall matrix     -----
+        !-------------------------------
+        block
```
jvdp1: I suggest that each `block` has its own `subroutine`. I would facilitate the review and maintainance IMHO.

jvdp1 (approved):
Thank you @loiseaujc . Overall LGTM. I just have some minor suggestions.

doc/specs/stdlib_linalg.md
perazz: Exactly - I had not reviewed in depth enough to see that you were correctly catching a documentation mismatch.

src/lapack/stdlib_linalg_lapack_aux.fypp
```diff
@@ -1462,6 +1463,27 @@ module stdlib_linalg_lapack_aux
...
+        ! Process output
+        select case (info)
+            case(0)
+                ! Success
+            case(-1)
+                err = linalg_state_type(this, LINALG_VALUE_ERROR, 'invalid matrix size m=', m)
+            case(-2)
+                err = linalg_state_type(this, LINALG_VALUE_ERROR, 'invalid matrix size n=', n)
+            case(-4)
+                err = linalg_state_type(this, LINALG_VALUE_ERROR, 'invalid matrix size a=', [m, n])
```
perazz:
```suggestion
                err = linalg_state_type(this, LINALG_VALUE_ERROR, 'invalid matrix shape a=', [m, n])
```

src/lapack/stdlib_linalg_lapack_aux.fypp
```diff
@@ -1462,6 +1463,27 @@ module stdlib_linalg_lapack_aux
...
+            case(-7)
+                err = linalg_state_type(this, LINALG_VALUE_ERROR, 'invalid input for lwork=', lwork)
+            case default
+                err = linalg_state_type(this, LINALG_VALUE_ERROR, 'catastrophic error')
```
perazz:
```suggestion
                err = linalg_state_type(this, LINALG_INTERNAL_ERROR, 'catastrophic error')
```

src/stdlib_linalg.fypp
```diff
@@ -614,6 +614,23 @@ module stdlib_linalg
...
+          ${rt}$, intent(inout), target :: a(:, :)
+          !> Orthogonal matrix Q ([m, m] or [m, k] if reduced)
+          ${rt}$, intent(out), contiguous, target :: q(:, :)
+          !> Upper triangular matrix R ([m, n] or [k, n] if reduced)
+          ${rt}$, intent(out), contiguous, target :: r(:, :)
+          !> Pivots.
+          integer(ilp), intent(out) :: pivots(:)
+          !> [optional] Can A data be overwritten and destroyed?
+          logical(lk), optional, intent(in) :: overwrite_a
+          !> [optional] Provide pre-allocated workspace, size to be checked with pivoting_qr_space.
```
perazz:
```suggestion
          !> [optional] Provide pre-allocated workspace, size to be checked with qr_space.
```

src/stdlib_linalg_qr.fypp
```diff
@@ -220,4 +220,201 @@ submodule (stdlib_linalg) stdlib_linalg_qr
...
+        ${rt}$, intent(inout), target :: a(:, :)
+        !> Orthogonal matrix Q ([m, m] or [m, k] if reduced)
+        ${rt}$, intent(out), contiguous, target :: q(:, :)
+        !> Upper triangular matrix R ([m, n] or [k, n] if reduced)
+        ${rt}$, intent(out), contiguous, target :: r(:, :)
+        !> Pivots.
+        integer(ilp), intent(out) :: pivots(:)
+        !> [optional] Can A data be overwritten and destroyed?
+        logical(lk), optional, intent(in) :: overwrite_a
+        !> [optional] Provide pre-allocated workspace, size to be checked with pivoting_qr_space.
```
perazz:
```suggestion
        !> [optional] Provide pre-allocated workspace, size to be checked with qr_space.
```

example/linalg/example_pivoting_qr_space.f90
```diff
@@ -0,0 +1,25 @@
+! QR example with pre-allocated storage
```
perazz:
```suggestion
! Pivoting QR example with pre-allocated storage
```

perazz (approved):
Thank you for this implementation @loiseaujc, great work. 
I stand by @jvdp1's comments, and I added a few edits/typos. 

Overall, looks good to me with the above.

src/stdlib_linalg_lapack.fypp
loiseaujc: Yeah sure. I copied what looked like the standard way to do it in the `stdlib_linalg_lapack.fypp` file, eg.

```
#:for ik,it,ii in LINALG_INT_KINDS_TYPES
#ifdef STDLIB_EXTERNAL_LAPACK${ii}$
               pure recursive subroutine cgeqrt3( m, n, a, lda, t, ldt, info )
                    import sp,dp,qp,${ik}$,lk
                    implicit none
                    integer(${ik}$), intent(out) :: info
                    integer(${ik}$), intent(in) :: lda,m,n,ldt
                    complex(sp), intent(inout) :: a(lda,*)
                    complex(sp), intent(out) :: t(ldt,*)
               end subroutine cgeqrt3
#else
               module procedure stdlib${ii}$_cgeqrt3
#endif
#ifdef STDLIB_EXTERNAL_LAPACK${ii}$
               pure recursive subroutine dgeqrt3( m, n, a, lda, t, ldt, info )
                    import sp,dp,qp,${ik}$,lk
                    implicit none
                    integer(${ik}$), intent(out) :: info
                    integer(${ik}$), intent(in) :: lda,m,n,ldt
                    real(dp), intent(inout) :: a(lda,*)
                    real(dp), intent(out) :: t(ldt,*)
               end subroutine dgeqrt3
#else
               module procedure stdlib${ii}$_dgeqrt3
#endif
...
```

but simply lumped everything into a single `#ifdef` block. Are there any technical reasons why you guys went this way initially rather than fypp-ified the whole prefix thing ?

test/linalg/test_linalg_pivoting_qr.fypp
loiseaujc: Will do right away.

src/stdlib_linalg_lapack.fypp
perazz: Lumping is also a valid approach imho. The original treatment was per-routine because not all generic LAPACK interfaces had exactly the same templatable s/d/c/z pattern. In this case it applies, so what you did looks good, but it could be even more templated as @jvdp1 suggested: something on the lines of 
```Fortran
#:for ik,it,ii in LINALG_INT_KINDS_TYPES
#:for rk,rt,ri in ALL_KINDS_TYPES
#:if rk in ["sp", "dp"]
#ifdef STDLIB_EXTERNAL_LAPACK${ii}$
               pure recursive subroutine ${ri}$geqrt3( m, n, a, lda, t, ldt, info )
                    import sp,dp,qp,${ik}$,lk
                    implicit none
                    integer(${ik}$), intent(out) :: info
                    integer(${ik}$), intent(in) :: lda,m,n,ldt
                    ${rt}$, intent(inout) :: a(lda,*)
                    ${rt}$, intent(out) :: t(ldt,*)
               end subroutine ${ri}$geqrt3
#else
#:endif
               module procedure stdlib${ii}$_${ri}$geqrt3
#:if rk in ["sp", "dp"]
#endif
#:endif
#:endfor
#:endfor
````

loiseaujc: Alright, job's done for `*GEQP3`.

loiseaujc:
Working on it. I don't think it is a tolerance issue though. The tolerance I use in the tests for now is `tol = 100 * sqrt(epsilon(1.0_wp))`. For single precision, that is about $10^{-2}$, and $10^{-6}$ for double precision. The comparison is done element-wise.

PR #1049, opened 2025-10-24
Modularized bitsets

jvdp1:
Based on #1031, #1033

Here is a proposition to modularize `stdlib_bitsets` in `cmake`.
Furthermore, compiling `stdlib_bitsets` can be now set off with `cmake`

\cc: @eduardz1 @jalvesz @perazz

### DIFF
--- a/config/fypp_deployment.py
+++ b/config/fypp_deployment.py
@@ -25,8 +25,8 @@ def pre_process_fypp(args):
     kwd.append("-DPROJECT_VERSION_MAJOR="+str(args.vmajor))
     kwd.append("-DPROJECT_VERSION_MINOR="+str(args.vminor))
     kwd.append("-DPROJECT_VERSION_PATCH="+str(args.vpatch))
-    if args.with_bitsets:
-        kwd.append("-DWITH_BITSETS=True")
+    if args.no_bitsets:
+        kwd.append("-DWITH_BITSETS=False")
     if args.with_qp:
         kwd.append("-DWITH_QP=True")
     if args.with_xdp:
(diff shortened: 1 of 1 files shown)


jvdp1:
> Here is an idea: In the `include` or `src` folder we add a file `macros.h`. Within, it can be defined:
> 
> ```c++
> #ifdef STDLIB_WITH_BITSET
> #define WITH_BITSET 1
> #else
> #define WITH_BITSET 0
> #endif
> ```
> 
> This file can be included elsewhere with `#include 'macros.h'` to use `#ifdef WITH_BITSET`. Like this, a default can be defined within and the user can override if needed with a flag `-DSTDLIB_WITH_BITSET` (or the opposite, depending which default to set).

Thank you. I tried with this strategy, but then I was challenged with the `fypp` loop approach, such as:
```fortran
    #:for k1, t1 in IRB_KIND_TYPES
      module procedure :: swap_${k1}$
   #:endfor
````
With `fypp`, it is easy to exclude or include the support for DT `bitset_type` when defining the variable `IRB_KIND_TYPES`. However, it is not so straightforward with the combinasion of `fypp` and `cpp`.

An ugly (at least for me) solution could be:
```fortran
    #:for k1, t1 in IRB_KIND_TYPES
    #:if 'bitset' in t1
#ifdef WITH_BITSET
    #:endif
      module procedure :: swap_${k1}$
    #:if 'bitset' in t1
#end
    #:endif
   #:endfor
```
I have the feeling that such an approach will result in a code even more difficult to follow and maintain.

All suggestions are welcome.

jvdp1:
> I tried to give it some more thoughts but indeed I can't think of a more elegant way of handling this. Your solution is very good, just wanted to check if we could rely on c-preprocessing which is much more standard.

Too bad! I would also prefer a c-preprocessing approach.
Let wait a few days. Maybe one of us will find a good idea!

> In any case, I would say, given that both GNU and Intel compilers can handle all the features and are the most common compilers, to keep it activated by default and let users remove it with fypp for other compilers which can't compile these modules.

I agree. At least it will provide some flexibility to the users.

> I would like to keep the idea of a macro header file open for future ideas though, I think it might facilitate many future hurdles.

Definitively!

jvdp1:
> One solution to combine fypp and cpp preprocessing for this module and keep it "readable" would be :
> 
> * On the implementation modules like sort or math: keep the bitset family separated:
>   Don't mix `#:set IRB_KIND_TYPES = IRB_KIND_TYPES + BITSET_KINDS_TYPES` such that we could "simply" state:
> 
> ```fortran
> module stdlib_math
>     use stdlib_kinds, only: int8, int16, int32, int64, sp, dp, xdp, qp
>     use stdlib_optval, only: optval
> #if WITH_BITSETS
>     use stdlib_bitsets, only: bitset_64, bitset_large
> #endif
> 
>     implicit none
>     private
> ...
>     interface swap
>     #:for k1, t1 in IR_KINDS_TYPES
>     module procedure :: swap_${k1}$
>     #:endfor
> #if WITH_BITSETS
>     #:for k1, t1 in BITSET_KINDS_TYPES
>     module procedure :: swap_${k1}$
>     #:endfor
> #endif
>     ...
>     end interface
> ```

I am afraid that this approach will consist in duplicating codes. For example, the code for the implementation of `swap` will require the same approach resulting in copying the same code.
Or do I miss something?

PR #1050, opened 2025-10-30
Modularized bitset cpp

jvdp1:
Based on #1031, #1033
Closes #1049 

Here is a proposition to modularize `stdlib` using `cpp` directives (instead of `fypp`) based on [this comment](https://github.com/fortran-lang/stdlib/pull/1049#issuecomment-3467894958)

\cc: @eduardz1 @jalvesz @perazz

### DIFF
--- a/include/common.fypp
+++ b/include/common.fypp
@@ -151,7 +151,7 @@ $:"s" if cmplx=="c" else "d" if cmplx=="z" else "x" if cmplx=="y" else "q" if cm
 #:set BITSET_INIT = ["" for k in BITSET_KINDS]
 
 #! Bitset CPP directive to be considered during templating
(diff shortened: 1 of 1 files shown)


jalvesz:
Sorry @jvdp1, I was out of office for a while and couldn't check everything before leaving.

>However, I think that the file macros.inc is not required if the directive
#ifdef STDLIB_BITSET
is replaced by
#ifndef STDLIB_NO_BITSET

You are absolutely write about this. What would you prefer to stick with? What I do find better for readability is formulating in "positive" terms. Also, while not mandatory, if we stick with using the include approach, it would be maybe good to have it explicitly indicating the values:
```
#ifdef STDLIB_NO_BITSET
#define STDLIB_BITSET 0
#else
#define STDLIB_BITSET 1
#endif
```
Which would lead to the use of `#if STDLIB_BITSET ... ; #else ... ; #endif`

PR #1054, opened 2025-11-04
Fix(tridiagonal_matrices): spmv y vector overwrite and missing error check in impure init functions

Mahmood-Sinan:
Summary:
This pr fixes two issues within tridiagonal_matrices module.
1. Inside the impure functions for initializing `tridiagonal_matrices`, the matrix fields were assigned even if there was an error(for example, when the sizes of du and dl vectors differed from dv by more than one). 
2. Inside spmv subroutine, the y vector was forcefully being zeroed instead of using the passed argument.

Changes:
1. Added an `if (err0%ok())` check in the impure initialization functions.
2. Removed the unintended `y = 0.0_${k1}$` in the spmv subroutine.

### DIFF
--- a/src/stdlib_specialmatrices_tridiagonal.fypp
+++ b/src/stdlib_specialmatrices_tridiagonal.fypp
@@ -69,9 +69,9 @@ submodule (stdlib_specialmatrices) tridiagonal_matrices
             call linalg_error_handling(err0)
         endif
         ! Matrix elements.
-        A%dl = [(dl, i = 1, n-1)]
-        A%dv = [(dv, i = 1, n)]
-        A%du = [(du, i = 1, n-1)]
+        allocate( A%dl(n-1), source = dl )
+        allocate( A%dv(n), source= dv )
+        allocate( A%du(n-1), source = du )
     end function
 
     module function initialize_tridiagonal_impure_${s1}$(dl, dv, du, err) result(A)
@@ -135,9 +135,9 @@ submodule (stdlib_specialmatrices) tridiagonal_matrices
             ! Description of the matrix.
             A%n = n
             ! Matrix elements.
-            A%dl = [(dl, i = 1, n-1)]
-            A%dv = [(dv, i = 1, n)]
-            A%du = [(du, i = 1, n-1)]
+            allocate( A%dl(n-1), source = dl )
+            allocate( A%dv(n), source= dv )
+            allocate( A%du(n-1), source = du )
         endif
     end function
     #:endfor
--- a/test/linalg/test_linalg_specialmatrices.fypp
+++ b/test/linalg/test_linalg_specialmatrices.fypp
@@ -76,7 +76,8 @@ contains
 
                     y1 = 0.0_wp
                     call random_number(y2)
-                    y1 = alpha * matmul(Amat, x) + beta * y2 ; call spmv(A, x, y2, alpha=alpha, beta=beta)
+                    y1 = alpha * matmul(Amat, x) + beta * y2
+                    call spmv(A, x, y2, alpha=alpha, beta=beta)
                     call check(error, all_close(y1, y2), .true.)
                     if (allocated(error)) return
                 end do


jalvesz:
I guess we could add an optional error output argument to check for the admissibility of `op`. Regarding `alpha` and `beta`, they could in principle take on any real (or complex) value, so I don't think it is necessary to be so strict as to check on their value.

src/stdlib_specialmatrices_tridiagonal.fypp
```diff
@@ -124,16 +126,19 @@ submodule (stdlib_specialmatrices) tridiagonal_matrices
...
-        ! Matrix elements.
-        A%dl = [(dl, i = 1, n-1)]
-        A%dv = [(dv, i = 1, n)]
-        A%du = [(du, i = 1, n-1)]
+
+        if(err0%ok()) then
+            ! Description of the matrix.
+            A%n = n
+            ! Matrix elements.
+            A%dl = [(dl, i = 1, n-1)]
```
jalvesz: I know that the brackets syntax looks elegant but it does 2 allocations plus an assignment. Preferer the more efficient syntax:
```Fortran
allocate( A%dl(n-1), source = dl )
allocate( A%dv(n), source= dv )
allocate( A%du(n-1), source = du )
```

test/linalg/test_linalg_specialmatrices.fypp
```diff
@@ -56,13 +60,27 @@ contains
...
+
+            ! Test y = alpha * A @ x + beta * y for alpha,beta in {-1,0,1}
+            do i = 1, 3
+                do j = 1,3
+                    alpha = coeffs(i)
+                    beta = coeffs(j)
+
+                    y1 = 0.0_wp
+                    call random_number(y2)
+                    y1 = alpha * matmul(Amat, x) + beta * y2 ; call spmv(A, x, y2, alpha=alpha, beta=beta)
```
jalvesz: Please use two lines for this. `call`s should always be in their own line. Same for array assignments. The use  of ";" to mix in a single line multiple executions is recommended only for very basic scalar operations.

jalvesz (changes requested):

loiseaujc:
I've just tried with `openblas 0.3.30` and it follows the `lapack` documentation. Any value of  `alpha` not in $(-1, 0, 1)$ is treated as 0, and any value of `beta` not in the same set is treated as 1.

Mahmood-Sinan:
Thank you for the suggestions. I have made the changes. Please let me know if anything else needs adjustment.

jalvesz:
Thanks @Mahmood-Sinan for these fixes, LGTM!

@loiseaujc I would propose that this PR is kept as is in terms of scope and then open 2 different ones:
PR1. Add an optional `err` output to the `spmv` kernels for sparse and special matrices checking only on the admissibility of the `op` argument.
PR2. As is, these higher level `spmv` are supposed to be general with respect to `alpha` and `beta`, but indeed blas\lapack only provide specialized ones. I would propose to included within the tridiagonal module an extention to `*LAGTM` to cover for the general case and that within the `spmv` kernel a check is carried out to switch between the lapack backed `*LAGTM` for the -1,0,1 cases, else use the a `stdlib` internal generic version:
```Fortran
pure subroutine stdlib_glagtm_sp( trans, n, nrhs, alpha, dl, d, du, x, ldx, beta,b, ldb )
     !! GLAGTM performs a matrix-vector product of the form
     !! B := alpha * A * X + beta * B
     !! where A is a tridiagonal matrix of order N, B and X are N by NRHS
     !! matrices, and alpha and beta are real scalars
           use stdlib_blas_constants_sp, only: negone, zero, half, one, two, three, four, eight, ten, czero, chalf, cone, cnegone
           ! Scalar Arguments 
           integer, parameter :: ilp = int32
           integer, parameter :: wp = sp
           character, intent(in) :: trans
           
           integer(ilp), intent(in) :: ldb, ldx, n, nrhs
           real(wp), intent(in) :: alpha, beta
           ! Array Arguments 
           real(wp), intent(inout) :: b(ldb,*)
           real(wp), intent(in) :: d(*), dl(*), du(*), x(ldx,*)
        ! =====================================================================
           
           ! Local Scalars 
           integer(ilp) :: i, j
           ! Executable Statements 
           if( n==0 )return
           ! multiply b by beta if beta/=0
           if( beta==zero ) then
            b(1:n,1:nrhs) = zero
           else
            b(1:n,1:nrhs) = beta * b(1:n,1:nrhs)
           end if

           if( trans(1:1) == 'N' ) then
              ! compute b := b + alpha * a*x
              do j = 1, nrhs
                 if( n==1_ilp ) then
                    b( 1_ilp, j ) = b( 1_ilp, j ) + alpha*d( 1_ilp )*x( 1_ilp, j )
                 else
                    b( 1_ilp, j ) = b( 1_ilp, j ) + alpha*d( 1_ilp )*x( 1_ilp, j ) + alpha*du( 1_ilp )*x( 2_ilp, j )
                    b( n, j ) = b( n, j ) + alpha*dl( n-1 )*x( n-1, j ) + alpha*d( n )*x( n, j )
                    do i = 2, n - 1
                       b( i, j ) = b( i, j ) + alpha*dl( i-1 )*x( i-1, j ) + alpha*d( i )*x( i, j )  &
                                 + alpha*du( i )*x( i+1, j )
                    end do
                 end if
              end do
           else
              ! compute b := b + alpha * a**t*x
              do j = 1, nrhs
                 if( n==1_ilp ) then
                    b( 1_ilp, j ) = b( 1_ilp, j ) + alpha*d( 1_ilp )*x( 1_ilp, j )
                 else
                    b( 1_ilp, j ) = b( 1_ilp, j ) + alpha*d( 1_ilp )*x( 1_ilp, j ) + alpha*dl( 1_ilp )*x( 2_ilp, j )
                    b( n, j ) = b( n, j ) + alpha*du( n-1 )*x( n-1, j ) + alpha*d( n )*x( n, j )
                    do i = 2, n - 1
                       b( i, j ) = b( i, j ) + alpha*du( i-1 )*x( i-1, j ) + alpha*d( i )*x( i, j ) &
                                 + alpha*dl( i )*x( i+1, j )
                    end do
                 end if
              end do
           end if
           return
end subroutine stdlib_glagtm_sp
```

Mahmood-Sinan:
Thanks for approving. And @loiseaujc Yes, I’d be happy to work on the generalized LAGTM extension. I’ll likely need some help or pointers as I go.

PR #1056, opened 2025-11-12
Add default (mold, array_size) overloads for rvs_normal

### LINKED ISSUE #963: Make `loc` and `scale` arguments of rvs_normal optional when returning an array
I wish the `loc` and `scale` arguments of `rvs_normal` were optional when generating a 1D array of variates, so that one could write

`x = rvs_normal(array_size=n)`

instead of

`x = rvs_normal(0.0_dp, 1.0_dp, array_size=n)`

That is how numpy.random.normal works.

gururaj1512:
Resolves #963

- Add convenience overloads for rvs_normal that accept (mold, array_size) and return standard-normal arrays (loc=0, scale=1) for both real and complex kinds.
- Add deterministic tests that compare the new (mold, array_size) call with the explicit (loc=0, scale=1, array_size) call to ensure identical behavior.

### DIFF
--- a/doc/specs/stdlib_stats_distribution_normal.md
+++ b/doc/specs/stdlib_stats_distribution_normal.md
@@ -23,7 +23,7 @@ With two arguments, the function returns a normal distributed random variate \(N
 
 With three arguments, the function returns a rank-1 array of normal distributed random variates.
 
-With two arguments `array_size` and `mold`, the function returns a rank-1 array of standard normal distributed random variates \(N(0,1)\), where the `mold` argument is used only to determine the output type and kind.
+With one or two arguments where the first is `array_size`, the function returns a rank-1 array of standard normal distributed random variates \(N(0,1)\). The `mold` argument determines the output type and kind; it is optional only for `real(dp)` (and defaults to `real(dp)` when omitted), but required for all other types.
 
 @note
 The algorithm used for generating exponential random variates is fundamentally limited to double precision.[^1]
@@ -32,7 +32,7 @@ The algorithm used for generating exponential random variates is fundamentally l
 
 `result = ` [[stdlib_stats_distribution_normal(module):rvs_normal(interface)]] `([loc, scale] [[, array_size]])`
 
-`result = ` [[stdlib_stats_distribution_normal(module):rvs_normal(interface)]] `(array_size, mold)`
+`result = ` [[stdlib_stats_distribution_normal(module):rvs_normal(interface)]] `(array_size [, mold])`
 
 ### Class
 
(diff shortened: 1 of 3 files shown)


gururaj1512:
@jvdp1, Can you have a look. Needs review...!

src/stdlib_stats_distribution_normal.fypp
```diff
@@ -34,6 +34,10 @@ module stdlib_stats_distribution_normal
         #:for k1, t1 in RC_KINDS_TYPES
             module procedure rvs_norm_array_${t1[0]}$${k1}$  !3 dummy variables
         #:endfor
+
+        #:for k1, t1 in RC_KINDS_TYPES
+            module procedure rvs_norm_array_default_${t1[0]}$${k1}$  !2 dummy variables (mold, array_size)
```
jvdp1: This can be included in the previous `fypp` loop (Line 35).

src/stdlib_stats_distribution_normal.fypp
```diff
@@ -238,6 +242,82 @@ contains
 
     #:endfor
 
+    #:for k1, t1 in REAL_KINDS_TYPES
+        impure function rvs_norm_array_default_${t1[0]}$${k1}$ (mold, array_size) result(res)
```
jvdp1:
```suggestion
        impure function rvs_norm_array_default_${t1[0]}$${k1}$ (array_size, mold) result(res)
```
`mold` is usually the last argument.

src/stdlib_stats_distribution_normal.fypp
```diff
@@ -238,6 +242,82 @@ contains
...
+                        hz = dist_rand(1_int32)
+                        iz = iand(hz, 127)
+                        if (abs(hz) < kn(iz)) then
+                            re = hz*wn(iz)
+                            exit L1
+                        end if
+                    end do L1
+                end if
+                res(i) = re  ! Default: loc=0, scale=1, so re*1 + 0 = re
+            end do
```
jvdp1:
```suggestion
            res = rvs_norm_array_default_${t1[0]}$${k1}$(0._${k1}, 1._${k1}$, array_size)
```
I propose to call the procedure, instead of repeating the code, except if there is a good reason to repeat it (I didn't look carefully to the code; if the code needs to be (partly) repeated, the reason should be provided in the comments)..

jvdp1 (commented):
Thank you @gururaj1512 . I have some suggestions below.

jvdp1:
> @jvdp1, I've refactored as per your suggestion.

Thank you. LGTM. Could you also update the specs, please?

doc/specs/stdlib_stats_distribution_normal.md
```diff
@@ -23,13 +23,17 @@ With two arguments, the function returns a normal distributed random variate \(N
...
+With two arguments `array_size` and `mold`, the function returns a rank-1 array of standard normal distributed random variates \(N(0,1)\), where the `mold` argument is used only to determine the output type and kind.
+
 @note
 The algorithm used for generating exponential random variates is fundamentally limited to double precision.[^1]
 
 ### Syntax
 
 `result = ` [[stdlib_stats_distribution_normal(module):rvs_normal(interface)]] `([loc, scale] [[, array_size]])`
 
+`result = ` [[stdlib_stats_distribution_normal(module):rvs_normal(interface)]] `(array_size, mold)`
```
perazz:
```suggestion
`result = ` [[stdlib_stats_distribution_normal(module):rvs_normal(interface)]] `(array_size [, mold])`
```

`mold` is an optional argument

doc/specs/stdlib_stats_distribution_normal.md
```diff
@@ -40,13 +44,15 @@ Elemental function (passing both `loc` and `scale`).
...
+`array_size`: optional argument has `intent(in)` and is a scalar of type `integer`. When used with `loc` and `scale`, specifies the size of the output array. When used with `mold`, must be provided as the first argument.
+
+`mold`: optional argument has `intent(in)` and is a scalar of type `real` or `complex`. Used only to determine the type and kind of the output; its value is not referenced. When provided, generates standard normal variates \(N(0,1)\).
 
 `loc` and `scale` arguments must be of the same type.
 
 ### Return value
 
-The result is a scalar or rank-1 array, with a size of `array_size`, and the same type as `scale` and `loc`. If `scale` is non-positive, the result is `NaN`.
+The result is a scalar or rank-1 array, with a size of `array_size`, and the same type as `scale` and `loc` (or same type and kind as `mold` when using the `array_size, mold` form). If `scale` is non-positive, the result is `NaN`.
```
perazz:
```suggestion
The result is a scalar or rank-1 array, with a size of `array_size`, and the same type as `scale` and `loc` (or same type and kind as `mold` when using the `(array_size, mold)` form). If `scale` is non-positive, the result is `NaN`.
```

doc/specs/stdlib_stats_distribution_normal.md
gururaj1512: I kept mold as a required argument in the (array_size, mold) interface because making it optional would eliminate the type disambiguation mechanism. Without mold, the compiler cannot determine whether to generate real(sp), real(dp), complex(sp), or other kind variants.

perazz: Sorry @gururaj1512 I thought that was just a typo in the description. What I wanted to suggest is it would be good if the API for `mold` remains similar to other places in the library. I.e., to keep `mold` optional and return a `double precision` real result in the default case: 

https://github.com/fortran-lang/stdlib/blob/03736faab793ee52fd88e79f6d21d8bda1d294ed/src/stdlib_linalg.fypp#L1789-L1796

gururaj1512: This seems correct. I will refactor according to it.

gururaj1512:
@perazz, I have refactored the implementations as suggested. Ready for review.

PR #1059, opened 2025-11-21
Update macOS architecture and OS version in README.md

### LINKED ISSUE #1058: macOS CI runner changed from x86_64 to Arm64
In the supported compilers table in README.md, the architecture for macOS 14 is listed as x86_64. However, the ```macos-14``` CI runner that was added in #1042 is [Arm64-based](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories). Should the table be updated to reflect this, and is there a preferred label for this architecture, Arm64 or AArch64? If an x86_64-based runner is still wanted, then ```macos-15-intel```  could be used instead of the soon-to-be-removed ```macos-13```.

edvinolo:
Closes #1058. The macos-14 CI runner is based on Arm64, so the compiler table in README.md has been updated to reflect this. Also, the macOS version number has been updated. @jvdp1

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ The following combinations are tested on the default branch of stdlib:
 Name | Version | Platform | Architecture
 --- | --- | --- | ---
 GCC Fortran | 10, 11, 12, 13, 14 | Ubuntu 24.04.3 LTS | x86_64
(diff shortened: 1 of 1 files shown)


jvdp1 (approved):
Thank you @edvinolo for updating the README

Issue #1063, opened 2025-11-30
Regression on Darwin: `Undefined symbols for architecture ppc: ___stdlib_string_type_MOD_new_string`


barracuda156:
### Description

Current master fails to build  on powerpc-apple-darwin with a linking error:
```
Undefined symbols for architecture ppc:
  "___stdlib_string_type_MOD_new_string", referenced from:
      ___stdlib_io_MOD_get_line_string in stdlib_io.f90.o
ld: symbol(s) not found for architecture ppc
```

### Expected Behaviour

The build should succeed, like it was earlier.

### Version of stdlib

d57c77690288aa11690344b92e6edfc414b151b7

### Platform and Architecture

macOS 10.6.8 / ppc

### Additional Information

```
[...]

PR #1064, opened 2025-11-30
fix Issue #1063

perazz:
The LAPACK library fails to link with an undefined symbol error:

  Undefined symbols for architecture ppc:
    "___stdlib_string_type_MOD_new_string", referenced from:
        ___stdlib_io_MOD_get_line_string in stdlib_io.f90.o

When the LAPACK module was modularized in #1033 (commit 80a8164c), the stdlib_string_type_constructor.fypp submodule was not included in src/lapack/CMakeLists.txt.

The dependency chain is:
- stdlib_io.fypp uses string_type() constructor in get_line_string
- stdlib_string_type.fypp declares the new_string interface
- stdlib_string_type_constructor.fypp provides the implementation (submodule)

The LAPACK library included the first two but was missing the third, resulting in an undefined symbol at link time.

cc: @barracuda156 @jvdp1

### DIFF
--- a/src/lapack/CMakeLists.txt
+++ b/src/lapack/CMakeLists.txt
@@ -7,6 +7,9 @@ set(lapack_fppFiles
     ../stdlib_io.fypp
     ../stdlib_ascii.fypp
     ../stdlib_string_type.fypp
+    ../stdlib_string_type_constructor.fypp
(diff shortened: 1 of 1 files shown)


perazz:
I am not sure this will be a good fix - some of those source files may now appear as duplicates

PR #1066, opened 2025-12-07
Modularized stdlib_stats

jvdp1:
Here is an attempt to modularize further `stdlib` with making `stdlib_stats` optional


jalvesz (approved):
LGTM @jvdp1, thank you!

PR #1068, opened 2025-12-09
feat: Add generalized lagtm routine supporting arbitrary values for alpha and beta

Mahmood-Sinan:
This PR adds a generalized `lagtm` routine that supports arbitrary values of `alpha` and `beta` for the operation:

~~~
y = alpha * A * x + beta * y
~~~

where `A` is a tridiagonal matrix. The existing LAPACK `lagtm` restricts `alpha` and `beta` to the set `{ -1, 0, 1 }`, which limits its applicability in several higher-level routines.
The routine can be called as follows:
~~~
glagtm(trans, n, nrhs, alpha, dl, d, du, x, ldx, beta, b, ldb)
~~~
which is similar to the LAPACK version:
~~~
lagtm( trans, n, nrhs, alpha, dl, d, du, x, ldx, beta,b, ldb )
~~~
except that alpha and beta can be arbitrary(including complex values).

This new subroutine is placed inside
`src/lapack_extended/stdlib_extended_lapack_base.fypp` (interface) and
`src/lapack_extended/stdlib_extended_lapack.fypp` (implementation).


jalvesz:
Hi @Mahmood-Sinan, this looks pretty neat! the only thing I would suggest would be to keep the files and folders naming somewhat coherent with respect to the name `lapack_extended`.

src/stdlib_specialmatrices_tridiagonal.fypp
```diff
@@ -172,16 +174,35 @@ submodule (stdlib_specialmatrices) tridiagonal_matrices
         beta_  = 0.0_${k1}$ ; if (present(beta))  beta_  = beta
         op_    = "N"        ; if (present(op))    op_    = op
 
+        is_alpha_special = (alpha_ ==  1.0_${k1}$  .or. alpha_ ==  0.0_${k1}$  .or. alpha_ == -1.0_${k1}$)
+        is_beta_special  = (beta_  ==  1.0_${k1}$  .or. beta_  ==  0.0_${k1}$  .or. beta_  == -1.0_${k1}$)
```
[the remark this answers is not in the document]
jalvesz: An easier solution is to use `stdlib_constants` where constants zero and one have been added for all kinds. So simply replacing the real values with the respective constants would do the job.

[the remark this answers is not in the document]
Mahmood-Sinan: > An easier solution is to use `stdlib_constants` where constants zero and one have been added for all kinds. So simply replacing the real values with the respective constants would do the job.

Another solution would be to wrap those two lines and the declaration of `is_alpha_special` and `is_beta_special` in an `#:if t1.startswith('real')` block. Then we can avoid the checks for `is_alpha_special` and `is_beta_special` in case of complex types. This works because for complex types we always call glagtm regardless of the values of `alpha` and `beta`.
Let me know which approach you’d prefer.

Mahmood-Sinan:
> Hi @Mahmood-Sinan, this looks pretty neat! the only thing I would suggest would be to keep the files and folders naming somewhat coherent with respect to the name `lapack_extended`.

Sure, I will do that in the next commit.

src/stdlib_specialmatrices_tridiagonal.fypp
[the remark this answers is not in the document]
jalvesz: Sounds good to me!

Mahmood-Sinan:
I’ve applied the changes. Please let me know if anything else is needed.

jalvesz:
I don't really understand why the codecov jobs are failing. It would be good to understand this, maybe @loiseaujc could have an insight here?

src/lapack_extended/stdlib_lapack_extended.fypp
```diff
@@ -0,0 +1,85 @@
...
+    implicit none
+contains
+    #:for ik,it,ii in LINALG_INT_KINDS_TYPES
+    #:for k1,t1,s1 in KINDS_TYPES
+    pure module subroutine stdlib${ii}$_glagtm_${s1}$(trans, n, nrhs, alpha, dl, d, du, x, ldx, beta, b, ldb)
+        character, intent(in) :: trans
+        integer(${ik}$), intent(in) :: ldb, ldx, n, nrhs
+        ${t1}$, intent(in) :: alpha, beta
+        ${t1}$, intent(inout) :: b(ldb,*)
+        ${t1}$, intent(in) :: d(*), dl(*), du(*), x(ldx,*)
```
jvdp1, line 16: Can we modernize this interface?
```suggestion
        ${t1}$, intent(inout) :: b(:,:)
        ${t1}$, intent(in) :: d(:), dl(:), du(:), x(:,:)
```
What is the reason to keep it?

jvdp1 (approved):
Overall LGTM. I have only a minor question

src/lapack_extended/stdlib_lapack_extended.fypp
Mahmood-Sinan: @jvdp1 Thanks for the review. I kept the LAPACK style interface(`*`) to stay close to `lagtm`, but I agree that assumed shape arrays would be more modern. I'm happy to switch to that, if you prefer.

jvdp1: @perazz @jalvesz what do you think about that?

jalvesz: Such change should not be taken lightly. While assumed-shape is considered more safe than assumed-size for the simple reason that it enforces bound checking and helps with runtime problem detection, it also implies an API behavioral change on what is allowed or not to do with the interface.

With assumed-size declaration, one can have a 1D working array passed to the function which will be reinterpreted as a 2D array internally. with assumed-shape, this is no longer possible as ranks should match between caller and callee.

If this routine is intended to closely match but extending the capabilities of `*lagtm`, then I would suggest not changing the array declaration style.

jvdp1: > Such change should not be taken lightly. 

I fully agree with you. As I don't have a full overview of all the old and newLAPACK features, a second opinion is needed.

> If this routine is intended to closely match but extending the capabilities of `*lagtm`, then I would suggest not changing the array declaration style.

I agree with this justification.

@jalvesz @Mahmood-Sinan I consider this suggestion has being resolved. The PRcan be merged IMHO.

jvdp1:
@loiseaujc @jalvesz @Mahmood-Sinan it seems that the covecod failed because the Fortran example is not covered ([report here](https://app.codecov.io/gh/fortran-lang/stdlib/pull/1068?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=fortran-lang)). how should we deal with that?

Mahmood-Sinan:
> @loiseaujc @jalvesz @Mahmood-Sinan it seems that the covecod failed because the Fortran example is not covered ([report here](https://app.codecov.io/gh/fortran-lang/stdlib/pull/1068?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=fortran-lang)). how should we deal with that?

The example `example_specialmatrices_cdp_spmv` can be removed if you prefer, because there is already another example `example_specialmatrices_dp_spmv` for `spmv` interface. I kept it because `example_specialmatrices_dp_spmv` calls LAPACK `lagtm` and `example_specialmatrices_cdp_spmv` calls lapack_extended `glagtm` internally.

PR #1073, opened 2025-12-17
fix: remove parent module use statement in iterative solver submodules

edvinolo:
Remove the ```use stdlib_linalg_iterative_solvers``` statement in the iterative solver implementation submodules. The use statement is not needed since a submodule has access to its parent module via host association. There is also a paragraph in the standard that might disallow this, see e.g. the discussion here: https://github.com/llvm/llvm-project/issues/169863. Flang gives an error for these lines, although apparently not gfortran or ifx.  

Along the same lines, the submodules also contain use statements that are already present in the parent module and should therefore be redundant; these could also be removed if you agree.

(@jalvesz @jvdp1 since you have worked on the iterative solvers)


jalvesz:
Brilliant! thanks for the info, looking forward for progress on this

PR #1079, opened 2025-12-29
fix(stdlib_system): use STARTUPINFO for windows stdin redirection

JAi-SATHVIK:
### **Problem:-**

The previous implementation attempted to redirect standard input by appending the shell operator `< filename` directly to the command string `(e.g., cmd < input.txt).`

The Problem: The Windows API function CreateProcess does not spawn a shell (like cmd.exe). Consequently, it does not recognize or interpret shell operators.

The Result: The < character and the filename were passed as literal arguments to the child program. This caused the child program to receive "garbage" arguments while the actual stdin stream remained unassigned.

### **Solution:-**

Removed Incorrect Syntax: The code no longer appends `< filename` to the command line string.

Native Handle Management: Utilizes CreateFile to open a legitimate, read-only handle to the specified input file (or NUL if no file is provided).

Process Inheritance:  The handle is assigned to the hStdInput field of the STARTUPINFO structure.

The STARTF_USESTDHANDLES flag is enabled to instruct the OS to use the provided handles.

Resource Cleanup: Proper logic was added to ensure the file handle is closed safely after the process is launched, preventing resource leaks.

### DIFF
--- a/src/stdlib_system_subprocess.c
+++ b/src/stdlib_system_subprocess.c
@@ -37,6 +37,7 @@ void process_create_windows(const char* cmd, const char* stdin_stream,
     HANDLE hStdout = NULL, hStderr = NULL, hStdin = NULL;
     SECURITY_ATTRIBUTES sa = { sizeof(SECURITY_ATTRIBUTES), NULL, TRUE };
     FILE* stdin_fp = NULL;
+    char* full_cmd = NULL;
     
     // Initialize null handle
     (*pid) = 0;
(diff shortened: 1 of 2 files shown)


perazz (commented):
Thank you for the catch! LGTM overall. Before merging, I suggest:
- Make stdin redirection work in all cases: on Windows the child can only use hStdInput if the underlying handle is allowed to be “passed” to the child process. Please open both the real input file and the fallback NUL handle in a way that allows passing to the child, and ensure CreateProcess is called with handle passing enabled.
- Test: add a small Windows test that redirects stdin from a file and verifies the child reads expected content.
- Cleanup/leak: close any opened handles on all early-return/error paths to avoid leaks.

perazz (approved):

jalvesz (approved):

PR #1081, opened 2026-01-04
Additional modularization of stdlib

jvdp1:
**Update**:

* All files in `src` (except `stdlib_version.fypp`) were moved to the follwing subfolders (a library is created within each subfolder): `ansi`, ` bitsets`, `core`, `hashmaps`, ` io`,  `lapack_extended`  `linalg_core`  `math`  `selection`  `specialfunctions`  `stats`, `stringlist`  `system` `array`  `blas`     `constants`       `hash`  `intrinsics`  `lapack`  `linalg`           `logger`       `quadrature` `sorting`    `specialmatrices`  `strings`, `sparse` and `linalg_iterative`
* This structure closely follows  the structure in `test` and `examples`. 
* The folders `core` and `linalg_core` contain modules that are required by various parts of `stdlib` and were created to avoid interdependencies.
* The file `stdlib_version.fypp` was maintained in `src` to avoid to modify `cmake/stdlib.cmake` in this PR. It could be moved later to the subfolder `core`.
* Optional compilaton (like for `stdlib_ansi` and stdlib_hashmaps` will be added in a following PR.

Depends on #1088

### DIFF
--- a/.github/workflows/ci_modular.yml
+++ b/.github/workflows/ci_modular.yml
@@ -1,4 +1,4 @@
-name: CI
+name: ci_modular
 
 on: [push, pull_request]
 
--- a/.github/workflows/fpm-build.yml
+++ b/.github/workflows/fpm-build.yml
@@ -1,0 +1,41 @@
+name: fpm-build
+
+on: [push, pull_request]
+
+jobs:
+  fpm-build:
+    name: Run tests the fpm branches
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        toolchain: [{compiler: gcc, version: '13' }]
+        fpm_versions: [v0.12.0, current]
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2.3.1
+
+      - name: Set up Python 3.x
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.x
+
+      - name: Install requirements
+        run: pip install --upgrade -r config/requirements.txt
+
+      - uses: fortran-lang/setup-fortran@main
+        id: setup-fortran
+        with:
+          compiler: ${{ matrix.toolchain.compiler }}
+          version: ${{ matrix.toolchain.version }}
+
+      - name: Setup Fortran Package Manager
+        uses: fortran-lang/setup-fpm@v9
+        with:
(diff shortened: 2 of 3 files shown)


jvdp1:
> @jvdp1 I wonder if it would be advisable given all different options being added to include in the readme.md a table listing all possible macros/flags. I asked copilot to generate a first version that would look like this:

Thank you @jalvesz .I agree that such a table is needed. I will open a new PR based on your first version.

jvdp1:
Thank you @jalvesz for the review.

I am currently playing with this branch. I just noted that the lib `libfortran_stdlib.a` created and installed with `CMake` only contains the file related with `stdlib_version.fypp`. It means that the user has to mention all other libraries for the linker. I think that it makes things difficult to handle with, e.g., Makefile.
Furthermore, the current structure of the package installed by `CMake` does not correspond to the config templates.

jalvesz:
@jvdp1 quick question, I built and installed locally this branch under windows. I see the following archive:
image
From where I understand that only certain "modules" got actually separated into their own folder while all others just got sent to `fortran_stdlib`. Is this the expected archive architecture? I was expecting to see each of the new "modules" as their own independent folder. At this point I guess it is not that relevant. I'm also wondering why `test-drive` is being archived with `stdlib` when it is supposed to be a dependency only for testing the library.

jalvesz:
I think this PR is not only ready but we should consider trying to get it merged before more PRs are stacked. It is a pivotal and required housekeeping of the internal architecture and packaging.

jvdp1:
> I think this PR is not only ready but we should consider trying to get it merged before more PRs are stacked. It is a pivotal and required housekeeping of the internal architecture and packaging.

Thank you @jalvesz . @perazz Are you also happy with this PR? If yes, I think we can merge it.

I think it will be good to create a new version (I can open a PR for that) after this PR as it will modify quite significantly the structure of `stdlib`

jvdp1:
> @jvdp1 the modularization looks absolutely good to me. 

Thank you. Then I propose that we merge it.

>I have a question about the CI: would it be possible to require stdlib to pass a check on the stdlib-fpm repository? 

I can do it in another PR. However, `fpm` `v0.12.0` is currently used for the code coverage. So, do you mean a more recent version?

>Some of the recent updates seem to have broken fpm builds again:

Do you mean recent updates of `stdlib` or of `fpm`?

Issue #1037, opened 2025-10-06
enhance `coo2csr` procedure by allowing to internally sort the coo matrix


jalvesz:
### Motivation

In https://github.com/fortran-lang/stdlib/pull/1034#discussion_r2404572171 a discussion about sparse transformation for linear systems was opened.

Currently, if one first creates a `COO` matrix, it is requested to:
`call coo2sorted( COO,sort_data)` : sort and ensure unique (i,j,v) entries. This procedure reallocates the COO matrix to remove duplicates.
`call coo2csr(COO,CSR)` : assumes COO is sorted and unique

It could be possible to enhance the second call with a boolean flag to carry out both steps with a single call.

PR #1082, opened 2026-01-04
feat (sparse): add binary operators overloading

jalvesz:
This PR proposes support for binary operators `+, -, *, /` for all sparse matrices defined in the `stdlib_sparse_kinds` module.

- [x] Implementation with `pure` functions
- [x] Unit testing (`ELL` and `SELLC` skipped because of explicit zeros causing NaNs at division)
- [x] documentation

Additional fixes needed for this PR:
* Include support for `from_ijv` for the `CSC` type
* fix COO sorting call during `coo2csc` in case of a rectangular matrix.

### DIFF
--- a/src/linalg/stdlib_linalg.fypp
+++ b/src/linalg/stdlib_linalg.fypp
@@ -876,8 +876,7 @@ module stdlib_linalg
     !! This interface provides methods for computing the determinant of a matrix.
     !! Supported data types include `real` and `complex`.
     !! 
-    !!@note The provided functions are intended for square matrices only.          
-    !!@note BLAS/LAPACK backends do not currently support extended precision (``xdp``).
+    !!@note The provided functions are intended for square matrices only.
     !! 
     !!### Example
     !!
@@ -898,10 +897,8 @@ module stdlib_linalg
     !!```
     !!     
     #:for rk,rt in RC_KINDS_TYPES
-    #:if rk!="xdp" 
     module procedure stdlib_linalg_${rt[0]}$${rk}$determinant
     module procedure stdlib_linalg_pure_${rt[0]}$${rk}$determinant
-    #:endif
     #:endfor
   end interface det
 
@@ -920,7 +917,6 @@ module stdlib_linalg
     !! Supported data types include real and complex.
     !!
     !!@note The provided functions are intended for square matrices.
-    !!@note BLAS/LAPACK backends do not currently support extended precision (``xdp``).
     !!
     !!### Example
     !!
@@ -935,15 +931,12 @@ module stdlib_linalg
     !!```
     !     
     #:for rk,rt in RC_KINDS_TYPES
-    #:if rk!="xdp"
     module procedure stdlib_linalg_pure_${rt[0]}$${rk}$determinant
-    #:endif
     #:endfor        
   end interface operator(.det.)
 
   interface
     #:for rk,rt in RC_KINDS_TYPES
-    #:if rk!="xdp" 
     module function stdlib_linalg_${rt[0]}$${rk}$determinant(a,overwrite_a,err) result(det)
         !> Input matrix a[m,n]
         ${rt}$, intent(inout), target :: a(:,:)
@@ -960,7 +953,6 @@ module stdlib_linalg
         !> Matrix determinant
         ${rt}$ :: det                
     end function stdlib_linalg_pure_${rt[0]}$${rk}$determinant
-    #:endif
     #:endfor
   end interface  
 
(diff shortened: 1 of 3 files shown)


src/stdlib_sparse_conversion.fypp
```diff
@@ -258,9 +259,10 @@ contains
         temp(1,1:COO%nnz) = COO%index(2,1:COO%nnz)
         temp(2,1:COO%nnz) = COO%index(1,1:COO%nnz)
         allocate(data, source = COO%data )
+        
         nnz = COO%nnz
-        call sort_coo_unique_${s1}$( temp, data, nnz, COO%nrows, COO%ncols )
-
+        call sort_coo_unique_${s1}$( temp, data, nnz, COO%ncols, COO%nrows )
+
```
jvdp1:
```suggestion

```

src/stdlib_sparse_conversion.fypp
```diff
@@ -270,7 +272,7 @@ contains
             allocate( CSC%colptr(CSC%ncols+1) , source = 0 )
             allocate( CSC%data(CSC%nnz) , source = data(1:COO%nnz) )
         end if
-
+
```
jvdp1:
```suggestion

```

src/stdlib_sparse_conversion.fypp
```diff
@@ -740,6 +742,40 @@ contains
     end subroutine
     #:endfor
 
+    #:for k1, t1, s1 in (KINDS_TYPES)
+    subroutine csc_from_ijv_${s1}$(CSC,row,col,data,nrows,ncols)
+        type(CSC_${s1}$_type), intent(inout) :: CSC
```
jvdp1: should it be:

```suggestion
        type(CSC_${s1}$_type), intent(out) :: CSC
```

doc/specs/stdlib_sparse.md
```diff
@@ -361,4 +361,39 @@ If the `diagonal` array has not been previously allocated, the `diag` subroutine
...
+
+- Matrix scalar operators : 
+
+`B = alpha + A`
+
+`B = A - alpha`
+
+`B = alpha * A`
+
+`B = A / alpha`
```
jvdp1: Just to keep the same order:

```suggestion
`B = A + alpha`

`B = A - alpha`

`B = A * alpha`

`B = A / alpha`
```

doc/specs/stdlib_sparse.md
```diff
@@ -361,4 +361,39 @@ If the `diagonal` array has not been previously allocated, the `diag` subroutine
...
+
+*Note*: scalar addition and subtraction operators perform element-wise operations only on the stored (non-zero) values, not on the full mathematical matrix. Meaning, the sparsity pattern is preserved.
```
jvdp1: Is it not the case for scalar multiplication/division?

jvdp1 (commented):
Thank you. Useful additions. I left a few comments.

doc/specs/stdlib_sparse.md
jalvesz: Yes, but it is natural that the multiplication and division do not modify the sparsity pattern as zeros remain zeros. Addition and subtraction on the other hand would in principle create a dense matrix because all zeros would be filled. That's why the note is useful for the first two and redundant for the latter.

doc/specs/stdlib_sparse.md
jalvesz: I wrote different orders to show that both are permitted. Should I maybe add it with an "or" next by?

jvdp1: I would suggest to add a "or" to avoid confusion.

doc/specs/stdlib_sparse.md
jvdp1: OK, I see. In my mind, it was obvious that scalar addition and substratction to a sparse matrix do not modify its sparsity pattern. Hence, my confusion. But your note is indeed relevant!

src/stdlib_sparse_conversion.fypp
jalvesz: all `<>_from_ijv_<>` procedures define `intent(inout)`. If we change here, we might want to change them all ?

jvdp1: If in practice, it is always `intent(out)`, I think that it would be cleaner to change them all. However, I did not check the whole code. It could be also done in another PR.

jalvesz: Let's save it for a PR maybe together with #1037

PR #1083, opened 2026-01-05
Addition of tables summarizing preprocessing macros and flags

jvdp1:
As suggested by @jalvesz in [his comment](https://github.com/fortran-lang/stdlib/pull/1081#issuecomment-3708510700), here is a PR to introduce tables that summarize fypp and C-preprocessing flags and macros.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,10 +257,9 @@ with `prog` being the name of the example program (e.g., `example_sort`).
 `stdlib` uses two preprocessing steps:
 
 - *fypp* for meta-programming (templating and feature selection)
-- *C* preprocessing (compiler flag `-cpp`/`-fpp`) for conditional compilation
+- *C* preprocessing (activated through compiler flags like `-cpp`(GNU) or `-fpp`(Intel) or use of uppercase file suffix .F90) for conditional compilation
 
-*fypp* preprocessing macros and flags are supported only through `CMake` and the
-`python` script `config/fypp_deployment.sh`.
+*fypp* preprocessing macros and flags are supported through `CMake` or the `python` script `config/fypp_deployment.py`.
 *C* preprocessing macros and flags are supported through `CMake` and `fpm`.
 
 The table below lists all *fypp* preprocessing macros and flags currently used by `stdlib`:


README.md
```diff
@@ -252,6 +252,46 @@ fpm run --example prog
 with `prog` being the name of the example program (e.g., `example_sort`).
 
 
+### Preprocessing macros and flags
+
+`stdlib` uses two preprocessing steps:
+
+- *fypp* for meta-programming (templating and feature selection)
+- *C* preprocessing (compiler flag `-cpp`/`-fpp`) for conditional compilation
```
jalvesz:
```suggestion
- *C* preprocessing (activated through compiler flags like `-cpp`(GNU) or `-fpp`(Intel) or use of uppercase file suffix .F90) for conditional compilation
```

README.md
```diff
@@ -252,6 +252,46 @@ fpm run --example prog
...
+
+*fypp* preprocessing macros and flags are supported only through `CMake` and the
+`python` script `config/fypp_deployment.sh`.
```
jalvesz:
```suggestion
*fypp* preprocessing macros and flags are supported through `CMake` or the `python` script `config/fypp_deployment.py`.
```

jalvesz (changes requested):
Thanks for adding the table @jvdp1, here a couple of minor edits

jalvesz (approved):

perazz:
I agree that the positive form is definitely superior @jvdp1, I also personally think the `STDLIB_WITH_BITSETS` is also more clear than `STDLIB_WITH_BITSETS == 1`, so my vote would definitely go for the boolean form

jalvesz:
If I'm not understanding correctly the discussion please let me know. Maybe one way to simply things would be to unify in a single macro variable per library using only the positive version:

```Fortran
!--------- macro
#ifndef STDLIB_BITSET
#define STDLIB_BITSET 1
#endif

!--------- usage
#if STDLIB_BITSET
print *, "hello"
#else
print *, "goodbye"
#endif

end
```
* With `-cpp` gives `hello`
* With `-cpp -DSTDLIB_BITSET=1` gives `hello`
* With `-cpp -DSTDLIB_BITSET=0` gives `goodbye`

I think it would be less confusing as there is just one variable to manage.

perazz:
> Either with CMake or FPM, one would just modify to change the default behavior.

This was not possible with FPM as it did not parse macros, at least until [recently](https://github.com/fortran-lang/fpm/pull/1218)

jvdp1:
@perazz @jalvesz I removed the references to t`STDLIB_BITSET` and `STDLIB_STATS` from the tables. This PR can be reviewed and merged if there are no additional comments.

Issue #1080, opened 2025-12-29
Implement Principal Component Analysis (PCA) module


JAi-SATHVIK:
### **WHY NEEDED**
Principal Component Analysis is the workhorse of dimensionality reduction in modern scientific computing (e.g., climate patterns, fluid dynamics modes, machine learning preprocessing). Currently, users must manually combine stdlib_stats_cov, stdlib_stats_mean, and stdlib_linalg_svd to perform this common task. Adding a dedicated PCA module would significantly lower the barrier to entry for data science workflows in Fortran, bringing it closer to ecosystems like scikit-learn or MATLAB.

### **PLAN**
I have analyzed the current capabilities of stdlib_stats and stdlib_linalg and found we have all the prerequisites. I propose implementing a pca derived type or set of routines that offers:

**Dual Algorithms:**
**Eigendecomposition of Covariance Matrix:** Leveraging existing stdlib_stats_cov and stdlib_linalg_eig.
SVD Method: Performing SVD directly on the centered data matrix (often more numerically stable) using stdlib_linalg_svd.
**API Design:**
**pca_fit(X, n_components):** Computes the principal components (eigenvectors), singular values, and explained variance ratios.
**pca_transform(X, pca_model):** Projects new data into the reduced dimensional space.
**pca_inverse_transform():** Reconstructs original data from the reduced space.
**Integration:** This module would sit naturally in stdlib_stats, acting as a bridge between the statistics and linear algebra modules.

### Prior Art

https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html
https://www.mathworks.com/help/stats/pca.html

jalvesz:
Excellent idea @JAi-SATHVIK if you have a plan for it you can start a draft PR to polish the design.

My advice, lets try to iterate first on the high level API which is the one thing that once settles is harder to change, and then polish the inner details.

Given the dependency on both the stats and linalg modules, just be careful with the dependencies as there is an ongoing effort to modularize the build. As of now the stats module does not depend on either blas or lapack nor the higher level linalg APIs. But indeed a PCA module would make sense to live closer to the stats module rather than linalg conceptually speaking.

Poke @jvdp1 @perazz @loiseaujc

JAi-SATHVIK:
Thank you @jalvesz  

[...]

PR #1086, opened 2026-01-06
Implement Principal Component Analysis (PCA) module

JAi-SATHVIK:
#1080

### DIFF
--- a/src/stats/stdlib_stats_pca.fypp
+++ b/src/stats/stdlib_stats_pca.fypp
@@ -129,10 +129,10 @@ contains
       components = zero
       singular_values = zero
 
-      ! Input validation: check for empty arrays
-      if (n < 1 .or. p < 1) then
+      ! Input validation: check for empty arrays or non-positive dimensions
+      if (n < 1 .or. p < 1 .or. nc < 1) then
          err0 = linalg_state_type("pca", LINALG_VALUE_ERROR, &
-                "Input array must have at least 1 observation and 1 feature")
+                "Input array and output components must have at least 1 observation, feature, and component")
          call err0%handle(err)
          return
       end if
@@ -221,6 +221,9 @@ contains
       nc = size(components, 1, kind=ilp)
 
       ! Validate dimensions
+      if (nc < 1) then
+         call error_stop("ERROR (pca_transform): number of components must be at least 1")
+      end if
       if (size(components, 2, kind=ilp) /= p) then
          call error_stop("ERROR (pca_transform): components columns must match x columns")
       end if
@@ -258,6 +261,9 @@ contains
       p = size(components, 2, kind=ilp)
 
       ! Validate dimensions
+      if (nc < 1) then
+         call error_stop("ERROR (pca_inverse_transform): number of components must be at least 1")
+      end if
       if (size(components, 1, kind=ilp) /= nc) then
          call error_stop("ERROR (pca_inverse_transform): components rows must match x_reduced columns")
       end if


src/stats/stdlib_stats.fypp
```diff
@@ -639,4 +641,63 @@ module stdlib_stats
     #:endfor
   end interface moment
 
+
+  #! Note: PCA uses SVD and EIGH which rely on LAPACK. LAPACK backends do not support extended (xdp) or
```
jalvesz: This is not accurate. `stdlib`'s modernized BLAS/LAPACK do support all kinds. The only caveat is that when linking against an optimized library such as OpenBLAS or MKL, they only provide support for simple and double precision, so for extended and quadruple, the internal `stdlib` version will be used.

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,172 @@
...
+      allocate(x_centered(n, size(x, 2, kind=ilp)))
+      if (present(x_mean)) then
+         do i = 1, n
+            x_centered(i, :) = x(i, :) - x_mean
+         end do
+      else
+         x_centered = x
+      end if
+      
+      res = matmul(x_centered, transpose(components))
```
jalvesz: I would advise to switch to using `gemm` instead to avoid temporal array allocations and ensuring use of optimized backends when linking against.

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,172 @@
...
+      end if
+      
+      if (present(err)) err = err0
+      
+    end subroutine pca_${k1}$
+  #:endfor
+
+
+  #:for k1, t1 in REAL_KINDS_TYPES
+    module function pca_transform_${k1}$(x, components, x_mean) result(res)
```
jalvesz: Generally speaking, please preferer subroutines when returning arrays. Subroutines give more control and predictability on memory allocation/recycling. Functions (depending on the compiler options) might or might not create temp arrays or reallocate the left-hand-side of the assignment. You can always create a function on top of the subroutine to have a functional style, but for performance and safety subroutines are more commendable.

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,172 @@
...
+  #:for k1, t1 in REAL_KINDS_TYPES
+    module function pca_inverse_transform_${k1}$(x_reduced, components, x_mean) result(res)
+      ${t1}$, intent(in) :: x_reduced(:,:)
+      ${t1}$, intent(in) :: components(:,:)
+      ${t1}$, intent(in), optional :: x_mean(:)
+      ${t1}$ :: res(size(x_reduced,1), size(components,2))
+
+      integer(ilp) :: i, n
+      n = size(x_reduced, 1, kind=ilp)
+      res = matmul(x_reduced, components)
```
jalvesz: also here, please prefer `gemm`

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,172 @@
...
+      k = size(components, 1, kind=ilp)
+      
+      method_ = optval(method, "svd")
+      
+      ! 1. Calculate and optionally return mean
+      allocate(mu(p))
+      mu = mean(x, dim=1)
+      if (present(x_mean)) x_mean = mu
+      
+      err0 = linalg_state_type("pca", LINALG_SUCCESS)
```
perazz: It is not necessary to initialize the state handler - leads to unnecessary overhead
```suggestion
```

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,172 @@
...
+
+            allocate(lambda(p), idx(p), vectors(p, p))
+            c = cov(x, dim=1)
+            call eigh(c, lambda, vectors=vectors, err=err0)
+            
+            if (err0%ok()) then
+               ! Sort eigenvalues and vectors in descending order
+               do j = 1, p
+                  idx(j) = j
+               end do
```
perazz: Use array operations where straightforward

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,172 @@
...
+               ! Simple bubble sort
+               do i = 1, p-1
+                  do j = i+1, p
+                     if (lambda(idx(i)) < lambda(idx(j))) then
+                        m = idx(i)
+                        idx(i) = idx(j)
+                        idx(j) = m
+                     end if
+                  end do
+               end do
```
perazz: Use `stdlib` functions from the sorting module where already provided

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,172 @@
...
+                     singular_values(i) = 0.0_${k1}$
+                  end if
+               end do
+            end if
+         end block
+      else
+         err0 = linalg_state_type("pca", LINALG_ERROR, "Unknown method: "//method_)
+      end if
+      
+      if (present(err)) err = err0
```
perazz: please fix state handling: should `error stop` on missing argument + error 
```suggestion
      call err0%handle(err)
```

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,172 @@
...
+      
+      n = size(x, 1, kind=ilp)
+      p = size(x, 2, kind=ilp)
+      k = size(components, 1, kind=ilp)
+      
+      method_ = optval(method, "svd")
+      
+      ! 1. Calculate and optionally return mean
+      allocate(mu(p))
+      mu = mean(x, dim=1)
```
jvdp1:
```suggestion
      allocate(mu, source = mean(x, dim=1)
```

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,172 @@
...
+
+contains
+
+  #:for k1, t1 in REAL_KINDS_TYPES
+    module subroutine pca_${k1}$(x, components, singular_values, x_mean, &
+                                method, overwrite_x, err)
+      ${t1}$, intent(inout), target :: x(:,:)
+      ${t1}$, intent(out) :: components(:,:)
+      real(${k1}$), intent(out) :: singular_values(:)
+      ${t1}$, intent(out), optional :: x_mean(:)
```
jvdp1, line 113: Do we expect that the user provide an array of the right length? OR should it be allocatable?

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,172 @@
...
+      n = size(x, 1, kind=ilp)
+      p = size(x, 2, kind=ilp)
+      k = size(components, 1, kind=ilp)
+      
+      method_ = optval(method, "svd")
+      
+      ! 1. Calculate and optionally return mean
+      allocate(mu(p))
+      mu = mean(x, dim=1)
+      if (present(x_mean)) x_mean = mu
```
jvdp1: If `x_mean` is not allocatable, maybe:
```suggestion
      if (present(x_mean)) x_mean(1:size(mu)) = mu
```
A check on the size of `x_mean` migth be needed.

test/stats/test_pca.f90
```diff
@@ -0,0 +1,76 @@
...
+
+    real(sp), parameter :: sptol = 1000 * epsilon(1._sp)
+    real(dp), parameter :: dptol = 1000 * epsilon(1._dp)
+
+    call test_pca_sp()
+    call test_pca_dp()
+
+contains
+
+    subroutine test_pca_sp()
```
jvdp1: Could the `sp` and `dp` versions be generated with `fypp`?

jvdp1:
> New fails are happening because of the dependency on the sorting module. This exposes an issue with the modularization of the library. The sorting module being at the root of src is not visible by the stats module within its own independent folder... 

In this case, add `../stdlib_sorting.fypp` in the `CMakeLists.txt` should be enough (similarly to `stdlib_string_type.fypp` already present in `CMakeLists.txt`)

> we might need to reconsider next steps.

However, I agree with that. When I was working on #1081, I started to get "faked" circular dependencies.

> One idea would be to make this library (pca) not a submodule of stats but a module in itself at the root of src such that it can use easily any other module such as stats, blas/lapack and sorting. Or there might be another approach to think about

If the CMake file is correctly written, a submodule "pca" should not be a problem.
However, in terms of efficiency for the stats module (but also for other modules, like stdlib_linalg), it might be good to use modules instead of submodules: as `fpm` compiles only what is needed, if a user only needs `mean`, then the blas and lapack modules are currently not compiled (if I am correct). However, if the pca procedures are added as submodules of the stats module, then the blas and lapack will be compiled, even if only the procedure `mean` is used by the user. Similar "issues" can easily happen for `stdlib_linalg`.

jvdp1:
> I’ve tested the fix, and as suspected, it confirms that the current structure cannot handle these dependencies.

It can handle it AFAIK, but I agree that it is complex. I will open [this PR](https://github.com/JAi-SATHVIK/stdlib/pull/1) against your branch with a proposed solution.
 
> Adding the sorting source to the stats directory caused a `fatal error: the compiler cannot locate stdlib_bitsets.mod` (which sorting depends on), and we are seeing undefined references to linalg SVD/EIGH procedures.
> 
> To resolve this cleanly and avoid 'faking' the dependency tree, We can proceed with: Move the PCA implementation to the root as src/stdlib_pca.fypp. Update the root CMakeLists.txt to include it as a standalone module. So, This will allow PCA to correctly link against linalg, sorting, and stats using the standard library infrastructure.
> 
> @jalvesz @jvdp1 @perazz Any approach you would like to mention?

Let try to fix this as it is proposed now (that is, using a submodule for PCA), as proposed by @jalvesz

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,191 @@
...
+      n = size(x, 1, kind=ilp)
+      p = size(x, 2, kind=ilp)
+      k = size(components, 1, kind=ilp)
+      
+      method_ = optval(method, "svd")
+      
+      ! 1. Calculate mean using intrinsic sum (avoids submodule dependency issues)
+      allocate(mu(p))
+      do j = 1, p
+         mu(j) = sum(x(:, j)) / real(n, ${k1}$)
```
loiseaujc: I may be late to the party, but are there any reasons why you do not use the `mean` function from `stdlib_stats`?

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,191 @@
...
+         ! 2. Center data and call SVD with temporaries for robustness
+         block
+            ${t1}$, allocatable :: s_tmp(:), vt_tmp(:,:)
+            n_s = min(n, p)
+            allocate(s_tmp(n_s), vt_tmp(n_s, p))
+
+            if (optval(overwrite_x, .false.)) then
+               do i = 1, n
+                  x(i, :) = x(i, :) - mu
+               end do
```
loiseaujc: I guess you could replace this with

```fortran
do concurrent (i=1:n)
     x(i, :) = x(i, :) - mu
enddo
```

since all iterations are independent from one another.

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,191 @@
...
+               call svd(x, s_tmp, vt=vt_tmp, overwrite_a=.true., full_matrices=.false., err=err0)
+            else
+               block
+                  ${t1}$, allocatable :: x_centered(:,:)
+                  allocate(x_centered(n, p))
+                  do i = 1, n
+                     x_centered(i, :) = x(i, :) - mu
+                  end do
```
loiseaujc: Likewise, you can replace the standard `do` with a `do concurrent`.

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,191 @@
...
+      if (present(x_mean)) x_mean = mu
+
+      if (method_ == "svd") then
+         ! 2. Center data and call SVD with temporaries for robustness
+         block
+            ${t1}$, allocatable :: s_tmp(:), vt_tmp(:,:)
+            n_s = min(n, p)
+            allocate(s_tmp(n_s), vt_tmp(n_s, p))
+
+            if (optval(overwrite_x, .false.)) then
```
loiseaujc: For the sake of clarity, I would probably define an `overwrite_x_` variable defined as `overwrite_x_ = optval(overwrite_x, .false.)` and then use and `if (overwrite_x_)`. Doesn't change anything in the code logic, but I find it a bit more readable.

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,191 @@
...
+               x_centered(i, :) = x(i, :) - mu
+            end do
+            
+            ! Compute covariance matrix: C = X^T * X / (n-1)
+            scale_factor = 1.0_${k1}$ / real(max(n-1, 1), ${k1}$)
+            do i = 1, p
+               do j = 1, p
+                  c(i, j) = dot_product(x_centered(:, i), x_centered(:, j)) * scale_factor
+               end do
+            end do
```
loiseaujc: You could directly compute the covariance matrix as `C = X.T @ X / (n-1)` or `C = X @ X.T / (n-1)` (not sure what your convention is for the data layout) using a call to `gemmtr` (it only computes the triangular part of `C` since it is known to be symmetric). That would very likely improve the computational performances.

loiseaujc:
Thank you for your contribution. Here are my two cents regarding your code:

- the `pca` function is quite long (100-ish) lines with a somewhat complicated logic depending on which algorithm is being used (`eigh` or `svd`) and whether or not `mu` is passed. I would suggest to break it down into more manageable units avoiding code duplication. You could have some helper functions/subroutines like `center_data(x, mu)` (with `x` and `mu` being `intent(inout)`), and `pca_eigh`/`pca_svd` low-level drivers which you then call depending on `method`. That would make it easier to add additional algorithms (e.g. iterative svd for large scale problems).

- to avoid code duplication, try to re-use what already is present in `stdlib` (e.g. `mean`, `stdlib_dot_product`, etc) and make use of `blas` and `lapack` functionalities as much as possible for increased computational performances. It is fine though in general to have a first draft making use of simple `matmul` and simple handcrafted implementations to make sure things work, but eventually `blas`/`lapack` is the way to go for serious number crunching.

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,220 @@
...
+contains
+
+  ! Helper subroutine: Centers data in-place by subtracting the mean from each row
+  #:for k1, t1 in PCA_KINDS_TYPES
+    pure subroutine center_data_${k1}$(x, mu)
+      ${t1}$, intent(inout) :: x(:,:)
+      ${t1}$, intent(in) :: mu(:)
+      integer(ilp) :: i, n
+      n = size(x, 1, kind=ilp)
+      do i = 1, n
```
jalvesz: This will lead to the second index to be addressed for each `i`: no good for cache efficiency. In Fortran the left-most index is the one changing the fastest (opposed to C-based languages where the right most index is the one changing the fastest in terms of memory alignment).

A more robust and efficient loop would be, given that every iteration is independent:
```fortran
do concurrent( j=1:m, i=1:n )
    x(i, j) = x(i, j) - mu(j)
end do
```

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,220 @@
...
+      p = size(components, 2, kind=ilp)
+      
+      ! x_reconstructed = x_reduced * components using GEMM
+      ! GEMM: C = alpha * op(A) * op(B) + beta * C
+      ! x_reconstructed(n, p) = x_reduced(n, nc) * components(nc, p)
+      call gemm('N', 'N', n, p, nc, alpha, x_reduced, n, components, nc, beta, x_reconstructed, n)
+      
+      if (present(x_mean)) then
+         do i = 1, n
+            x_reconstructed(i, :) = x_reconstructed(i, :) + x_mean
```
jalvesz: Same as before:
```fortran
do concurrent( j=1:m, i=1:n )
    x_reconstructed(i, j) = x_reconstructed(i, j) + x_mean(j)
end do
```

JAi-SATHVIK: Loops now use `do concurrent( j=1:m, i=1:n )` with element wise indexing

src/stats/stdlib_stats.fypp
```diff
@@ -639,4 +641,62 @@ module stdlib_stats
     #:endfor
   end interface moment
 
+
+  #! Note: PCA is limited to single (sp) and double (dp) precision because external
+  #! optimized BLAS/LAPACK libraries (OpenBLAS, MKL) only support these precisions.
+  #! Extended (xdp) and quadruple (qp) precision are not supported for PCA.
```
jvdp1: Internal `stdlib` BLAS/LAPACK could be used for `xdp` and `qp`. Are there any reasons to not use them?

src/stats/CMakeLists.txt
```diff
@@ -29,4 +36,9 @@ set(stats_fppFiles
     stdlib_stats_var.fypp
 )
 
+set(f90Files
+    ../stdlib_sorting_radix_sort.f90
+    )
+
 configure_stdlib_target(stats "" stats_fppFiles stats_cppFiles)
+target_link_libraries(stats PUBLIC blas lapack $<$<NOT:$<BOOL:${STDLIB_NO_BITSET}>>:bitsets>)
```
jvdp1: Can you rebase your branch against the fortran-lang/stdlib/master, please? These macros have been changed recently.

```suggestion
target_link_libraries(stats PUBLIC blas lapack $<$<BOOL:${STDLIB_BITSET}>:bitsets>)
```

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,237 @@
...
+  ! Helper subroutine: Centers data in-place by subtracting the mean from each row
+  #:for k1, t1 in PCA_KINDS_TYPES
+    subroutine center_data_${k1}$(x, mu)
+      ${t1}$, intent(inout) :: x(:,:)
+      ${t1}$, intent(in) :: mu(:)
+      integer(ilp) :: i, j, n, m
+      n = size(x, 1, kind=ilp)
+      m = size(x, 2, kind=ilp)
+      do j = 1, m
+         do i = 1, n
```
jvdp1:
```suggestion
      m = size(x, 1, kind=ilp)
      n = size(x, 2, kind=ilp)
      do j = 1, n
         do i = 1, m
```

I guess it is a matter of taste

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,237 @@
...
+      allocate(vt_tmp(n_s, p))
+
+      call svd(x_centered, s_tmp, vt=vt_tmp, overwrite_a=.true., full_matrices=.false., err=err)
+
+      if (err%ok()) then
+         m = size(components, 1, kind=ilp)
+         if (n_s < m) m = n_s
+         do i = 1, m
+            components(i, :) = vt_tmp(i, :)
+         end do
```
jvdp1: The current notation might not be optimal for cache-efficiency. Maybe:

```suggestion
        components(1:m, :) = vt_tmp(1:m, :)
```

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,237 @@
...
+         
+         m = size(singular_values, 1, kind=ilp)
+         if (n_s < m) m = n_s
+         do i = 1, m
+            singular_values(i) = s_tmp(i)
+         end do
```
jvdp1:
```suggestion
        singular_values(1:m) = s_tmp(1:m)
```

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,237 @@
...
+         do i = j+1, p
+            c(i, j) = c(j, i)
+         end do
+      end do
+
+      call eigh(c, lambda, vectors=vectors, err=err)
+
+      if (err%ok()) then
+         ! Sort eigenvalues in descending order
+         lambda_copy = -lambda
```
jvdp1: Allocate when it is needed/closer to the use of the variable. Some arrays don't need to be allocated if `.not.err%ok()`

```suggestion


      ! Compute covariance matrix using BLAS syrk: C = (1/(n-1)) * X^T * X
      scale_factor = 1.0_${k1}$ / real(max(n-1, 1), ${k1}$)
      alpha = real(scale_factor, ${k1}$)
      beta = 0.0_${k1}$
      allocate(c(p, p), source=0.0_${k1}$)
      call syrk('U', 'T', p, n, alpha, x_centered, n, beta, c, p)
      
      ! Fill lower triangle from upper triangle (syrk only fills upper)
      do j = 1, p-1
         do i = j+1, p
            c(i, j) = c(j, i)
         end do
      end do

      allocate(lambda(p))
      allocate(vectors(p, p))
      call eigh(c, lambda, vectors=vectors, err=err)

      if (err%ok()) then
         ! Sort eigenvalues in descending order
         allocate(lambda_copy, source=-lambda)
         allocate(idx(p))

```

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,237 @@
...
+      real(${k1}$), intent(out) :: singular_values(:)
+      ${t1}$, intent(out), optional :: x_mean(:)
+      character(*), intent(in), optional :: method
+      logical, intent(in), optional :: overwrite_x
+      type(linalg_state_type), intent(out), optional :: err
+
+      type(linalg_state_type) :: err0
+      integer(ilp) :: n, p
+      ${t1}$, allocatable :: mu(:), x_centered(:,:)
+      character(16) :: method_
```
jvdp1: Could it be:
```suggestion
      character(len=:), allocatable :: method_
```

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,237 @@
...
+
+      n = size(x, 1, kind=ilp)
+      p = size(x, 2, kind=ilp)
+      method_ = adjustl(optval(method, "svd"))
```
jvdp1: If `method_` is `allocatable`:
```suggestion
      method_ = trim(adjustl(optval(method, "svd")))
```

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,237 @@
...
+
+      ! Calculate mean along dimension 1 (column means)
+      allocate(mu(p))
+      mu = sum(x, dim=1) / real(n, ${k1}$)
```
jvdp1: Matter of taste?
```suggestion
      allocate(mu, source = sum(x, dim=1) / real(n, ${k1}$)
```

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,237 @@
...
+      if (present(x_mean)) x_mean = mu
+
+      ! Method dispatch using if-else for maximum compiler stability
+      if (trim(method_) == "svd") then
+         if (optval(overwrite_x, .false.)) then
+            call center_data_${k1}$(x, mu)
+            call pca_svd_driver_${k1}$(x, n, p, components, singular_values, err0)
+         else
+            allocate(x_centered(n, p))
+            x_centered = x
```
jvdp1:
```suggestion
            allocate(x_centered, source = x)
```

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,237 @@
...
+      integer(ilp) :: n, p, nc
+      ${t1}$ :: alpha, beta
+      ${t1}$, allocatable :: x_centered(:,:)
+      
+      n = size(x, 1, kind=ilp)
+      p = size(x, 2, kind=ilp)
+      nc = size(components, 1, kind=ilp)
+      
+      allocate(x_centered(n, p))
+      x_centered = x
```
jvdp1:
```suggestion
      allocate(x_centered, source = x)
```

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,237 @@
...
+      if (present(x_mean)) call center_data_${k1}$(x_centered, x_mean)
+      
+      ! x_transformed = x_centered * components^T using GEMM
+      alpha = 1.0_${k1}$
+      beta = 0.0_${k1}$
+      call gemm('N', 'T', n, nc, p, alpha, x_centered, n, components, nc, beta, x_transformed, n)
```
jvdp1:
```suggestion
      call gemm('N', 'T', n, nc, p, one_${k1}$, x_centered, n, components, nc, zero_${k1}$, x_transformed, n)
```
with `one` and `zero` defined in `stdlib_blas_constants`

jvdp1 (commented):
Thank you @JAi-SATHVIK . Here are some additional suggestions.

src/stats/stdlib_stats.fypp
JAi-SATHVIK: You're right! 
I reviewed  `stdlib_linalg_svd.fypp `and  `stdlib_linalg_eigenvalues.fypp`
both use RC_KINDS_TYPES which includes xdp/qp when enabled. The internal reference LAPACK does support extended precision. I'll extend PCA to include these precisions and remove the limiting comment.

test/stats/test_pca.f90
JAi-SATHVIK: Yes, I'll convert t`est_pca.f90` to `test_pca.fypp`
 and use `#:for k1, t1 in PCA_KINDS_TYPES` to generate both `sp `and `dp `test subroutines. This will reduce duplication and make it easier to add `xdp`/`qp `tests

src/stats/stdlib_stats_pca.fypp
JAi-SATHVIK: Following the existing `stdlib_linalg ` convention (e.g., svd, eigh), output arrays like components, `singular_values`, and `x_mean ` are user-provided with fixed size. The user is expected to allocate them with the correct dimensions before calling `pca`. I'll add size validation with clear error messages if the arrays are too small. This matches how `svd `handles `s(:)` and `u(:,:)`.

src/stats/stdlib_stats_pca.fypp
JAi-SATHVIK: Following the `stdlib_linalg ` convention, I'll keep `x_mean ` as user-provided (not allocatable). I'll add:

A size check: if `size(x_mean) < p,` return an error.
Bounds-safe assignment: `x_mean(1:p)` = `mu`
This matches how `svd `validates `s(:) `dimensions

jvdp1:
> Hi, @jvdp1
> 
> I've addressed the previous review comments (code cleanup, precision support, etc.), but I'm consistently hitting an Internal Compiler Error (Segmentation Fault) with the Intel `ifx ` compiler in the CI. The error points to the generated `stdlib_stats_pca` submodule. I suspect this is triggered by the `allocate(var, source=expr) `constructs. Is there a preferred workaround for these types of compiler bugs in stdlib?
> 
> I propose refactoring the remaining `allocate(..., source=...)` calls to use explicit allocation and assignment (e.g., `allocate(a(n)); a = expr`). Please let me know if you'd prefer I proceed with this workaround or if there are other potential causes I should investigate.

I am surprised about this, because `allocate(..., source=...)` is widely used in `stdlib` in other modules. Currently, I have no idea what could be other causes. I will investigate that issue a bit.

jalvesz:
Here is my take with respect to `allocate(var, source=...)` it works properly when the source term is either a scalar value or a known array. Expressions involving reductions and transformation of array data within this source assignment should not be "one-lined" like that, it is difficult to debug and prone to creating temporary arrays.

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,247 @@
...
+            call pca_svd_driver_${k1}$(x_centered, n, p, components, singular_values, err0)
+         end if
+
+      case ("eig", "cov")
+         allocate(x_centered, source=x)
+         call center_data_${k1}$(x_centered, mu)
+         call pca_eigh_driver_${k1}$(x_centered, n, p, components, singular_values, err0)
+
+      case default
+         err0 = linalg_state_type("pca", LINALG_ERROR, "Unknown method: "//method_)
```
jvdp1: @JAi-SATHVIK : Commenting this line solves the issue with `ifx` on my computer. I will investigate the reason.

\cc @jalvesz

jvdp1:
```suggestion
         err0 = linalg_state_type("pca", LINALG_ERROR, "Unknown method: ", method_)
```

JAi-SATHVIK: Thanks @jvdp1 for identifying the root cause!

I've applied your suggested fix replacing the string concatenation `(//method_)` with comma-separated arguments in the `linalg_state_type`

src/stats/stdlib_stats.fypp
```diff
@@ -639,4 +641,67 @@ module stdlib_stats
...
+
+  #! PCA supports all real kinds available in stdlib's internal BLAS/LAPACK.
+  #! When WITH_XDP or WITH_QP are enabled, extended precision is also supported.
+  #:set PCA_KINDS_TYPES = [("sp", "real(sp)"), ("dp", "real(dp)")]
+  #:if WITH_XDP
+  #:set PCA_KINDS_TYPES = PCA_KINDS_TYPES + [("xdp", "real(xdp)")]
+  #:endif
+  #:if WITH_QP
+  #:set PCA_KINDS_TYPES = PCA_KINDS_TYPES + [("qp", "real(qp)")]
+  #:endif
```
jvdp1:
```suggestion
```
Is this needed? `REAL_KINDS_TYPES` could be used instead of `PCA_KINDS_TYPES`

src/stats/stdlib_stats.fypp
```diff
@@ -639,4 +641,67 @@ module stdlib_stats
...
+    !! version: experimental
+    !!
+    !! Principal Component Analysis (PCA)
+    !! ([Specification](../page/specs/stdlib_stats.html#pca))
+    #:for k1, t1 in PCA_KINDS_TYPES
+      module subroutine pca_${k1}$(x, components, singular_values, x_mean, &
+                                  method, overwrite_x, err)
+        ${t1}$, intent(inout) :: x(:,:)
+        ${t1}$, intent(out) :: components(:,:)
+        real(${k1}$), intent(out) :: singular_values(:)
```
jvdp1:
```suggestion
        ${t1}$, intent(out) :: singular_values(:)
```
Is this correct? Otherwise, why should it be always `real()`?

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,247 @@
+#:include "common.fypp"
+#! PCA supports all real kinds available in stdlib's internal BLAS/LAPACK.
+#! When WITH_XDP or WITH_QP are enabled, extended precision is also supported.
+#:set PCA_KINDS_TYPES = [("sp", "real(sp)"), ("dp", "real(dp)")]
+#:if WITH_XDP
+#:set PCA_KINDS_TYPES = PCA_KINDS_TYPES + [("xdp", "real(xdp)")]
+#:endif
+#:if WITH_QP
+#:set PCA_KINDS_TYPES = PCA_KINDS_TYPES + [("qp", "real(qp)")]
+#:endif
```
jvdp1: Could you use `REAL_KINDS_TYPES` instead of `PCA_KINDS_TYPE`?

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,247 @@
...
+  #:endfor
+
+  ! SVD-based PCA driver: computes principal components via SVD of centered data
+  #:for k1, t1 in PCA_KINDS_TYPES
+    subroutine pca_svd_driver_${k1}$(x_centered, n, p, components, singular_values, err)
+      use stdlib_blas_constants_${k1}$, only: zero
+      ${t1}$, intent(inout) :: x_centered(:,:)
+      integer(ilp), intent(in) :: n, p
+      ${t1}$, intent(out) :: components(:,:)
+      real(${k1}$), intent(out) :: singular_values(:)
```
jvdp1:
```suggestion
      ${t1}$, intent(out) :: singular_values(:)
```

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,247 @@
...
+      real(${k1}$), allocatable :: lambda(:), lambda_copy(:)
+      ${t1}$, allocatable :: c(:,:), vectors(:,:)
+
+      ! Initialize outputs to zero to prevent uninitialized memory access
+      components = zero
+      singular_values = zero
+
+      ! Compute covariance matrix using BLAS syrk: C = (1/(n-1)) * X^T * X
+      scale_factor = 1.0_${k1}$ / real(max(n-1, 1), ${k1}$)
+      alpha = scale_factor
```
jvdp1:
```suggestion
      alpha = 1.0_${k1}$ / real(max(n-1, 1), ${k1}$)
```
is the variable `scale_factor` needed?

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,247 @@
...
+      beta = zero
+      allocate(c(p, p))
+      c = zero
+      call syrk('U', 'T', p, n, alpha, x_centered, n, beta, c, p)
```
jvdp1, line 83:
```suggestion
      allocate(c(p, p))
      c = zero
      call syrk('U', 'T', p, n, alpha, x_centered, n, zero, c, p)
```

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,247 @@
...
+      allocate(lambda(p))
+      allocate(vectors(p, p))
+      call eigh(c, lambda, vectors=vectors, err=err)
+
+      if (err%ok()) then
+         ! Sort eigenvalues in descending order
+         allocate(lambda_copy(p))
+         lambda_copy = -lambda
+         allocate(idx(p))
+         call sort_index(lambda_copy, idx)
```
jvdp1: why did you not use `reverse` in `sort_index`?

```suggestion
         allocate(idx(p))
         call sort_index(lambda, idx, reverse=.true.)
```

If so, in the following code, `lambda(idx(i))` should be replaced by `lambda(i)`

src/stats/stdlib_stats_pca.fypp
jalvesz: in this case `allocate(c(p,p), source=zero)` is recommended since is just a scalar value. (what should be avoided is when the assignment uses intermediate array transformations)

src/stats/stdlib_stats_pca.fypp
jalvesz: In principle it should be possible to implement PCA also for complex numbers. But there a few attention points, so it requires a bit careful points beyond just allowing complex data: https://chatgpt.com/share/696e9b1b-9e9c-800f-8898-e44d2a9f976f. If it can be covered that would be indeed very interesting.

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,233 @@
...
+      beta = zero
+      allocate(c(p, p), source=zero)
+      call syrk('U', 'T', p, n, alpha, x_centered, n, beta, c, p)
+
+      ! Fill lower triangle from upper triangle (syrk only fills upper)
+      do j = 1, p-1
+         do i = j+1, p
+            c(i, j) = c(j, i)
+         end do
+      end do
```
loiseaujc: I think you don't actually need to fill-in the lower triangular part of the covariance matrix. The `eigh` subroutine is dedicated for symmetric matrices and accepts the `upper_a` boolean flag (default to `.true.` I guess) so that the underlying `lapack` driver `heev` works only with the appropriate upper or lower triangular part.

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,233 @@
...
+
+      allocate(lambda(p))
+      allocate(vectors(p, p))
+      call eigh(c, lambda, vectors=vectors, err=err)
+
+      if (err%ok()) then
+         ! Sort eigenvalues in descending order
+         allocate(idx(p))
+         call sort_index(lambda, idx, reverse=.true.)
```
loiseaujc: You don't need to sort the eigenvalues out of `eigh`. The `lapack` convention for `heev` is to return the eigenvalues in ascending order. You thus only need to reverse the eigenvalues array and do the left-right flip of the eigenvector array.

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,233 @@
...
+      
+      ! x_reconstructed = x_reduced * components using GEMM
+      call gemm('N', 'N', n, p, nc, one, x_reduced, n, components, nc, zero, x_reconstructed, n)
+      
+      if (present(x_mean)) then
+         do j = 1, p
+            do i = 1, n
+               x_reconstructed(i, j) = x_reconstructed(i, j) + x_mean(j)
+            end do
+         end do
```
loiseaujc, line 286: Any reasons why you use nested `do` loops instead of

```fortran
do concurrent(i=1:n, j=1:p)
    x_reconstructed(i, j) = x_reconstructed(i, j) + x_mean(j)
end do
```

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,233 @@
...
+      ${t1}$, intent(inout) :: x(:,:)
+      ${t1}$, intent(in) :: mu(:)
+      integer(ilp) :: i, j, m, n
+      m = size(x, 1, kind=ilp)
+      n = size(x, 2, kind=ilp)
+      do j = 1, n
+         do i = 1, m
+            x(i, j) = x(i, j) - mu(j)
+         end do
+      end do
```
loiseaujc, line 25: Any reasons why you use nested `do` loops instead of

```fortran
do concurrent(i=1:m, j=1:n)
    x(i, j) = x(i, j) - mu(j)
end do
```

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,233 @@
...
+         ! Assign sorted results with safety bounds checks
+         m = min(size(components, 1, kind=ilp), p)
+         m = min(m, size(singular_values, 1, kind=ilp))
+         do i = 1, m
+            components(i, :) = vectors(:, idx(i))
+            if (lambda(i) > 0.0_${k1}$) then
+               singular_values(i) = sqrt(lambda(i) * real(n-1, ${k1}$))
+            else
+               singular_values(i) = 0.0_${k1}$
+            end if
```
loiseaujc: You replaced the `if`/`else` condition with

```fortran
singular_values(i) = merge(sqrt(lambda(i)*(n-1)), 0.0_${k1}$, lambda(i) > 0.0_${k1}$)
```

or since you simply need to flip the eigenvector arrays (no need for `idx`), I'm pretty sure you can get rid of the `do` loop and replace it with

```fortran
where (lambda > 0.0_${rk}$) singular_values = sqrt(lambda*(n-1))
```

or something along these lines.

src/stats/stdlib_stats_pca.fypp
JAi-SATHVIK: I have actually updated the implementation to use `mean(x, 1)` from `stdlib_stats`.

src/stats/stdlib_stats_pca.fypp
JAi-SATHVIK: I have updated all centering and reconstruction loops to use the 2-dimensional `do concurrent(i=1:m, j=1:n)` syntax.

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,224 @@
...
+      components = zero
+      singular_values = zero
+
+      ! Compute covariance matrix using BLAS syrk: C = (1/(n-1)) * X^T * X
+      alpha = 1.0_${k1}$ / real(max(n-1, 1), ${k1}$)
+      beta = zero
+      allocate(c(p, p), source=zero)
+      call syrk('U', 'T', p, n, alpha, x_centered, n, beta, c, p)
+
+      ! Fill lower triangle from upper triangle (syrk only fills upper)
```
loiseaujc: This comment no longer applies.

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,224 @@
...
+      allocate(lambda(p))
+      allocate(vectors(p, p))
```
loiseaujc: You can coalesce these two lines into

```fortran
allocate(lambda(p), vectors(p, p))
```

This is essentially a matter of personal preferences. No big deal.

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,224 @@
...
+         m = min(m, size(singular_values, 1, kind=ilp))
+         
+         ! Components are eigenvectors as rows (transpose of vectors columns)
+         components(1:m, :) = transpose(vectors(:, 1:m))
+         
+         ! Convert eigenvalues to singular values: s = sqrt(lambda * (n-1))
+         where (lambda(1:m) > 0.0_${k1}$)
+            singular_values(1:m) = sqrt(lambda(1:m) * real(n-1, ${k1}$))
+         elsewhere
+            singular_values(1:m) = 0.0_${k1}$
```
loiseaujc: Since `singular_values` is initialized to `zero` on line 76, I don't think you need the `elsewhere` statement. I might be wrong though.

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,224 @@
...
+      ${t1}$, allocatable :: mu(:), x_centered(:,:)
+      character(len=:), allocatable :: method_
+
+      n = size(x, 1, kind=ilp)
+      p = size(x, 2, kind=ilp)
+      method_ = trim(adjustl(optval(method, "svd")))
+
+      ! Calculate mean along dimension 1 (column means) using stdlib mean
+      allocate(mu(p))
+      mu = mean(x, 1)
```
loiseaujc, line 166: I think you can get rid of the `allocate` statement. `mu = mean(x, 1)` will directly allocate `mu` and fill it with `mean(x, 1)`.

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,224 @@
...
+      end if
+
+      ! Method dispatch
+      select case (method_)
+      case ("svd")
+         if (optval(overwrite_x, .false.)) then
+            call center_data_${k1}$(x, mu)
+            call pca_svd_driver_${k1}$(x, n, p, components, singular_values, err0)
+         else
+            allocate(x_centered, source=x)
```
loiseaujc: You can keep it simpler with just `x_centered = x`.

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,224 @@
...
+            call center_data_${k1}$(x_centered, mu)
+            call pca_svd_driver_${k1}$(x_centered, n, p, components, singular_values, err0)
+         end if
+
+      case ("eig", "cov")
+         allocate(x_centered, source=x)
```
loiseaujc: Same thing, `x_centered = x` is just as readable.

loiseaujc:
I don't quite like `sqrt(epsilon(1.0_${rk}$))` either (although it is a simple solution). I dislike even more artificially increasing the constant. A better way would be to look at error estimates for matrix-matrix multiplication in floating points. I would have to double check some of my old lecture notes but it would basically involve the product of the Frobenius norms of the different factors. A simple remedy would thus to normalize all the matrices to have a Frobenius norm of one. I'll create a local branch on my fork and test this to see if I get rid of the failure.

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,233 @@
...
+         m = min(size(components, 1, kind=ilp), p)
+         m = min(m, size(singular_values, 1, kind=ilp))
+         
+         ! Components are eigenvectors as rows (transpose of vectors columns)
+         components(1:m, :) = transpose(vectors(:, 1:m))
+         
+         ! Convert eigenvalues to singular values: s = sqrt(lambda * (n-1))
+         where (lambda(1:m) > 0.0_${k1}$)
+            singular_values(1:m) = sqrt(lambda(1:m) * real(n-1, ${k1}$))
+         end where
```
loiseaujc: I think you can write

```fortran
where (lambda(1:m) > 0.0_${k1}$) singular_values(1:m) = sqrt(lambda(1:m) * real(n-1, ${k1}$))
```

on a single line. It essentially is a matter of personal preferences though, no big deal.

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,233 @@
...
+
+      if (err%ok()) then
+         ! LAPACK returns eigenvalues in ascending order.
+         ! Flip them to get descending order for PCA.
+         lambda = lambda(p:1:-1)
+         vectors = vectors(:, p:1:-1)
+
+         ! Assign results with safety bounds checks
+         m = min(size(components, 1, kind=ilp), p)
+         m = min(m, size(singular_values, 1, kind=ilp))
```
loiseaujc: `min` can take multiple arguments. If I understand the logic correctly, you could merge these two lines into

```fortran
m = min(size(components, 1, kind=ilp), size(singular_values, 1, kind=ilp), p)
```

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,233 @@
...
+         return
+      end if
+
+      ! Convert method to lowercase for case-insensitive comparison
+      method_ = trim(adjustl(optval(method, "svd")))
+      do i = 1, len(method_)
+         if (method_(i:i) >= 'A' .and. method_(i:i) <= 'Z') then
+            method_(i:i) = achar(iachar(method_(i:i)) + 32)
+         end if
+      end do
```
loiseaujc: You could use the function `to_lower` from the `stdlib_ascii` module I guess here.

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,285 @@
...
+  #:endfor
+
+
+  #:for k1, t1, ri, cpp in REAL_KINDS_TYPES
+    module subroutine pca_transform_${k1}$(x, components, x_mean, x_transformed)
+      use stdlib_blas_constants_${k1}$, only: one, zero
+      ${t1}$, intent(in) :: x(:,:)
+      ${t1}$, intent(in) :: components(:,:)
+      ${t1}$, intent(in), optional :: x_mean(:)
+      ${t1}$, intent(out) :: x_transformed(:,:)
```
jalvesz, line 213: optional arguments shall **always** be the right-most arguments.
Please change here and adapt the documentation accordingly. No need for additional extra comments on positioning of arguments.
https://github.com/fortran-lang/stdlib/pull/1086/changes#r2834349838 This proposal by Github copilot was actually very misleading, it should have suggested to fix the api, not to hot-patch the documentation.

Please also check any other occurance of misplaced optional arguments.

doc/specs/stdlib_stats.md
```diff
@@ -283,6 +283,120 @@ If `mask` is specified, the result is the _k_-th  (central) moment of all elemen
...
+The input matrix `x` has shape `(m, n)`, where `m` is the number of observations and `n` is the number of features.
+The subroutine computes the principal components (loadings), the singular values, and optionally the feature means.
+
+Two methods are supported:
+- `"svd"`: (Default) Computes PCA via Singular Value Decomposition of the centered data. This is generally more numerically stable.
+- `"eig"` or `"cov"`: Computes PCA via Eigendecomposition of the covariance matrix.
+
+### Syntax
+
+`call ` [[stdlib_stats(module):pca(interface)]] `(x, components, singular_values [, x_mean [, method [, overwrite_x [, err]]]])`
```
jalvesz:
```suggestion
`call ` [[stdlib_stats(module):pca(interface)]] `(x, components, singular_values [, x_mean, method, overwrite_x, err])`
```

src/stats/stdlib_stats_pca.fypp
```diff
@@ -0,0 +1,226 @@
...
+      type(linalg_state_type), intent(out), optional :: err
+
+      type(linalg_state_type) :: err0
+      integer(ilp) :: n, p, i
+      ${t1}$, allocatable :: mu(:), x_centered(:,:)
+      character(len=:), allocatable :: method_
+
+      n = size(x, 1, kind=ilp)
+      p = size(x, 2, kind=ilp)
+
```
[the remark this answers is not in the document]
jvdp1: @JAi-SATHVIK could you add the suggested checks, please? Or close this comment if already done?

JAi-SATHVIK:
@jvdp1 I have added the requested shape and dimension checks in `pca`, `pca_transform`, and `pca_inverse_transform`.

PR #1087, opened 2026-01-08
Add validation for op argument in spmv routine for special matrices

Mahmood-Sinan:
This pr adds validation for the optional argument `op` in `stdlib_specialmatrices`.
When op is present and not one of `N`, `T`, or `C`, a `LINALG_VALUE_ERROR` is raised using` linalg error handling`.

### DIFF
--- a/src/specialmatrices/stdlib_specialmatrices_tridiagonal.fypp
+++ b/src/specialmatrices/stdlib_specialmatrices_tridiagonal.fypp
@@ -174,8 +174,8 @@ submodule (stdlib_specialmatrices) tridiagonal_matrices
         #:endif
 
         if(present(op)) then
-            if(.not.(op == "N" .or. op == "T" .or. op == "C")) then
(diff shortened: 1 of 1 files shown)


loiseaujc:
If I recall correctly, the sparse matrix-vector product accepts `H` as input as well for the complex-conjugate matrix-vector product. Maybe you could extend the validation to include it and internally handles the fact that `H` and `C` are equivalent.

Mahmood-Sinan:
> If I recall correctly, the sparse matrix-vector product accepts `H` as input as well for the complex-conjugate matrix-vector product. Maybe you could extend the validation to include it and internally handles the fact that `H` and `C` are equivalent.

@loiseaujc I have made the changes. The routine now accepts `H` and maps it internally to `C` before calling `lagtm`, since only `C` is supported there.

loiseaujc:
LGTM.

PR #1088, opened 2026-01-08
Simplification of CPP macros for modularizing stdlib

jvdp1:
Simplification of CPP macros for modularizing stdlib based on the discussion in #1083


perazz:
yes: fpm <=v0.12.0 will not read any macros, so everything works provided that all modules are included in the fpm version.

PR #1094, opened 2026-01-21
CI: addition of fpm-build for testing the latest version of fpm

jvdp1:
Following the recent issues with `fpm` and `stdlib-fpm`, here is an additional CI process to test the latest version of `fpm` with `stdlib-fpm`.

### DIFF
--- a/.github/workflows/fpm-build.yml
+++ b/.github/workflows/fpm-build.yml
@@ -32,7 +32,7 @@ jobs:
           version: ${{ matrix.toolchain.version }}
 
       - name: Setup Fortran Package Manager
-        uses: fortran-lang/setup-fpm@v8
+        uses: fortran-lang/setup-fpm@v9
         with:
           fpm-version: ${{ matrix.fpm_versions }}
 
--- a/.github/workflows/fpm-deployment.yml
+++ b/.github/workflows/fpm-deployment.yml
@@ -32,7 +32,7 @@ jobs:
           version: ${{ matrix.toolchain.version }}
 
       - name: Setup Fortran Package Manager
-        uses: fortran-lang/setup-fpm@v8
+        uses: fortran-lang/setup-fpm@v9
         with:
           fpm-version: v0.12.0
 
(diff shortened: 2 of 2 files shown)


.github/workflows/fpm-build.yml
```diff
@@ -0,0 +1,41 @@
...
+jobs:
+  fpm-build:
+    name: Run tests the fpm branches
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        toolchain: [{compiler: gcc, version: '13' }]
+        fpm_versions: [v0.12.0, current]
```
jvdp1, line 14: @perazz 

This does not work because there is no `fpm-current-linux-x86_64` in `https://github.com/fortran-lang/fpm/releases/download/current/`

Instead it should be `fpm-linux-x86_64`.

Is there another way to load the current version of `fpm`?

perazz: Right: I believe this is currently not possible with `setup-fpm`, but definitely worthwhile of adding. What name would be best? `latest` is supported and represents the latest release (v0.12.0). I like `current`. @copilot what is a good name for a "trunk" version that uses the working main branch? 

Another option is to release fpm v0.13.0 soon - I've been working on it and should be ready within the next week or two

jvdp1: > Right: I believe this is currently not possible with `setup-fpm`, but definitely worthwhile of adding. What name would be best? `latest` is supported and represents the latest release (v0.12.0). I like `current`.

I also like how `current` and `latest` are defined.

> Another option is to release fpm v0.13.0 soon - I've been working on it and should be ready within the next week or two

Indeed, but I am afraid that we will have a similar issue when a few additional commits will be merged on top of `v0.13.0`. 


Alternativetly, we could modify the code of `setup-fpm` [here](https://github.com/fortran-lang/setup-fpm/blob/fdf027c7f6b1c39f67ce98d1833f6feda7b682df/index.js#L153), with an exception for `current`. what do you think about this idea?

perazz: `setup-fpm@v9` is now released with your support for the `current` version

jvdp1: Updated in https://github.com/fortran-lang/stdlib/pull/1094/commits/d10059c78af0b98fe1ead9daa809f4a9aa42e837

PR #1096, opened 2026-01-23
feat(linalg): Add weighted and generalized least squares solvers

### LINKED ISSUE #1047: More least-squares variants
### Motivation

#### Weighted least-squares

```fortran
```

$$
\begin{aligned}
\end{aligned}
$$

```fortran
```

### Prior Art

### Additional Information

_No response_

aamrindersingh:
Resolves #1047

This PR adds two interfaces to `stdlib_linalg` for weighted least-squares problems. The `weighted_lstsq` interface handles diagonal weight matrices where each observation has a different importance, common in heteroscedastic regression and outlier downweighting. Usage: `x = weighted_lstsq(w, A, b)` where `w` is a vector of positive weights. It transforms to ordinary least-squares via row scaling and uses LAPACK's `GELSD`. The `generalized_lstsq` interface handles correlated errors described by a symmetric positive definite covariance matrix: `x = generalized_lstsq(W, A, b)`. This minimizes the Mahalanobis distance `(Ax-b)^T W^{-1} (Ax-b)` using LAPACK's `GGGLM`.

The key design decisions are:-

(1) `weighted_lstsq` validates that all weights are positive since negative weights have no statistical meaning and would cause numerical issues with `sqrt(w)`.
(2) `generalized_lstsq` always copies `W` internally to protect user data from `GGGLM`'s destructive operations. This applies regardless of the `prefactored_w` flag , the input `W` matrix is never modified.
(3) Both interfaces follow the `overwrite_a` pattern from `solve_lu` where `copy_a` defaults to `.true.` to preserve `A` unless the user explicitly opts into destruction for performance.
(4) A `handle_ggglm_info` error handler was added to parse LAPACK return codes, consistent with existing `handle_gelsd_info` and `handle_gglse_info` patterns.

Testing includes:-
- Basic WLS solves with uniform weights
- Verification that non-uniform weights change solutions
- Error handling for negative weights
- Basic GLS solves with correlated errors
- Verification that GLS with identity covariance equals OLS
- All real types (`sp`/`dp`/`qp`/`xdp`) covered following existing `lstsq` patterns

### DIFF
--- a/src/linalg/stdlib_linalg_least_squares.fypp
+++ b/src/linalg/stdlib_linalg_least_squares.fypp
@@ -590,9 +590,11 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
 
         ! Local variables
         type(linalg_state_type) :: err0
-        integer(ilp) :: m, n, i
+        integer(ilp) :: m, n, j
         logical(lk) :: copy_a
-        ${rt}$, allocatable :: a_scaled(:,:), b_scaled(:)
+        ${rt}$, pointer :: amat(:,:)
+        ${rt}$, allocatable, target :: amat_alloc(:,:)
+        ${rt}$, allocatable :: b_scaled(:)
         real(${rk}$), allocatable :: sqrt_w(:)
         character(*), parameter :: this = 'weighted_lstsq'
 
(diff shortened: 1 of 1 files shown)


doc/specs/stdlib_linalg.md
```diff
@@ -976,6 +976,96 @@ call [stdlib_linalg(module):constrained_lstsq_space(interface)]`(a,  c,  lwork [
 `c`: Shall be a rank-2 `real` or `complex` array of the same kind as `a` defining the linear equality constraints. It is an `intent(in)` argument.
 `lwork`: Shall be an `integer` scalar returning the optimal size required for the workspace array to solve the constrained least-squares problem.
 
+## `weighted_lstsq` - Computes the weighted least squares solution to a linear matrix equation {#weighted-lstsq}
```
jvdp1, line 979: Thank you @aamrindersingh . Could you split this PR in two smaller one (`weighted_lstsq` and `generalized_lstsq`), please? Or are these two implementations dependent of each other?

aamrindersingh: No, they are not dependent on each other , Will start on splitting them into two PRs.
Thanks

aamrindersingh: @jvdp1, Below are the Two PR's

- #1104 :- `weighted_lstsq`
- #1105  :-  `generalized_lstsq`

PR #278, opened 2020-12-22
Probability Distribution and Statistical Functions -- Gamma Distribution Module

Jim-215-Fisher:
This is the second round of probability distribution and statistical functions, continuing from #240. Since the whole modular structure has been changed, each distribution will have its own PR.
This PR contains gamma distributions. Files uploaded are:

- src/stdlib_stats_distribtuion_gamma.fypp
- src/CMakeLists.txt
- src/tests/test_distribution_gamma.f90
- src/tests/CMakeLists.txt
- doc/specs/stdlib_stats_distribution_gamma.md
- doc/specs/index.md

[Link](https://dl.acm.org/doi/pdf/10.1145/358407.358414) to the open-access article on which this PR is based. 
P.369 (7/10) has a summary of the final algorithm.

PR #1099, opened 2026-01-24
Probability Distribution -- Gamma Distribution 

Calluumm:
Adds a Gamma Distribution to Stats
This is a modernisation/continuance of #278 all credit to the original author.

Comparatively to #278;
- Tests have been standardised to current stdlib standards
- Documentation lightly amended 
- Filepaths adjusted for current structure

ctest distribution_gamma passed fine
the function itself was previously tested against SciPy so I did again it's still accurate, I further tested against [FSML](https://fsml.mutz.science/page/api/dst.html#gamma-distribution)'s function and again it was identical.
specialfunctions did have to be linked into stats for the lower incomplete gamma function, if this isn't standard it could be duplicated into stats but I'm not sure what the practice is there for this lib

### DIFF
--- a/src/stats/stdlib_stats_distribution_gamma.fypp
+++ b/src/stats/stdlib_stats_distribution_gamma.fypp
@@ -6,6 +6,7 @@ Module stdlib_stats_distribution_gamma
     use ieee_arithmetic, only: ieee_value, ieee_quiet_nan, ieee_is_nan
     use stdlib_kinds, only : sp, dp, xdp
     use stdlib_error, only : error_stop
+    use stdlib_optval, only : optval
     use stdlib_stats_distribution_uniform, only : uni=>rvs_uniform
     use stdlib_stats_distribution_normal, only : rnor=>rvs_normal
     use stdlib_specialfunctions_gamma, only : regamma_p=>regularized_gamma_p
@@ -157,11 +158,7 @@ contains
         ${t1}$ :: res
         ${t1}$ :: loc_
 
-        if(present(loc)) then
-            loc_ = loc
-        else
-            loc_ = 0.0_${k1}$
-        end if
+        loc_ = optval(loc, 0.0_${k1}$)
 
         if(rate <= 0.0_${k1}$) call error_stop("Error(gamma_dist_rvs): Gamma" &
         //" distribution rate parameter must be greater than zero")
@@ -183,11 +180,7 @@ contains
         ${t1}$ :: res
         ${t1}$ :: loc_
 
-        if(present(loc)) then
-            loc_ = loc
-        else
-            loc_ = cmplx(0.0_${k1}$, 0.0_${k1}$, kind=${k1}$)
-        end if
+        loc_ = optval(loc, cmplx(0.0_${k1}$, 0.0_${k1}$, kind=${k1}$))
 
         res = cmplx(gamma_dist_rvs_r${k1}$(shape%re, rate%re) + loc_%re,                 &
                     gamma_dist_rvs_r${k1}$(shape%im, rate%im) + loc_%im, kind=${k1}$)
@@ -207,11 +200,7 @@ contains
         integer :: i
         ${t1}$ :: loc_
 
-        if(present(loc)) then
-            loc_ = loc
-        else
-            loc_ = 0.0_${k1}$
-        end if
+        loc_ = optval(loc, 0.0_${k1}$)
 
         do i = 1, array_size
 
@@ -236,11 +225,7 @@ contains
         integer :: i
         ${t1}$ :: loc_
 
-        if(present(loc)) then
-            loc_ = loc
-        else
-            loc_ = cmplx(0.0_${k1}$, 0.0_${k1}$, kind=${k1}$)
-        end if
+        loc_ = optval(loc, cmplx(0.0_${k1}$, 0.0_${k1}$, kind=${k1}$))
 
         do i = 1, array_size
 
@@ -263,11 +248,7 @@ contains
         ${t1}$ :: xs
         ${t1}$ :: loc_
 
-        if(present(loc)) then
-            loc_ = loc
-        else
-            loc_ = 0.0_${k1}$
-        end if
+        loc_ = optval(loc, 0.0_${k1}$)
 
         xs = x - loc_
 
@@ -336,11 +317,7 @@ contains
         real(${k1}$) :: res
         ${t1}$ :: loc_
 
-        if(present(loc)) then
-            loc_ = loc
-        else
-            loc_ = cmplx(0.0_${k1}$, 0.0_${k1}$, kind=${k1}$)
-        end if
+        loc_ = optval(loc, cmplx(0.0_${k1}$, 0.0_${k1}$, kind=${k1}$))
 
         res = gamma_dist_pdf_r${k1}$(x%re, shape%re, rate%re, loc_%re)
         res = res * gamma_dist_pdf_r${k1}$(x%im, shape%im, rate%im, loc_%im)
@@ -361,11 +338,7 @@ contains
         real(${k1}$) :: res
         ${t1}$ :: loc_
 
-        if(present(loc)) then
-            loc_ = loc
-        else
-            loc_ = cmplx(0.0_${k1}$, 0.0_${k1}$, kind=${k1}$)
-        end if
+        loc_ = optval(loc, cmplx(0.0_${k1}$, 0.0_${k1}$, kind=${k1}$))
 
         res = gamma_dist_pdf_r${k1}$(x%re, shape%re, rate%re, loc_%re)
         res = res * gamma_dist_pdf_r${k1}$(x%im, shape%im, rate%im, loc_%im)
@@ -386,11 +359,7 @@ contains
         real(${k1}$) :: res, xs
         ${t1}$ :: loc_
 
-        if(present(loc)) then
-            loc_ = loc
-        else
-            loc_ = 0.0_${k1}$
-        end if
+        loc_ = optval(loc, 0.0_${k1}$)
 
         if(rate <= 0.0_${k1}$ .or. shape <= 0.0_${k1}$) then
             res = ieee_value(1.0_${k1}$, ieee_quiet_nan)
@@ -438,11 +407,7 @@ contains
         real(${k1}$) :: res
         ${t1}$ :: loc_
 
-        if(present(loc)) then
-            loc_ = loc
-        else
-            loc_ = cmplx(0.0_${k1}$, 0.0_${k1}$, kind=${k1}$)
-        end if
+        loc_ = optval(loc, cmplx(0.0_${k1}$, 0.0_${k1}$, kind=${k1}$))
 
         res = gamma_dist_cdf_r${k1}$(x%re, shape%re, rate%re, loc_%re)
 
@@ -464,11 +429,7 @@ contains
         real(${k1}$) :: res
         ${t1}$ :: loc_
 
-        if(present(loc)) then
-            loc_ = loc
-        else
-            loc_ = cmplx(0.0_${k1}$, 0.0_${k1}$, kind=${k1}$)
-        end if
+        loc_ = optval(loc, cmplx(0.0_${k1}$, 0.0_${k1}$, kind=${k1}$))
 
         res = gamma_dist_cdf_${t1[0]}$${k1}$(x, shape, rate, loc_)
         err = 0


src/stats/stdlib_stats_distribution_gamma.fypp
```diff
@@ -0,0 +1,317 @@
+#:set WITH_QP = False
+#:set WITH_XDP = False
+#:include "common.fypp"
+#:set RC_KINDS_TYPES = REAL_KINDS_TYPES + CMPLX_KINDS_TYPES
```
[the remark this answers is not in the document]
Calluumm: specialfunctions lower_incomplete_gamma has support for xdp but not quadrouple point, implementing that would be an entirely different change. Would leaving the override flag there be fine or do I just remove support for qp in gamma distribution too for now.

test/stats/test_distribution_gamma.fypp
```diff
@@ -0,0 +1,267 @@
+#:set WITH_QP = False
+#:set WITH_XDP = False
+#:include "common.fypp"
+#:set RC_KINDS_TYPES = REAL_KINDS_TYPES + CMPLX_KINDS_TYPES
```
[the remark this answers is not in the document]
Calluumm: Same issue as the actual module

src/stats/stdlib_stats_distribution_gamma.fypp
```diff
@@ -0,0 +1,317 @@
...
+
+        if(x <= 0.0_${k1}$) call error_stop("Error(gamma_dist_pdf): Gamma"     &
+            //" distribution variate x must be greater than zero")
+
+        if(x == 0.0_${k1}$) then
+
+            if(shape <= 1.0_${k1}$) then
+
+                res = 1.0_${k1}$
+
```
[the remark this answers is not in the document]
Calluumm: PDF needs to be error stopped at 0. If it returned a 0 it would be wrong, density is singular at 0 where shape is less than 1, equals rate where shape equals 1 and is 0 when shape is greater than 1. By mathematical definition gamma PDF only applies for x greater than 0

test/stats/test_distribution_gamma.fypp
```diff
@@ -0,0 +1,267 @@
...
+        real(dp) :: chisq, expct
+
+        print *, ""
+        print *, "Test gamma random generator with chi-squared"
+
+        freq = 0
+        do i = 1, num
+            j = 1000 * gamma_cdf(rgamma(2.0_dp, 1.5_dp), 2.0_dp, 1.5_dp)
+            freq(j) = freq(j) + 1
+        end do
```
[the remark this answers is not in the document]
Calluumm: I could reduce to 1 million but the other distribution functions don't so it wouldn't match standards, current test time is ~3s, knocking a 0 off would yeah reduce it by ~10 fold I just don't know what's expected

src/stats/stdlib_stats_distribution_gamma.fypp
```diff
@@ -0,0 +1,317 @@
...
+                res = 0.0_${k1}$
+
+            endif
+
+        else
+
+            res = exp((shape - 1._${k1}$) * log(x) - x * rate + shape *        &
+                log(rate)) / gamma(shape)
+
```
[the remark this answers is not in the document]
Calluumm: Added, not sure if further reliance on specialfunctions will bite back when changes are tried to make to it in the future say to impliment qp

src/stats/stdlib_stats_distribution_gamma.fypp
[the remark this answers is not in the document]
Calluumm: still true even after changing lower_complete_gamma for regular gamma + log gamma

test/stats/test_distribution_gamma.fypp
```diff
@@ -0,0 +1,267 @@
...
+    use stdlib_error, only : check
+    use stdlib_random, only : random_seed
+    use stdlib_stats_distribution_gamma, only : rgamma => rvs_gamma,           &
+                              gamma_pdf => pdf_gamma, gamma_cdf => cdf_gamma
+
+    implicit none
+    #:for k1, t1 in GAMMA_REAL_KINDS_TYPES
+    ${t1}$, parameter :: ${k1}$tol = 1000 * epsilon(1.0_${k1}$)
+    #:endfor
+    logical ::  warn = .true.
```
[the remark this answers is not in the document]
Calluumm: This doesn't match the standard of the existing functions

test/stats/test_distribution_gamma.fypp
```diff
@@ -0,0 +1,267 @@
...
+    #:else
+      #! for complex type
+        gshape = (2.0_${k1}$, 0.7_${k1}$); rate = (0.8_${k1}$, 1.2_${k1}$)
+    #:endif
+
+        x1 = rgamma(gshape, rate)
+        x2 = reshape(rgamma(gshape, rate, 12), [3,4])
+
+        res(1:3) = gamma_pdf(x1, gshape, rate)
+        res(4:15) = reshape(gamma_pdf(x2, gshape, rate), [12])
```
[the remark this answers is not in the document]
Calluumm: Seems unnecessary and non-standard

test/stats/test_distribution_gamma.fypp
```diff
@@ -0,0 +1,267 @@
...
+    #:endfor
+
+contains
+
+    subroutine test_gamma_random_generator
+        integer, parameter :: num = 10000000, array_size = 1000
+        integer :: i, j, freq(0:array_size-1)
+        real(dp) :: chisq, expct
+
+        print *, ""
```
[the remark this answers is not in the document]
Calluumm: Other distribution functions use 10 million

src/stats/stdlib_stats_distribution_gamma.fypp
```diff
@@ -0,0 +1,318 @@
...
+                           kind=${k1}$)
+
+        end do
+    end function gamma_dist_rvs_array_${t1[0]}$${k1}$
+
+    #:endfor
+
+
+    #:for k1, t1 in GAMMA_REAL_KINDS_TYPES
+    impure elemental function gamma_dist_pdf_${t1[0]}$${k1}$(x, shape, rate)   &
```
jalvesz: Is there a reason for so many of these functions to be `impure` ? here for instance, `ieee_value` and `log_gamma` can be used within `elemental` functions. I think it would be good to do a pass to double check the attributes and if they can be promoted to `elemental`.

src/stats/stdlib_stats_distribution_gamma.fypp
```diff
@@ -0,0 +1,318 @@
...
+    ! Variables", G. Marsaglia & W. W. Tsang, ACM Transactions on Mathematical
+    ! Software, 26(3), 2000, p. 363
+    !
+        ${t1}$, intent(in) :: shape
+        ${t1}$ :: res
+        ${t1}$ :: x, v, u, zz, d, c
+        ${t1}$, parameter :: sq = 0.0331_${k1}$
+
+
+        if(shape <= 0.0_${k1}$) call error_stop("Error(gamma_dist_rvs): Gamma"  &
```
jalvesz: Wouldn't it be better for a library function to simply return a nan from `ieee_arithmetic` instead of error stopping? That's something an user can catch. As long as the documentation is clear about the behavior it seems cleaner.

Calluumm: Absolutely! changed to fit

src/stats/stdlib_stats_distribution_gamma.fypp
Calluumm: I left them impure from the original PRs structure. But, the functions largely rely on regamma_p from specialfunctions; that itself is impure elemental, we'd have to change that first if we wanted to change it here.

jalvesz: I took a quick glance at the special functions gamma module. It seems the `impure` attribute was set indeed to enable error stop with a message. This boils down to what should be the design pattern to adopt for the library. Most of those functions seems that would perfectly fit `elemental` functions. Not that it should be changed but it might be worth discussing.

Calluumm: The convention isn't something I'm super familiar with I'm happy to change it to be either.

jalvesz: My proposal to discuss would be:
* move numerical kernels to `elemental` or `pure` procedures when possible, making sure that NaN is returned for an invalid input.
* Add in the same module an impure version which calls the base kernel but also has an extra `intent(out)` error argument

idea sketch
```Fortran
module
use, intrinsic :: ieee_arithmetic
implicit none
private

interface my_fun
 module procedure :: my_func_base
 module procedure :: my_func_impure
end interface
public :: my_fun

contains

elemental function my_func_base(x, shape, rate) result(res)
    real, intent(in) :: x, shape, rate
    real :: res
    ! .... crunch numbers
end module

impure elemental function my_func_impure(x, shape, rate, err) result(res)
    real, intent(in) :: x, shape, rate
    integer, intent(out) :: err !> or type(state_type), intent(out) :: err from https://stdlib.fortran-lang.org/page/specs/stdlib_error_state_type.html
    real :: res
    if( ... condition for bad inputs ... ) err = ! some non zero argument or fixed code to retrieve an error message from a static table
   res = my_func_base( x, shape, rate )
end module

```

Any thoughts @sebastian-mutz, @jvdp1, @perazz ?

sebastian-mutz: I like your proposal.

Procedures like distribution functions lend themselves well to be `elemental` (not "just" `pure`) and I can see a lot of merit in this – thinking about larger stats/ML applications in an HPC setting, for example. Keeping the core procedures `elemental` and creating an `impure` wrapper to handle invalid input and give user feedback (error messages) is the exact approach I took with the [FSML distribution module](https://github.com/sebastian-mutz/fsml/blob/master/src/fsml_dst.f90). 

To clarify: Are you proposing checking the validity of input in the `impure` **and** `elemental` functions, and handling invalid inputs differently (through NaN returns and error messages, respectively)?

jalvesz: Given the `elemental` procedures have to check in any case to properly return NaN, the `impure` version could simply catch the NaN and return an appropriate error in that case.

sebastian-mutz: That sounds good to me.

doc/specs/stdlib_stats_distribution_gamma.md
```diff
@@ -0,0 +1,167 @@
...
+
+## `rvs_gamma` - gamma distribution random variates
+
+### Status
+
+Experimental
+
+### Description
+
+The gamma distribution is a continuous probability distribution widely used in modeling waiting times, queue lengths, and other non-negative quantities. It is parameterized by a shape parameter (\\(k\\) or \\(\\alpha\\)) that controls the distribution's form and a rate parameter (\\(\\lambda\\) or \\(\\beta\\)) that controls its scale. The location parameter (\\(\\text{loc}\\)) shifts the distribution.
```
sebastian-mutz: Minor language suggestion: `It is parameterized by a shape parameter [...]` => `It is defined by a shape parameter [...]`

`or \\(\\beta\\))` can be deleted. `Lambda` is common, and beta isn't referred to anymore in the rest of the doc. Note also that `beta` is usually used to describe the `scale` parameter (the inverse of the rate) and therefore isn't the equivalent.

doc/specs/stdlib_stats_distribution_gamma.md
```diff
@@ -0,0 +1,167 @@
...
+
+Experimental
+
+### Description
+
+The probability density function (pdf) of the single real variable gamma distribution is:
+
+$$ f(x)= \frac{\lambda^{k}}{\Gamma (k)}(x-\text{loc})^{k-1}e^{-\lambda (x-\text{loc})} ,\quad x>\text{loc},\ k>0,\ \lambda>0 $$
+
+where \\(k\\) is the shape parameter, \\(\\lambda\\) is the rate parameter, and \\(\\Gamma(k)\\) is the gamma function.
```
sebastian-mutz, line 79: The location parameter should also be mentioned here.

doc/specs/stdlib_stats_distribution_gamma.md
```diff
@@ -0,0 +1,167 @@
...
+
+Experimental
+
+### Description
+
+Cumulative distribution function (cdf) of the single real variable gamma distribution:
+
+$$ F(x)= \frac{\gamma (k, \lambda (x-\text{loc}))}{\Gamma (k)},\quad x>\text{loc},\ k>0,\ \lambda>0 $$
+
+where \\(\\gamma(k, z)\\) is the lower incomplete gamma function and \\(\\Gamma(k)\\) is the gamma function. This is often referred to as the regularized gamma P function.
```
sebastian-mutz: The loc parameter should be explained here.

doc/specs/stdlib_stats_distribution_gamma.md
```diff
@@ -0,0 +1,167 @@
...
+
+### Status
+
+Experimental
+
+### Description
+
+Cumulative distribution function (cdf) of the single real variable gamma distribution:
+
+$$ F(x)= \frac{\gamma (k, \lambda (x-\text{loc}))}{\Gamma (k)},\quad x>\text{loc},\ k>0,\ \lambda>0 $$
```
sebastian-mutz: For PDF and CDF, you currently include the `loc` shift in the equations.

A case can be made for or against that. On one hand, it is one of the arguments. On the other, it's not a standard parameter for the distributions and lengthens equations (perhaps needlessly?) For reference, [SciPy lists the equations without `loc` ](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.gamma.html) and I used the same approach with [FSML](https://fsml.mutz.science/page/api/dst.html#fsml_gamma_pdf). Whatever is decided on, it should then be kept consistent throughout stdlib stats.

Any opinions on how this should be done for `stdlib`? @jalvesz

Calluumm: The existing ones don't so it's probably best i put this in line with those, but I'll wait to see; I should've cleaned up the other doc fixes in the commit underneath

sebastian-mutz: Since the description is a more high-level introduction to the method/mathematics rather than the specific implementation (the numerically more stable implementation using `logs` also isn't part of the description), I lean more towards skipping the `loc` parameter here.

jalvesz: Hi @Calluumm @sebastian-mutz, I skimmed through the docs of matlab https://www.mathworks.com/help/stats/gamma-distribution.html and also on wikipedia https://en.wikipedia.org/wiki/Gamma_distribution and indeed I didn't see immediately references to the shift `loc` parameter. The clearest doc was indeed the SciPy one cited by @sebastian-mutz 

Now, going through the current implementation, `loc` is a mandatory argument, I didn't see any `optional` formulation. So, in that case it does deserve to appear clearly in the docs and refer to the common use case.

This opens another door: would making these parameters `optional` make things even more clear for both the implementer and consumer of the procedure? It seems that SciPy does handle it as optional arguments. For instance, when I see `cdf(x, a, loc=0, scale=1)` in their doc, I understand that `x` and `a` are mandatory but `loc` and `scale` are optional with clear defaults.

sebastian-mutz: Hi @Calluumm and @jalvesz. 

As a non-`optional` parameter, it does need to be stated clearly what it does, but I'd follow SciPy's approach and leave it out of the equation in the description. This could easily create some confusion. 

The `loc` parameter isn't something you'd typically come across in maths/stats books/resources; not as "loc" anyway. It's a parameter added for computational convenience and consistency across distribution. For example, normal distribution functions usually include `loc` and `scale` with their common mathematical notations (e.g.,   `mu` and `sigma` in [Matlab's normpdf](https://www.mathworks.com/help/stats/normpdf.html)). Giving it more general names and consistently making them (optional) arguments for different distribution seems to be SciPy's approach and I see the merit int this.

For `stdlib`, making it `optional` and falling back to defaults (standardised version with loc=0; scale=1) if not passed is arguably the best path in the long term. It allows users with a maths/stats background to use it more intuitively without having to bother with `loc`, while it's still there for convenience and those used to working with it (from SciPy, for example). This would mean making the `loc` and `scale` parameter optional for other distributions too (for consistency).

doc/specs/stdlib_stats_distribution_gamma.md
```diff
@@ -72,17 +74,17 @@ Experimental
 
 The probability density function (pdf) of the single real variable gamma distribution is:
 
-$$ f(x)= \frac{\lambda^{k}}{\Gamma (k)}(x-\text{loc})^{k-1}e^{-\lambda (x-\text{loc})} ,\quad x>\text{loc},\ k>0,\ \lambda>0 $$
+$$ f(x)= \frac{\lambda^{k}}{\Gamma (k)}x^{k-1}e^{-\lambda x} ,\quad x>0,\ k>0,\ \lambda>0 $$
 
-where \\(k\\) is the shape parameter, \\(\\lambda\\) is the rate parameter, \\(\\text{loc}\\) is the location parameter, and \\(\\Gamma(k)\\) is the gamma function.
+where \\(k\\) is the shape parameter, \\(\\lambda\\) is the rate parameter, and \\(\\Gamma(k)\\) is the gamma function.
```
sebastian-mutz, line 80: I think it's better to remove loc from the equations, as you've done (more familiar way to parameterise the function). However, I'd add a note here, so there's no discrepancy between description and syntax:

"
An optional `loc` parameter specifies the location (shift) of the distribution.
"

doc/specs/stdlib_stats_distribution_gamma.md
```diff
@@ -122,17 +124,17 @@ Experimental
 
 Cumulative distribution function (cdf) of the single real variable gamma distribution:
 
-$$ F(x)= \frac{\gamma (k, \lambda (x-\text{loc}))}{\Gamma (k)},\quad x>\text{loc},\ k>0,\ \lambda>0 $$
+$$ F(x)= \frac{\gamma (k, \lambda x)}{\Gamma (k)},\quad x>0,\ k>0,\ \lambda>0 $$
 
-where \\(\\text{loc}\\) is the location parameter, \\(\\gamma(k, z)\\) is the lower incomplete gamma function, and \\(\\Gamma(k)\\) is the gamma function. This is often referred to as the regularized gamma P function.
+where \\(\\gamma(k, z)\\) is the lower incomplete gamma function, and \\(\\Gamma(k)\\) is the gamma function. This is often referred to as the regularized gamma P function.
```
sebastian-mutz, line 132: Same as above. Simply add:

An optional `loc` parameter specifies the location (shift) of the distribution.

src/stats/stdlib_stats_distribution_gamma.fypp
```diff
@@ -0,0 +1,481 @@
...
+        ${t1}$, intent(in), optional :: loc
+        ${t1}$ :: res
+        ${t1}$ :: loc_
+
+        if(present(loc)) then
+            loc_ = loc
+        else
+            loc_ = 0.0_${k1}$
+        end if
+
```
sebastian-mutz, line 162: Generally, it's more readable and little safer (to maintain) to always initialise and overwrite if argument is passed (the original assignment effectively costs nothing):
```Fortran
loc_ = 0.0_${k1}$
if (present(loc)) loc_ = loc
```
However, in this case, I'd recommend you use stdlib's [optval](https://stdlib.fortran-lang.org/page/specs/stdlib_optval.html). That's what it's for, and it's also used in other stdlib modules like [math_diff](https://github.com/fortran-lang/stdlib/blob/master/src/math/stdlib_math_diff.fypp). This would then simplify to:

```Fortran
loc_ = optval( loc,  0.0_${k1}$ )
```

src/stats/stdlib_stats_distribution_gamma.fypp
```diff
@@ -0,0 +1,481 @@
...
+        ${t1}$, intent(in) :: shape, rate
+        ${t1}$, intent(in), optional :: loc
+        ${t1}$ :: res
+        ${t1}$ :: loc_
+
+        if(present(loc)) then
+            loc_ = loc
+        else
+            loc_ = cmplx(0.0_${k1}$, 0.0_${k1}$, kind=${k1}$)
+        end if
```
sebastian-mutz: Use `optval(x, default)`.

src/stats/stdlib_stats_distribution_gamma.fypp
```diff
@@ -0,0 +1,481 @@
...
+        ${t1}$, intent(in), optional :: loc
+        ${t1}$ :: res(array_size)
+        integer :: i
+        ${t1}$ :: loc_
+
+        if(present(loc)) then
+            loc_ = loc
+        else
+            loc_ = 0.0_${k1}$
+        end if
```
sebastian-mutz: Use `optval(x, default)`.

src/stats/stdlib_stats_distribution_gamma.fypp
```diff
@@ -0,0 +1,481 @@
...
+        ${t1}$, intent(in), optional :: loc
+        ${t1}$ :: res(array_size)
+        integer :: i
+        ${t1}$ :: loc_
+
+        if(present(loc)) then
+            loc_ = loc
+        else
+            loc_ = cmplx(0.0_${k1}$, 0.0_${k1}$, kind=${k1}$)
+        end if
```
sebastian-mutz: Use `optval(x, default)`.

src/stats/stdlib_stats_distribution_gamma.fypp
```diff
@@ -0,0 +1,481 @@
...
+        ${t1}$, intent(in), optional :: loc
+        real(${k1}$) :: res
+        ${t1}$ :: xs
+        ${t1}$ :: loc_
+
+        if(present(loc)) then
+            loc_ = loc
+        else
+            loc_ = 0.0_${k1}$
+        end if
```
sebastian-mutz: Use `optval(x, default)`.

src/stats/stdlib_stats_distribution_gamma.fypp
```diff
@@ -0,0 +1,481 @@
...
+        ${t1}$, intent(in) :: x, shape, rate
+        ${t1}$, intent(in), optional :: loc
+        real(${k1}$) :: res
+        ${t1}$ :: loc_
+
+        if(present(loc)) then
+            loc_ = loc
+        else
+            loc_ = cmplx(0.0_${k1}$, 0.0_${k1}$, kind=${k1}$)
+        end if
```
sebastian-mutz: Use `optval(x, default)`.

src/stats/stdlib_stats_distribution_gamma.fypp
```diff
@@ -0,0 +1,481 @@
...
+        integer, intent(out) :: err
+        ${t1}$, intent(in), optional :: loc
+        real(${k1}$) :: res
+        ${t1}$ :: loc_
+
+        if(present(loc)) then
+            loc_ = loc
+        else
+            loc_ = cmplx(0.0_${k1}$, 0.0_${k1}$, kind=${k1}$)
+        end if
```
sebastian-mutz: Use `optval(x, default)`.

src/stats/stdlib_stats_distribution_gamma.fypp
```diff
@@ -0,0 +1,481 @@
...
+        ${t1}$, intent(in) :: x, shape, rate
+        ${t1}$, intent(in), optional :: loc
+        real(${k1}$) :: res, xs
+        ${t1}$ :: loc_
+
+        if(present(loc)) then
+            loc_ = loc
+        else
+            loc_ = 0.0_${k1}$
+        end if
```
sebastian-mutz: Use `optval(x, default)`.

src/stats/stdlib_stats_distribution_gamma.fypp
```diff
@@ -0,0 +1,481 @@
...
+        ${t1}$, intent(in) :: x, shape, rate
+        ${t1}$, intent(in), optional :: loc
+        real(${k1}$) :: res
+        ${t1}$ :: loc_
+
+        if(present(loc)) then
+            loc_ = loc
+        else
+            loc_ = cmplx(0.0_${k1}$, 0.0_${k1}$, kind=${k1}$)
+        end if
```
sebastian-mutz: Use `optval(x, default)`.

src/stats/stdlib_stats_distribution_gamma.fypp
```diff
@@ -0,0 +1,481 @@
...
+        integer, intent(out) :: err
+        ${t1}$, intent(in), optional :: loc
+        real(${k1}$) :: res
+        ${t1}$ :: loc_
+
+        if(present(loc)) then
+            loc_ = loc
+        else
+            loc_ = cmplx(0.0_${k1}$, 0.0_${k1}$, kind=${k1}$)
+        end if
```
sebastian-mutz: Use `optval(x, default)`.

sebastian-mutz (approved):
LGTM, @Calluumm. Let's give it a few days in case @jalvesz or anyone else has any comments.

PR #1104, opened 2026-01-24
feat(linalg): add weighted least squares solver

### LINKED ISSUE #1047: More least-squares variants

aamrindersingh:
Resolves #1047 (1 of 2)

This PR adds the `weighted_lstsq` interface to stdlib_linalg for weighted least squares problems. It handles diagonal weight matrices where each observation has a different importance, common in heteroscedastic regression and outlier downweighting. 

The key design decisions are:

1. Validates that all weights are positive since negative weights have no statistical meaning and would cause numerical issues with `sqrt(w)`.
2. Follows the `overwrite_a` pattern from `solve_lu` where `copy_a` defaults to `.true.` to preserve A unless the user explicitly opts into destruction for performance.
3. Uses column major loops for cache friendly memory access in the row scaling transformation.
4. Uses allocatable arrays instead of automatic arrays to avoid stack overflow on large problems.

Testing includes:

- Basic WLS solves with uniform weights
- Verification that non uniform weights change solutions  
- Error handling for negative weights
- All real types (sp/dp/qp/xdp) covered following existing `lstsq` patterns


test/linalg/test_linalg_lstsq.fypp
```diff
@@ -97,6 +100,98 @@ module test_linalg_least_squares
...
+        real(${rk}$) :: w(m)
+        ${rt}$, allocatable :: x(:)
+
+        ! Simple test case
+        A(:,1) = 1.0_${rk}$
+        A(:,2) = [1.0_${rk}$, 2.0_${rk}$, 3.0_${rk}$, 4.0_${rk}$]
+        b = [2.0_${rk}$, 4.0_${rk}$, 5.0_${rk}$, 4.0_${rk}$]
+        w = [1.0_${rk}$, 1.0_${rk}$, 1.0_${rk}$, 1.0_${rk}$]  ! Uniform weights = OLS
+
+        x = weighted_lstsq(w, A, b, err=state)
```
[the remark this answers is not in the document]
aamrindersingh: @jvdp1, Added a test verifying this just to be safe but i just checked existing lstsq/solve_lstsq tests do not check this either

test/linalg/test_linalg_lstsq.fypp
```diff
@@ -97,6 +100,98 @@ module test_linalg_least_squares
...
+
+        call check(error, state%ok(), 'weighted_lstsq failed: '//state%print())
+        if (allocated(error)) return
+        
+        call check(error, size(x)==n, 'weighted_lstsq: wrong solution size')
+        if (allocated(error)) return
+
+        ! Verify residual is small
+        call check(error, norm2(matmul(A,x) - b) < 2.0_${rk}$, 'weighted_lstsq: residual too large')
```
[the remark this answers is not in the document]
aamrindersingh: Done

test/linalg/test_linalg_lstsq.fypp
```diff
@@ -97,6 +100,98 @@ module test_linalg_least_squares
...
+    !-------------------------------------------------------------
+
+    #:for rk,rt,ri in REAL_KINDS_TYPES
+    !> Test basic weighted least-squares
+    subroutine test_weighted_lstsq_${ri}$(error)
+        type(error_type), allocatable, intent(out) :: error
+
+        type(linalg_state_type) :: state
+        integer(ilp), parameter :: m = 4, n = 2
+        real(${rk}$), parameter :: tol = 100*sqrt(epsilon(0.0_${rk}$))
```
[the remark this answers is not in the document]
aamrindersingh: Done

src/linalg/stdlib_linalg_least_squares.fypp
```diff
@@ -564,4 +564,114 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+        do j = 1, n
+            amat(:, j) = sqrt_w(:) * a(:, j)
+        end do
+
+        ! Scale b
+        allocate(b_scaled(m))
+        b_scaled = sqrt_w * b
+
+        ! Solve transformed OLS problem
+        x = stdlib_linalg_${ri}$_lstsq_one(amat, b_scaled, cond=cond, overwrite_a=.true., rank=rank, err=err)
```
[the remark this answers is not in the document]
aamrindersingh: suggested fix has a bug: `if (.not. allocated(x)) `is dead code since x is already allocated at line 605. Instead, just call solve_lstsq_one directly.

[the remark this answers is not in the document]
aamrindersingh: Done with this Fix

src/linalg/stdlib_linalg_least_squares.fypp
```diff
@@ -564,4 +564,114 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+            amat(:, j) = sqrt_w(:) * a(:, j)
+        end do
+
+        ! Scale b
+        allocate(b_scaled(m))
+        b_scaled = sqrt_w * b
+
+        ! Solve transformed OLS problem
+        call stdlib_linalg_${ri}$_solve_lstsq_one(amat, b_scaled, x, cond=cond, overwrite_a=.true., rank=rank, err=err)
+
```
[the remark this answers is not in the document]
aamrindersingh: @jvdp1 Need some inputs on this :-
right now weighted_lstsq validation errors say "weighted_lstsq" but internal errors say "lstsq". Should I change it so all errors just say "lstsq" for consistency or keep it as is?

src/linalg/stdlib_linalg_least_squares.fypp
```diff
@@ -564,4 +564,114 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+            if (present(rank)) rank = 0
+            return
+        end if
+
+        ! Can A be overwritten? By default, do not overwrite
+        if (present(overwrite_a)) then
+            copy_a = .not.overwrite_a
+        else
+            copy_a = .true._lk
+        end if
```
loiseaujc: I think you can replace these lines with

```fortran
copy_a = .not. optval(overwrite_a, .false._lk)
```

src/linalg/stdlib_linalg_least_squares.fypp
```diff
@@ -564,4 +564,114 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+
+        ! Compute sqrt of weights
+        allocate(sqrt_w(m))
+        sqrt_w = sqrt(w)
```
loiseaujc, line 688: You can combine these two lines into

```fortran
allocate(sqrt_w(m), source=sqrt(w))
```

src/linalg/stdlib_linalg_least_squares.fypp
```diff
@@ -564,4 +564,114 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+            allocate(amat_alloc(m, n))
+            amat => amat_alloc
+        else
+            amat => a
+        end if
+
+        ! Scale A column-wise (cache-friendly: column-major order)
+        do j = 1, n
+            amat(:, j) = sqrt_w(:) * a(:, j)
+        end do
```
loiseaujc: This is a multiplication by a diagonal matrix. You can use the `lascl2` function from `lapack` for that purpose.

src/linalg/stdlib_linalg_least_squares.fypp
```diff
@@ -564,4 +564,114 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+
+        ! Scale b
+        allocate(b_scaled(m))
+        b_scaled = sqrt_w * b
```
loiseaujc, line 702: You can fuse these two lines into

```fortran
allocate(b_scaled(m), source=sqrt_w * b)
```

or even simply 

```fortran
b_scaled = sqrt_w * b
```

src/linalg/stdlib_linalg_least_squares.fypp
loiseaujc: Or even simply `sqrt_w = sqrt(w)`.

src/linalg/stdlib_linalg_least_squares.fypp
aamrindersingh: Yes, Fixed
Will make sure to use this for other issues as well. 
Thanks

src/linalg/stdlib_linalg_least_squares.fypp
aamrindersingh: Done

src/linalg/stdlib_linalg_least_squares.fypp
aamrindersingh: @loiseaujc ,
Looked into using `lascl2 ` for diagnol matrix multiplication but it does not appear to be wrapped in stdlib LAPACK interface 
Shall I proceed with opening a separate issue/PR for adding `lascl2`?

loiseaujc: @perazz : I've never really asked, but which version of `lapack`did you use for the modernization? It seems like some of `blas`-like function in [`lapack v3.12.1`](https://www.netlib.org/lapack/explore-html/d7/df0/group__blas2__like__grp.html) are no present in our port.

src/linalg/stdlib_linalg_least_squares.fypp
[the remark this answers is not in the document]
loiseaujc: I ain't too familiar with how the `linalg_error_type` system works. @perazz : any input on this?

src/linalg/stdlib_linalg.fypp
```diff
@@ -679,6 +680,48 @@ module stdlib_linalg
...
+        real(${rk}$), optional, intent(in) :: cond
+        !> [optional] Can A data be overwritten and destroyed?
+        logical(lk), optional, intent(in) :: overwrite_a
+        !> [optional] Return rank of A
+        integer(ilp), optional, intent(out) :: rank
+        !> [optional] state return flag. On error if not requested, the code will stop
+        type(linalg_state_type), optional, intent(out) :: err
+        !> Result array x[n]
+        ${rt}$, allocatable :: x(:)
+    end function stdlib_linalg_${ri}$_weighted_lstsq
```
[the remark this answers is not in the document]
jalvesz: `target` is to be used only if needed in the case one needs to set a pointer on top. If that's not the case, then please don't use it.

src/linalg/stdlib_linalg_least_squares.fypp
jalvesz: A plain nested `do` loop is also fine here.

@loiseaujc the reference lapack was 3.10

loiseaujc: `dlascl2` was introduced in 3.11 apparently, so that's the reason why. In any case, the code is as simple as

```fortran
do concurrent( i=1:m, j=1:n)
    x(i, j) = d(i) * x(i, j)
end do
```

You can probably keep it as is for the time being because it's so simple. Eventually we'll probably have to update the `lapack` module to 3.12 and we'll change these few lines at this time. May be you can simple add a comment mentioning `dlascl2` so that we easily grep it and retrieve it whenever `lapack` will have been updated.

perazz: Another option is to add a templated `lascl2` implementation from LAPACK 3.11+ into our LAPACK module: it is only one routine, it should be manageable

loiseaujc: Sure enough. The code is extremely simple:

```fortran
*  =====================================================================
      SUBROUTINE dlascl2( M, N, D, X, LDX )
*
*  -- LAPACK computational routine --
*  -- LAPACK is a software package provided by Univ. of Tennessee,    --
*  -- Univ. of California Berkeley, Univ. of Colorado Denver and NAG Ltd..--
*
*     .. Scalar Arguments ..
      INTEGER            M, N, LDX
*     ..
*     .. Array Arguments ..
      DOUBLE PRECISION   D( * ), X( LDX, * )
*     ..
*
*  =====================================================================
*
*     .. Local Scalars ..
      INTEGER            I, J
*     ..
*     .. Executable Statements ..
*
      DO j = 1, n
         DO i = 1, m
            x( i, j ) = x( i, j ) * d( i )
         END DO
      END DO
 
      RETURN
      END
```

I'm commuting right now and, out of curiosity, have started to look at the changelogs for `lapack 3.10.1` and `lapack 3.11.0`. Nothing too complicated as far as I can see. May a bit tedious to port all of these changes, but fairly easy nonetheless.

perazz: > The code is extremely simple:

so many such cases; perhaps our implementation should have `do concurrent` internally. However, if libraries are going to pick it up in the future as part of the "external" API, maybe the future-proof choice is to use the same API?

loiseaujc: Using `do concurrent` internally makes sense. When you say the same api, you mean the same signature (i.e. `dlascl2( M, N, D, X, LDX )`) ?

perazz: yep, same signature as the official API

test/linalg/test_linalg_lstsq.fypp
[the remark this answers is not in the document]
perazz: [9666b45](https://github.com/fortran-lang/stdlib/pull/1104/commits/9666b4543df572617bf5ea615021a8d84e953c89)

perazz (commented):
Here is a summary of the outstanding edits/suggestions from the discussion above: 

1a) add a procedure `update_location` to https://github.com/fortran-lang/stdlib/blob/9591c0d4c1ddc4236ab44a3bbc67d0b54fdef004/src/core/stdlib_error.fypp#L61 to update the location of the error message
1b) in `linalg_error_handling`, add a third optional parameter `where_at` and if `(present(where_at))` update the location of the error message;
1c) in `call stdlib_linalg_${ri}$_solve_lstsq_one(... err = err0)` -> replace err with err0 and then call the error handler `(err0, err, where_at='weighted_lstsq`)
2) create a templated `*lascl2` procedure, should be in the LAPACK's "blas-like" module

src/core/stdlib_error.fypp
```diff
@@ -252,6 +255,15 @@ contains
 
     end subroutine error_handling
 
+    !> Update the location of the error message
+    pure subroutine state_update_location(this, where_at)
+         class(state_type), intent(inout) :: this
+         character(len=*), intent(in) :: where_at
+
```
perazz: `where_at` could be `optional`. If not `present`, return

src/lapack/stdlib_lapack_blas_like_l2.fypp
```diff
@@ -5681,4 +5681,55 @@ submodule(stdlib_lapack_base) stdlib_lapack_blas_like_l2
...
+           integer(${ik}$), intent(in) :: m, n, ldx
+           ! Array Arguments
+           real(${rk}$), intent(in) :: d(*)
+           real(${rk}$), intent(inout) :: x(ldx,*)
+        ! =====================================================================
+           ! Local Scalars
+           integer(${ik}$) :: i, j
+           ! Executable Statements
+           do j = 1, n
+              do i = 1, m
```
perazz: I believe the previous comments suggested `do concurrent`

src/linalg_core/stdlib_linalg_state.fypp
```diff
@@ -73,17 +73,25 @@ module stdlib_linalg_state
...
          type(linalg_state_type),optional,intent(out) :: ierr_out
+         character(len=*),optional,intent(in) :: where_at
 
          character(len=:),allocatable :: err_msg
 
          if (present(ierr_out)) then
              ! Return error flag
              ierr_out = ierr
+             ! Update location if requested
+             if (present(where_at)) call ierr_out%update_location(where_at)
```
perazz: You can remove `if (present())` when that argument is optional in update_location

src/linalg_core/stdlib_linalg_state.fypp
```diff
@@ -73,17 +73,25 @@ module stdlib_linalg_state
...
          elseif (ierr%error()) then
-             err_msg = ierr%print()
+             if (present(where_at)) then
+                 err_msg = '['//trim(where_at)//'] returned '//ierr%print_msg()
+             else
+                 err_msg = ierr%print()
+             end if
```
perazz: Duplicate location print: location handing is already in `pring_msg`

perazz (changes requested):
LGTM overall, except a few tiny necessary edits, thank you.

example/io/example.dat
```diff
@@ -1,3 +1 @@
- 1.00000000E+00  1.00000000E+00
- 1.00000000E+00  1.00000000E+00
- 1.00000000E+00  1.00000000E+00
+This is an example for open
```
perazz: Unrelated change: please restore the original file

aamrindersingh: Ran build and tests , will restore this to original 
Thanks

src/lapack/stdlib_lapack_blas_like_l2.fypp
```diff
@@ -5681,4 +5681,51 @@ submodule(stdlib_lapack_base) stdlib_lapack_blas_like_l2
 
 
 #:endfor
+
+! LASCL2 performs diagonal scaling on a matrix: X(i,:) = D(i) * X(i,:)
```
jvdp1, line 5685: Do we need these subroutines as it is only to encapsulate a `do concurrent`?

src/linalg/stdlib_linalg_least_squares.fypp
```diff
@@ -564,4 +564,109 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+        ! Handle A matrix: either copy or use original
+        if (copy_a) then
+            allocate(amat_alloc(m, n), source=a)
+            amat => amat_alloc
+        else
+            amat => a
+        end if
+
+        ! Scale rows of A by sqrt(w) using LAPACK's lascl2
+        call lascl2(m, n, sqrt_w, amat, m)
```
jvdp1, line 699: This could be replaced by the `do concurrent`. Or what is the reason to call this procedure?

doc/specs/stdlib_linalg.md
```diff
@@ -976,6 +976,52 @@ call [stdlib_linalg(module):constrained_lstsq_space(interface)]`(a,  c,  lwork [
 `c`: Shall be a rank-2 `real` or `complex` array of the same kind as `a` defining the linear equality constraints. It is an `intent(in)` argument.
 `lwork`: Shall be an `integer` scalar returning the optimal size required for the workspace array to solve the constrained least-squares problem.
 
+## `weighted_lstsq` - Computes the weighted least squares solution to a linear matrix equation {#weighted-lstsq}
```
jvdp1, line 979: I suggest to also introduce the subroutine version of this function (as it is the case for other procedures).

src/lapack/stdlib_lapack_blas_like_l2.fypp
perazz: We had a previous discussion above with @loiseaujc , it seems like this procedure is part of the "new" LAPACK 3.12.0+ API, so we kind of decided that irrespective of the "simple" implementation, it was a good idea to maintain the same API, for when external libraries provide that

jvdp1: Sorry, I just read the discussion now.
I agree that it is good to maintain the same API as in the "new" lapack, and it is good to introduce `lascl2` in `stdlib`. 
However, I wonder if it is wise to call it in a newly introduced procedure of `stdlib`. IMO, it adds complexity to understand the code, without real benefits.

doc/specs/stdlib_linalg.md
loiseaujc: Modulo the subroutine version, I think you are pretty much done @aamrindersingh.

aamrindersingh: Done @loiseaujc @jvdp1

example/linalg/example_weighted_lstsq.f90
```diff
@@ -0,0 +1,26 @@
...
+  A(:,2) = [1.0_dp, 2.0_dp, 3.0_dp, 4.0_dp]
+
+  ! Observations (one outlier at position 3)
+  b = [2.1_dp, 4.0_dp, 10.0_dp, 7.9_dp]
+
+  ! Weights: downweight the outlier
+  w = [1.0_dp, 1.0_dp, 0.1_dp, 1.0_dp]
+
+  ! Solve weighted least-squares
+  x = weighted_lstsq(w, A, b)
```
jvdp1, line 21: Could we have a single example file?

src/linalg/stdlib_linalg.fypp
```diff
@@ -679,6 +681,90 @@ module stdlib_linalg
...
+    !!
+    !!@note The solution is based on LAPACK's `*GELSD` after applying diagonal weights.
+    !!@warning Avoid extreme weight ratios (e.g., max(w)/min(w) > 1e6) as this may 
+    !!         cause loss of precision in the SVD-based solver.
+    !!
+    #:for rk,rt,ri in RC_KINDS_TYPES
+    module function stdlib_linalg_${ri}$_weighted_lstsq(w,a,b,cond,overwrite_a,rank,err) result(x)
+        !> Weight vector (must be positive, always real)
+        real(${rk}$), intent(in) :: w(:)
+        !> Input matrix a[m,n]
```
jvdp1: here and elsewhere: could you use the fortran notation, please?
```suggestion
        !> Input matrix a(m,n)
```

src/linalg/stdlib_linalg_least_squares.fypp
```diff
@@ -564,4 +564,154 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+        !> [optional] cutoff for rank evaluation: singular values s(i)<=cond*maxval(s) are considered 0.
+        real(${rk}$), optional, intent(in) :: cond
+        !> [optional] Can A data be overwritten and destroyed?
+        logical(lk), optional, intent(in) :: overwrite_a
+        !> [optional] Return rank of A
+        integer(ilp), optional, intent(out) :: rank
+        !> [optional] state return flag. On error if not requested, the code will stop
+        type(linalg_state_type), optional, intent(out) :: err
+        !> Result array x[n]
+        ${rt}$, allocatable, target :: x(:)
```
jvdp1:
```suggestion
        ${rt}$, allocatable :: x(:)
```

src/linalg/stdlib_linalg.fypp
```diff
@@ -679,6 +681,90 @@ module stdlib_linalg
...
+        !> [optional] cutoff for rank evaluation: singular values s(i)<=cond*maxval(s) are considered 0.
+        real(${rk}$), optional, intent(in) :: cond
+        !> [optional] Can A data be overwritten and destroyed?
+        logical(lk), optional, intent(in) :: overwrite_a
+        !> [optional] Return rank of A
+        integer(ilp), optional, intent(out) :: rank
+        !> [optional] state return flag. On error if not requested, the code will stop
+        type(linalg_state_type), optional, intent(out) :: err
+        !> Result array x[n]
+        ${rt}$, allocatable, target :: x(:)
```
jvdp1:
```suggestion
        ${rt}$, allocatable :: x(:)
```

src/linalg/stdlib_linalg_least_squares.fypp
```diff
@@ -564,4 +564,114 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+        real(${rk}$), optional, intent(in) :: cond
+        !> [optional] Can A data be overwritten and destroyed?
+        logical(lk), optional, intent(in) :: overwrite_a
+        !> [optional] Return rank of A
+        integer(ilp), optional, intent(out) :: rank
+        !> [optional] state return flag. On error if not requested, the code will stop
+        type(linalg_state_type), optional, intent(out) :: err
+        !> Result array x[n]
+        ${rt}$, allocatable :: x(:)
+
```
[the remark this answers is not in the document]
jvdp1: I am not sure I agree with copilot here. If I understand it well, the `target` attribute is only within the procedure, and is useless in this case.

jvdp1 (commented):
thank you @aamrindersingh . Overall it looks good. Here are some minor suggestions.

perazz:
@aamrindersingh do you still plan to complete @jvdp1's outstanding reviews? it seems like they're mostly cosmetic, should be easy to deploy.

example/linalg/example_weighted_lstsq.f90
aamrindersingh: Done

src/linalg/stdlib_linalg.fypp
aamrindersingh: Fixed in all

PR #1105, opened 2026-01-24
feat(linalg): add generalized least squares solver

### LINKED ISSUE #1047: More least-squares variants

aamrindersingh:
Resolves #1047 (2 of 2)

This PR adds the `generalized_lstsq` interface to `stdlib_linalg` for least-squares problems with correlated errors described by a symmetric positive definite covariance matrix. Usage: `x = generalized_lstsq(W, A, b)`. This minimizes the Mahalanobis distance `(Ax-b)^T W^{-1} (Ax-b)` using LAPACK's `GGGLM`.

The key design decisions are:

1. `generalized_lstsq` always copies `W` internally to protect user data from `GGGLM`'s destructive operations. This applies regardless of the `prefactored_w` flag, the input `W` matrix is never modified.
2. The interface follows the `overwrite_a` pattern from `solve_lu` where `copy_a` defaults to `.true.` to preserve `A` unless the user explicitly opts into destruction for performance.
3. A `handle_ggglm_info` error handler was added to parse LAPACK return codes, consistent with existing `handle_gelsd_info` and `handle_gglse_info` patterns.

Testing includes:

- Basic GLS solves with correlated errors
- Verification that GLS with identity covariance equals OLS
- All real types (sp/dp/qp/xdp) covered following existing `lstsq` patterns


src/linalg/stdlib_linalg_least_squares.fypp
```diff
@@ -564,4 +564,150 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+        if (m < n) then
+            err0 = linalg_state_type(this, LINALG_VALUE_ERROR, &
+                   'GGGLM requires m >= n (overdetermined or square):', m, '<', n)
+            call linalg_error_handling(err0, err)
+            return
+        end if
+
+        ! Process options
+        is_prefactored = .false._lk
+        if (present(prefactored_w)) is_prefactored = prefactored_w
```
loiseaujc: You can use

```fortran
is_prefactored = optval(prefactored_w, .false._lk)
```

src/linalg/stdlib_linalg_least_squares.fypp
```diff
@@ -564,4 +564,150 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+
+        if (present(overwrite_a)) then
+            copy_a = .not.overwrite_a
+        else
+            copy_a = .true._lk
+        end if
```
loiseaujc, line 668: Same as for the weighted least-squares, I think you can collapse these lines to

```fortran
copy_a = .not. optval(overwrite_a, .false._lk)
```

src/linalg/stdlib_linalg_least_squares.fypp
```diff
@@ -564,4 +564,150 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+
+        ! Handle covariance/Cholesky factor
+        ! ALWAYS copy W because GGGLM modifies it (protects user's data)
+        allocate(lmat(m, m), source=w)
+
+        if (.not. is_prefactored) then
+            ! Compute Cholesky factorization: W = L * L^T (real) or W = L * L^H (complex)
+            call potrf('L', m, lmat, m, info)
+            if (info /= 0) then
```
loiseaujc: Please use the function `handle_potrf_info` from the `stdlib_linalg_lapack_aux` module to handle the return flag of `potrf`.

```fortran
call handle_potrf_info(this, whatever_the_other_args_are, err0)
call linalg_error_handling(err0, err)
```

loiseaujc: Note by the way that you could directly use the `cholesky` subroutine from the `stdlib_linalg_module` for that purpose.

src/linalg/stdlib_linalg_least_squares.fypp
```diff
@@ -564,4 +564,150 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+        lda = m
+        ldb = m
+
+        ! Zero out upper triangle of lmat (GGGLM reads full matrix, 
+        ! but potrf only sets lower triangle)
+        do j = 1, m
+            do i = 1, j - 1
+                lmat(i, j) = zero
+            end do
+        end do
```
loiseaujc: You can replace these lines with

```fortran
do concurrent(i=1:m, j=1:m, i < j)
    lmat(i, j) = zero
end do
```

Note that, had you used the `cholesky` subroutine, it admits an `other_zeroed` flag to do just that.

loiseaujc (changes requested):
Again, I took a quick glance at your code and will try to go deeper into it by the end of the week. Looks pretty good so far.

src/linalg/stdlib_linalg_least_squares.fypp
aamrindersingh: Done

src/linalg/stdlib_linalg_least_squares.fypp
aamrindersingh: Done
 using stdlib cholesky subroutine , Much cleaner now

src/linalg/stdlib_linalg_least_squares.fypp
aamrindersingh: Yes Done

src/linalg/stdlib_linalg_least_squares.fypp
aamrindersingh: Done

src/linalg/stdlib_linalg_least_squares.fypp
```diff
@@ -564,4 +564,135 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+                if (copy_a) deallocate(amat_alloc)
+                deallocate(lmat)
+                call linalg_error_handling(err0, err)
+                return
+            end if
+        else
+            ! User provided pre-factored L: zero out upper triangle for GGGLM
+            do concurrent(i=1:m, j=1:m, i < j)
+                lmat(i, j) = zero
+            end do
```
loiseaujc: Since you use `cholesky(lmat, lower=.true., other_zeroed=.true.)`, you do not need to zero-out `lmat` again. `cholesky` does it for you.

src/lapack/stdlib_linalg_lapack_aux.fypp
```diff
@@ -1654,4 +1655,24 @@ module stdlib_linalg_lapack_aux
...
+        character(len=*), intent(in) :: this
+        integer(ilp), intent(in) :: info, m, n, p
+        type(linalg_state_type), intent(out) :: err
+        ! Process output.
+        select case (info)
+            case (0)
+                ! Success.
+                err%state = LINALG_SUCCESS
+            case (:-1)
+                err = linalg_state_type(this, LINALG_VALUE_ERROR, 'Invalid argument at position', -info)
```
loiseaujc: Please expend the `select case` to return exactly which input argument is wrong rather than lump them all in `case (:-1)`.

loiseaujc (changes requested):

src/linalg/stdlib_linalg_least_squares.fypp
aamrindersingh: @loiseaujc 
the `do concurrent `zeroing is in the else branch and it only runs when `prefactored_w=.true.`

src/lapack/stdlib_linalg_lapack_aux.fypp
aamrindersingh: Done , expanded to specific cases for M, N, P, LDA, LDB, and LWORK.

src/linalg/stdlib_linalg_least_squares.fypp
aamrindersingh: Went Through other stdlib functions , Should I remove the else branch and trust user provides proper lower triangular L?

loiseaujc: Two possibilities there:

1. Dumb-proof the code and effectively lower the upper triangular part of the user provided matrix just to make sure.

or

2. Specify clearly in the specs that, if already prefactored, the upper triangular part of $W$ needs to be zero.

@perazz @jalvesz @jvdp1 : What would be your take on this and user-friendliness in general for such issues?

jalvesz: Since `lmat` is an internal allocatable variable, the simplest solution would be:
Allocate `lmat` accordingly with `source=zero`, assign the  lower symmetric values from `w` systematically. Just check `if (.not. is_prefactored)` to perform the cholesky factorization.

src/lapack/stdlib_linalg_lapack_aux.fypp
```diff
@@ -1654,4 +1655,34 @@ module stdlib_linalg_lapack_aux
...
+                ! Success.
+                err%state = LINALG_SUCCESS
+            case (-1)
+                err = linalg_state_type(this, LINALG_VALUE_ERROR, 'Invalid number of rows for A and B, m=', m)
+            case (-2)
+                err = linalg_state_type(this, LINALG_VALUE_ERROR, 'Invalid number of columns for A, n=', n)
+            case (-3)
+                err = linalg_state_type(this, LINALG_VALUE_ERROR, 'Invalid number of columns for B, p=', p)
+            case (-5)
+                err = linalg_state_type(this, LINALG_VALUE_ERROR, 'Invalid leading dimension for A, lda < m=', m)
```
perazz: Please pass `lda`, `ldb` to this routine so the
```suggestion
                err = linalg_state_type(this, LINALG_VALUE_ERROR, 'Invalid leading dimension for A, lda=',lda,' is < m=', m)
```

src/linalg/stdlib_linalg_least_squares.fypp
```diff
@@ -564,4 +564,135 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+        if (m < 1 .or. n < 1) then
+            err0 = linalg_state_type(this, LINALG_VALUE_ERROR, 'Invalid matrix size a(m, n) =', [m, n])
+            call linalg_error_handling(err0, err)
+            return
+        end if
+
+        ! Validate sizes
+        if (size(w, 1, kind=ilp) /= m .or. size(w, 2, kind=ilp) /= m) then
+            err0 = linalg_state_type(this, LINALG_VALUE_ERROR, &
+                   'Covariance matrix must be square m×m:', [size(w, 1, kind=ilp), size(w, 2, kind=ilp)])
```
perazz:
```suggestion
                   'Covariance matrix must be square m×m:', shape(w, kind=ilp))
```

src/linalg/stdlib_linalg_least_squares.fypp
```diff
@@ -564,4 +564,135 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+        ! Handle A matrix
+        if (copy_a) then
+            allocate(amat_alloc(m, n), source=a)
+            amat => amat_alloc
+        else
+            amat => a
+        end if
+
+        ! Handle covariance/Cholesky factor
+        ! ALWAYS copy W because GGGLM modifies it (protects user's data)
```
perazz: Should we provide an `overwrite_w` logical argument, the same way we do for `overwrite_a`? @loiseaujc that would allow to avoid internal allocation if the user can afford that. 

We have strived to allow users to pass all necessary variables as optional arguments and avoid internal allocations wherever possible, here we have at least 4 separate allocations at every call. Perhaps this could be avoided?

src/linalg/stdlib_linalg_least_squares.fypp
perazz: @loiseaujc I agree: when `w` is a user-provided prefactored matrix, we should not modify that. It is the user's responsibility to provide a correct input matrix (perhaps we could state that in the documentation).  It would be also worthwhile to check if DGGGLM actually does access the upper triangular part or not - perhaps this zeroing is not even necessary

perazz (commented):
Thanks for this PR @aamrindersingh, I have added some comments.

src/linalg/stdlib_linalg_least_squares.fypp
loiseaujc: `xGGGLM` solves the following constrained quadratic program

```
!>         minimize || y ||_2   subject to   d = A*x + B*y
!>             x
```

where `A` and `B` are generic matrices. The generalized least-squares problem can be recast in this form where `B` is a square root of the symmetric positive definite weight matrix `W`. Since `xGGGLM` works for arbitrary matrices `B`, we indeed do need to zero-out the upper triangular part if `B` is a Cholesky factor.

But I actually agree that we should not touch the user-provided matrix. One simple reason is that the Cholesky factor is not the only way to express the matrix square-root ! Maybe for some reason, user has computed it with `svd`. It would be a valid input and the routine should still work.

loiseaujc: @aamrindersingh : Following with previous comment, could you add a test where the matrix square root of `W` is computed using its svd followed by taking the non-negative square-root of the singular values ? If I'm correct, it should return the same solution as if the Cholesky factor of `W` was used.

src/linalg/stdlib_linalg_least_squares.fypp
loiseaujc: Yep, that would seem consistent with the rest of the `stdlib_linalg` stuff. As usual, `overwrite_w` would default to `.false.` just to make sure only users who know what they're doing are actually overwriting.

src/linalg/stdlib_linalg_least_squares.fypp
aamrindersingh: Done

src/lapack/stdlib_linalg_lapack_aux.fypp
aamrindersingh: Done

src/linalg/stdlib_linalg_least_squares.fypp
aamrindersingh: Done, 
Removed the upper triangle zeroing and added an eigendecomposition based sqrt test. Confirms both give identical solutions as expected. Used eigh since it is equivalent to SVD for SPD matrices.

src/linalg/stdlib_linalg_least_squares.fypp
aamrindersingh: @perazz @loiseaujc 
Added overwrite_w (default .false.), matching overwrite_a pattern. Both main matrix allocations are now user controllable.
Are any other changes required?

test/linalg/test_linalg_lstsq.fypp
```diff
@@ -139,6 +143,172 @@ module test_linalg_least_squares
...
+        ! Design matrix
+        A(:,1) = 1.0_${rk}$
+        A(:,2) = [1.0_${rk}$, 2.0_${rk}$, 3.0_${rk}$, 4.0_${rk}$]
+        b = [2.0_${rk}$, 4.0_${rk}$, 5.0_${rk}$, 4.0_${rk}$]
+
+        ! Identity covariance
+        W = 0.0_${rk}$
+        do i = 1, m
+            W(i,i) = 1.0_${rk}$
+        end do
```
loiseaujc: You can use the `eye` function from `stdlib_linalg`.

test/linalg/test_linalg_lstsq.fypp
```diff
@@ -139,6 +143,172 @@ module test_linalg_least_squares
...
+        if (allocated(error)) return
+
+        ! Compute eigendecomposition: W = U * diag(lambda) * U^T
+        call eigh(W, lambda, vectors=U, err=state)
+        call check(error, state%ok(), 'eigh failed: '//state%print())
+        if (allocated(error)) return
+
+        ! Compute matrix square root: sqrt(W) = U * diag(sqrt(lambda)) * U^T
+        ! This is a DENSE matrix (not triangular like Cholesky factor)
+        sqrt_W = matmul(U, matmul(diag(sqrt(lambda)), transpose(U)))
```
loiseaujc: You only test for real matrices at the moment, but whenever complex matrices will be handled, be careful that the transpose operation on `U` will need to be replaced with the hermitian. Save yourself some trouble possibly and use the `hermitian` function from `stdlib_linalg` just to be on the safe side.

doc/specs/stdlib_linalg.md
```diff
@@ -976,6 +976,52 @@ call [stdlib_linalg(module):constrained_lstsq_space(interface)]`(a,  c,  lwork [
...
+### Arguments
+
+`w`: Shall be a rank-2 `real` or `complex` symmetric (real) or Hermitian (complex) positive definite array containing the covariance matrix, or a valid matrix square root (e.g., Cholesky factor, SVD-based) if `prefactored_w=.true.`. It is an `intent(inout)` argument.
+
+`a`: Shall be a rank-2 `real` or `complex` array containing the coefficient matrix. The matrix must have full column rank. It is an `intent(inout)` argument.
+
+`b`: Shall be a rank-1 array of the same kind as `a`, containing the right-hand-side vector. It is an `intent(in)` argument.
+
+`prefactored_w` (optional): Shall be an input `logical` flag. If `.true.`, `w` is assumed to contain a matrix square root \( B \) such that \( W = B \cdot B^T \). This can be a Cholesky factor or any other valid square root (e.g., SVD-based). Default: `.false.`. This is an `intent(in)` argument.
+
```
loiseaujc, line 1004: You could mention here that, if the Cholesky factor is used, user need to have zeroed-out the other triangular part.

loiseaujc (commented):
Here are some more minor comments. Really close.

aamrindersingh:
Done @loiseaujc

doc/specs/stdlib_linalg.md
jvdp1: why asking that to the user, and not to do it internally?
If  asked to the user, should there be an internal check to ensure that it is really the case?

doc/specs/stdlib_linalg.md
```diff
@@ -976,6 +976,52 @@ call [stdlib_linalg(module):constrained_lstsq_space(interface)]`(a,  c,  lwork [
...
+`overwrite_a` (optional): Shall be an input `logical` flag. If `.true.`, input matrix `a` will be used as temporary storage and overwritten. This avoids internal data allocation. Default: `.false.`. This is an `intent(in)` argument.
+
+`overwrite_w` (optional): Shall be an input `logical` flag. If `.true.`, input matrix `w` will be used as temporary storage and overwritten. This avoids internal data allocation. Default: `.false.`. This is an `intent(in)` argument.
+
+`err` (optional): Shall be a `type(linalg_state_type)` value. This is an `intent(out)` argument.
+
+### Return value
+
+Returns an array value of the same kind as `a`, containing the generalized least-squares solution.
```
jvdp1:
```suggestion
Returns a rank-1 array of the same kind as `a`, containing the generalized least-squares solution.
```

src/linalg/stdlib_linalg.fypp
```diff
@@ -679,6 +680,45 @@ module stdlib_linalg
...
+    !!### Description
+    !!
+    !! This interface provides methods for computing generalized least-squares
+    !! with a symmetric (real) or Hermitian (complex) positive definite covariance matrix.
+    !! Supported data types include `real` and `complex`.
+    !!
+    !!@note The solution is based on LAPACK's `*GGGLM` routine.
+    !!
+    #:for rk,rt,ri in RC_KINDS_TYPES
+    module function stdlib_linalg_${ri}$_generalized_lstsq(w,a,b,prefactored_w,overwrite_a,overwrite_w,err) result(x)
```
jvdp1, line 702: Is there a corresponding subroutine? (similarly to `constrained_lstsq` and `solve_constrained_lstsq`)

src/linalg/stdlib_linalg_least_squares.fypp
```diff
@@ -564,4 +564,139 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+        !> [optional] Is W already a matrix square root (e.g., Cholesky factor)? Default: .false.
+        logical(lk), optional, intent(in) :: prefactored_w
+        !> [optional] Can A data be overwritten and destroyed?
+        logical(lk), optional, intent(in) :: overwrite_a
+        !> [optional] Can W data be overwritten and destroyed? Default: .false.
+        logical(lk), optional, intent(in) :: overwrite_w
+        !> [optional] state return flag. On error if not requested, the code will stop
+        type(linalg_state_type), optional, intent(out) :: err
+        !> Result array x[n]
+        ${rt}$, allocatable :: x(:)
```
jvdp1, line 589: Could we avoid to ask this one to be allocatable:

```suggestion
        ${rt}$ :: x(size(a, 2, kind=ilp))
```

src/linalg/stdlib_linalg_least_squares.fypp
```diff
@@ -564,4 +564,139 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+                call linalg_error_handling(err0, err)
+                return
+            end if
+        end if
+        ! If prefactored_w=.true., user provides a valid matrix square root B where W = B*B^T.
+        ! This can be a Cholesky factor OR any other valid square root (e.g., SVD-based).
+        ! We do not modify the user's input.
+
+        ! Prepare for GGGLM
+        allocate(d(m), source=b)
```
jvdp1:
```suggestion
        allocate(d, source=b)
```

jvdp1 (commented):
Overall it looks good to me. Here are a few suggestions.

doc/specs/stdlib_linalg.md
loiseaujc: The Cholesky factors are not the only matrix square-root one can use. If you use `svd` and take the square-root of the singular values, it'll be a valid square-root although it won't have the upper or lower triangular structure of the Cholesky factor. In order to check internally whether the upper (lower) triangular part has been zeroed-out by the user, we would need to know for sure that the pre-factored `W` matrix has been obtained with Cholesky. And even there, we would need to know whether `U` or `L` is being used.

src/linalg/stdlib_linalg_least_squares.fypp
loiseaujc: The interface for the function `lstsq` reads

```fortran
      module function stdlib_linalg_${ri}$_lstsq_${ndsuf}$(a,b,cond,overwrite_a,rank,err) result(x)
         !> Input matrix a[n,n]
         ${rt}$, intent(inout), target :: a(:,:)
         !> Right hand side vector or array, b[n] or b[n,nrhs]
         ${rt}$, intent(in) :: b${nd}$
         !> [optional] cutoff for rank evaluation: singular values s(i)<=cond*maxval(s) are considered 0.
         real(${rk}$), optional, intent(in) :: cond
         !> [optional] Can A,b data be overwritten and destroyed?
         logical(lk), optional, intent(in) :: overwrite_a
         !> [optional] Return rank of A
         integer(ilp), optional, intent(out) :: rank
         !> [optional] state return flag. On error if not requested, the code will stop
         type(linalg_state_type), optional, intent(out) :: err
         !> Result array/matrix x[n] or x[n,nrhs]
         ${rt}$, allocatable, target :: x${nd}$
      end function stdlib_linalg_${ri}$_lstsq_${ndsuf}$
```

having the return vector `x` as `allocatable`. The function form of the constrained least-square also. I guess it makes sense to have it `allocatable` for the function form, although it would need to be pre-allocated for the `subroutine` form.

doc/specs/stdlib_linalg.md
aamrindersingh: The flag exists for two reasons == performance (skip redundant O(m^3) Cholesky when reusing the same W across multiple solves) and flexibility (users can pass non-Cholesky square roots like SVD based U*diag(sqrt(lambda))*U^H)

Moreover,  an internal check would cost O(m^3) , defeating the purpose , and checked LAPACK, it itself follows the same trust model for POTRS which assumes you have already called POTRF

src/linalg/stdlib_linalg_least_squares.fypp
aamrindersingh: Done

src/linalg/stdlib_linalg.fypp
aamrindersingh: Added 👍

doc/specs/stdlib_linalg.md
jvdp1: OK. thank you for the explanation. Could we then add something like that to warn the user that is up to its responsability:
"It is the user's responsibility to ensure that the prefactored matrix B is valid."

example/linalg/example_generalized_lstsq.f90
```diff
@@ -0,0 +1,28 @@
...
+  A(:,1) = 1.0_dp
+  A(:,2) = [1.0_dp, 2.0_dp, 3.0_dp]
+
+  ! Observations
+  b = [1.0_dp, 2.1_dp, 2.9_dp]
+
+  ! Covariance matrix (correlated errors)
+  W(1,:) = [1.0_dp, 0.5_dp, 0.25_dp]
+  W(2,:) = [0.5_dp, 1.0_dp, 0.5_dp]
+  W(3,:) = [0.25_dp, 0.5_dp, 1.0_dp]
```
jvdp1:
```suggestion
  W = reshape([1.0_dp, 0.5_dp, 0.25_dp, 0.5_dp, 1.0_dp, 0.5_dp, 0.25_dp, 0.5_dp, 1.0_dp], [m, m])
```

example/linalg/example_generalized_lstsq.f90
```diff
@@ -0,0 +1,28 @@
+! Generalized least-squares solver with correlated errors
+program example_generalized_lstsq
+  use stdlib_linalg_constants, only: dp
+  use stdlib_linalg, only: generalized_lstsq
+  implicit none
+
+  real(dp) :: A(3,2), b(3), W(3,3)
```
jvdp1:
```suggestion
  integer, parameter ::  m=3
  real(dp) :: A(m,2), b(m), W(m,m)
```

example/linalg/example_solve_generalized_lstsq.f90
```diff
@@ -0,0 +1,27 @@
...
+  ! Observations
+  b = [1.0_dp, 2.1_dp, 2.9_dp]
+
+  ! Covariance matrix (correlated errors)
+  W(1,:) = [1.0_dp, 0.5_dp, 0.25_dp]
+  W(2,:) = [0.5_dp, 1.0_dp, 0.5_dp]
+  W(3,:) = [0.25_dp, 0.5_dp, 1.0_dp]
+
+  ! Solve generalized least-squares
+  call solve_generalized_lstsq(W, A, b, x)
```
jvdp1: could this be combined in a single example file?

src/linalg/stdlib_linalg.fypp
```diff
@@ -40,6 +40,8 @@ module stdlib_linalg
   public :: lstsq_space
   public :: constrained_lstsq
   public :: constrained_lstsq_space
+  public :: generalized_lstsq
+  public :: solve_generalized_lstsq
```
jvdp1: could this be moved to `solve_constrainted_lstsq`?

src/linalg/stdlib_linalg.fypp
```diff
@@ -679,6 +681,84 @@ module stdlib_linalg
...
+        ${rt}$, intent(inout), target :: a(:,:)
+        !> Right hand side vector b[m]
+        ${rt}$, intent(in) :: b(:)
+        !> [optional] Is W already a matrix square root (e.g., Cholesky factor)? Default: .false.
+        logical(lk), optional, intent(in) :: prefactored_w
+        !> [optional] Can A data be overwritten and destroyed?
+        logical(lk), optional, intent(in) :: overwrite_a
+        !> [optional] Can W data be overwritten and destroyed? Default: .false.
+        logical(lk), optional, intent(in) :: overwrite_w
+        !> [optional] state return flag. On error if not requested, the code will stop
```
jvdp1:
```suggestion
        !> [optional] state return flag. On error if not requested, the code will trigger an error stop
```

src/linalg/stdlib_linalg.fypp
```diff
@@ -679,6 +681,84 @@ module stdlib_linalg
...
+    !!
+    !! This interface provides methods for computing generalized least-squares
+    !! with a symmetric (real) or Hermitian (complex) positive definite covariance matrix.
+    !! Supported data types include `real` and `complex`.
+    !!
+    !!@note The solution is based on LAPACK's `*GGGLM` routine.
+    !!
+    #:for rk,rt,ri in RC_KINDS_TYPES
+    module subroutine stdlib_linalg_${ri}$_solve_generalized_lstsq(w,a,b,x,prefactored_w,overwrite_a,overwrite_w,err)
+        !> Covariance matrix W[m,m] (symmetric/Hermitian positive definite) or its matrix square root
```
jvdp1: Use the Fortran notation (here and elsewhere)
```suggestion
        !> Covariance matrix W(m,m) (symmetric/Hermitian positive definite) or its matrix square root
```

src/linalg/stdlib_linalg.fypp
```diff
@@ -679,6 +681,84 @@ module stdlib_linalg
...
+        ${rt}$, intent(inout), target :: w(:,:)
+        !> Input matrix a[m,n]
```
jvdp1:
```suggestion
        !> Input matrix a(m,n)
```

src/linalg/stdlib_linalg.fypp
```diff
@@ -679,6 +681,84 @@ module stdlib_linalg
...
+        ${rt}$, intent(inout), target :: a(:,:)
+        !> Right hand side vector b[m]
+        ${rt}$, intent(in) :: b(:)
+        !> Solution vector x[n]
+        ${rt}$, intent(out) :: x(:)
+        !> [optional] Is W already a matrix square root (e.g., Cholesky factor)? Default: .false.
+        logical(lk), optional, intent(in) :: prefactored_w
+        !> [optional] Can A data be overwritten and destroyed?
```
jvdp1, line 752:
```suggestion
        !> (optional) Can A data be overwritten and destroyed?
```

src/linalg/stdlib_linalg_least_squares.fypp
```diff
@@ -564,4 +564,165 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+    !-------------------------------------------------------------
+    !-----     Generalized Least-Squares Solver              -----
+    !-------------------------------------------------------------
+
+    #:for rk,rt,ri in RC_KINDS_TYPES
+    ! Generalized least-squares function: thin wrapper that allocates x and delegates to subroutine
+    module function stdlib_linalg_${ri}$_generalized_lstsq(w,a,b,prefactored_w,overwrite_a,overwrite_w,err) result(x)
+        !> Covariance matrix W[m,m] (symmetric/Hermitian positive definite) or its matrix square root
+        ${rt}$, intent(inout), target :: w(:,:)
+        !> Input matrix a[m,n]
```
jvdp1:
```suggestion
        !> Input matrix a(m,n)
```

src/linalg/stdlib_linalg_least_squares.fypp
```diff
@@ -564,4 +564,165 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+        !> [optional] Is W already a matrix square root (e.g., Cholesky factor)? Default: .false.
+        logical(lk), optional, intent(in) :: prefactored_w
+        !> [optional] Can A data be overwritten and destroyed?
+        logical(lk), optional, intent(in) :: overwrite_a
+        !> [optional] Can W data be overwritten and destroyed? Default: .false.
+        logical(lk), optional, intent(in) :: overwrite_w
+        !> [optional] state return flag. On error if not requested, the code will stop
+        type(linalg_state_type), optional, intent(out) :: err
+        !> Result array x[n]
+        ${rt}$, allocatable, target :: x(:)
```
jvdp1:
```suggestion
        ${rt}$, allocatable :: x(:)
```

src/linalg/stdlib_linalg_least_squares.fypp
```diff
@@ -564,4 +564,165 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+
+        ! Validate sizes
+        if (size(w, 1, kind=ilp) /= m .or. size(w, 2, kind=ilp) /= m) then
+            err0 = linalg_state_type(this, LINALG_VALUE_ERROR, &
+                   'Covariance matrix must be square m×m:', shape(w, kind=ilp))
+            call linalg_error_handling(err0, err)
+            return
+        end if
+
+        if (size(b, kind=ilp) /= m) then
```
jvdp1, line 649: A check on the size of `x` should be added too

src/linalg/stdlib_linalg_least_squares.fypp
```diff
@@ -564,4 +564,165 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+
+        ! Workspace query
+        allocate(work(1))
+        call ggglm(m, n, p, amat, lda, lmat, ldb, d, x, y, work, -1_ilp, info)
+        lwork = ceiling(real(work(1), kind=${rk}$), kind=ilp)
+        deallocate(work)
+        allocate(work(lwork))
+
+        ! Solve GLS via GGGLM
+        call ggglm(m, n, p, amat, lda, lmat, ldb, d, x, y, work, lwork, info)
```
jvdp1, line 719:
```suggestion

        ! Workspace query
        allocate(work(1))
        call ggglm(m, n, p, amat, m, lmat, m, d, x, y, work, -1_ilp, info)
        lwork = ceiling(real(work(1), kind=${rk}$), kind=ilp)
        deallocate(work)
        allocate(work(lwork))

        ! Solve GLS via GGGLM
        call ggglm(m, n, p, amat, m, lmat, m, d, x, y, work, lwork, info)
```

test/linalg/test_linalg_lstsq.fypp
```diff
@@ -139,6 +145,206 @@ module test_linalg_least_squares
...
+        A(:,1) = 1.0_${rk}$
+        A(:,2) = [1.0_${rk}$, 2.0_${rk}$, 3.0_${rk}$]
+
+        ! Observations
+        b = [1.0_${rk}$, 2.1_${rk}$, 2.9_${rk}$]
+
+        ! SPD covariance matrix (correlated errors)
+        W(1,:) = [2.0_${rk}$, 1.0_${rk}$, 0.5_${rk}$]
+        W(2,:) = [1.0_${rk}$, 2.0_${rk}$, 1.0_${rk}$]
+        W(3,:) = [0.5_${rk}$, 1.0_${rk}$, 2.0_${rk}$]
```
jvdp1: Use `reshape`

jvdp1 (commented):
Thank you @aamrindersingh . Here are some additional comments. It is taking a good shape!

doc/specs/stdlib_linalg.md
loiseaujc: Sure enough !

loiseaujc (commented):
It looks pretty good to me. Following @jvdp1 comment, just try to be consistent between `[optional]` and `(optional)` in the documentation (the parentheses form being preferred I guess).

src/linalg/stdlib_linalg_least_squares.fypp
```diff
@@ -564,4 +564,172 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+        !> [optional] state return flag. On error if not requested, the code will stop
+        type(linalg_state_type), optional, intent(out) :: err
+
+        ! Local variables
+        type(linalg_state_type) :: err0
+        integer(ilp) :: m, n, p, lda, ldb, lwork, info
+        logical(lk) :: copy_a, copy_w, is_prefactored
+        ${rt}$, pointer :: amat(:,:), lmat(:,:)
+        ${rt}$, allocatable, target :: amat_alloc(:,:), lmat_alloc(:,:)
+        ${rt}$, allocatable :: d(:), y(:), work(:)
```
perazz, line 626: Two work allocations will be too expensive. let's use a static variable for work size query
```suggestion
        ${rt}$, allocatable :: d(:), y(:), work(:)
        ${rt}$ :: work_size(1)
```

src/linalg/stdlib_linalg_least_squares.fypp
perazz: size check is on line 662, a bit below this

src/linalg/stdlib_linalg_least_squares.fypp
```diff
@@ -564,4 +564,172 @@ submodule (stdlib_linalg) stdlib_linalg_least_squares
...
+
+        lda = m
+        ldb = m
+
+        ! Workspace query
+        allocate(work(1))
+        call ggglm(m, n, p, amat, lda, lmat, ldb, d, x, y, work, -1_ilp, info)
+        lwork = ceiling(real(work(1), kind=${rk}$), kind=ilp)
+        deallocate(work)
+        allocate(work(lwork))
```
perazz: Avoid double allocation here
```suggestion
        call ggglm(m, n, p, amat, lda, lmat, ldb, d, x, y, work_size, -1_ilp, info)
        lwork = ceiling(real(work_size(1), kind=${rk}$), kind=ilp)
        allocate(work(lwork))
```

src/linalg/stdlib_linalg_least_squares.fypp
perazz: I added static work size variable

perazz (changes requested):
Couple comments to remove an unnecessary allocation. lgtm otherwise

loiseaujc:
@aamrindersingh : Are still planning to wrap up this PR or should we take over ?

perazz:
@copilot resolve the merge conflicts in this pull request. 
@loiseaujc applied them manually - hopefully they don't break anything!

PR #1106, opened 2026-01-25
fix (linalg_iterative): remove pure attribute from stdlib_linop inner_product 

jalvesz:
The `pure` attribute for the `inner_product` type bound procedure makes it impossible to embed mpi calls such as `mpi_allreduce` required for solving in a spmd famework.


doc/specs/stdlib_linalg_iterative_solvers.md
```diff
@@ -57,7 +57,7 @@ Proxy procedure for the `dot_product`.
 
 ###### Class
 
-Pure function
+function
```
jvdp1:
```suggestion
Function
```

PR #1110, opened 2026-01-30
Reduce occurrences of intrinsic stdlib_matmul test failures

loiseaujc:
Following the discussion in #1086 about the intrinsic test failure, this PR implements the following changes to the `test_matmul` (intrinsic module) :
- All matrices used in the test have now been normalized to have unit-norm (1-norm).
- The error check has been replaced from element-wise checks to 1-norm norm of the difference.

As far as I can recall, the error estimates are basically correct, although sharper ones could be used but I haven't been able to find my old lecture notes yet about matrix-matrix multiplications and floating point error analysis. In the mean time, this should (hopefully) reduce the number of times the CI fails on these tests. Something similar could be done with the `pseudoinverse` tests which do fail quite regularly (the single precision mostly I believe).

ping: @perazz, @jvdp1, @jalvesz

### DIFF
--- a/test/intrinsics/test_intrinsics.fypp
+++ b/test/intrinsics/test_intrinsics.fypp
@@ -292,14 +292,14 @@ subroutine test_matmul(error)
         call random_number(z)
 
         ! Normalize matrices.
-        x = x / mnorm(x, "fro")
-        y = y / mnorm(y, "fro")
-        z = z / mnorm(z, "fro")
+        x = x / mnorm(x, 1)
+        y = y / mnorm(y, 1)
+        z = z / mnorm(z, 1)
 
         r = stdlib_matmul(x, y, z) ! the optimal ordering would be (x(yz))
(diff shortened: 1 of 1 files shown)


loiseaujc:
> From what I understand, we're switching from a max absolute value to an average column absolute value, looks perfectly reasonable to me. LGTM @loiseaujc

Indeed. Basically, for the product of two matrices, it is relatively to show that

$$
\vert\vert AB - \mathrm{fl}(AB) \vert\vert_1 \leq n \cdot \varepsilon \cdot \vert\vert A \vert\vert_1 \cdot \vert\vert B \vert\vert_1
$$

where $\varepsilon$ is the floating-point unit and $\mathrm{fl}( \cdot )$ denotes the actual floating-point operation. The maximum of the absolute value of the difference is not a matrix norm, and I can't say anything about how it should behave in general.

PR #1111, opened 2026-02-02
fix (sparse): ensure proper assembly when filling in symmetric matrices with blocks

jalvesz:
This PR fixes the `add_block` type-bound-procedure for assembling sparse matrices in the cases the matrix is symmetric.
A check on the storage attribute combined with the current row and col enables deciding to skip or not the (row,col) pair for assembly.

### DIFF
--- a/test/linalg/test_linalg_sparse.fypp
+++ b/test/linalg/test_linalg_sparse.fypp
@@ -484,6 +484,7 @@ contains
 
         end block
         #:endfor
+        #:endfor
 
     end subroutine
 
@@ -574,8 +575,9 @@ contains
             y = 0._wp
             call spmv( CSC_low, x, y ) 
             call check(error, all(y == y_ref) , "error in low CSC ${s1}$ spmv" )
-        #:endfor
         end block
+        #:endfor
+        
     end subroutine
     
 end module


test/linalg/test_linalg_sparse.fypp
```diff
@@ -368,7 +369,7 @@ contains
 
             call check(error, all(CSR%data == COO%data) )
             if (allocated(error)) return
-
+
```
jvdp1:
```suggestion

```

test/linalg/test_linalg_sparse.fypp
```diff
@@ -383,6 +384,98 @@ contains
         #:endfor
     end subroutine
 
+    subroutine test_add_block_symmetric_skip(error)
+        !> Error handling
+        type(error_type), allocatable, intent(out) :: error
+        #:for k1, t1, s1 in (KINDS_TYPES)
+        block
```
jvdp1, line 495: Since all meesages below are the same, I suggest to move the `fypp` loop around the subroutine, instead of around the block. It will be easier track which test fails.
Or the type should be added in the error messages.

jvdp1 (approved):
Overall LGTM. I have a minor suggestion

test/linalg/test_linalg_sparse.fypp
jalvesz: kind suffix added to the error message here https://github.com/fortran-lang/stdlib/pull/1111/commits/c20e9d018132756d5754a4990891dcf8d8d860b6

src/sparse/stdlib_sparse_kinds.fypp
```diff
@@ -296,6 +296,11 @@ contains
     ! data accessors
     !==================================================================
 
+    logical(int8) elemental function skip(sym,row,col)
```
perazz: Please use the `stdlib_kinds, only: lk` logical kind

perazz (commented):
Thank you for this PR. I also second @jvdp1's observation that there should be one routine per kind, rather than one scope per kind. LGTM otherwise.

src/sparse/stdlib_sparse_kinds.fypp
jalvesz: I use the smallest logical kind on purpose for efficiency.

jvdp1: Could it be defined in `stdlib_kinds`, mainy as `lk_int8`? Or could `c_bool` be used (note that I don't like this idea)?

jalvesz: Options to define a `lk_int8` parameter:
```Fortran
integer, parameter :: lk_int8 = 1 
integer, parameter :: lk_int8 = int8
integer, parameter :: lk_int8 = c_bool 
integer, parameter :: lk_int8 = selected_logical_kind(1) !> only available from the F2023 standard (too new)
```
The second one could be a good option.

perazz: These tiny edits are meant for consistency with the rest of the library. Performance-wise, I believe in most cases (-O2 at least) the private function will be inlined. Regarding the `kind`, although they may work, both `1` and `int8` refer to `integers`. On the other hand, `c_bool` is already part of the `stdlib_kinds`

jalvesz: It is not only a style edit, it does have consequences, even if they might not seem very important. `logical`s in Fortran can have different sizes, just like `integer`s, which mean that one does ask the compiler to use different register sizes depending on the requested size. Since the default Fortran size for logical is 32bits (8bytes) to correspond to the same size of the default `integer`, one does end up either implying usage of unnecessary extra memory or a couple of additional operations w.r.t the resulting ASM code.

I played a bit with this on CE https://godbolt.org/z/7aK8KvWrn, I don't want to overblow this thread so I won't put all the details but if one goes and play within, changing the compiler, it is possible to see ever so slightly differences between the ASM code for these two versions (I tried with gfortran, ifort and ifx with -O3)
```Fortran
logical(int8) elemental function skip_1(sym,row,col)
    integer(ilp), intent(in) :: sym, row, col
    skip_1 = (sym == sparse_lower .and. row < col) .or. (sym == sparse_upper .and. row > col)
end function

logical(int32) elemental function skip_2(sym,row,col)
    integer(ilp), intent(in) :: sym, row, col
    skip_2 = (sym == sparse_lower .and. row < col) .or. (sym == sparse_upper .and. row > col)
end function
```
Now, this can be anecdotal, even transparent, depending on the use case. But just as when defining `integer`s one voluntarily sets a storage size and the compiler "respects" one whishes, it does happens with Fortran `logical`s.

Maybe what we are missing is to add in the kinds module, not only one `lk` parameter. But maybe a couple of options. I agree that `int#` is for integers (even if in practical terms the values are the same). So maybe `lk8`, `lk16`, `lk32`, `lk64`, and a `lk` with the default pointing to `lk32`.

jalvesz: https://github.com/fortran-lang/stdlib/pull/1111/commits/86967982c67e43621ac68b7b5398c90d198a57e4 introduces the basic logical kind parameters (`lk8`, `lk16`, `lk32`, `lk64`) as fallbacks for [selected_logical_kinds](https://gcc.gnu.org/onlinedocs/gcc-15.1.0/gfortran/SELECTED_005fLOGICAL_005fKIND.html) which is too new (F2023) to be used effectively by in-use compiler versions.

perazz: I don't think we should introduce new “logical kind” parameters like `1, 2, 4, 8` (non-portable) or using logical(int8) (for integers) here. Pre-F2023 there is no standard, portable way to request an 8-bit LOGICAL kind (except `c_bool`). Reusing `int8` / integer literal kind values as `LOGICAL` kind parameters is not something stdlib should rely on across compilers.

For this PR, please keep it simple and portable: `use stdlib_kinds, only: lk` (or plain default `logical`) for skip, and keep the change scoped to the sparse fix + tests. If we want to support multiple logical kinds in `stdlib_kinds`, that should definitely be a separate design discussion/issue/PR with a clear portability story and compiler coverage.

jalvesz: I'll keep it with `c_bool` which is standard-portable for reducing storage footprint, because the Fortran standard defines this kind specifically for interoperability with C’s `_Bool`, which has a defined size and representation. This avoids non-portable bit assumptions while providing a known, minimal footprint for logical values.

jvdp1: `lk` or `c_bool` are fine for me. 

Just a question that will show my ignorance. When `skip` is declared as `c_bool`, what happens with the following:
```fortran
        skip = (sym == sparse_lower .and. row < col) .or. (sym == sparse_upper .and. row > col)
```
does the compiler directly return a `c_bool`, or will it first return a default `logical`, and then convert it to `c_bool?

jvdp1: No simple answer, as it seems to depend on the compiler and optimization levels. A general answer would be that with no optimization (-O0), there is indeed a copy from `lk` to `c_bool`, while such copies do no appear with the highest optimization level (-O3)

jalvesz: Indeed, that's the key issue, the standard leaves the Fortran developer to it's own because it is purposefully vague in this respect. So each compiler is free to do as it can/wants. From -O2/-O3 it normally behaves as expected.

jvdp1:
@jalvesz can this PR be merged?

PR #1119, opened 2026-02-13
Add stdlib_spatial module with Kabsch–Umeyama vector alignment algorithm

### LINKED ISSUE #1051: kabsh umeyama algorithm for vectors aligment
### Motivation

I would like to propose to add the the Kabsh-Umeyama algorithm (https://en.wikipedia.org/wiki/Kabsch_algorithm) in a (yet to be) `stdlib_spatial` module for vector alignment purposes.

A first API implementation combining several approaches including the use of weights:

```Fortran
module test
    use stdlib_kinds, only: dp
    use stdlib_intrinsics, only: stdlib_sum
    use stdlib_linalg, only: svd, det, eye, trace

contains

subroutine kabsh_umeyama( R, t, c, lrms, d, N, P, Q, W )
    integer, parameter :: wp = dp
    integer, intent(in) :: d, N
    real(wp), intent(out):: R(d,d) !> Rotation matrix
    real(wp), intent(out):: t(d)     !> Translation vector
    real(wp), intent(out):: c         !> Scaling factor
    real(wp), intent(out):: lrms    !> least root mean square
    real(wp), intent(in) :: P(d,N)  !> Reference point cloud
    real(wp), intent(in) :: Q(d,N) !> Target point cloud
    real(wp), intent(in), optional :: W(N) !> Weights
    
    ! -- Internal Variables
    integer :: i, j, point
    real(wp) :: centroid_P(d), centroid_Q(d)
    real(wp) :: covariance(d,d), S(d), U(d,d), Vt(d,d), B(d,d)
    real(wp) :: vec(d), variance_p, vp, vq, isum_w
    !---------------------------------------
    ! Compute centroids
    isum_w = 1._wp / N
    if(present(W)) isum_w = 1._wp / stdlib_sum(W)
    
    centroid_P = 0._wp; centroid_Q = 0._wp
    if(present(W))then
        do concurrent( point = 1:N )
            centroid_P(1:d) = centroid_P(1:d) + P(1:d,point)*W(point)
            centroid_Q(1:d) = centroid_Q(1:d) + Q(1:d,point)*W(point)
        end do
    else
        do concurrent( point = 1:N )
            centroid_P(1:d) = centroid_P(1:d) + P(1:d,point)
            centroid_Q(1:d) = centroid_Q(1:d) + Q(1:d,point)
        end do
    end if
    centroid_P = centroid_P * isum_w
    centroid_Q = centroid_Q * isum_w

    ! Compute the covariance matrix
    covariance = 0._wp; variance_p = 0._wp
    if(present(W))then
        do concurrent( point = 1:N, j = 1:d, i = 1:d )
            vp = ( P(i,point)-centroid_P(i) ) * W(point)
            vq = Q(j,point)-centroid_Q(j)
            covariance(i,j) = covariance(i,j) + vp*vq
            variance_p = variance_p + vp**2
        end do
    else
        do concurrent( point = 1:N, j = 1:d, i = 1:d )
            vp = P(i,point)-centroid_P(i)
            vq = Q(j,point)-centroid_Q(j)
            covariance(i,j) = covariance(i,j) + vp*vq
            variance_p = variance_p + vp**2
        end do
    end if
    covariance = covariance * isum_w
    variance_p = variance_p * isum_w / d

    ! SVD
    call svd(covariance, S, U, Vt )
    
    ! Optimal rotation
    R = matmul( U , Vt )
    
    ! Validate right-handed coordinate system
    B = eye(d,d)
    B(d,d) = sign( 1._wp, det(R))

    R = matmul( U , matmul( B , Vt ) )
    
    c = variance_p / (sum(S(1:d-1))+B(d,d)*S(d))
    
    ! Optimal translation
    t = centroid_P - c * matmul(R , centroid_Q )
    
    ! RMSD
    lrms = 0._wp
    do concurrent( point = 1:N )
        vec(1:d) = t(1:d) + c * matmul(R , Q(1:d,point) )
        vec(1:d) = vec(1:d) - P(1:d,point)
        lrms  = lrms  + dot_product(vec,vec)
    end do
    lrms  = sqrt( rmsd * isum_w )

end subroutine

end module

program main
use test

integer, parameter :: wp = 8
integer, parameter :: d = 2
real(wp), allocatable :: P(:,:)
real(wp), allocatable :: Q(:,:)
real(wp) :: R(d,d), t(d), vec(d), c, lrms 
integer :: i, N
N = 7
P = reshape( &
    [23._wp, 178._wp,&
     66._wp, 173._wp,&
     88._wp, 187._wp,&
    119._wp, 202._wp,&
    122._wp, 229._wp,&
    170._wp, 232._wp,&
    179._wp, 199._wp] , [d,N] )

Q = reshape( & 
    [232._wp, 38._wp,&
     208._wp, 32._wp,&
     181._wp, 31._wp,&
     155._wp, 45._wp,&
     142._wp, 33._wp,&
     121._wp, 59._wp,&
     139._wp, 69._wp] , [d,N] )

call kabsh_umeyama(R, t, c, lrms, d, N, P, Q)
print *, 'RMSD        :', lrms 
print *, 'Translation :', t
print *, 'Scaling     :', c
print *, 'Rotation    :'
do i = 1, d
    print *, R(i,:)
end do

end program
```
gives:
```shell
 LRMS        :   16.242818369005477     
 Translation :   271.33459510449791        396.07800316838268     
 Scaling     :   1.4616613091002035     
 Rotation    :
 -0.81034281019830057       0.58595608193782001     
 -0.58595608193782023      -0.81034281019830057     
```

### Prior Art

* "Aligning point patterns with Kabsch–Umeyama algorithm": https://zpl.fi/aligning-point-patterns-with-kabsch-umeyama-algorithm/?utm_source=chatgpt.com
* SciPy: https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.align_vectors.html
* Implementations in (Numpy, Torch, Jax): https://hunterheidenreich.com/posts/kabsch-algorithm/
* MATLAB : https://www.mathworks.com/matlabcentral/fileexchange/25746-kabsch-algorithm?utm_source=chatgpt.com
* Eigen (C++) : https://libeigen.gitlab.io/eigen/docs-nightly/Umeyama_8h_source.html?utm_source=chatgpt.com
* R : https://rdrr.io/cran/pracma/man/procrustes.html?utm_source=chatgpt.com

### Additional Information

_No response_

Mahmood-Sinan:
This PR introduces a new `stdlib_spatial` module and implements Kabsch-Umeyama algorithm for vector alignment which calculates the optimal transformation between two given sets of points P and Q as:
P ≈ c R Q + t
where:
- `R` is an optimal rotation matrix
- `c` is an optional uniform scaling factor
- `t` is a translation vector
This algorithm minimizes the the root-mean-square deviation(`rmsd`) between the two given sets of points.

The algorithm:

- Computes centroids and covariance matrix
- Uses SVD to obtain the optimal rotation
- Supports optional scaling
- Returns RMSD
- Includes randomized tests

Fixes https://github.com/fortran-lang/stdlib/issues/1051
I am opening this as a draft to gain feedback and suggestions.
Thanks to @jalvesz for the idea and for the guidance during the discussion.


src/spatial/stdlib_spatial_kabsch_umeyama.fypp
```diff
@@ -0,0 +1,173 @@
...
+                #:if t.startswith('complex')
+                R(i,j) = stdlib_dot_product_kahan(conjg(U(i,:)), Vt(:, j))
+                #:else
+                R(i,j) = stdlib_dot_product_kahan(U(i,:), Vt(:, j))
+                #:endif
+            end do
+        end do
+
+        ! Scaling factor
+        c = variance_p / (sum(S(1:d)))
```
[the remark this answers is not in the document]
Mahmood-Sinan: The implementation follows the covariance convention `H = (P - c_P) * (Q - c_Q)^T` , under which the scale naturally depends on var(P) not var(Q)

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
```diff
@@ -0,0 +1,173 @@
...
+        ${t}$, intent(out) :: t(:)
+        !> Scale factor
+        ${t}$, intent(out) :: c
+        !> Root-mean-square deviation
+        real(${k}$), intent(out) :: rmsd
+        !> Optional weights
+        ${t}$, intent(in), optional :: W(:)
+        !> Enable scaling
+        logical, intent(in), optional :: scale
+
```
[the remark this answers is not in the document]
Mahmood-Sinan: Updated.

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
```diff
@@ -0,0 +1,173 @@
...
+        ! Optimal rotation matrix.
+        do i = 1,d
+            do j = 1,d
+                #:if t.startswith('complex')
+                R(i,j) = stdlib_dot_product_kahan(conjg(U(i,:)), Vt(:, j))
+                #:else
+                R(i,j) = stdlib_dot_product_kahan(U(i,:), Vt(:, j))
+                #:endif
+            end do
+        end do
```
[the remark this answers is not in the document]
Mahmood-Sinan: Updated

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
```diff
@@ -0,0 +1,201 @@
...
+            call error_stop("array sizes do not match")
+        end if
+        if (present(W)) then
+            if (size(W) /= size(P,dim=2)) then
+                call error_stop("array sizes do not match")
+            end if
+        end if
+        d = size(P,dim=1)
+        N = size(P,dim=2)
+        scale_ = .true.
```
jalvesz: `optval` can be used to initialize this value. Also, optional parameters initialization are better placed at the begining of the procedure.

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
```diff
@@ -0,0 +1,201 @@
...
+        if(present(scale)) scale_ = scale
+
+        sum_w = one_${k}$ / N
+        if(present(W)) sum_w = one_${k}$ / stdlib_sum_kahan(W)
```
jalvesz: it would be better to compute `sum_w` in the following two steps:
```fortran
if(present(W)) then
    sum_w = stdlib_sum_kahan(W)
else
    sum_w = real( N , kind = ${k}$ )
end if
!> leave opportunity for future discussion on how to add spmd reduction needed here to reduce sum_w before division

sum_w = one_${k}$ / sum_w
```

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
```diff
@@ -0,0 +1,201 @@
...
+        ! Compute centroids of P and Q
+        if(present(W)) then
+            do i = 1, d
+                tmp_N(:) = W(:) * P(i,:)
+                c_P(i) = stdlib_sum_kahan(tmp_N)
+                tmp_N(:) = W(:) * Q(i,:)
+                c_Q(i) = stdlib_sum_kahan(tmp_N)
+            end do
+        else
+            do i = 1, d
```
jalvesz: `stdlib_sum_kahan` also accepts a `dim` parameter, so in this case the expresion can be simply:
```
c_P(1:d) = stdlib_sum_kahan(P, dim=2)
c_Q(1:d) = stdlib_sum_kahan(Q, dim=2)
```

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
```diff
@@ -0,0 +1,201 @@
...
+
+        ! SVD of covariance matrix H -> H = U * S * Vt
+        call svd(covariance, S, U, Vt)
+
+        ! Check for reflections in case of real entries.
+        #:if t.startswith('real')
+        reflect_ = det(matmul(U,Vt)) < zero_${s}$
+        #:endif
+
+        #:if t.startswith('real')
```
jalvesz: this fypp macro can be merged in the previous #:if

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
```diff
@@ -0,0 +1,201 @@
...
+        c_P = c_P * sum_w
+        c_Q = c_Q * sum_w
+
+        ! Compute covariance matrix H = (P - c_P) * (Q - c_Q)^T and variance of P
+        allocate(covariance(d,d), source=zero_${s}$)
+        allocate(tmp_d(d), source=zero_${s}$)
+        variance_p = zero_${k}$
+
+        if (present(W)) then
+            do point = 1, N
```
jalvesz, line 93: The loops in this section should be revisited. This algorithm is usually applied for N >> d, this means that the best approach in terms of memory access is to handle, for each point in `N` everything related to the topological dimension `d`. In that same line, `stdlib_sum` and `stdlib_dot_product` (and their kahan versions) are very interesting performance wise when handling large arrays in one go.

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
```diff
@@ -0,0 +1,201 @@
...
+                    covariance(i,j) = stdlib_sum_kahan(tmp_N)
+                    #:else
+                    tmp_N(:) = W(:) * (P(i,:) - c_P(i)) * (Q(j,:) - c_Q(j))
+                    covariance(i,j) = stdlib_sum_kahan(tmp_N)
+                    #:endif
+                end do
+            end do
+        else
+            ! Calculate variance by the formula (1/n)*sigma(P - c_P)^2
+            do point = 1, N
```
jalvesz, line 107: same as https://github.com/fortran-lang/stdlib/pull/1119/changes#r3075555666

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
```diff
@@ -0,0 +1,201 @@
...
+        #:if t.startswith('real')
+        reflect_ = det(matmul(U,Vt)) < zero_${s}$
+        #:endif
+
+        #:if t.startswith('real')
+        if(reflect_) Vt(d,:) = -Vt(d,:)
+        #:endif
+
+        ! Optimal rotation matrix.
+        do i = 1,d
```
jalvesz: Same comment as before. I'm worried that using `stdlib_dot_product_kahan` here will be detrimental performance wise in this loop construct. I think this could be better handled with blas `gemm`

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
```diff
@@ -0,0 +1,201 @@
...
+                c = variance_p / (sum(S(1:d-1)) - S(d))
+            else
+                c = variance_p / (sum(S(1:d)))
+            end if
+        #:else
+            c = variance_p / (sum(S(1:d)))
+        #:endif
+        else
+            c = one_${s}$
+        end if
```
jalvesz, line 152: a small suggestion to make this section easier to follow (includding aligning the fypp macro with the inner scope.
```fortran
        ! Scaling factor
        c = one_${s}$
        if(scale_) then
            #:if t.startswith('real')
            if(reflect_) then
                c = sum(S(1:d-1)) - S(d)
            else
                c = sum(S(1:d))
            end if
            #:else
            c = sum(S(1:d))
            #:endif
            c = variance_p / c
        end if
```

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
```diff
@@ -0,0 +1,201 @@
...
+
+        ! Translation vector t = c_P - c*R*c_Q
```
jalvesz, line 154: This operation might be better of being writtent using `gemv` (https://stdlib.fortran-lang.org/interface/gemv.html) with `alpha=-1` and `beta=1` (assuming `t` is first initialized to `c_P`

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
```diff
@@ -0,0 +1,201 @@
...
+            #:endif
+        end do
+
+        ! Compute RMSD
+        allocate(vec(d), source=zero_${s}$)
+        rmsd = zero_${k}$
+        rmsd_err = zero_${k}$
+        do point = 1, N
+            ! Calculate the k^th difference vector by the formula vec_k = c*R*Q_k + t - P_k
+            do i = 1, d
```
jalvesz: Same comment as before, `vec` can be computed using `gemv` instead

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
```diff
@@ -0,0 +1,201 @@
...
+                #:if t.startswith('complex')
+                vec(i) = c * stdlib_dot_product_kahan(conjg(R(i,1:d)), Q(1:d,point))
+                #:else
+                vec(i) = c * stdlib_dot_product_kahan(R(i,1:d), Q(1:d,point))
+                #:endif
+            end do
+            vec(1:d) = vec(1:d) + t(1:d) - P(1:d,point)
+            call kahan_kernel(real(stdlib_dot_product_kahan(vec,vec), kind=${k}$), rmsd, rmsd_err)
```
jalvesz: It is preferable to avoid passing expressions as arguments to procedures. This induces the creation of temporal variables. Better to declare the required variables, assign the expression and then pass the argument.

jalvesz (changes requested):

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
Mahmood-Sinan: I can reweite the covariance accumulation as a sum of outer products over the d-dimensional point vectors. However, in that formulation there is no natural dot-product reduction left for the covariance computation. It would be something like:
```fortran
do point = 1, N
    p(:) = P(:,point) - c_P(:)
    q(:) = Q(:,point) - c_Q(:)

    do j = 1, d
        do i = 1, d
            covariance(i,j) = covariance(i,j) + p(i) * q(j)
        end do
    end do
end do
```
Or one other way which might work will be using GEMM, but there we will have to calculate the whole centered matrices c_P and c_Q.
Let me know what you think

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
```diff
@@ -0,0 +1,194 @@
...
+        #:endif
+        logical :: scale_
+        #:if t.startswith('real')
+        logical :: reflect_
+        #:endif
+        real(${k}$) :: rmsd_err
+
+        scale_ = optval(scale, .true.)
+        ! Dimension checks
+        if(size(P,dim=1)/=size(Q,dim=1) .or. size(P,dim=1)/=size(R,dim=1) .or. size(P,dim=1)/=size(R,dim=2) &
```
loiseaujc: I guess you could condense the two `if` conditions into a single one as

```fortran
if (any(shape(P) /= shape(Q)) .or. any(shape(P) /= shape(R)) then
    call error_stop("array sizes do not match")
endif

File: src/spatial/stdlib_spatial_kabsch_umeyama.fypp [hunk 16]
```diff
@@ -0,0 +1,194 @@
...
+            sum_w = stdlib_sum_kahan(W)
+        else
+            sum_w = real(N, kind = ${k}$)
+        end if
+        !> leave opportunity for future discussion on how to add spmd reduction needed here to reduce sum_w before division
+
+        sum_w = one_${k}$ / sum_w
+        if(sum_w<zero_${k}$) call error_stop("Invalid weights: sum of weights must be positive")
+
+        allocate(c_P(d), source=zero_${s}$)
```
loiseaujc: All three lines could be condensed down to

```fortran
allocate(c_P(d), c_Q(d), tmp_N(N), source=zero${s}$)
```

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
```diff
@@ -0,0 +1,194 @@
...
+        allocate(tmp_N(N), source=zero_${s}$)
+
+        ! Compute centroids of P and Q
+        if(present(W)) then
+            do i = 1, d
+                tmp_N(:) = W(:) * P(i,:)
+                c_P(i) = stdlib_sum_kahan(tmp_N)
+                tmp_N(:) = W(:) * Q(i,:)
+                c_Q(i) = stdlib_sum_kahan(tmp_N)
+            end do
```
loiseaujc, line 79: Why do you need the `tmp_N` array here ?
Wouldn't

```fortran
do concurrent( i = 1:d )
    c_P(i) = stdlib_dot_product_kahan(W, P(i, :))
    c_Q(i) = stdlib_dot_product_kahan(W, Q(i, :))
enddo
```

be better?

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
```diff
@@ -0,0 +1,194 @@
...
+        allocate(tmp_d(d), source=zero_${s}$)
+        variance_p = zero_${k}$
+
+        if (present(W)) then
+            do point = 1, N
+                tmp_d = P(:, point) - c_P(:)
+                tmp_N(point) = stdlib_dot_product_kahan(tmp_d, tmp_d)
+            end do
+            tmp_N(:) = W(:) * tmp_N(:)
+            variance_p = stdlib_sum_kahan(tmp_N)
```
loiseaujc, line 104: I believe you can get rid of line 103 and simply have

```fortran
variance_p = stdlib_dot_product_kahan(W, tmp_N)
```

or would that break for complex values?

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
```diff
@@ -0,0 +1,194 @@
...
+            do j = 1, d
+                do i = 1, d
+                    #:if t.startswith('complex')
+                    tmp_N(:) = W(:) * (P(i,:) - c_P(i)) * conjg(Q(j,:) - c_Q(j))
+                    covariance(i,j) = stdlib_sum_kahan(tmp_N)
+                    #:else
+                    tmp_N(:) = W(:) * (P(i,:) - c_P(i)) * (Q(j,:) - c_Q(j))
+                    covariance(i,j) = stdlib_sum_kahan(tmp_N)
+                    #:endif
+                end do
```
loiseaujc: Again, you can probably get rid of `tmp_N` array here and directly have

```fortran
do concurrent( j=1:d, i=1:d, i <= j)
    covariance(i, j) = stdlib_dot_product_kahan( Q(j, :) - c_Q(j) , W * (P(i, :) - c_P(i)) )
    covariance(j, i) = covariance(i, j)
enddo
```

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
```diff
@@ -0,0 +1,194 @@
...
+                do i = 1, d
+                    #:if t.startswith('complex')
+                    tmp_N(:) = (P(i,:) - c_P(i)) * conjg(Q(j,:) - c_Q(j))
+                    covariance(i,j) = stdlib_sum_kahan(tmp_N)
+                    #:else
+                    tmp_N(:) = (P(i,:) - c_P(i)) * (Q(j,:) - c_Q(j))
+                    covariance(i,j) = stdlib_sum_kahan(tmp_N)
+                    #:endif
+                end do
+            end do
```
loiseaujc: Same as [here](https://github.com/fortran-lang/stdlib/pull/1119/changes#r3441064121) : you can get rid of the temporary `tpm_N` array and use directly `stdlib_dot_product_kahan` along with the `do concurrent`.

loiseaujc (changes requested):
I would have to deep dive into the actual algorithm to make sure everything is correct, but so far it looks pretty good to me. There are many little things though that could be improved for performance sake.

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
Mahmood-Sinan: For this algorithm, it is necessary for P and Q to have the same dimenstions, but not for R. Because the algorithm computes P ≈ c * R * Q + t. 
If P and Q is of dimension d\*N, R needs to be d\*d. So we cannot directly compare `shape(P)` with `shape(R)`. So I made a slight change to your suggestion and have improved this part.

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
Mahmood-Sinan: Updated

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
Mahmood-Sinan: The reason that I used a temp array here is because, the weights are real numbers for any kind of P and Q. So in case of complex points P and Q along with real weights W, we would have to pass that to  `stdlib_dot_product_kahan`. But `stdlib_dot_product_kahan` does not provide such a function which can accepts one real and one complex array.

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
Mahmood-Sinan: Same reason as https://github.com/fortran-lang/stdlib/pull/1119#discussion_r3469335559

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
Mahmood-Sinan: I have updated this. Also when I tried, the covariance matrix need not be symmtric, so I avoided symmetric condition and implemented the rest

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
Mahmood-Sinan: Updated

Mahmood-Sinan:
@loiseaujc Thank you for reviewing. I have made the changes. Please have a look.

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
jalvesz: The two inner loops could be replaced by the blas general rank-1 update `*ger` & `*gerc` 

https://github.com/fortran-lang/stdlib/blob/692632fc22a679a02c858945f9aa767d288b401f/src/blas/stdlib_linalg_blas.fypp#L467

https://github.com/fortran-lang/stdlib/blob/692632fc22a679a02c858945f9aa767d288b401f/src/blas/stdlib_linalg_blas.fypp#L503

Something like:
```fortran
do point = 1, N
    p(:) = P(:,point) - c_P(:)
    q(:) = Q(:,point) - c_Q(:)
    call ger(d, d, 1.0d0, p, 1, q, 1, covariance, d)
end do
```

Mahmood-Sinan: Thanks, I changed it

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
loiseaujc: Oh I see. In this case, wouldn't using the Fortran intrinsic `dot_product` be faster still as it'd bypass the temporary array completely anyway ?

Or could you define `W` to be a complex-valued array anyway ? If so, you could actually write `c_P = W.T @ P` and use `blas-2` matrix-vector product.

Mahmood-Sinan: Sorry for the late reply. `stdlib_dot_product_kahan` was used because the algorithm needed more accuracy while doing multiple addition and multiplications. I wanted to improve numerical accuracy by using compensated summation. I have tried defining `W` as a complex-valued array. But since weights are conceptually `real`, and also we have to check whether all the weights have zero imaginary value, I avoided it( I thought exposing weights as complex would be non meaningful). Another possibility would be to use a temporary `complex` weight array which has same values as user passed weights along with zero imaginary value part, then use this inside `stdlib_dot_product_kahan`.
Let me know what you think. @loiseaujc @jalvesz

src/spatial/stdlib_spatial_kabsch_umeyama.fypp
Mahmood-Sinan: This has been updated after introducing `gerc` and `ger`. Please have a look

PR #1122, opened 2026-02-15
Fix Division-by-Zero in Masked Statistical Functions

JAi-SATHVIK:
Masked statistical functions (mean, variance, correlation, and covariance) currently crash or return `NaN` when provided with an "all-false" mask. This is caused by a missing guard against zero counts in denominators specifically when `count(mask)` is 0 leading to critical division-by-zero errors during runtime.

### DIFF
--- a/src/stats/stdlib_stats_mean.fypp
+++ b/src/stats/stdlib_stats_mean.fypp
@@ -105,6 +105,7 @@ contains
         ${t1}$, intent(in) :: x${ranksuffix(rank)}$
         logical, intent(in) :: mask${ranksuffix(rank)}$
         ${t1}$ :: res
+        real(${k1}$) :: n
 
         n = real(count(mask, kind = int64), ${k1}$)
         if (n == 0._${k1}$) then
@@ -126,6 +127,7 @@ contains
         ${t1}$, intent(in) :: x${ranksuffix(rank)}$
         logical, intent(in) :: mask${ranksuffix(rank)}$
         real(dp) :: res
+        real(dp) :: n
 
         n = real(count(mask, kind = int64), dp)
         if (n == 0._dp) then


src/stats/stdlib_stats_corr.fypp
```diff
@@ -218,6 +218,11 @@ contains
               #:endif
                 mask_)
 
+              if (count(mask_) < 2) then
```
jalvesz: Instead of introducing another internal implicit loop (using `count`) it might be better to store the denominator in its own local variable and check whether the denominator is equal to zero or not to assign NaN. E.g:
```
denom = dot_product( centeri_, centeri_)*dot_product( centerj_, centerj_)
if (denom<epsilon(0._${k1}$)) then
     res(j, i) = ieee_value(1._${k1}$, ieee_quiet_nan)
else
     res(j, i) = dot_product( centerj_, centeri_)/sqrt(denom)
end if
```
same applies elsewhere

build_cmake/_deps/test-drive-src
jalvesz, line 1: could you please revert this change which seems to be a residual from a local commit ?

jalvesz (changes requested):

src/stats/stdlib_stats_corr.fypp
JAi-SATHVIK: Refactored the masked statistics implementations to improve performance by removing redundant passes over masks and data

**Correlation/Covariance:** Replaced the `count(mask) `pre-check with a direct check on the calculated denominator against `epsilon`, as suggested. This avoids an additional implicit loop and handles the `NaN `assignment more efficiently.
**Mean:** Optimized the masked mean routines by storing the `count `result in a local variable to prevent double-counting during the zero-element check and final division.

src/stats/stdlib_stats_corr.fypp
```diff
@@ -197,6 +197,7 @@ contains
       ${t1}$ :: centeri_(merge(size(x, 2), size(x, 1), mask = 1<dim))
       ${t1}$ :: centerj_(merge(size(x, 2), size(x, 1), mask = 1<dim))
       logical :: mask_(merge(size(x, 2), size(x, 1), mask = 1<dim))
+      real(${k1}$) :: denom
```
jvdp1, line 200:
```suggestion
      ${t1}$ :: denom
```

src/stats/stdlib_stats_corr.fypp
```diff
@@ -218,9 +219,12 @@ contains
               #:endif
                 mask_)
 
-              res(j, i) = dot_product( centerj_, centeri_)&
-               /sqrt(dot_product( centeri_, centeri_)*&
-                     dot_product( centerj_, centerj_))
+              denom = real(dot_product( centeri_, centeri_), ${k1}$)*real(dot_product( centerj_, centerj_), ${k1}$)
```
jvdp1, line 222:
```suggestion
              denom = dot_product( centeri_, centeri_)*dot_product( centerj_, centerj_)
```
Why do you convert the dot_product to a `real`?

src/stats/stdlib_stats_corr.fypp
```diff
@@ -218,9 +219,12 @@ contains
...
+              if (denom < epsilon(0._${k1}$)) then
+                res(j, i) = ieee_value(1._${k1}$, ieee_quiet_nan)
```
jvdp1, line 224: Is this ok with `complex` values?

src/stats/stdlib_stats_cov.fypp
```diff
@@ -181,28 +181,37 @@ contains
...
+      ${t1}$ :: mean_i, mean_j
 
       select case(dim)
         case(1)
           do i = 1, size(res, 2)
             do j = 1, size(res, 1)
              mask_ = mask(:, i) .and. mask(:, j)
-             centeri_ = merge( x(:, i) - mean(x(:, i), mask = mask_),&
+             n = count(mask_)
+             if (n < 2) then
```
jvdp1, line 192:
```suggestion
             if (n == 0) then
```
`n <2` does not match the explanation given for opening this PR. Should it be `n==0`?

src/stats/stdlib_stats_cov.fypp
```diff
@@ -181,28 +181,37 @@ contains
...
+               res(j, i) = ieee_value(1._${k1}$, ieee_quiet_nan)
+               cycle
+             end if
+
+             mean_i = sum(x(:, i), mask = mask_) / real(n, ${k1}$)
+             mean_j = sum(x(:, j), mask = mask_) / real(n, ${k1}$)
```
jvdp1, line 198: I suggest to use `stdlib` `mean` function.

src/stats/stdlib_stats_mean.fypp
```diff
@@ -105,8 +105,15 @@ contains
...
+        real(${k1}$) :: n
 
-        res = sum(x, mask) / real(count(mask, kind = int64), ${k1}$)
+        n = real(count(mask, kind = int64), ${k1}$)
+        if (n == 0._${k1}$) then
+          res = ieee_value(1._${k1}$, ieee_quiet_nan)
+          return
+        end if
+
+        res = sum(x, mask) / n
```
jvdp1, line 116: I am puzzle by this change as the original code will return NaN when count(mask)=0. So, what is the benefit of this change?

src/stats/stdlib_stats_mean.fypp
```diff
@@ -135,9 +149,24 @@ contains
...
+              res = sum(x, dim, mask) / n
+            end if
+          #:else
+            n = real(count(mask, dim), ${k1}$)
+            where (n > 0._${k1}$)
+              res = sum(x, dim, mask) / n
+            elsewhere
+              res = ieee_value(1._${k1}$, ieee_quiet_nan)
+            end where
+          #:endif
```
jvdp1, line 169: This code is more complex to undertand than the original one. So,what is its benefit?

src/stats/stdlib_stats_mean.fypp
```diff
@@ -135,9 +149,24 @@ contains
...
 
         if (dim >= 1 .and. dim <= ${rank}$) then
-          res = sum(x, dim, mask) / real(count(mask, dim), ${k1}$)
+          #:if rank == 1
+            n = real(count(mask), ${k1}$)
+            if (n == 0._${k1}$) then
+              res = ieee_value(1._${k1}$, ieee_quiet_nan)
+            else
+              res = sum(x, dim, mask) / n
+            end if
```
jvdp1, line 161: Why is this code duplicated for `dim==1`? the following one would do the job too.

jvdp1 (commented):
Thank you @JAi-SATHVIK for this PR. 
I am a bit puzzled by these changes. 

You mentioned the following reason to open this PR:

> Masked statistical functions (mean, variance, correlation, and covariance) currently crash or return NaN when provided with an "all-false" mask. 

It is expected that NaN is return with an all-false mask. Therefore your changes will result in the same output.
Furthermore, these scenarios are covered by the tests (or at least it should be)!

However, could you provide more information when it craches in this configuration?

> This is caused by a missing guard against zero counts in denominators specifically when count(mask) is 0 leading to critical division-by-zero errors during runtime.

Division-by-zero should return NaN is nothing is available in the numeretor or Inf if the numerator is different from 0. I believe that the code already handle this correctly.

PR #1118, opened 2026-02-13
build: fix absolute paths for external BLAS/LAPACK in pkg-config export

ijatinydv:
This PR addresses **Issue #1109**, preventing absolute filesystem paths to external BLAS/LAPACK libraries from being hardcoded into the generated fortran_stdlib.pc file.

### **Technical Changes**

1. **Target Interception**: Modified `config/export_pc.cmake` to add `BLAS::BLAS` and `LAPACK::LAPACK` to a skip list during recursive resolution.

2. **Vendor Discovery**: Implemented a probing mechanism using `pkg_check_modules` to identify known BLAS/LAPACK vendor package names (OpenBLAS, MKL, Netlib).

3. **Metadata Separation:** >     * Found packages are correctly added to `Requires.private`.

   - Added a `libs_to_linker_flags` fallback to convert absolute paths to portable `-l` flags (e.g., `-lopenblas`) if no `.pc` file is found.

4. **Template Update**: Updated `config/template.pc` to support `Requires.private` and `Libs.private` fields for better portability.

### **Verification Results (Windows 11, MinGW-W64)**

image

PR #1123, opened 2026-02-15
fix: prevent double-prefixing of absolute install paths in pkg-config

### LINKED ISSUE #1121: Fix double-prefixing of absolute install paths in pkg-config export
### Description

While finalizing #1118, I noticed that `libdir `and `includedir `are sometimes double-prefixed in the generated `.pc` file. This happens on distributions like Debian/Ubuntu where `GNUInstallDirs `provides absolute paths (e.g., `/usr/lib/x86_64-linux-gnu`). The result is a broken path like `libdir=${prefix}//usr/lib/x86_64-linux-gnu`.

### Expected Behaviour

The `pkg-config` variables should handle both relative and absolute paths correctly:

- **Relative**: Prepend `${prefix}/` (e.g., `libdir=${prefix}/lib`).

- **Absolute**: Use the path as-is (e.g., `libdir=/usr/lib/x86_64-linux-gnu`).

### Version of stdlib

8116e5c

### Platform and Architecture

Linux (Multiarch)

### Additional Information

I have a fix ready using `IS_ABSOLUTE `checks in `export_pc.cmake` and can submit a PR immediately if the approach is approved.

ijatinydv:
This PR fixes a bug in the generated `fortran_stdlib.pc` where `${prefix}/` was incorrectly prepended to installation paths that were already absolute (e.g., on multiarch Linux distributions).

**Changes**

- **Path Logic**: Implemented `IS_ABSOLUTE `checks in `config/export_pc.cmake` for `libdir`, `includedir`, and `moduledir`.

- **Template Update**: Refactored `config/template.pc` to use pre-computed `@PC_*@` variables, ensuring valid paths for both relative (relocatable) and absolute (system-wide) installations.

- **Extended Fix**: Identified and resolved the same double-prefixing issue in `moduledir `during local validation.

**Verification Results**
Validated across both standard and multiarch scenarios:

image

**Regression Testing**
Confirmed that BLAS/LAPACK dependency logic from #1118 is preserved in the `private `metadata fields.

Closes #1121

### DIFF
--- a/config/export_pc.cmake
+++ b/config/export_pc.cmake
@@ -183,6 +183,27 @@ endif()
 string(REPLACE ";" " " PC_REQUIRES_PRIVATE "${PC_REQUIRES_PRIVATE}")
 string(REPLACE ";" " " PC_LIBS_PRIVATE "${PC_LIBS_PRIVATE}")
 
+# Handle absolute vs relative install paths for libdir
+if(IS_ABSOLUTE "${CMAKE_INSTALL_LIBDIR}")
+    set(PC_LIBDIR "${CMAKE_INSTALL_LIBDIR}")
+else()
+    set(PC_LIBDIR "\${prefix}/${CMAKE_INSTALL_LIBDIR}")
+endif()
+
+# Handle absolute vs relative install paths for includedir
+if(IS_ABSOLUTE "${CMAKE_INSTALL_INCLUDEDIR}")
+    set(PC_INCLUDEDIR "${CMAKE_INSTALL_INCLUDEDIR}")
+else()
+    set(PC_INCLUDEDIR "\${prefix}/${CMAKE_INSTALL_INCLUDEDIR}")
+endif()
+
+# Handle absolute vs relative install paths for moduledir
+if(IS_ABSOLUTE "${CMAKE_INSTALL_MODULEDIR}")
+    set(PC_MODULEDIR "${CMAKE_INSTALL_MODULEDIR}")
+else()
+    set(PC_MODULEDIR "\${prefix}/${CMAKE_INSTALL_MODULEDIR}")
+endif()
+
 configure_file(
   "${CMAKE_CURRENT_SOURCE_DIR}/config/template.pc"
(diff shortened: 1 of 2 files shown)


config/export_pc.cmake
```diff
@@ -183,6 +183,27 @@ endif()
...
+
+# Handle absolute vs relative install paths for includedir
+if(IS_ABSOLUTE "${CMAKE_INSTALL_INCLUDEDIR}")
+    set(PC_INCLUDEDIR "${CMAKE_INSTALL_INCLUDEDIR}")
+else()
+    set(PC_INCLUDEDIR "\${prefix}/${CMAKE_INSTALL_INCLUDEDIR}")
+endif()
+
+# Handle absolute vs relative install paths for moduledir
+if(IS_ABSOLUTE "${CMAKE_INSTALL_MODULEDIR}")
```
jvdp1, line 201: I read the following for `IS_ABSOLUTE` : "On non-Windows hosts, any path that begins with a tilde (~) evaluates to true."

So, does it work if the path starts with `/`?

ijatinydv: > I read the following for `IS_ABSOLUTE` : "On non-Windows hosts, any path that begins with a tilde (~) evaluates to true."
> 
> So, does it work if the path starts with `/`?

Yes, it does. On POSIX systems, any path starting with `/` is treated as absolute by `IS_ABSOLUTE`.

I verified this during local testing with `/usr/lib/x86_64-linux-gnu`, where it correctly identified the path as absolute and successfully prevented the double-prefixing bug. It works as intended for standard Linux directory structures.

jvdp1: After testing it too, I confirmed it. Thank you @ijatinydv . I will approve and merge this PR, as it doesn't impact the code.

jvdp1 (approved):
LGTM. Thank you @ijatinydv

ijatinydv:
> LGTM. Thank you @ijatinydv

Thank you for the review and for merging the PR, @jvdp1 Glad to have this resolved.

PR #1125, opened 2026-02-20
FIX: Early exit fix

JAi-SATHVIK:
**Bug Overview: The **"Early Exit"** Logic Error**

The `median `and `median_mask ` functions in `stdlib `contained a critical logic error affecting multi-dimensional arrays (Rank > 1). When processing an array in slices (e.g., calculating the median of every column in a matrix), the code would mistakenly treat an edge case in one slice as a signal to terminate the entire function.

**Issue :-**

If a user attempted to calculate the median of a matrix where one column was fully masked or contained a `NaN `value:

Column 1: Calculated correctly.
Column 2 (The Edge Case): Corrected to `NaN`.
Columns 3+: Skipped entirely.
The function would exit early, returning a result array where the remaining elements were never calculated, containing "garbage" values from uninitialized memory. This resulted in silent data corruption because the program would continue running with invalid statistical data.

**The Solution: Rank-Conditional Loop Control**

Scalar Results (Rank 1): If the input is a simple 1D array, the function still uses `return `because there is only one result to produce.
Array Results (Rank > 1): If the input has higher dimensions, the code now uses `cycle`. This tells the program: "This specific slice is invalid, mark it `NaN `and move to the next one."

### DIFF
--- a/src/sparse/stdlib_sparse_constants.fypp
+++ b/src/sparse/stdlib_sparse_constants.fypp
@@ -2,7 +2,7 @@
 #:set R_KINDS_TYPES = list(zip(REAL_KINDS, REAL_TYPES, REAL_SUFFIX))
 #:set C_KINDS_TYPES = list(zip(CMPLX_KINDS, CMPLX_TYPES, CMPLX_SUFFIX))
 module stdlib_sparse_constants
-    use stdlib_kinds, only: int8, int16, int32, int64, sp, dp, xdp, qp
+    use stdlib_kinds, only: int8, int16, int32, int64, sp, dp, xdp, qp, c_bool
     use stdlib_constants
     implicit none
     public
--- a/src/sparse/stdlib_sparse_kinds.fypp
+++ b/src/sparse/stdlib_sparse_kinds.fypp
@@ -330,12 +330,17 @@ contains
     ! data accessors
     !==================================================================
 
+    logical(c_bool) elemental function skip(sym,row,col)
+        integer(ilp), intent(in) :: sym, row, col
+        skip = (sym == sparse_lower .and. row < col) .or. (sym == sparse_upper .and. row > col)
+    end function
+
     #:for k1, t1, s1 in (KINDS_TYPES)
     pure ${t1}$ function at_value_coo_${s1}$(self,ik,jk) result(val)
         class(COO_${s1}$_type), intent(in) :: self
         integer(ilp), intent(in) :: ik, jk
         integer(ilp) :: k, ik_, jk_
-        logical :: transpose
+        logical(c_bool) :: transpose
         ! naive implementation
         if( (ik<1 .or. ik>self%nrows) .or. (jk<1 .or. jk>self%ncols) ) then
             val = ieee_value( 0._${k1}$ , ieee_quiet_nan)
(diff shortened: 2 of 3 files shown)


src/stats/stdlib_stats_median.fypp
```diff
@@ -100,7 +100,7 @@ contains
                   if (any(ieee_is_nan(x_tmp))) then
                     res${reduce_subvector('j', rank, fi)}$ = &
                       ieee_value(1._${o1}$, ieee_quiet_nan)
-                    #:if fi == 1
+                    #:if rank == 1
```
jvdp1, line 103:
```suggestion
                    #:if fi == 1
```

Is this needed?

src/stats/stdlib_stats_median.fypp
```diff
@@ -234,7 +234,11 @@ contains
                 if (n == 0) then
                     res${reduce_subvector('j', rank, fi)}$ = &
                         ieee_value(1._${o1}$, ieee_quiet_nan)
-                    return
+                    #:if rank == 1
+                      return
+                    #:else
+                      cycle
```
jvdp1, line 240: Good catch!

test/stats/test_median.fypp
```diff
@@ -284,7 +284,31 @@ contains
         call check(error&
                     , any(ieee_is_nan(median(d3_${k1}$, 1, d3_${k1}$ > 0)))&
                     , 'median(d3_${k1}$, 1, d3_${k1}$ > 0): should contain at least 1 IEEE NaN')
+        if (allocated(error)) return
 
+        ! Regression: verify median_mask processes all slices when one
+        ! slice has zero valid elements (n==0). Previously, the function
+        ! used 'return' instead of 'cycle', aborting all subsequent slices.
+        block
+            ${t1}$ :: xm(3, 3)
```
jvdp1:
```suggestion
            integer, parameter :: n = 3
            ${t1}$ :: xm(n, n)
```

test/stats/test_median.fypp
```diff
@@ -284,7 +284,31 @@ contains
...
+            logical :: mm(3, 3)
+            real(dp) :: rm(3)
```
jvdp1:
```suggestion
            logical :: mm(n, n)
            real(dp) :: rm(n)
```

test/stats/test_median.fypp
```diff
@@ -284,7 +284,31 @@ contains
...
+            xm(:, 1) = [${t1}$ :: 1, 2, 3]
+            xm(:, 2) = [${t1}$ :: 4, 5, 6]
+            xm(:, 3) = [${t1}$ :: 7, 8, 9]
```
jvdp1:
```suggestion
            xm = reshape([${t1}$ :: 1, 2, 3, 4, 5, 6, 7 ,8 ,9], [n,n])
```

src/stats/stdlib_stats_median.fypp
JAi-SATHVIK: The change from `#:if fi == 1` to `#:if rank == 1 `is needed. The issue is that `fi `is the dimension index being reduced (iterating from 1 to `rank`), not the rank of the array.

Considering a rank 2 array with` dim=1 (fi == 1)`

```
  case(1)
  do j2 = 1, size(x, 2)   ! <-- loop exists
    x_tmp(:) = x(:, j2)
    if (any(ieee_is_nan(x_tmp))) then
      res(j2) = ieee_value(...)
      return   ! BUG: exits entire function, skipping j2+1 onward
    end if
  end do
```

  
  With `#:if fi == 1`, the code emits return here — but there is an enclosing do loop, so it should emit cycle instead. The return aborts the entire function.

With `#:if rank == 1`, the logic is:

Rank 1: No loop exists → return (correct, no loop to `cycle`)
Rank > 1: Loop always exists for any `fi `→ `cycle `(correct)

jvdp1: I see. You are right. Thank you for the explanation

jvdp1 (approved):
Thank you @JAi-SATHVIK for this fix. I approve it and can be merged IMO.

Issue #1055, opened 2025-11-12
logspace generates wrong results for integer arguments


cneiss:
### Description

The logspace routine seems to generate wrong results when called with integer arguments.
Here is an example code (essentially taken from the stdlib specs page [https://stdlib.fortran-lang.org/page/specs/stdlib_math.html#logspace-create-a-logarithmically-spaced-rank-one-array](https://stdlib.fortran-lang.org/page/specs/stdlib_math.html#logspace-create-a-logarithmically-spaced-rank-one-array)):
```

program example_logspace_int
  use stdlib_math, only: logspace
  use stdlib_kinds, only: dp
  implicit none

  integer, parameter :: start = 10
  integer, parameter :: end = 23
  integer, parameter :: n = 3

  real(dp) :: r(n) ! Integer values raised to real powers results in real values

  r = logspace(start, end, n)
  print *,  r
end program example_logspace_int
```
This produces 
`   1410065408.0000000        1874919424.0000000       -159383552.00000000`
as output, which is obviously wrong.
I've tested with gfortran-14 and Intel oneapi ifx 2025; does not seem to depend on the compiler.

### Expected Behaviour

The result should be
`   10000000000.000000        31622776601683792.        1.0000000000000001E+023`

### Version of stdlib

fffe0d7058180d22faba9a2d32377cd88f6aefb7

### Platform and Architecture

Linux (OpenSUSE Leap 15.6)

### Additional Information

Note: I can reproduce the numbers if I replace
`r = logspace(start, end, n)`
with
```
integer :: dum
dum = linspace(start, end, n)
r = 10**dum
```
where dum ist of type integer; with type real(dp) the correct results are obtained.
Hence, the error seems to be related to a type mismatch in the logspace implementation.

cneiss:
The problem in your second example is the line 
`   exponents = linspace(start, end, n)`
where exponents is of type integer, but linspace returns real (as it should).
This is exactly what seems to happen in the logspace routine as well.

It's not a compiler bug IMO.

PR #1129, opened 2026-02-23
Fix: logspace generate wrong result for integer arguments

MadhurKumar004:
This resolves the issue of logspace giving wrong result. Please review if the approach is right
Fix: https://github.com/fortran-lang/stdlib/issues/1055


jvdp1 (commented):
thank you @MadhurKumar004 . Overall it looks good to me. Could you also update the specs, please?

PR #1130, opened 2026-02-25
Fix test_hash_functions failures on big-endian architectures

### LINKED ISSUE #1128: stdlib fails test_hash_functions on big--endian systems
### Description

Hi, Debian developer here. I have been packaging stdlib for Debian, and 0.8.1 fails on `test_hash_functions` on s390x, powerpc64 , sparc64 (ie all big-endian) architectures.
Please see the logs at https://buildd.debian.org/status/package.php?p=fortran-stdlib for details.

Regards
Alastair McKinstry

### Expected Behaviour

Test passes

### Version of stdlib

0.8.1

### Platform and Architecture

Debian Linux: s390x, PPC, powerprc64, sparc64

### Additional Information

_No response_

ijatinydv:
Resolves Debian build failures on big-endian architectures (s390x, powerpc, sparc64) reported in issue #1128

**The Issue:**
The `test_hash_functions` suite was failing on big-endian (BE) machines. The root cause was a mismatch in design philosophies: the Fortran implementations of `nmhash` and `waterhash` explicitly normalize byte reads to Little-Endian to ensure consistent hashes across platforms, while the vendored C reference implementations (used as test oracles) rely on native memory layouts. Because of this, the C tests were generating different hashes on BE and causing the assertions to fail.

`pengyhash` and `SpookyV2` use native reads in both C and Fortran, so they were already perfectly aligned on BE.

**Changes in this PR:**
* **Test Infrastructure:** Replaced the hard `error stop` in `test_little_endian` with `skip_test` on BE systems so the rest of the suite can actually run.
* **C Reference Patches (Test-only):** Patched `waterhash.h` (added `__BYTE_ORDER__` swaps) and `nmhash.h` (fixed the 16-bit union multiplication pairing) so the C test code accurately mirrors the Fortran LE-normalization on BE platforms. 
* **Self-Consistency Tests:** Added two new pure-Fortran tests (`test_hash_determinism` and `test_hash_distribution`) that run on all platforms. This ensures our Fortran algorithms aren't mathematically degenerate on BE, independent of the C reference code.
* **Docs:** Added a small note to the API docs warning users that `pengyhash` and `SpookyV2` produce endian-dependent output and shouldn't be used for cross-architecture verification.

Since all C-level changes are guarded by preprocessor macros (`#if defined(__BYTE_ORDER__)` and `#if NMHASH_LITTLE_ENDIAN`), this is guaranteed to be a zero-regression change on x86/ARM hardware.

Closes #1128


jvdp1:
@ijatinydv Sorry for my delayed answer. I agree with your proposition. It seems it is the most appropriate approach currently

jvdp1:
LGTM. Thank you @ijatinydv !

Issue #1131, opened 2026-02-26
issue with gamma function fypp template


jalvesz:
### Description

Hi @Calluumm @sebastian-mutz,

After the merge of #1099 I hit an error when building with the flags: `-DWITH_QP=False -DWITH_XDP=False`. I think this is linked to the way the fypp templating was handled.

The error I get:
```bash
...stdlib/build/src/stats/stdlib_stats_distribution_gamma.f90:612:19:

  612 |         res = real(regamma_p(shape, xs * rate), kind=dp)
      |                   1
Error: There is no specific function for the generic 'regularized_gamma_p' at (1)
```

### Expected Behaviour

should build

### Version of stdlib

master branch

### Platform and Architecture

Windows

sebastian-mutz:
Hi @jalvesz (cc @Calluumm ).

Thanks for the issue. I had a quick look. While I can't reproduce your exact problem with the same flags, I run into a different one related to an the older [special_functions_gamma](https://github.com/fortran-lang/stdlib/blob/master/src/specialfunctions/stdlib_specialfunctions_gamma.fypp) module that `regamma_p` belongs to. In that module, the list of KINDS_TYPES is truncated as `REAL_KINDS_TYPES[:-1]`. It seems to assume an ordered list of kinds. If I understand this correctly, it would normally drop `qp`, but if you omit the kinds according to your flags, it would drop `dp` and only generate `sp` procedures. When the `dp` procedure is requested, it's not there.

PR #1133, opened 2026-02-28
fix: issue with gamma function fypp template

Calluumm:
Looks to fixing [issue 1131](https://github.com/fortran-lang/stdlib/issues/1131) 

uses explicit filtering instead of -1 and fixed a type derivation


src/specialfunctions/stdlib_specialfunctions_gamma.fypp
```diff
@@ -1,4 +1,10 @@
 #:include "common.fypp"
+#:set GAMMA_REAL_KINDS_TYPES = [item for item in REAL_KINDS_TYPES if item[0] in ('sp', 'dp', 'xdp')]
+#:set GAMMA_CMPLX_KINDS_TYPES = [item for item in CMPLX_KINDS_TYPES if item[0] in ('sp', 'dp', 'xdp')]
+#:set GAMMA_CI_KINDS_TYPES = INT_KINDS_TYPES + GAMMA_CMPLX_KINDS_TYPES
+#:set GAMMA_IDX_CMPLX_KINDS_TYPES = [(i, CMPLX_KINDS[i], CMPLX_TYPES[i], CMPLX_INIT[i]) for i in range(len(CMPLX_KINDS)) if CMPLX_KINDS[i] in ('sp', 'dp', 'xdp')]
+#:set GAMMA_IDX_REAL_KINDS_TYPES  = [(i, REAL_KINDS[i], REAL_TYPES[i], REAL_INIT[i]) for i in range(len(REAL_KINDS)) if REAL_KINDS[i] in ('sp', 'dp', 'xdp')]
+#! LEGACY BELOW
 #:set CI_KINDS_TYPES = INT_KINDS_TYPES + CMPLX_KINDS_TYPES
 #:set IDX_CMPLX_KINDS_TYPES = [(i, CMPLX_KINDS[i], CMPLX_TYPES[i], CMPLX_INIT[i]) for i in range(len(CMPLX_KINDS))]
 #:set IDX_REAL_KINDS_TYPES  = [(i, REAL_KINDS[i], REAL_TYPES[i], REAL_INIT[i]) for i in range(len(REAL_KINDS))]
```
sebastian-mutz: Once it's been updated, I don't think there's a need for this.

jalvesz:
I would like to suggest something slightly different:

Currently, the CMake builds enable by default `qp` ( and `xdp` with gfortran) as it checks if available in the compiler and then allows it. Since we only check with gfortran, ifort and ifx which have `qp` and `xdp` for the first. It all passed.

With FPM, we added a job to test for `xdp` and `qp`. Currently, the FPM jobs are slower than the CMake jobs. I would like to propose then to use that job to test "without" then. In this case, we should first make sure that it does fail with fpm when desactivating both kinds and without the current fix.

https://github.com/fortran-lang/stdlib/blob/ee330556a02987d006f1089c44837d73505e4436/.github/workflows/fpm-deployment.yml#L45C1-L47C67

Opinions @jvdp1 @perazz  ?

Otherwise we could add yet another job more or less in those lines @Calluumm, but I would not use `contains(matrix.gcc_v, '10')` but the latest GNU compilers.

PR #1134, opened 2026-03-02
fix(math): preserve non-differenced dimension shape in diff for 2D array

JAi-SATHVIK:
### **Bug:-**
In `stdlib_math_diff.fypp` , the `diff` function for 2D arrays (`diff_2_*`) incorrectly handles the edge case where the number of differences `n` exceeds or equals the size of the operated dimension (`size_work <= n_`). 

The code executes:
 `allocate(y(0, 0))`

This collapses both dimensions to zero, destroying the shape of the non-differenced dimension. 

For example, calling `diff(x, n=5, dim=1)` on a (3, 10) array returns a (0, 0) result instead of the correct  (0, 10).
This breaks any downstream code that relies on the preserved dimension size (e.g `size(y, 2)` returning 10).

### **Fix:-**
Replacing the blanket `allocate(y(0, 0))` with a dimension-aware allocation that preserves the unrelated dimension:
```
if (dim_ == 1) then
    allocate(y(0, size(x, 2)))
else
    allocate(y(size(x, 1), 0))
end if
```

This ensures only the differenced dimension shrinks to zero, while the other dimension retains its original size.

### DIFF
--- a/src/math/stdlib_math_diff.fypp
+++ b/src/math/stdlib_math_diff.fypp
@@ -97,7 +97,7 @@ contains
             case (2)
                 allocate(y(size(x, 1), 0))
             case default
-                error stop "diff_2: invalid dimension (dim_ must be 1 or 2)"
+                error stop "diff_2: internal error: invalid dimension (dim_ must be 1 or 2)"
             end select
             return
         end if


src/math/stdlib_math_diff.fypp
```diff
@@ -91,7 +91,14 @@ contains
...
         
         if (size_work <= n_) then
-            allocate(y(0, 0))
+            select case (dim_)
+            case (1)
+                allocate(y(0, size(x, 2)))
+            case (2)
+                allocate(y(size(x, 1), 0))
+            case default
+                error stop "diff_2: invalid dimension (dim_ must be 1 or 2)"
```
jvdp1: Based on the lines 70- this should never be reached. However, I agree it is good to keep it.  Should we maybe mention that it is an internal error?

jvdp1 (commented):
Overall it sounds good to me. Could you also update the specs, please?

src/math/stdlib_math_diff.fypp
JAi-SATHVIK: Agreed. Although `dim_` is validated to be 1 or 2 earlier in the function. I've updated the message to clarify that hitting this branch would be an internal error.

`error stop "diff_2: internal error: invalid dimension (dim_ must be 1 or 2)"`

PR #1148, opened 2026-03-17
feat: implementation of the stdlib_datetime module

JAi-SATHVIK:
#1147


doc/specs/stdlib_datetime.md
```diff
@@ -0,0 +1,239 @@
...
+title: datetime
+---
+
+[TOC]
+
+## Introduction
+
+The `stdlib_datetime` module provides types and procedures for date, time, and duration handling. It defines two primary derived types — `datetime_type` for representing specific points in time and `timedelta_type` for representing durations — along with arithmetic operators, comparison operators, ISO 8601 parsing/formatting, and calendar utilities.
+
+No C-bindings or parameterized derived types are required. The module is purely mathematical (calendar algorithms) and string manipulation, ensuring near-100% compiler compatibility.
```
jvdp1: What do you mean with "near-100$ compiler compatibility"?

doc/specs/stdlib_datetime.md
```diff
@@ -0,0 +1,239 @@
...
+| Component | Type | Default | Description |
+|-----------|------|---------|-------------|
+| `year` | `integer` | 1 | Year (1–9999) |
+| `month` | `integer` | 1 | Month (1–12) |
+| `day` | `integer` | 1 | Day (1–31) |
+| `hour` | `integer` | 0 | Hour (0–23) |
+| `minute` | `integer` | 0 | Minute (0–59) |
+| `second` | `integer` | 0 | Second (0–59) |
+| `millisecond` | `integer` | 0 | Millisecond (0–999) |
+| `utc_offset_minutes` | `integer` | 0 | UTC offset in minutes |
```
jvdp1: I am not familiar with this type of data. Why is the offset provided in minutes and not in a more accurate measure (e.g., at least seconds)?

doc/specs/stdlib_datetime.md
```diff
@@ -0,0 +1,239 @@
...
+
+### `timedelta_type`
+
+Represents a duration or interval. Normalized so that `seconds` is in [0, 86399] and `milliseconds` is in [0, 999]. The `days` component can be negative for negative durations.
+
+| Component | Type | Default | Description |
+|-----------|------|---------|-------------|
+| `days` | `integer` | 0 | Number of days (can be negative) |
+| `seconds` | `integer` | 0 | Seconds (0–86399) |
+| `milliseconds` | `integer` | 0 | Milliseconds (0–999) |
```
jvdp1: Can these be also negative?

doc/specs/stdlib_datetime.md
```diff
@@ -0,0 +1,239 @@
...
+
+## Constructors
+
+### `datetime` — Create from components
+
+#### Description
+
+Creates a `datetime_type` from individual components. All arguments are optional and default to the type's defaults.
```
jvdp1: What is the type's defaults?

doc/specs/stdlib_datetime.md
```diff
@@ -0,0 +1,239 @@
...
+
+`dt = ` [[stdlib_datetime(module):datetime(function)]] `([year] [, month] [, day] [, hour] [, minute] [, second] [, millisecond] [, utc_offset_minutes])`
+
+#### Arguments
+
+All arguments are optional with `intent(in)` and type `integer`.
+
+#### Return value
+
+A `datetime_type` value.
```
jvdp1:
```suggestion
A `datetime_type` value with components being equal to those provided or to the default ones.
```

doc/specs/stdlib_datetime.md
```diff
@@ -0,0 +1,239 @@
...
+
+Creates a normalized `timedelta_type`. Accepts mixed units (days, hours, minutes, seconds, milliseconds) and normalizes them.
+
+#### Syntax
+
+`td = ` [[stdlib_datetime(module):timedelta(function)]] `([days] [, hours] [, minutes] [, seconds] [, milliseconds])`
+
+#### Arguments
+
+All arguments are optional with `intent(in)` and type `integer`.
```
jvdp1: What is the default if no arguments are provided?

doc/specs/stdlib_datetime.md
```diff
@@ -0,0 +1,239 @@
...
+
+| Component | Type | Default | Description |
+|-----------|------|---------|-------------|
+| `days` | `integer` | 0 | Number of days (can be negative) |
+| `seconds` | `integer` | 0 | Seconds (0–86399) |
+| `milliseconds` | `integer` | 0 | Milliseconds (0–999) |
+
+## Constructors
+
+### `datetime` — Create from components
```
jvdp1, line 46: for all descriptions, version and type of procedures should be provided

doc/specs/stdlib_datetime.md
```diff
@@ -0,0 +1,239 @@
...
+
+#### Syntax
+
+`dt = ` [[stdlib_datetime(module):now_utc(function)]] `()`
+
+### `epoch` — Unix epoch
+
+#### Description
+
+Returns the Unix epoch: `1970-01-01T00:00:00Z`.
```
jvdp1: Also on Windows OS? Please specify it

src/datetime/stdlib_datetime.f90
```diff
@@ -0,0 +1,778 @@
...
+            write(buf(n+2:n+3), '(I2.2)') off_h
+            buf(n+4:n+4) = ':'
+            write(buf(n+5:n+6), '(I2.2)') off_m
+            n = n + 6
+        end if
+
+        str = buf(1:n)
+    end function format_datetime
+
+    function format_timedelta(td) result(str)
```
jvdp1: could this function be `pure`?

src/datetime/stdlib_datetime.f90
```diff
@@ -0,0 +1,778 @@
+module stdlib_datetime
+    !! version: experimental
+    !!
+    !! Date, time, and time interval handling for Fortran.
+    !! ([Specification](../page/specs/stdlib_datetime.html))
+    use iso_fortran_env, only: int64
```
jvdp1: could you use the kinds provided by stdlib_kinds?

src/datetime/stdlib_datetime.f90
```diff
@@ -0,0 +1,778 @@
...
+        td = ms_to_td(total_ms)
+    end function timedelta
+
+    function now() result(dt)
+        !! version: experimental
+        !!
+        !! Return the current local date and time.
+        type(datetime_type) :: dt
+        integer :: v(8)
+        call date_and_time(values=v)
```
jvdp1, line 156: Is the intrinsic `date_and_time` `pure`?

jvdp1 (commented):
Thank you @JAi-SATHVIK  for this PR. A very complete set of procedures to handle dates an times IMO.

doc/specs/stdlib_datetime.md
JAi-SATHVIK: Clarified in the spec. After normalization these are always non-negative
 only `days` can be negative for negative durations.

doc/specs/stdlib_datetime.md
JAi-SATHVIK: Fixed. Rephrased to specify that the module uses only standard Fortran intrinsics and calendar arithmetic.

doc/specs/stdlib_datetime.md
JAi-SATHVIK: Added a note to the docs. Minutes are used to stay consistent with ISO 8601 `(±HH:MM)` as all real-world offsets are multiples of minutes.

src/datetime/stdlib_datetime.f90
JAi-SATHVIK: No, it is impure. I have updated the documentation for `now()` to reflect that it is a non-pure function due to this intrinsic call.

src/datetime/stdlib_datetime.f90
JAi-SATHVIK: Fixed. Replaced the `iso_fortran_env` import with `stdlib_kinds` for `int64`.

src/datetime/stdlib_datetime.f90
JAi-SATHVIK: This procedure cannot be `PURE` because the Fortran standard prohibits the use of internal `WRITE` statements within pure subprograms.

doc/specs/stdlib_datetime.md
JAi-SATHVIK: Clarified specified that the epoch is a fixed mathematical date independent of the operating system.

doc/specs/stdlib_datetime.md
JAi-SATHVIK: if no arguments are provided all components default to `0`.

doc/specs/stdlib_datetime.md
JAi-SATHVIK: default values (Jan 1, Year 1, 00:00:00)

src/datetime/stdlib_datetime.f90
jvdp1: Internal `write` are allowed in `pure` procedures.
See an example here: https://github.com/fortran-lang/stdlib/blob/e33b3ebd7a01600b43f5ef5e67af24708dd7a454/src/strings/stdlib_strings_to_string.fypp#L20

Or does I miss something?

JAi-SATHVIK: my mistake! I've updated both `format_datetime` and `format_timedelta` to be pure. Thanks for the correction and the reference.

src/datetime/stdlib_datetime.f90
```diff
@@ -0,0 +1,777 @@
...
+    pure function format_datetime(dt) result(str)
+        !! version: experimental
+        !!
+        !! Format a datetime_type as an ISO 8601 string.
+        type(datetime_type), intent(in) :: dt
+        character(:), allocatable :: str
+        character(len=29) :: buf
+        integer :: n, off_h, off_m
+
+        write(buf(1:4),  '(I4.4)') dt%year
```
jalvesz: This kind of string buffer construction I have always found prone to errors.
Could you consider the use of `to_string` (https://stdlib.fortran-lang.org/page/specs/stdlib_strings.html#to_string) and Fortran's string concatenation operator `//` instead ?

jalvesz (changes requested):

doc/specs/stdlib_datetime.md
jvdp1: Good. Could you add that in tthe specs, please?

JAi-SATHVIK: I have updated the `stdlib_datetime.md` specification.
- `datetime`: Defaults to the type's initial values (`1-01-01T00:00:00Z`).
- `timedelta`: Defaults to `0` for all components.

test/datetime/test_datetime.f90
```diff
@@ -0,0 +1,571 @@
...
+                     test_midnight_rollover), &
+        new_unittest("year_boundary", &
+                     test_year_boundary), &
+        new_unittest("now_returns_valid", &
+                     test_now_valid)]
+end subroutine collect_datetime
+
+subroutine test_leap_year_basic(error)
+    type(error_type), allocatable, intent(out) :: error
+    call check(error, is_leap_year(2000), .true.)
```
jalvesz: I think the tests would be clearer to read and more helpful for debuging if you switch to the following pattern:
```fortran
! call check(error, condition_evaluation , "some comment if something goes wrong")
!> E.g:
call check(error, is_leap_year(2000) == .true. ,  "2000 should be a leap year" )
```
for logical functions you could simply use `is_leap_year(xxxx)` or `.not.is_leap_year(xxxx)`

jalvesz (changes requested):

test/datetime/test_datetime.f90
JAi-SATHVIK: I have changed every `check` call to `check(error, condition, "message")` with a failure message for debugging.

jalvesz (approved):
LGTM @JAi-SATHVIK thanks for this nice new functionality!

jvdp1 (approved):
Thank you @JAi-SATHVIK for this nice addition to `stdlib`!

cyrilgandon:
Hi, thanks for the great work on this PR, it's really appreciated.

I noticed the mention of potential `date_type` and `time_type` sub-types:

> "Smaller sub-types if separating only the calendar date from the clock time is deemed necessary."

Do you currently plan to introduce these types as part of a follow-up, or is the intention to keep a single `datetime_type` for now?

I'm asking because separating date and time can be quite useful in practice, but I understand it also comes with additional API complexity.

Thanks!

JAi-SATHVIK:
Hi @cyrilgandon, thanks for the feedback!

I used a unified `datetime_type `for now to build a stable foundation with a simpler API. I do plan to add `date_type `and `time_type` in a follow up PR once the core logic is tested. Feel free to share any specific use cases you have in mind

cyrilgandon:
Hi, thanks for the clarification!

One concrete use case on my side: I work on STICS, a crop modeling system with a daily time step. We only manipulate calendar dates (year, month, day), and time of day is irrelevant.

In that context, a dedicated date_type would make the code clearer and avoid carrying unused time fields around.

Looking forward to the follow-up PR 🙂

PR #1150, opened 2026-03-20
ci: add validation step for generated pkg-config files

srinjoy933:
This pull request introduces an automated validation step within the CMake CI workflow (.github/workflows/CI.yml) to ensure the structural integrity of generated pkg-config exports. Building on recent fixes for absolute pathing in .pc files, this addition prevents future regressions by strictly testing the generated fortran_stdlib.pc file immediately following the cmake --install step. The workflow now ensures pkg-config is present on the runner, maps the PKG_CONFIG_PATH to the local _dist installation directory, and executes both a strict --validate check and a test of the --cflags and --libs flag resolution.
Solve #1149


.github/workflows/CI.yml
```diff
@@ -113,3 +113,25 @@ jobs:
     - name: Install project
       if: ${{ contains(matrix.build, 'cmake') }}
       run: cmake --install ${{ env.BUILD_DIR }}
+
+    - name: Validate generated pkg-config file
+      if: ${{ contains(matrix.build, 'cmake') }}
+      shell: bash
+      run: |
+        # Ensure pkg-config is available
```
jvdp1: Is there not "action" defined somewhere to setup `pkg-config`?
It would help to avoid these if-else conditions

.github/workflows/CI.yml
```diff
@@ -113,3 +113,25 @@ jobs:
...
+        fi
+        
+        # Point to the installation directory specified by CMAKE_INSTALL_PREFIX
+        export PKG_CONFIG_PATH="$PWD/_dist/lib/pkgconfig:$PWD/_dist/share/pkgconfig:$PKG_CONFIG_PATH"
+        
+        # Run the validation
+        pkg-config --validate fortran_stdlib
+        
+        # Test flags to ensure no errors are thrown
+        pkg-config --cflags --libs fortran_stdlib
```
jvdp1: If I am correct, this will return the paths, but it won't check if the paths are correct. what is the expectation?

jvdp1 (commented):
Thank you @srinjoy933 . I left a few comments.

.github/workflows/CI.yml
srinjoy933: Hi @jvdp1, thank you for the review! It turns out pkg-config is pre-installed on the default GitHub Ubuntu and macOS runners, so I have completely removed the if-else installation block to keep the workflow clean.

.github/workflows/CI.yml
srinjoy933: You are correct that --cflags --libs only prints the paths but doesn't verify them. To actually validate the paths (and catch any double-prefixing bugs), I've updated the script to extract the libdir and includedir variables from pkg-config and explicitly verify that those directories exist on the filesystem.

.github/workflows/CI.yml
```diff
@@ -5,7 +5,6 @@ on: [push, pull_request]
 env:
   CMAKE_BUILD_PARALLEL_LEVEL: "2" # 2 cores on each GHA VM, enable parallel builds
   CTEST_OUTPUT_ON_FAILURE: "ON" # This way we don't need a flag to ctest
-  CTEST_PARALLEL_LEVEL: "2"
```
jvdp1, line 8: why is this removed?

.github/workflows/CI.yml
```diff
@@ -106,10 +105,47 @@ jobs:
...
         --test-dir ${{ env.BUILD_DIR }}
-        --parallel
         --output-on-failure
         --no-tests=error
 
     - name: Install project
       if: ${{ contains(matrix.build, 'cmake') }}
       run: cmake --install ${{ env.BUILD_DIR }}
+
+    - name: Validate generated pkg-config file
```
jvdp1, line 117: Thank you @srinjoy933 . It is a nice improvment, and the paths are now checked.

In the past, we got some issues with the pkg-config files (e..g, issues with the order of the libs, ...; see e.g. #1102). 
I wonder if it would be approriate to add a specific CI yml file for testing pkg-config files (even for compilation).

.github/workflows/CI.yml
srinjoy933: Hi @jvdp1! I removed the parallel testing flags to bypass a race condition in the upstream test suite that was causing the CI to crash on this branch.Specifically, the I/O example tests (example_open, example_savetxt, and example_loadtxt) all hardcode their read/write operations to the exact same file (example/io/example.dat). When CTest runs them in parallel, they step on each other (e.g., loadtxt tries to read floats while example_open is writing English text to the same file), which causes an <input conversion error>.

Removing the parallel flag forces CTest to run them sequentially, which immediately fixed the CI.

.github/workflows/CI.yml
srinjoy933: Thank you @jvdp1 ! I completely agree. Setting up a dedicated .yml workflow that actually compiles a dummy Fortran program using the generated pkg-config flags is the best way to catch those linking order bugs like #1102.
Since this current PR is stable and successfully catches the basic path/double-prefixing errors, would you prefer to merge this ? I would be more than happy to open a fresh issue and a follow-up PR dedicated specifically to building out that standalone pkg-config compilation CI workflow.

.github/workflows/CI.yml
srinjoy933: Hey @jvdp1 @jalvesz is removing this  part will create any problem? I checked and tested and according to me it will not create any problem as such, that's why I suggest that it can be removed

jvdp1: I never like to remove a feature (here parallelization) to solve another problem (here with I/O tests). It will only hide the underlying problem that will re-appear later. Therefore, I suggest to add the parallelization feature and try to solve the underlying problem in another PR

srinjoy933:
hey @jvdp1 as you have suggested I reverted the  parallelization feature and will create another pr to solve the test problem. Please review this pr once . Thanks!

srinjoy933:
hey @jvdp1,  any more changes that you want for this pr or is this good to go to be merged?

.github/workflows/CI.yml
```diff
@@ -106,10 +106,49 @@ jobs:
       run: >-
         ctest
         --test-dir ${{ env.BUILD_DIR }}
-        --parallel
+          --parallel
```
jvdp1:
```suggestion
        --parallel
```

PR #1155, opened 2026-03-25
feat (linalg): permutation matrices

### LINKED ISSUE #891: Linear algebra support for permutation matrices
### Motivation

It is well known that it is inefficient to store and multiply by dense permutation matrices. Instead permutation matrices should be stored as vectors which give information on how to organize the rows/columns of a matrix depending on whether left or right multiplication by the permutation matrix is desired.

Procedures would include:

- Inverting a permutation
- Multiplication by a permutation matrix on the right (column permutation)
- Multiplication by a permutation matrix on the left (row permutation)
- Support for "finalized and nonfinalized" permutations (see below for more details)

### Prior Art

Julia has implementations for nonfinalized permutation matrix multiplication, see permutecols! and permuterows! in https://github.com/JuliaLang/julia/blob/master/base/combinatorics.jl. As far as I know, LAPACK only has an implementation for nonfinalized left permutation matrix multiplication (row permutation), see dlaswp https://www.netlib.org/lapack/explore-html/d1/d7e/group__laswp_ga5d3ea3e3cb61e32750bf062a2446aa33.html#ga5d3ea3e3cb61e32750bf062a2446aa33.

Some outputs of LAPACK routines contain finalized permutations, see the jpvt argument of dgeqp3 https://www.netlib.org/lapack/explore-html/d0/dea/group__geqp3_gae96659507d789266660af28c617ef87f.html#gae96659507d789266660af28c617ef87f.

The difference between what I have been calling a finalized and nonfinalized permutation matrix is that a finalized permutation matrix tells you exactly where each column/row should end up after multiplication, while a nonfinalized permutation matrix just tells you what columns/rows to switch and where (with the possibility of moving a single column/row multiple times).

I have my own implementations of some multiplication routines, but their performance is questionable. I can provide them if they would be of use.

### Additional Information

_No response_

edish-github:
## Summary

Fixes #891. 

Currently, applying permutation matrices requires storing them as dense matrices, wasting $O(n^2)$ memory and costing $O(n^2 k)$ time during multiplication.

This PR introduces a `permutation_type` type that stores the permutation as an integer array, bringing storage down to $O(n)$ and row/column operations down to $O(n)$. 

### Changes
- Added `permutation_type` supporting both direct mapping (`finalized`) and LAPACK swap sequences (`ipiv` from `dgetrf`).
- Added standard operations: `identity`, `invert`, `compose`, and array conversions.
- Added `permutation_apply_left` (row swaps) and `permutation_apply_right` (column swaps).
- Added 12 unit tests using `testdrive` covering all real, complex, and integer kinds..

### Benchmark
I ran a quick sanity check to compare dense `matmul` vs the new vector approach on a 5000x100 matrix using `gfortran -O3`:

```text
n=5000, k=100
Dense approach (P_dense @ A):  0.155507 s
Vector approach (row copy):    0.000703 s
Speedup: ~ 221x

### DIFF
--- a/src/linalg/stdlib_linalg.fypp
+++ b/src/linalg/stdlib_linalg.fypp
@@ -70,6 +70,19 @@ module stdlib_linalg
   public :: is_hermitian
   public :: is_triangular
   public :: is_hessenberg
+
+  ! Permutation matrix operations
+  public :: permutation_type
+  public :: permutation_identity
+  public :: permutation_init
+  public :: permutation_is_valid
+  public :: permutation_invert
+  public :: permutation_compose
+  public :: permutation_from_lapack
+  public :: permutation_to_matrix
+  public :: permutation_apply_left
+  public :: permutation_apply_right
+  public :: permutation_apply_vector
   
   ! Export linalg error handling
   public :: linalg_state_type, linalg_error_handling
@@ -2103,6 +2116,132 @@ module stdlib_linalg
     end subroutine stdlib_linalg_${ri}$_expm
     #:endfor
   end interface matrix_exp
+
+  !> Permutation type: stores permutation as integer vector
+  type :: permutation_type
+    integer(ilp), allocatable :: perm(:)
+    integer(ilp) :: n = 0
+    logical :: finalized = .true.
+  end type
+
+  interface permutation_identity
+    !! version: experimental
+    !!
+    !! Creates an identity permutation of size n
+    module function permutation_identity(n) result(p)
+      integer(ilp), intent(in) :: n
+      type(permutation_type) :: p
+    end function
+  end interface
+
+  interface permutation_init
+    !! version: experimental
+    !!
+    !! Creates a permutation from a given integer vector
+    module function permutation_init(perm_vec, finalized) result(p)
+      integer(ilp), intent(in) :: perm_vec(:)
+      logical, intent(in), optional :: finalized
+      type(permutation_type) :: p
+    end function
+  end interface
+
+  interface permutation_is_valid
+    !! version: experimental
+    !!
+    !! Checks if a permutation is valid (each integer 1..n appears exactly once)
+    module function permutation_is_valid(p) result(valid)
+      type(permutation_type), intent(in) :: p
+      logical :: valid
+    end function
+  end interface
+
+  interface permutation_invert
+    !! version: experimental
+    !!
+    !! Inverts a finalized permutation
+    module function permutation_invert(p) result(inv)
+      type(permutation_type), intent(in) :: p
+      type(permutation_type) :: inv
+    end function
+  end interface
+
+  interface permutation_compose
+    !! version: experimental
+    !!
+    !! Composes two finalized permutations: result = p1 * p2
+    module function permutation_compose(p1, p2) result(p)
+      type(permutation_type), intent(in) :: p1, p2
+      type(permutation_type) :: p
+    end function
+  end interface
+
+  interface permutation_from_lapack
+    !! version: experimental
+    !!
+    !! Converts LAPACK pivot vector to finalized permutation
+    module function permutation_from_lapack(ipiv, n, err) result(p)
+      integer(ilp), intent(in) :: ipiv(:)
+      integer(ilp), intent(in) :: n
+      type(linalg_state_type), intent(out), optional :: err
+      type(permutation_type) :: p
+    end function
+  end interface
+
+  interface permutation_apply_left
+    !! version: experimental
+    !!
+    !! Applies permutation from the left: B = P * A (row permutation)
+    #:for k1, t1 in RCI_KINDS_TYPES
+    module subroutine permutation_apply_left_${t1[0]}$${k1}$ (p, A, B, err)
+      type(permutation_type), intent(in) :: p
+      ${t1}$, intent(in) :: A(:,:)
+      ${t1}$, intent(out) :: B(:,:)
+      type(linalg_state_type), intent(out), optional :: err
+    end subroutine
+    #:endfor
+  end interface
+
+  interface permutation_apply_right
+    !! version: experimental
+    !!
+    !! Applies permutation from the right: B = A * P (column permutation)
+    #:for k1, t1 in RCI_KINDS_TYPES
+    module subroutine permutation_apply_right_${t1[0]}$${k1}$ (p, A, B, err)
+      type(permutation_type), intent(in) :: p
+      ${t1}$, intent(in) :: A(:,:)
+      ${t1}$, intent(out) :: B(:,:)
+      type(linalg_state_type), intent(out), optional :: err
+    end subroutine
+    #:endfor
+  end interface
+
+  interface permutation_apply_vector
+    !! version: experimental
+    !!
+    !! Applies permutation to a vector: b = P * a
+    #:for k1, t1 in RCI_KINDS_TYPES
+    module subroutine permutation_apply_vector_${t1[0]}$${k1}$ (p, a, b, err)
+      type(permutation_type), intent(in) :: p
+      ${t1}$, intent(in) :: a(:)
+      ${t1}$, intent(out) :: b(:)
+      type(linalg_state_type), intent(out), optional :: err
+    end subroutine
+    #:endfor
+  end interface
+
+  interface permutation_to_matrix
+    !! version: experimental
+    !!
+    !! Converts permutation to dense matrix
+    #:for k1, t1 in RCI_KINDS_TYPES
+    module subroutine permutation_to_matrix_${t1[0]}$${k1}$ (p, A, err)
+      type(permutation_type), intent(in) :: p
+      ${t1}$, intent(out) :: A(:,:)
+      type(linalg_state_type), intent(out), optional :: err
+    end subroutine
+    #:endfor
+  end interface
+
 contains
 
 
--- a/src/linalg/stdlib_linalg.fypp
+++ b/src/linalg/stdlib_linalg.fypp
@@ -75,7 +75,7 @@ module stdlib_linalg
   public :: permutation_type
   public :: permutation_identity
   public :: permutation_init
-  public :: permutation_is_valid
+  public :: is_valid_permutation
   public :: permutation_invert
   public :: permutation_compose
   public :: permutation_from_lapack
@@ -2128,7 +2128,7 @@ module stdlib_linalg
     !! version: experimental
     !!
     !! Creates an identity permutation of size n
-    module function permutation_identity(n) result(p)
+    pure module function permutation_identity(n) result(p)
       integer(ilp), intent(in) :: n
       type(permutation_type) :: p
     end function
@@ -2138,18 +2138,18 @@ module stdlib_linalg
     !! version: experimental
     !!
     !! Creates a permutation from a given integer vector
-    module function permutation_init(perm_vec, finalized) result(p)
+    pure module function permutation_init(perm_vec, finalized) result(p)
       integer(ilp), intent(in) :: perm_vec(:)
       logical, intent(in), optional :: finalized
       type(permutation_type) :: p
     end function
   end interface
 
-  interface permutation_is_valid
+  interface is_valid_permutation
     !! version: experimental
     !!
     !! Checks if a permutation is valid (each integer 1..n appears exactly once)
-    module function permutation_is_valid(p) result(valid)
+    pure module function is_valid_permutation(p) result(valid)
       type(permutation_type), intent(in) :: p
       logical :: valid
     end function
--- a/src/linalg/stdlib_linalg_permutation.fypp
+++ b/src/linalg/stdlib_linalg_permutation.fypp
@@ -1,0 +1,254 @@
+!> Permutation matrix operations using efficient vector storage.
+!>
+!> Instead of storing permutations as dense n×n matrices (O(n²) memory),
+!> we represent them as integer vectors of length n (O(n) memory).
+!> All operations (inversion, left/right multiplication, composition)
+!> run in O(n) time per column/row rather than O(n²).
+!>
+!> Two formats are supported:
+!>   - Finalized: perm(i) = j means row i maps to row j (direct mapping)
+!>   - LAPACK pivot: ipiv(i) = j means swap rows i and j at step i
+!>
+!> Reference: Issue #891 (https://github.com/fortran-lang/stdlib/issues/891)
+
+#:include "common.fypp"
+#:set REAL_KINDS_TYPES = [("sp","real(sp)"), ("dp","real(dp)")]
+#:set INT_KINDS_TYPES = [("int32","integer(int32)"), ("int64","integer(int64)")]
+
+module stdlib_linalg_permutation
+  use stdlib_kinds, only: sp, dp, int32, int64
+  implicit none
+  private
+
+  !> Permutation type: stores permutation as integer vector
+  type, public :: permutation_t
+    integer, allocatable :: perm(:)
+    integer :: n = 0
+    logical :: finalized = .true.
+  end type
+
+  ! Public procedures
+  public :: permutation_identity
+  public :: permutation_init
+  public :: permutation_is_valid
+  public :: permutation_invert
+  public :: permutation_compose
+  public :: permutation_from_lapack
+  public :: permutation_to_matrix
+  public :: permutation_apply_left
+  public :: permutation_apply_right
+  public :: permutation_apply_vector
+
+  ! Generic interfaces for multi-type support
+  interface permutation_apply_left
+#:for rk, rt in REAL_KINDS_TYPES
+    module procedure permutation_apply_left_${rk}$
+#:endfor
+  end interface
+
+  interface permutation_apply_right
+#:for rk, rt in REAL_KINDS_TYPES
+    module procedure permutation_apply_right_${rk}$
+#:endfor
+  end interface
+
+  interface permutation_apply_vector
+#:for rk, rt in REAL_KINDS_TYPES
+    module procedure permutation_apply_vector_${rk}$
+#:endfor
+  end interface
+
+  interface permutation_to_matrix
+#:for rk, rt in REAL_KINDS_TYPES
+    module procedure permutation_to_matrix_${rk}$
+#:endfor
+  end interface
+
+contains
+
+  !> Create an identity permutation of size n
+  function permutation_identity(n) result(p)
+    integer, intent(in) :: n
+    type(permutation_t) :: p
+    integer :: i
+
+    p%n = n
+    p%finalized = .true.
+    allocate(p%perm(n))
+    do i = 1, n
+      p%perm(i) = i
+    end do
+  end function
+
+  !> Create a permutation from a given integer vector
+  function permutation_init(perm_vec, finalized) result(p)
+    integer, intent(in) :: perm_vec(:)
+    logical, intent(in), optional :: finalized
+    type(permutation_t) :: p
+
+    p%n = size(perm_vec)
+    allocate(p%perm(p%n), source=perm_vec)
+    if (present(finalized)) then
+      p%finalized = finalized
+    else
+      p%finalized = .true.
+    end if
+  end function
+
+  !> Check if a permutation is valid (each integer 1..n appears exactly once)
+  function permutation_is_valid(p) result(valid)
+    type(permutation_t), intent(in) :: p
+    logical :: valid
+    logical, allocatable :: seen(:)
+    integer :: i
+
+    valid = .false.
+    if (.not. allocated(p%perm)) return
+    if (p%n < 1) return
+    if (size(p%perm) /= p%n) return
+
+    allocate(seen(p%n), source=.false.)
+    do i = 1, p%n
+      if (p%perm(i) < 1 .or. p%perm(i) > p%n) return
+      if (seen(p%perm(i))) return
+      seen(p%perm(i)) = .true.
+    end do
+    valid = .true.
+  end function
+
+  !> Invert a finalized permutation
+  !> If perm(i) = j, then inv(j) = i
+  recursive function permutation_invert(p) result(inv)
+    type(permutation_t), intent(in) :: p
+    type(permutation_t) :: inv
+    integer :: i
+
+    inv%n = p%n
+    inv%finalized = .true.
+    allocate(inv%perm(p%n))
+
+    if (p%finalized) then
+      ! Direct inversion: swap keys and values
+      do i = 1, p%n
+        inv%perm(p%perm(i)) = i
+      end do
+    else
+      ! For LAPACK format: first finalize, then invert
+      inv = permutation_invert(permutation_finalize(p))
+    end if
+  end function
+
+  !> Convert LAPACK pivot vector (swap sequence) to finalized permutation
+  !> ipiv(i) = j means "swap rows i and j" applied at step i
+  function permutation_from_lapack(ipiv, n) result(p)
+    integer, intent(in) :: ipiv(:)
+    integer, intent(in) :: n
+    type(permutation_t) :: p
+    integer :: i, tmp
+
+    p%n = n
+    p%finalized = .true.
+    allocate(p%perm(n))
+
+    ! Start with identity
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: src/linalg/stdlib_linalg_permutation.fypp [hunk 1]
```diff
@@ -0,0 +1,255 @@
...
+!> run in O(n) time per column/row rather than O(n²).
+!>
+!> Two formats are supported:
+!>   - Finalized: perm(i) = j means row i maps to row j (direct mapping)
+!>   - LAPACK pivot: ipiv(i) = j means swap rows i and j at step i
+!>
+!> Reference: Issue #891 (https://github.com/fortran-lang/stdlib/issues/891)
+
+#:include "common.fypp"
+#:set REAL_KINDS_TYPES = [("sp","real(sp)"), ("dp","real(dp)")]
```
jalvesz: Support for all real and complex kinds should be added for such functionality. This can be done quite easily with stdlib's fypp templates. Please look at other modules in the library, specially those in the linalg folder, to see how that can be done. Do not filter kinds unless there is a very good reason for it, if the motivation was LAPACK, stdlib's modernized BLAS/LAPACK covers all kinds thanks to fypp.

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,255 @@
...
+#:set INT_KINDS_TYPES = [("int32","integer(int32)"), ("int64","integer(int64)")]
```
jalvesz: What's the purpose of defining integer kinds here in this manner ? if the reason is to define the integer size for the matrix maximun storage size, you should use `ilp` from `stdlib_linalg_constants` and procedure to use `integer(ilp)` instead of `integer`.
Now, if the reason is to support integer matrices, then you can use something like `#:set RCI_KINDS_TYPES = REAL_KINDS_TYPES + CMPLX_KINDS_TYPES + INT_KINDS_TYPES` for the template.

test/linalg/test_linalg_permutation.fypp
```diff
@@ -0,0 +1,278 @@
...
+    do i = 1, 3
+      do j = 1, 3
+        A(i,j) = real(i*10 + j, dp)
+      end do
+    end do
+
+    call permutation_apply_left(p, A, B)
+    do i = 1, 3
+      do j = 1, 3
+        call check(error, abs(A(i,j) - B(i,j)) < 1.0e-10_dp, "identity noop")
```
jalvesz: errors checks should be done on full vectors or full matrices, not individual entries. you can use either `all` or a check on the L2 norm of the difference. This is much more robust.

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,255 @@
...
+
+module stdlib_linalg_permutation
+  use stdlib_kinds, only: sp, dp, int32, int64
+  implicit none
+  private
+
+  !> Permutation type: stores permutation as integer vector
+  type, public :: permutation_t
```
jalvesz: stdlib's convention is to use `<>_type` for public derived types.

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,255 @@
...
+  public :: permutation_compose
+  public :: permutation_from_lapack
+  public :: permutation_to_matrix
+  public :: permutation_apply_left
+  public :: permutation_apply_right
+  public :: permutation_apply_vector
+
+  ! Generic interfaces for multi-type support
+  interface permutation_apply_left
+#:for rk, rt in REAL_KINDS_TYPES
```
jalvesz: please indent at the internal interface scope level

jalvesz (changes requested):

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,269 @@
...
+  public :: permutation_compose
+  public :: permutation_from_lapack
+  public :: permutation_to_matrix
+  public :: permutation_apply_left
+  public :: permutation_apply_right
+  public :: permutation_apply_vector
+
+  ! Generic interfaces for multi-type support
+  interface permutation_apply_left
+#:for k1, t1 in RCI_KINDS_TYPES
```
jalvesz: Please indent the fypp preprocessing macros to the entity scope.
```
interface xyz
      #:for ...
      module procedure xyz_<> ...
      #:endfor 
```
Please check across the PR, not only this line.

test/linalg/test_linalg_permutation.fypp
```diff
@@ -0,0 +1,297 @@
...
+    type(error_type), allocatable, intent(out) :: error
+    type(permutation_type) :: p
+    real(dp) :: a(4), b(4)
+
+    a = [10.0_dp, 20.0_dp, 30.0_dp, 40.0_dp]
+    p = permutation_init([3, 1, 4, 2])
+    call permutation_apply_vector(p, a, b)
+
+    ! b(1) = a(3) = 30, b(2) = a(1) = 10, b(3) = a(4) = 40, b(4) = a(2) = 20
+    call check(error, abs(b(1) - 30.0_dp) < 1.0e-10_dp, "vec(1)")
```
jalvesz: When possible, please check on whole arrays, not individual terms
```
call check(error, abs( all( b - [val1,val2,...])  < tol ), "some error comment" )
```
you can also store the reference values in a reference array and use it for the error check

jalvesz (changes requested):

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,269 @@
...
+!>     (i.e., after applying the permutation on the left, B(i,:) = A(perm(i),:))
+!>   - LAPACK pivot: ipiv(i) = j means swap rows i and j at elimination step i
+!>
+!> Reference: Issue #891 (https://github.com/fortran-lang/stdlib/issues/891)
+
+#:include "common.fypp"
+#:set RC_KINDS_TYPES = REAL_KINDS_TYPES + CMPLX_KINDS_TYPES
+#:set RCI_KINDS_TYPES = RC_KINDS_TYPES + INT_KINDS_TYPES
+
+module stdlib_linalg_permutation
```
loiseaujc: I would suggest making this contribution a `submodule` of `linalg` rather than a standalone `module`. It is just standard linear algebra after all. What's your take @jalvesz ?

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,269 @@
...
+    integer(ilp), intent(in) :: n
+    type(permutation_type) :: p
+    integer(ilp) :: i
+
+    p%n = n
+    p%finalized = .true.
+    allocate(p%perm(n))
+    do i = 1, n
+      p%perm(i) = i
+    end do
```
loiseaujc: MInor detail, but you could use the `arange` function from `stdlib_math` to collapse these four lines into a single one.

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,269 @@
...
+    logical, intent(in), optional :: finalized
+    type(permutation_type) :: p
+
+    p%n = size(perm_vec)
+    allocate(p%perm(p%n), source=perm_vec)
+    if (present(finalized)) then
+      p%finalized = finalized
+    else
+      p%finalized = .true.
+    end if
```
loiseaujc: You can use  `optval` from the `stdlib_optval` module.

```
p%finalize = optval(finalize, .true.)
```

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,269 @@
...
+
+    p%n = n
+    p%finalized = .true.
+    allocate(p%perm(n))
+
+    ! Start with identity
+    do i = 1, n
+      p%perm(i) = i
+    end do
+
```
loiseaujc, line 113: Same as before. You can use `arange` to collapse these lines into a single one.

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,269 @@
...
+    ! Replay swap sequence forward
+    do i = 1, min(n, int(size(ipiv), ilp))
+      if (ipiv(i) < 1 .or. ipiv(i) > n) error stop "permutation_from_lapack: Invalid LAPACK pivot index out of bounds"
```
loiseaujc: Please use `linalg_state_type` and `linalg_handling_error` for error handling.

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,269 @@
...
+    type(permutation_type), intent(in) :: p
+    type(permutation_type) :: fp
+
+    if (p%finalized) then
+      fp = p
+      return
+    end if
+    ! Treat swap sequence as LAPACK pivots
+    fp = permutation_from_lapack(p%perm, p%n)
+  end function
```
loiseaujc, line 141: Prefer

```
if (p%finalized) then
    fp = p
else
    fp = permutation_from_lapack(p%perm, p%n)
endif
```

over what you wrote. Whenever possible, the function should return only at the end.

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,269 @@
...
+    integer(ilp) :: i
+
+    p%n = p1%n
+    p%finalized = .true.
+    allocate(p%perm(p%n))
+
+    do i = 1, p%n
+      p%perm(i) = p2%perm(p1%perm(i))
+    end do
+  end function
```
loiseaujc, line 157: Wouldn't `p%perm = p2%perm(p1%perm)`  be equally valid and more succinct ?

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,269 @@
...
+
+  !> Apply permutation from the left: B = P * A (row permutation)
+  !> Row i of B comes from row perm(i) of A
+  subroutine permutation_apply_left_${t1[0]}$${k1}$ (p, A, B)
+    type(permutation_type), intent(in) :: p
+    ${t1}$, intent(in) :: A(:,:)
+    ${t1}$, intent(out) :: B(:,:)
+    integer(ilp) :: i
+
+    if (.not. p%finalized) error stop "Permutation must be finalized before applying."
```
loiseaujc: Use `linalg_state_type` and `linalg_error_handling`.

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,269 @@
...
+    do i = 1, p%n
+      B(i, :) = A(p%perm(i), :)
+    end do
+  end subroutine
```
loiseaujc, line 182: Each iteration of the loop are independent.

```
do concurrent(i=1:p%n)
    B(i, :) = A(p%perm(i), :)
end do
```

would do.

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,269 @@
...
+    type(permutation_type), intent(in) :: p
+    ${t1}$, intent(in) :: a(:)
+    ${t1}$, intent(out) :: b(:)
+    integer(ilp) :: i
+
+    if (.not. p%finalized) error stop "Permutation must be finalized before applying."
+    do i = 1, p%n
+      b(i) = a(p%perm(i))
+    end do
+  end subroutine
```
loiseaujc, line 221: `b = a(p%perm)` Keep it simple.

loiseaujc (changes requested):
Here is my first pass.

I do have one major question though for @jalvesz, @jvdp1 and @perazz : at the moment, the `permutation_apply_*` are performing out-of-place computations. In many situations however, I guess permutations should be applied in-place for performance-sake. How about having a `subroutine permutation_apply_*(A, p)` performing in-place, and a `function permutation_apply_*(A, p) result(B)` for out-of-place (I know subs and functions can't have same name, just lazy here) ?

src/linalg/stdlib_linalg_permutation.fypp
jalvesz: Yes, agree!

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,244 @@
...
+  end function
+
+  !> Create a permutation from a given integer vector
+  module function permutation_init(perm_vec, finalized) result(p)
+    integer(ilp), intent(in) :: perm_vec(:)
+    logical, intent(in), optional :: finalized
+    type(permutation_type) :: p
+
+    p%n = size(perm_vec)
+    allocate(p%perm(p%n), source=perm_vec)
```
jvdp1:
```suggestion
    allocate(p%perm, source=perm_vec)
```

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,244 @@
...
+    type(permutation_type) :: p
+
+    p%n = n
+    p%finalized = .true.
+    allocate(p%perm(n))
+    p%perm = arange(1_ilp, n)
+  end function
+
+  !> Create a permutation from a given integer vector
+  module function permutation_init(perm_vec, finalized) result(p)
```
jvdp1: could these functions be `pure`?

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,244 @@
...
+    else
+      fp = permutation_finalize(p)
+    end if
+
+    inv%n = fp%n
+    inv%finalized = .true.
+    allocate(inv%perm(fp%n))
+
+    ! Direct inversion: swap keys and values
+    do i = 1, fp%n
```
jvdp1: could you use do concurrent?

test/linalg/test_linalg_permutation.fypp
```diff
@@ -0,0 +1,274 @@
...
+    call check(error, all(abs(B - B_ref) < 1.0e-10_dp), "apply_right whole array")
+  end subroutine
+
+  !> Test 8: Vector permutation
+  subroutine test_apply_vector(error)
+    type(error_type), allocatable, intent(out) :: error
+    type(permutation_type) :: p
+    real(dp) :: a(4), b(4)
+
+    a = [10.0_dp, 20.0_dp, 30.0_dp, 40.0_dp]
```
jvdp1: these could be `parameter`

test/linalg/test_linalg_permutation.fypp
```diff
@@ -0,0 +1,274 @@
...
+    type(permutation_type) :: p1, p2, composed
+    real(dp) :: a(4), b1(4), b2(4), b_composed(4)
+
+    p1 = permutation_init([2_ilp, 3_ilp, 4_ilp, 1_ilp])
+    p2 = permutation_init([2_ilp, 1_ilp, 4_ilp, 3_ilp])
+    composed = permutation_compose(p1, p2)
+
+    ! Verify: composed applied to vector = p1 applied to (p2 applied to vector)
+    a = [10.0_dp, 20.0_dp, 30.0_dp, 40.0_dp]
+    b1 = 0.0_dp; b2 = 0.0_dp; b_composed = 0.0_dp ! Initialize to avoid garbage
```
jvdp1: avoid multiple assignements on a same line. It might complexify debugging

test/linalg/test_linalg_permutation.fypp
```diff
@@ -0,0 +1,274 @@
...
+
+  !> Test 12: Identity permutation is a no-op
+  subroutine test_identity_noop(error)
+    type(error_type), allocatable, intent(out) :: error
+    type(permutation_type) :: p
+    real(dp) :: A(3,3), B(3,3)
+    integer(ilp) :: i, j
+
+    p = permutation_identity(3)
+    do i = 1, 3
```
jvdp1: do concurrent could be used here

test/linalg/test_linalg_permutation.fypp
```diff
@@ -0,0 +1,274 @@
...
+  subroutine test_types_support(error)
+    type(error_type), allocatable, intent(out) :: error
+    type(permutation_type) :: p
+    complex(dp) :: c_in(3), c_out(3)
+    integer(int32) :: i_in(3), i_out(3)
+
+    p = permutation_init([3_ilp, 1_ilp, 2_ilp])
+
+    ! Complex support
+    c_in = [(1.0_dp, -1.0_dp), (2.0_dp, -2.0_dp), (3.0_dp, -3.0_dp)]
```
jvdp1: could be defined as `parameter`

src/linalg/stdlib_linalg.fypp
```diff
@@ -70,6 +70,19 @@ module stdlib_linalg
   public :: is_hermitian
   public :: is_triangular
   public :: is_hessenberg
+
+  ! Permutation matrix operations
+  public :: permutation_type
+  public :: permutation_identity
+  public :: permutation_init
+  public :: permutation_is_valid
```
jvdp1: I wonder if the names are fine, e.g., should it be `is_valid_permutation` (to be similar to other functions)?

jvdp1 (commented):
Thank you @edish-github for this PR.
Overall it looks good. However, I wonder if a DT is appropriate for this feature.
Providing some of the functions that would require the permutation vector would be enough for most cases.

edish-github:
Thanks for the detailed review @jvdp1! Addressed in the latest commit:
- Pure on the identity/init/validity procs, do concurrent in permutation_invert and test_identity_noop, parameter on the test constants, the allocate shortening, and split the multi-assignment. Left is_valid_permutation's loops as plain do since they return early on invalid entries.

- Renamed permutation_is_valid → is_valid_permutation to match is_hermitian/is_triangular/etc and also left the rest of the family as permutation_* — those are operations rather than property checks, and a couple (invert, compose) would collide with existing public names.

And about the DT question,  the finalized flag is doing real work: the type carries both a direct 1..n mapping and a LAPACK swap sequence under one API with the permutation_from_lapack bridging the two and a plain-vector version would push that bookkeeping onto every caller but yeah happy to revisit if you have a specific case in mind where the DT gets in the way.

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,244 @@
...
+contains
+
+  !> Create an identity permutation of size n
+  pure module function permutation_identity(n) result(p)
+    integer(ilp), intent(in) :: n
+    type(permutation_type) :: p
+
+    p%n = n
+    p%finalized = .true.
+    allocate(p%perm(n))
```
loiseaujc: `arange` should already allocate internally its result so you can remove line 33.

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,244 @@
...
+        seen(p%perm(i)) = .true.
+      end do
+    else
+      ! For LAPACK pivots, just check bounds (duplicates are valid for swapping)
+      do i = 1, size(p%perm)
+        if (p%perm(i) < 1 .or. p%perm(i) > p%n) return
+      end do
+    end if
+    valid = .true.
+  end function
```
loiseaujc, line 73: This function has multiple `return` points.
A better practice is to have only one.
I guess you could probably rewrite it along the lines of

```fortran
valid = .false.
if ( (allocated(p%perm)) .and. (p%n >= 1) .and. (size(p%perm) == p%n)) then
    if (p%finalized) then
        ...
    else
        ...
    endif
endif
```

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,244 @@
...
+  module function permutation_invert(p) result(inv)
+    type(permutation_type), intent(in) :: p
+    type(permutation_type) :: inv, fp
+    integer(ilp) :: i
+
+    if (p%finalized) then
+      fp = p
+    else
+      fp = permutation_finalize(p)
+    end if
```
loiseaujc, line 86: I suppose you could condense these lines down to

```fortran
fp = merge(p, permutation_finalize(p), p%finalized)
```

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,244 @@
...
+    integer(ilp), intent(in) :: ipiv(:)
+    integer(ilp), intent(in) :: n
+    type(linalg_state_type), intent(out), optional :: err
+    type(permutation_type) :: p
+    type(linalg_state_type) :: err0
+    integer(ilp) :: i, tmp
+
+    p%n = n
+    p%finalized = .true.
+    allocate(p%perm(n))
```
loiseaujc: Same as before, `arange` already handles the allocation internally so Line 111 is not needed.

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,244 @@
...
+
+    ! Start with identity
+    p%perm = arange(1_ilp, n)
+
+    ! Replay swap sequence forward
+    do i = 1, min(n, int(size(ipiv), ilp))
```
loiseaujc:
```fortran
do i = 1, min(n, size(ipiv, kind=ilp))
```

should be enough. No need for the `int`.

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,244 @@
...
+  function permutation_finalize(p) result(fp)
+    type(permutation_type), intent(in) :: p
+    type(permutation_type) :: fp
+
+    if (p%finalized) then
+      fp = p
+    else
+      ! Treat swap sequence as LAPACK pivots
+      fp = permutation_from_lapack(p%perm, p%n)
+    end if
```
loiseaujc, line 140: Not entirely sure, but couldn't these lines be reduced to

```fortran
fp = merge(p, permutation_from_lapack(p%perm, p%n), p%finazlize)
```

or would it incur additional computation to evaluate both possibilities?

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,244 @@
...
+  !> Apply p2 first, then p1
+  module function permutation_compose(p1, p2) result(p)
+    type(permutation_type), intent(in) :: p1, p2
+    type(permutation_type) :: p
+
+    p%n = p1%n
+    p%finalized = .true.
+    allocate(p%perm(p%n))
+
+    p%perm = p2%perm(p1%perm)
```
loiseaujc, line 156: I don't think the `allocate` is needed here since the `=` should handle everything right away.
May be add a check to ensure that `size(p1) == size(p2)` ?

src/linalg/stdlib_linalg_permutation.fypp
```diff
@@ -0,0 +1,244 @@
...
+    type(linalg_state_type) :: err0
+    integer(ilp) :: i
+
+    if (.not. p%finalized) then
+      err0 = linalg_state_type("permutation_to_matrix", LINALG_VALUE_ERROR, "Permutation must be finalized to convert to matrix.")
+      call linalg_error_handling(err0, err)
+      return
+    end if
+    A = 0
+    do i = 1, p%n
```
loiseaujc:
```fortran
do concurrent( i = 1:p%n)
```

loiseaujc (changes requested):
Hej,

Sorry for the long delay, I have been extremely busy the past two months.
If you are still working on this, I have just added some comments.
It looks pretty good overall.

edish-github:
@loiseaujc  Thanks for the review! I just pushed a commit addressing your feedback.

A quick summary of the changes:
Stripped out the redundant allocate() calls since arange and the assignment operators handle them under the hood.
Refactored is_valid_permutation to use a single exit point so it's easier to read.
Updated the loops to do concurrent and fixed the int() wrappers to use kind=ilp.

One quick note on the merge() suggestion: I actually kept the if/else blocks for permutation_invert and permutation_finalize because fortran's merge evaluates both arguments eagerly, using it would force the code to run calls (like permutation_from_lapack) on every pass.

PR #1171, opened 2026-03-31
feat: add base64 support

### LINKED ISSUE #916: base64
### Motivation

[Base64](https://en.wikipedia.org/wiki/Base64) is an encoding scheme to translate binary data into a set of 64 characters (the Base64 alphabet). It is often used to transfer binary data over channels which only support text. One notable use case of Base64 is in the [VTK XML format](https://docs.vtk.org/en/latest/design_documents/VTKFileFormats.html#xml-file-formats). 

A Fortran interface might look like:
```fortran
! for T in {real, integer, logical, complex}
function base64_encode(data) result(str)
   type(T), intent(in) :: data(..)
   character(len=:), allocatable :: str
end function

function base64_decode(str) result(res)
   character(len=:), allocatable :: res    ! Treated as a byte-array
end function
```
The decoding phase would need to be followed by `transfer` or internal I/O to recover the particular numerical type. Alternatively one could decode directly to a given type by using a `mold` argument to retrieve a particular type or use a generic subroutine interface.

### Prior Art

Available in multiple third-party libraries:
- Fortran: [BeFor64](https://github.com/szaghi/BeFoR64), [demo_base64.f90](https://fortran-lang.discourse.group/t/small-fortran-tools/8793/9)
- C and C++: [base64](https://github.com/tobiaslocker/base64) (1), [base64](https://github.com/aklomp/base64) (2), [base64](https://github.com/zhicheng/base64) (3), [simdutf](https://github.com/simdutf/simdutf), ...

Also available in standard libraries of other programming languages:
- MATLAB: [`matlab.net.base64encode`](https://www.mathworks.com/help/matlab/ref/matlab.net.base64encode.html), [`matlab.net.base64decode`](https://www.mathworks.com/help/matlab/ref/matlab.net.base64decode.html)
- Python: [`b64encode`](https://docs.python.org/3/library/base64.html#base64.b64encode), [`b64decode`](https://docs.python.org/3/library/base64.html#base64.b64decode)
- Julia: [Base64](https://docs.julialang.org/en/v1/stdlib/Base64/)

The best third-party libraries use SIMD instructions to make the encoding/decoding fast. Details can be found in the works of Muła and Lemire (https://doi.org/10.1145/3132709, https://doi.org/10.1002/spe.2777).

### Additional Information

In principle could also be exposed as a language feature: https://github.com/j3-fortran/fortran_proposals/issues/330

RatanKokal:
Adds support for base64

Solves issue: https://github.com/fortran-lang/stdlib/issues/916

This PR provides a portable, branchless scalar implementation. A follow-up PR is in development to add an AVX2-accelerated backend for supported x86_64 architectures.


src/strings/stdlib_base64.fypp
```diff
@@ -0,0 +1,324 @@
+! SPDX-Identifier: MIT
+#:include "common.fypp"
+
```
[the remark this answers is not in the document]
RatanKokal: fixed

src/strings/stdlib_base64.fypp
```diff
@@ -0,0 +1,324 @@
...
+            
+            if (val >= 0) then
+                ! If we see a valid char AFTER padding has started (e.g., "AA=A"), it's invalid
+                if (pad_chars > 0) then
+                    res = ""
+                    return
+                end if
+                
+                n_chars = n_chars + 1
+                buffer = ior(ishft(buffer, 6), val)
```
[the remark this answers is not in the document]
RatanKokal: fixed

src/strings/stdlib_base64.fypp
```diff
@@ -0,0 +1,324 @@
...
+        allocate(bytes((len(str) * 3) / 4 + 3)) 
+        
+        buffer = 0_int32
+        bits = 0
+        p = 1
+        n_chars = 0
+        pad_chars = 0
+
+        do i = 1, len(str)
+            val = DEC_LUT(iachar(str(i:i)))
```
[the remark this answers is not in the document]
RatanKokal: fixed

test/string/test_base64.f90
```diff
@@ -0,0 +1,154 @@
...
+        testsuite = [ &
+            new_unittest("base64-known-vectors", test_known_vectors), &
+            new_unittest("base64-decode-whitespace", test_decode_whitespace), &
+            new_unittest("base64-decode-invalid", test_decode_invalid), &
+            new_unittest("base64-roundtrip-int32", test_roundtrip_int32), &
+            new_unittest("base64-roundtrip-real", test_roundtrip_real), &
+            new_unittest("base64-roundtrip-complex", test_roundtrip_complex), &
+            new_unittest("base64-roundtrip-logical", test_roundtrip_logical), &
+            new_unittest("base64-rank0", test_rank0_encode) &
+            ]
```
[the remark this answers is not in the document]
RatanKokal: added

test/string/test_base64.f90
```diff
@@ -0,0 +1,154 @@
...
+    subroutine test_roundtrip_logical(error)
+        type(error_type), allocatable, intent(out) :: error
+        logical(lk) :: vals(5), got(5)
+        character(len=:), allocatable :: enc, dec
+
+        vals = [.true._lk, .false._lk, .true._lk, .true._lk, .false._lk]
+
+        enc = base64_encode(vals)
+        dec = base64_decode(enc)
+
```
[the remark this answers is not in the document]
RatanKokal: fixed

src/strings/stdlib_base64_decode.fypp
```diff
@@ -0,0 +1,207 @@
...
+        allocate(character(len=(len(str) / 4) * 3 + 3) :: res)
+        
+        call base64_decode_into(str, res, decoded_len, error_flag)
+        
+        if (error_flag) then
+            deallocate(res)
+            allocate(character(len=0) :: res)
+        else
+            res = res(1:decoded_len)
+        end if
```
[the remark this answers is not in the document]
RatanKokal: fixed

src/strings/stdlib_base64_encode.fypp
```diff
@@ -0,0 +1,314 @@
...
+        integer :: str_len
+        integer :: i, j
+        integer(int32) :: triplet 
+
+        str = ''
+        encoded_len = 0
+        error_flag = .false.
+
+        nbytes = size(bytes, kind=c_size_t)
+        if (nbytes == 0) return
```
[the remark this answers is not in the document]
RatanKokal: implemented a single checker

src/strings/stdlib_base64_encode.fypp
```diff
@@ -0,0 +1,314 @@
+! SPDX-Identifier: MIT
+#:include "common.fypp"
+
+module stdlib_base64_encode
+    use stdlib_kinds, only: sp, dp, xdp, qp, int8, int16, int32, int64, lk
+    use, intrinsic :: iso_c_binding, only: c_ptr, c_f_pointer, c_loc, c_size_t, c_int8_t, c_bool
```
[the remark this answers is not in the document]
RatanKokal: solved

src/strings/stdlib_base64_decode.fypp
```diff
@@ -0,0 +1,207 @@
...
+        error_accum = 0
+
+        ! Fortran Scalar Decoder Loop
+        do i = 1, slen - 7, 4
+            if (clean_mode) then
+                v0 = int(DT(iand(int(iachar(str(i:i)), int32), 255_int32)), int32)
+                v1 = int(DT(iand(int(iachar(str(i+1:i+1)), int32), 255_int32)), int32)
+                v2 = int(DT(iand(int(iachar(str(i+2:i+2)), int32), 255_int32)), int32)
+                v3 = int(DT(iand(int(iachar(str(i+3:i+3)), int32), 255_int32)), int32)
+            else
```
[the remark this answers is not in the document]
RatanKokal: fixed

src/strings/stdlib_base64_decode.fypp
```diff
@@ -0,0 +1,207 @@
...
+            allocate(filtered(len(str)))
+            j = 0
+            do i = 1, len(str)
+                c = iand(int(iachar(str(i:i)), int32), 255_int32)
+                
+                ! ALWAYS write the byte to the next available slot
+                filtered(j + 1) = int(c, int8)
+                
+                ! Add 0 (if space) or 1 (if valid) to the counter
+                j = j + IS_VAL(c)
```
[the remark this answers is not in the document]
RatanKokal: fixed

src/strings/stdlib_base64_encode.fypp
```diff
@@ -0,0 +1,314 @@
...
+#:for k1, t1, _, _ in REAL_KINDS_TYPES
+    function base64_encode_real_${k1}$(data) result(str)
+        ${t1}$, intent(in), target, contiguous :: data(..)
+        character(len=:), allocatable, target :: str
+        integer(int8), pointer :: raw(:)
+        integer(int64) :: nbytes
+        nbytes = int(size(data), int64) * storage_size(data) / 8
+        if (nbytes == 0) then; str = ""; return; end if
+        call c_f_pointer(c_loc(data), raw, [nbytes])
+        str = base64_encode_bytes(raw)
```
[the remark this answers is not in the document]
RatanKokal: revamped implementation to allow the c_loc in the faster subroutine only

RatanKokal:
@jvdp1 Could you please have a look? I have tackled all possible issues. 

Working on a performance improvement using avx2 specific instructions which was the ultimate goal of the resources shared in the issue. Upon completion, would increase speeds drastically.

Since fortran is depndent on autovectorizers, it may not always lead to the optimall register management as can be controlled using C intrinsics.

Will still have a architecture fallback. Architecture specific kernels can be added to increase performance in future.

src/strings/stdlib_base64.fypp
```diff
@@ -0,0 +1,22 @@
...
+
+module stdlib_base64
+    use stdlib_base64_encode, only: base64_encode, base64_encode_into
+    use stdlib_base64_decode, only: base64_decode, base64_decode_into
+    
+    implicit none
+    private
+
+    public :: base64_encode, base64_encode_into
+    public :: base64_decode, base64_decode_into
```
jvdp1: Why did you use this approach instead of using submodules?

src/strings/stdlib_base64_decode.fypp
```diff
@@ -0,0 +1,212 @@
...
+
+module stdlib_base64_decode
+    use stdlib_kinds, only: int8, int32
+    implicit none
+    private
+
+    public :: base64_decode
+    public :: base64_decode_into
+
+    integer(int8), parameter :: DT(0:255) = int( [ &
```
jvdp1: What are these number? Please provide some info.

src/strings/stdlib_base64_decode.fypp
```diff
@@ -0,0 +1,212 @@
...
+        -1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1, &
+        -1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1, &
+        -1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1, &
+        -1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1, &
+        -1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1, &
+        -1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1  &
+    ], int8)
+
+    ! 0 for whitespace/control chars, 1 for valid chars
+    integer(int32), parameter :: IS_VAL(0:255) = int([ &
```
jvdp1: what are these values?

src/strings/stdlib_base64_decode.fypp
```diff
@@ -0,0 +1,212 @@
...
+            decoded_len = (slen / 4) * 3 - n_pad
+            if (len(res) < decoded_len) then; error_flag = .true.; decoded_len = 0; return; end if
+        end if
+
+        error_accum = 0
+
+        ! Fortran Scalar Decoder Loop
+        do i = 1, slen - 7, 4
+            if (clean_mode) then
+                raw_c = int(iachar(str(i:i)), int32);     unicode_accum = ior(unicode_accum, raw_c); v0 = int(DT(iand(raw_c, 255_int32)), int32)
```
jvdp1: I suggest to avoid multiple loc on a same raw. Debugging is often more complex.

```suggestion
                raw_c = int(iachar(str(i:i)), int32)
                unicode_accum = ior(unicode_accum, raw_c)
                v0 = int(DT(iand(raw_c, 255_int32)), int32)
```

src/strings/stdlib_base64_encode.fypp
```diff
@@ -0,0 +1,203 @@
...
+
+module stdlib_base64_encode
+    use stdlib_kinds, only: sp, dp, xdp, qp, int8, int16, int32, int64, lk
+    use, intrinsic :: iso_c_binding, only: c_size_t, c_bool
+    implicit none
+    private
+
+    public :: base64_encode, base64_encode_into
+
+    character(len=64), parameter :: ALPHABET = &
```
jvdp1: this could be imported from `stdlib_ascii`

jvdp1 (changes requested):
thank you @RatanKokal for this PR. Overall it looks already very good to me.

Maybe some general comments:
* I suggest to use `state_type` (provided by `stdlib_error` to handle errors
* I think that some procedures could be declared `pure`
* Is there any reason to prefer `module` instead of using `submodule`?

src/strings/stdlib_base64_decode.fypp
RatanKokal: fixed, was surpassing the 132 char limit too in the build process

src/strings/stdlib_base64_encode.fypp
RatanKokal: fixed

src/strings/stdlib_base64_decode.fypp
RatanKokal: `DT` is an O(1) branchless lookup table that maps the 256 ASCII characters directly to their 6-bit Base64 integer values. By deliberately mapping all invalid characters to `-1`, we can safely decode the input without any `if` statements and validate the entire stream at the end using a single bitwise `OR` reduction, which significantly boosts loop vectorization.

I have added a short line about it in the file too.

src/strings/stdlib_base64_decode.fypp
RatanKokal: `IS_VAL` is a branchless stream-compaction mask used during the despacing step, mapping whitespace/control characters to `0` and valid Base64 characters to `1`. This allows the pointer to advance mathematically (`j = j + IS_VAL(c)`) rather than conditionally, keeping the CPU pipeline full by avoiding branch mispredictions on unpredictable whitespace.

I have added a short line about it in the file too.

src/strings/stdlib_base64.fypp
RatanKokal: There was no specific technical reason. I originally used the separate module approach for organization. However, after looking into the benefits of submodules for reducing compilation cascades and maintaining a cleaner public interface, I have refactored the implementation to use them.

RatanKokal:
Hey @jvdp1, could you please have a look. I have encorporated all the feedback.

src/core/stdlib_ascii.fypp
```diff
@@ -65,6 +65,8 @@ module stdlib_ascii
     character(len=*), public, parameter :: letters = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz" !! A .. Za .. z
     character(len=*), public, parameter :: uppercase = letters(1:26) !! A .. Z
     character(len=*), public, parameter :: lowercase = letters(27:) !! a .. z
+    character(len=*), public, parameter :: base64_alphabet = &
+        "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/" !! RFC 4648 Base64 alphabet
```
jvdp1:
```suggestion
        letters//digits//"+/" !! RFC 4648 Base64 alphabet
```

src/strings/stdlib_base64.fypp
```diff
@@ -0,0 +1,177 @@
...
+!> This module provides procedures to safely encode and decode data
+!> using the standard Base64 encoding scheme defined in RFC 4648.
+!>
+!> @note
+!> **Experimental:** This API is currently considered experimental and 
+!> may be subject to change in future releases of `stdlib`.
+
+module stdlib_base64
+    use stdlib_kinds, only: sp, dp, xdp, qp, int8, int16, int32, int64, lk
+    use, intrinsic :: iso_c_binding, only: c_size_t, c_bool
```
jvdp1: Are these not provided by `stdlib_kinds`, please check.

src/strings/stdlib_base64_decode.fypp
```diff
@@ -0,0 +1,216 @@
...
+contains
+
+    pure module subroutine base64_decode_into(str, res, decoded_len, err_state, skip_despace)
+        character(len=*), intent(in)  :: str
+        character(len=*), intent(out) :: res
+        integer, intent(out) :: decoded_len
+        type(state_type), intent(out) :: err_state
+        logical, intent(in), optional :: skip_despace
+
+        integer(int8), allocatable :: filtered(:)
```
jvdp1, line 16:
```suggestion
        integer(int8) :: filtered(lem(str))
```

src/strings/stdlib_base64_decode.fypp
```diff
@@ -0,0 +1,216 @@
...
+
+            decoded_len = (slen / 4) * 3 - n_pad
+            if (len(res) < decoded_len) then
+                err_state = state_type("base64_decode_into", STDLIB_VALUE_ERROR, &
+                    "Output buffer too small")
+                decoded_len = 0
+                return
+            end if
+        else
+            allocate(filtered(len(str)))
```
jvdp1, line 54:
```suggestion
```

src/strings/stdlib_base64_decode.fypp
```diff
@@ -0,0 +1,216 @@
...
+            j = 0
+            do i = 1, len(str)
+                raw_c = int(iachar(str(i:i)), int32)
+                unicode_accum = ior(unicode_accum, raw_c)
+                c = iand(raw_c, 255_int32)
+                filtered(j + 1) = int(c, int8)
+                j = j + IS_VAL(c)
```
jvdp1: `c` will be added all cases, and potentially overwritten if is_val ==0. I find the following cleaner:
```suggestion
                if (is_val == 0) cycle)
                j = j + 1
                filtered(j) = int(c, int8)

```

Note that in this case, `is_val` could be ` `logical`

src/strings/stdlib_base64_encode.fypp
```diff
@@ -0,0 +1,207 @@
+! SPDX-Identifier: MIT
+#:include "common.fypp"
+
+submodule(stdlib_base64) stdlib_base64_encode
+    use stdlib_error, only: state_type, STDLIB_SUCCESS, STDLIB_VALUE_ERROR
+
+contains
+
+#:set RANKS = range(16)
```
jvdp1: I suggest to use the same strategy as in `stdlib_stats`. some compilers do not support 15 dims.

src/strings/stdlib_base64_encode.fypp
```diff
@@ -0,0 +1,207 @@
...
+
+#:for k1, t1, _, _ in REAL_KINDS_TYPES
+    module function base64_encode_real_${k1}$(data) result(str)
+        ${t1}$, intent(in), target, contiguous :: data(..)
```
jvdp1: Why is `target` mentioned?

src/strings/stdlib_base64_encode.fypp
```diff
@@ -0,0 +1,207 @@
...
+        rank(${r}$)
+            str = base64_encode_bytes(transfer(data, [0_int8], int(nbytes)))
+        #:endfor
+        end select
+    end function base64_encode_int_${k1}$
+#:endfor
+
+#:for k1, t1, _ in CMPLX_KINDS_TYPES
+    module function base64_encode_cmplx_${k1}$(data) result(str)
+        ${t1}$, intent(in), target, contiguous :: data(..)
```
jvdp1:
```suggestion
        ${t1}$, intent(in), contiguous :: data(..)
```

src/strings/stdlib_base64_encode.fypp
```diff
@@ -0,0 +1,207 @@
...
+            str = base64_encode_bytes(transfer(data, [0_int8], int(nbytes)))
+        #:endfor
+        end select
+    end function base64_encode_logical_${k1}$
+#:endfor
+
+    module function base64_encode_bytes(bytes) result(str)
+        integer(int8), intent(in), target, contiguous :: bytes(:)
+        character(len=:), allocatable, target :: str
+        integer(c_size_t) :: nbytes
```
jvdp1: Could you add a comment to explain why nbytes must be of type `c_size_t`?

test/string/test_base64.f90
```diff
@@ -0,0 +1,208 @@
...
+        call check(error, all(got == vals))
+    end subroutine test_roundtrip_int32
+
+    subroutine test_roundtrip_real(error)
+        type(error_type), allocatable, intent(out) :: error
+
+        real(dp) :: vals(4), got(4)
+        character(len=:), allocatable :: enc, dec
+
+        vals = [1.5_dp, -2.25_dp, 0.125_dp, 9.0_dp]
```
jvdp1: could be declared a `parameter`

test/string/test_base64.f90
```diff
@@ -0,0 +1,208 @@
...
+        call check(error, all(got == vals))
+    end subroutine test_roundtrip_real
+
+    subroutine test_roundtrip_complex(error)
+        type(error_type), allocatable, intent(out) :: error
+
+        complex(dp) :: vals(3), got(3)
+        character(len=:), allocatable :: enc, dec
+
+        vals = [cmplx(1.0_dp, 2.0_dp, dp), cmplx(-3.0_dp, 0.5_dp, dp), cmplx(0.0_dp, -4.0_dp, dp)]
```
jvdp1: could be declared as `parameter`

jvdp1 (commented):
Thank you @RatanKokal . this PR is taking a good shape. Below are a few comments. Could you also add the specs, please?

src/core/stdlib_ascii.fypp
RatanKokal: applied

src/strings/stdlib_base64.fypp
RatanKokal: c_bool was part
c_size_t was not

src/strings/stdlib_base64_decode.fypp
RatanKokal: actually for larger srings, I think its safer to allocate on heap to prevent stack overflow

src/strings/stdlib_base64_decode.fypp
RatanKokal: actually for larger srings, I think its safer to allocate on heap to prevent stack overflow

src/strings/stdlib_base64_decode.fypp
RatanKokal: Initial idea was branchless solely for perf reasons. Will check the branch prediction quality and change if no significant performance difference is observed

src/strings/stdlib_base64_encode.fypp
RatanKokal: fixed

src/strings/stdlib_base64_encode.fypp
RatanKokal: target is a compiler hint, ennsuring maximum memory efficiency by forcing pass-by-reference.

src/strings/stdlib_base64_encode.fypp
RatanKokal: added

test/string/test_base64.f90
RatanKokal: applied

test/string/test_base64.f90
RatanKokal: applied

RatanKokal:
Hi @jvdp1,
I have implemented most of the changes as per the comments.
Also updated the specs.

src/strings/stdlib_base64.fypp
```diff
@@ -0,0 +1,177 @@
...
+
+    implicit none
+    private
+
+    public :: base64_encode, base64_encode_into
+    public :: base64_decode, base64_decode_into
+
+    ! Branchless RFC 4648 decode map: byte -> 6-bit value, invalid -> -1.
+    ! The main loop OR-reduces values and checks once at the end.
+    integer(int8), parameter :: DT(0:255) = int( [ &
```
jalvesz: Can you please provide a more explicit name to this string list. `base64_dictionary` or something of the sorts.

src/strings/stdlib_base64.fypp
```diff
@@ -0,0 +1,177 @@
...
+    !> Use this form when convenience matters more than avoiding allocations.
+    !>
+    !> Example:
+    !>```fortran
+    !> character(len=:), allocatable :: encoded
+    !> integer(int8) :: bytes(3) = [77_int8, 97_int8, 110_int8]
+    !> encoded = base64_encode(bytes)  ! "TWFu"
+    !>```
+    interface base64_encode
+#:for k1, t1, _, _ in REAL_KINDS_TYPES
```
jalvesz: Please align fypp macros to the inner scope where they take effect.

src/strings/stdlib_base64_encode.fypp
jalvesz: Where in the fortran standard does it specify such conclusion?
`target` has a very specific purpose: signaling the compiler that a variable can be "pointed to" by a pointer variable. Telling the compiler that a variable is pontetially a target, can prevent agressive optimization.

Also, in Fortran, arguments are passed by reference.

Please remove this attribute.

src/strings/stdlib_base64_encode.fypp
jalvesz: I think this simply requires `integer(int64)`, there is no need to invoke iso_c_binding here.

src/strings/stdlib_base64_decode.fypp
```diff
@@ -0,0 +1,217 @@
...
+                return
+            end if
+        end if
+
+        error_accum = 0
+
+        ! Fortran Scalar Decoder Loop
+        do i = 1, slen - 7, 4
+            if (clean_mode) then
+                raw_c = int(iachar(str(i:i)), int32)
```
jalvesz: Couple of comments:
* `int(iachar(str(i:i)), int32)` is not required => `iachar(str(i:i), int32)`. iachar takes a kind optional argument for that exact purpose.
* All these expressions could be perfectly vectorized and make the whole code more compact:
```fortran
            if (clean_mode) then
                raw_c(1:4) = iachar(str(i:i+3), int32)
                v(1:4) = int(DT(iand(raw_c, 255_int32)), int32)
            else
                v(1:4) = int(DT(iand(int(filtered(i), int32), 255_int32)), int32))
            end if
            ... etc
```
You could also parametrize the vector register size with : 
```fortran
integer, parameter :: chunk_size = 4
integer(int32) :: v(chunk_size)
... etc
```

src/strings/stdlib_base64_encode.fypp
jalvesz: I agree, target is not needed here, please remove.
The `target` attribute in fortran has a very specific use and it is to signal that a variable will be pointed to.

src/strings/stdlib_base64_encode.fypp
```diff
@@ -0,0 +1,209 @@
...
+        rank(${r}$)
+            str = base64_encode_bytes(transfer(data, [0_int8], int(nbytes)))
+        #:endfor
+        end select
+    end function base64_encode_cmplx_${k1}$
+#:endfor
+
+#:for k1, t1 in LOG_KINDS_TYPES
+    module function base64_encode_logical_${k1}$(data) result(str)
+        ${t1}$, intent(in), target, contiguous :: data(..)
```
jalvesz:
```suggestion
        ${t1}$, intent(in), contiguous :: data(..)
```

src/strings/stdlib_base64_encode.fypp
```diff
@@ -0,0 +1,209 @@
...
+        #:for r in RANKS
+        rank(${r}$)
+            str = base64_encode_bytes(transfer(data, [0_int8], int(nbytes)))
+        #:endfor
+        end select
+    end function base64_encode_logical_${k1}$
+#:endfor
+
+    module function base64_encode_bytes(bytes) result(str)
+        integer(int8), intent(in), target, contiguous :: bytes(:)
```
jalvesz:
```suggestion
        integer(int8), intent(in), contiguous :: bytes(:)
```

src/strings/stdlib_base64_encode.fypp
```diff
@@ -0,0 +1,209 @@
...
+                iand(ishft(triplet, -12_int32), 63_int32) + 1 : iand(ishft(triplet, -12_int32), 63_int32) + 1)
+            str(j+2:j+2) = base64_alphabet( &
+                iand(ishft(triplet, -6_int32), 63_int32) + 1 : iand(ishft(triplet, -6_int32), 63_int32) + 1)
+            str(j+3:j+3) = "="
+        end if
+
+    end function base64_encode_bytes
+
+    pure module subroutine base64_encode_into(bytes, str, encoded_len, err_state)
+        integer(int8), intent(in), target, contiguous :: bytes(:)
```
jalvesz:
```suggestion
        integer(int8), intent(in), contiguous :: bytes(:)
```

src/strings/stdlib_base64_encode.fypp
```diff
@@ -0,0 +1,209 @@
...
+        character(len=*), intent(out), target :: str
```
jalvesz:
```suggestion
        character(len=*), intent(out) :: str
```

src/strings/stdlib_base64.fypp
```diff
@@ -0,0 +1,177 @@
...
+#:for k1, t1, _ in CMPLX_KINDS_TYPES
+        module function base64_encode_cmplx_${k1}$(data) result(str)
+            ${t1}$, intent(in), target, contiguous :: data(..)
+            character(len=:), allocatable :: str
+        end function base64_encode_cmplx_${k1}$
+#:endfor
+
+#:for k1, t1 in LOG_KINDS_TYPES
+        module function base64_encode_logical_${k1}$(data) result(str)
+            ${t1}$, intent(in), target, contiguous :: data(..)
```
jalvesz:
```suggestion
            ${t1}$, intent(in), contiguous :: data(..)
```

doc/specs/stdlib_base64.md
```diff
@@ -0,0 +1,172 @@
...
+`complex`, `logical`, or `integer(int8)`). It is an `intent(in)` argument.
+
+#### Result value
+
+The result `res` is an allocatable character string (`character(len=:)`)
+containing the Base64 representation of the input `data`.
+
+#### Example
+
+```fortran
```
jalvesz, line 49: Please make examples executable programs and include them within the documentation by importing the file.
You can check other .md files to see how it is done.

jalvesz (changes requested):

src/strings/stdlib_base64_encode.fypp
jvdp1: I agree with @jalvesz : use `integer(int64)` instead of `c_size_t`

src/strings/stdlib_base64_encode.fypp
```diff
@@ -0,0 +1,209 @@
...
+        integer(int32) :: triplet
+
+        nbytes = size(bytes, kind=c_size_t)
+        str_len = 4 * ((int(nbytes) + 2) / 3)
+
+        allocate(character(len=str_len) :: str)
+        if (nbytes == 0) return
+
+        j = 1
+        do i = 1, int(nbytes) - 2, 3
```
jvdp1: usin`int(nbytes)` (that is converting `nbytes` to implicitely `int32`) could generate overflows

RatanKokal:
@jvdp1 @jalvesz I have incorporated the feedback/changes requested. Requesting you to please have a look

src/strings/stdlib_base64_decode.fypp
```diff
@@ -0,0 +1,186 @@
...
+        unicode_accum = 0
+
+        if (len(str) == 0) return
+
+        clean_mode = .false.
+        if (present(skip_despace)) clean_mode = skip_despace
+
+        if (clean_mode) then
+            slen = len(str)
+            if (mod(slen, 4) /= 0) then
```
jalvesz:
```suggestion
            if (mod(slen, chunk_size) /= 0) then
```

src/strings/stdlib_base64_decode.fypp
```diff
@@ -0,0 +1,186 @@
...
+                err_state = state_type("base64_decode_into", STDLIB_VALUE_ERROR, &
+                    "Input length must be a multiple of 4")
```
jalvesz:
```suggestion
                    "Input length must be a multiple of "//to_string(chunk_size))
```

src/strings/stdlib_base64_decode.fypp
```diff
@@ -0,0 +1,186 @@
...
+                return
+            end if
+
+            n_pad = 0
+            if (str(slen:slen) == '=') then
+                n_pad = 1
+                if (str(slen-1:slen-1) == '=') n_pad = 2
+            end if
+
+            decoded_len = (slen / 4) * 3 - n_pad
```
jalvesz:
```suggestion
            decoded_len = (slen / chunk_size) * (chunk_size-1) - n_pad
```

src/strings/stdlib_base64_decode.fypp
```diff
@@ -0,0 +1,186 @@
...
+        else
+            allocate(filtered(len(str)))
+            j = 0
+            do i = 1, len(str)
+                raw_c = iachar(str(i:i), kind=int32)
+                unicode_accum = ior(unicode_accum, raw_c)
+                c = iand(raw_c, 255_int32)
+                if (IS_VAL(c) == 0) cycle
+                    j = j + 1
+                    filtered(j) = int(c, int8)
```
jalvesz:
```suggestion
                j = j + 1
                filtered(j) = int(c, int8)
```

src/strings/stdlib_base64_decode.fypp
```diff
@@ -0,0 +1,186 @@
...
+            end do
+            slen = j
+
+            if (unicode_accum > 255_int32 .or. slen == 0 .or. mod(slen, 4) /= 0) then
```
jalvesz: Please use `chunk_size` in the code. The whole point of a parameter value is to automatically have effect on the code if this single line is changed. Please make sure all the code is conformant to being dependent on `chunk_size` with no free hanging values depending on it (some 4s and 3s that I see around).

jalvesz (changes requested):

RatanKokal:
@jalvesz I have used enc_chunk = 4 and raw_chunk = 3 now, 3 bytes of original data gets encoded as 4 bytes. This would give a even more clear picture in my opinion. Requesting you to please have a look.

src/strings/stdlib_base64_decode.fypp
RatanKokal: Unfortunately, to_string cannot be used here because base64_decode_into must remain a pure subroutine, and calling an impure function results in a compilation error. To remove the hardcoded number while preserving the procedure's purity, I defined an FYPP macro (#:set ENC_CHUNK = 4) and injected it directly into the string at compile time.

RatanKokal:
@jalvesz I have introduced two variables for chunk size, raw chunk and enc chunk which is 3 and 4 respectively since base64 converts 3 bytes of normal text in 4 bytes of encoded text

jalvesz:
@RatanKokal is there a reason for converting the size parameters to fypp parameters? This looks like an overkill complexity.

RatanKokal:
> @RatanKokal is there a reason for converting the size parameters to fypp parameters? This looks like an overkill complexity.

I get error in "Input length must be a multiple of "//to_string(chunk_size)) if not done this way. to_string cannot be used here because base64_decode_into must remain a pure subroutine, and calling an impure function results in a compilation error.

jalvesz:
https://github.com/fortran-lang/stdlib/blob/9a15c7772f1a76a6c497b9f3abb793841fc81f74/src/strings/stdlib_strings_to_string.fypp#L12

The procedure `to_string` is pure. I think the error you got was simply that the module containing it was not `use`d

RatanKokal:
@jalvesz Thanks for the correction. Made the changes and it does work now.

doc/specs/stdlib_base64.md
```diff
@@ -0,0 +1,179 @@
...
+`res =` [[stdlib_base64(module):base64_encode(interface)]] `(data)`
+
+#### Class
+
+Function.
+
+#### Argument
+
+`data`: shall be a contiguous array of an intrinsic type (`real`, `integer`,
+`complex`, `logical`, or `integer(int8)`). It is an `intent(in)` argument.
```
jvdp1: `integer` is mentioned twice. Does it mean that all kinds of `integer` are supported? (and what about other types?)
Maybe:

```suggestion
`data`: shall be a contiguous array of an intrinsic type (`real`, `integer`,
`complex`, `logical`) of any kind. It is an `intent(in)` argument.
```

doc/specs/stdlib_base64.md
```diff
@@ -0,0 +1,179 @@
...
+#### Status
+
+Experimental
+
+#### Description
+
+Encodes bytes into a caller-provided output buffer.
+
+This is the preallocated API for throughput-sensitive workflows. It does not
+allocate and reports status through `err_state`. On success,
```
jvdp1: The sentence is unclear. Does it or not report status?

doc/specs/stdlib_base64.md
```diff
@@ -0,0 +1,179 @@
...
+#### Class
+
+Pure module subroutine.
+
+#### Arguments
+
+`bytes`: shall be a contiguous array of type `integer(int8)`. It is an
+`intent(in)` argument.
+
+`str`: shall be an intrinsic character type. It is an `intent(out)` argument.
```
jvdp1: Is it `allocatable`?

src/strings/stdlib_base64.fypp
```diff
@@ -0,0 +1,176 @@
...
+!> This module provides procedures to safely encode and decode data
+!> using the standard Base64 encoding scheme defined in RFC 4648.
+!>
+!> @note
+!> **Experimental:** This API is currently considered experimental and 
+!> may be subject to change in future releases of `stdlib`.
+
+module stdlib_base64
+    use stdlib_kinds, only: sp, dp, xdp, qp, int8, int16, int32, int64, lk, c_bool
+    use stdlib_ascii, only: base64_dictionary => base64_alphabet
```
jvdp1: Why did you rename it? For readability, I suggest to avoid renaming variables, unless it improves readability, or it avoids issues likes name conflicts.

src/strings/stdlib_base64.fypp
```diff
@@ -0,0 +1,176 @@
...
+
+        module function base64_encode_bytes(bytes) result(str)
+            integer(int8), intent(in), contiguous :: bytes(:)
+            character(len=:), allocatable :: str
+        end function base64_encode_bytes
+
+        !> Encodes bytes into a caller-provided output buffer.
+        !>
+        !> This is the preallocated API for throughput-sensitive workflows.
+        !> It does not allocate and reports status through `err_state`.
```
jvdp1, line 133:
```suggestion
        !> It does not allocate `str` and reports status through `err_state`.
```
or

```suggestion
        !> `str` must be of a defined length. It reports status through `err_state`.
```

src/strings/stdlib_base64_decode.fypp
```diff
@@ -0,0 +1,189 @@
...
+            j = 0
+            do i = 1, len(str)
+                raw_c = iachar(str(i:i), kind=int32)
+                unicode_accum = ior(unicode_accum, raw_c)
+                c = iand(raw_c, 255_int32)
+                if (IS_VAL(c) == 0) cycle
+                j = j + 1
+                filtered(j) = int(c, int8)
+            end do
+            slen = j
```
jvdp1, line 66: some entries of `filtered` might be undefined due to `cycle`. I suggest to mention it in the specs.

src/strings/stdlib_base64_decode.fypp
```diff
@@ -0,0 +1,189 @@
...
+        error_accum = 0
+
+        ! Fortran Scalar Decoder Loop
+        do i = 1, slen - (enc_chunk * 2 - 1), enc_chunk
+            if (clean_mode) then
+                raw_c4 = [(iachar(str(k:k), kind=int32), k = i, i + enc_chunk - 1)]
+                unicode_accum = ior(unicode_accum, ior(raw_c4(1), ior(raw_c4(2), ior(raw_c4(3), raw_c4(4)))))
+                v = int(DT(iand(raw_c4, 255_int32)), int32)
+            else
+                v = int(DT(iand(int(filtered(i:i+enc_chunk-1), int32), 255_int32)), int32)
```
jvdp1, line 98: `DT` (declared as `integer(int8))` seems to be always converted to `int32`. For readability and efficiency, I suggest to declared `DT` as `integer(int32)`

src/strings/stdlib_base64_encode.fypp
```diff
@@ -0,0 +1,211 @@
+! SPDX-Identifier: MIT
+#:include "common.fypp"
+
+submodule(stdlib_base64) stdlib_base64_encode
+    use stdlib_error, only: state_type, STDLIB_SUCCESS, STDLIB_VALUE_ERROR
+    
+    integer, parameter :: enc_chunk = 4
+    integer, parameter :: raw_chunk = 3
```
jvdp1: Since these variables are declared in both modules, I suggest to define them only once in the parent module.

test/string/test_base64.f90
```diff
@@ -0,0 +1,209 @@
...
+        call check(error, base64_encode([int(77, int8), int(97, int8)]) == "TWE=")
+        if (allocated(error)) return
+        call check(error, base64_encode([int(77, int8)]) == "TQ==")
+        if (allocated(error)) return
+
+        call check(error, base64_decode("TWFu") == "Man")
+        if (allocated(error)) return
+        call check(error, base64_decode("TWE=") == "Ma")
+        if (allocated(error)) return
+        call check(error, base64_decode("TQ==") == "M")
```
jvdp1: It would be good to add some message in `check`. It will be helpful to identify which check is failing, in case one of them will fail. Currenty, the developer won't be able to identify it without additional work.

test/string/test_base64.f90
```diff
@@ -0,0 +1,209 @@
...
+        call check(error, base64_decode("AA=A") == "")
+    end subroutine test_decode_invalid
+
+    subroutine test_roundtrip_int32(error)
+        type(error_type), allocatable, intent(out) :: error
+
+        integer(int32) :: vals(4), got(4)
+        character(len=:), allocatable :: enc, dec
+
+        vals = [1_int32, -2_int32, 1024_int32, -4096_int32]
```
jvdp1: could it be declared as`parameter`?

jvdp1 (commented):
thank you @RatanKokal . Overall it looks good to me. Here are some suggestions

test/string/test_base64.f90
RatanKokal: Yes done

test/string/test_base64.f90
RatanKokal: Added

src/strings/stdlib_base64_encode.fypp
RatanKokal: Agreed, done

src/strings/stdlib_base64_decode.fypp
RatanKokal: Sure

src/strings/stdlib_base64_decode.fypp
RatanKokal: I have added comment in the code, since filtered is dev facing

src/strings/stdlib_base64.fypp
RatanKokal: Made these lines more clear

doc/specs/stdlib_base64.md
RatanKokal: No

doc/specs/stdlib_base64.md
RatanKokal: Made it clear, it reports status

src/strings/stdlib_base64.fypp
RatanKokal: I was requested to have a more explicit name like dictionary, and I wanted to have some resemblance with the original RFQ. I have updated this in my latest commit to use base64_dictionary everywhere.

doc/specs/stdlib_base64.md
RatanKokal: Updated the docs to make it clear. Yeah everything is supported.

RatanKokal:
@jvdp1 Thanks for the feedback. I have incorporated all of that in my latest commit. Requesting you to have a look.
Thanks.

RatanKokal:
@jalvesz Are there any more suggestions, or are we good to go with this one?

PR #1179, opened 2026-04-11
Add date_type and time_type sub-types to stdlib_datetime

JAi-SATHVIK:
### Follows up on issue #1147 (https://github.com/fortran-lang/stdlib/issues/1147#issuecomment-4189478629)

**I added two new types to make working with dates and times easier:**

**date_type:** For calendar dates only (Year, Month, Day).
**time_type:** For clock times only (Hours, Minutes, Seconds).

### DIFF
--- a/src/datetime/stdlib_datetime.f90
+++ b/src/datetime/stdlib_datetime.f90
@@ -25,7 +25,7 @@ module stdlib_datetime
         !! version: experimental
         !!
         !! Represents a specific point in time.
-        integer :: year        = 1   !! Year (1-9999)
+        integer :: year        = 1   !! Year
         integer :: month       = 1   !! Month (1-12)
         integer :: day         = 1   !! Day (1-31)
         integer :: hour        = 0   !! Hour (0-23)
@@ -50,7 +50,7 @@ module stdlib_datetime
         !! version: experimental
         !!
         !! Represents a calendar date (no time-of-day component).
-        integer :: year  = 1   !! Year (1-9999)
+        integer :: year  = 1   !! Year
         integer :: month = 1   !! Month (1-12)
         integer :: day   = 1   !! Day (1-31)
     end type date_type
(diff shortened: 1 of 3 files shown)


JAi-SATHVIK:
Hi @jalvesz,
Addressed several datetime issues by enforcing strict ISO date parsing, clamping millisecond values to prevent rounding overflows, and ensuring date arithmetic accounts for full `timedelta` durations. 

Added new regression tests to verify midnight-crossing `timezone` comparisons and strict string validation, ensuring the module handles edge cases like sub-day components and trailing characters correctly.

jvdp1:
@JAi-SATHVIK thank you for this PR. However, Github shows commits that are already merged and are older thant this PR. Could you check that, please, to faciliate its review?

src/datetime/stdlib_datetime.f90
```diff
@@ -43,62 +46,137 @@ module stdlib_datetime
         integer :: milliseconds = 0   !! Milliseconds (0-999)
     end type timedelta_type
 
+    type :: date_type
+        !! version: experimental
+        !!
+        !! Represents a calendar date (no time-of-day component).
+        integer :: year  = 1   !! Year (1-9999)
```
jvdp1: are negative years not supported?

src/datetime/stdlib_datetime.f90
```diff
@@ -302,6 +582,90 @@ pure function td_ge(td1, td2) result(res)
         res = td_to_ms(td1) >= td_to_ms(td2)
     end function td_ge
 
+    pure function date_eq(d1, d2) result(res)
+        type(date_type), intent(in) :: d1, d2
+        logical :: res
+        res = days_from_civil(d1%year, d1%month, d1%day) &
+           == days_from_civil(d2%year, d2%month, d2%day)
```
jvdp1: Would it not be more efficient to compare the components directly, instead of converting it first?

src/datetime/stdlib_datetime.f90
```diff
@@ -489,7 +896,9 @@ function parse_datetime(str, stat) result(dt)
                 if (present(stat)) stat = 1
                 return
             end if
-            dt%millisecond = nint(ms_frac * 1000.0_dp)
+            ! Clamp to [0, 999]: nint can round up to 1000 for values
+            ! like .9996, keeping millisecond in the documented range.
+            dt%millisecond = min(999, max(0, nint(ms_frac * 1000.0_dp)))
```
jvdp1: `nint` is not standard. use int(..., kind=intxxx) instead.

src/datetime/stdlib_datetime.f90
```diff
@@ -43,62 +46,137 @@ module stdlib_datetime
         integer :: milliseconds = 0   !! Milliseconds (0-999)
     end type timedelta_type
 
+    type :: date_type
```
jvdp1, line 49: I wonder if `date_time_type` could be an extension of `date_type`. This would avoid the need of "duplicating" many procedures.

jvdp1 (commented):
thank you @JAi-SATHVIK . Here are some suggestions

src/datetime/stdlib_datetime.f90
JAi-SATHVIK: Interesting suggestion. While extending `date_type` would create an 'is-a' relationship, we would still need separate implementations for most procedures (like addition, subtraction, and full equality) to handle the sub-day components (time/milliseconds) correctly. For now, I've kept them separate to keep the types simple and consistent with the existing logic, but I'm open to refactoring this if you feel strongly about using inheritance here.

src/datetime/stdlib_datetime.f90
JAi-SATHVIK: Yes, now negative years are supported by the underlying algorithm

src/datetime/stdlib_datetime.f90
JAi-SATHVIK: yeah correct, refactored the date comparison functions (`==`,` /=`, `<`, `<=`, `>`, `>=`) to compare the year, month, and day components directly.

cyrilgandon:
Thanks for this PR, looking forward to using it!

One question: is there a reason the pure functions aren't marked `elemental`? For example `day_of_year` — I had to write my own wrapper just to apply it over an array:

```fortran
!! same as day_of_year, but elemental
pure elemental function elemental_day_of_year(dt) result(doy)
    type(datetime_type), intent(in) :: dt
    integer :: doy
    doy = day_of_year(dt)
end function elemental_day_of_year
```

I don't see an obvious reason why they couldn't be `elemental`, but maybe I'm missing something?

jalvesz:
Agree @cyrilgandon I would advise to make any function that can be elemental, made elemental.

PR #1184, opened 2026-05-02
spmv: low-level API based on existing high-level API

jvdp1:
Here is a proposition to add low-level API based on existing high-level API for `spmv`.

It is currently only for `COO`, to illustrate the proposition.

If there are no oppositions, I'll continue the implementation of low-level API for the other sparse matrices.
Some discussions will be needed about the names and API.


jalvesz:
I think this is a good idea @jvdp1 and was exactly one of the reasons for breaking the spmv into files per type. I've been pondering also about which would be the best strategy to interface such low level kernels with also a view into enabling linking against libraries such as MKL (https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2025-2/sparse-blas-level-1-routines.html), NVPL (https://docs.nvidia.com/nvpl/latest/sparse/storage_format/sparse_matrix.html) and AMD (https://rocm.docs.amd.com/projects/rocSPARSE/en/latest/) . They all follow a structuring close to Sparse BLAS level 1,2 and 3, but they do use Classes to abstract away the data handlers for the matrices. I've been wondering if using iso_c_binding it would be possible to have C-preprocessing build-time switchs to either use the internal low-level kernels of stdlib or to link against such libraries.

I guess the first thing that could be tried is to have a "sparse_blas" like folder with the level2 procedures in non-oop version. And from there figure out the next steps ?

jvdp1:
> I think this is a good idea @jvdp1 and was exactly one of the reasons for breaking the spmv into files per type. 

Good! I will push some low-level API, with some tests and specs.

>I've been pondering also about which would be the best strategy to interface such low level kernels with also a view into enabling linking against libraries such as MKL (https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2025-2/sparse-blas-level-1-routines.html), NVPL (https://docs.nvidia.com/nvpl/latest/sparse/storage_format/sparse_matrix.html) and AMD (https://rocm.docs.amd.com/projects/rocSPARSE/en/latest/) . They all follow a structuring close to Sparse BLAS level 1,2 and 3, but they do use Classes to abstract away the data handlers for the matrices. I've been wondering if using iso_c_binding it would be possible to have C-preprocessing build-time switchs to either use the internal low-level kernels of stdlib or to link against such libraries.
I guess the first thing that could be tried is to have a "sparse_blas" like folder with the level2 procedures in non-oop version. And from there figure out the next steps ?

I think that it would be nice to propose interfaces to these optimized libraries. C-preprocessing build-time switchs are probably the easiest and most flexible solution. But as a first step, I agree we should first focus on low-level API.

jalvesz:
Hi @jvdp1 I went through the changes (sorry for the delay) and everything looks good. I don't understand why so many jobs failed but seems to be related to the runners and nothing related to the code changes.

jalvesz:
@jvdp1 maybe updating this PR with the latest commit should retrigger the jobs properly

jalvesz (approved):

jvdp1:
@jalvez I slightly modified the APIs of the `spmv_kernel_' subroutines. I tried to uniformize them as mush as possible.
I still must finalize the specs.

doc/specs/stdlib_sparse.md
```diff
@@ -217,6 +217,94 @@ $$y=\alpha*op(M)*x+\beta*y$$
...
+
+`call ` [[stdlib_sparse_spmv(module):spmv_kernel_coo(interface)]] `(nrows,ncols,data,index,nnz,storage,vec_x,vec_y [,alpha,beta,op])`
+`call ` [[stdlib_sparse_spmv(module):spmv_kernel_csc(interface)]] `(nrows,ncols,data,colptr,row,storage,vec_x,vec_y [,alpha,beta,op])`
+`call ` [[stdlib_sparse_spmv(module):spmv_kernel_csr(interface)]] `(nrows,ncols,data,col,rowptr,storage,vec_x,vec_y [,alpha,beta,op])`
+`call ` [[stdlib_sparse_spmv(module):spmv_kernel_ell(interface)]] `(nrows,ncols,data,index,mnz_p_row,storage,vec_x,vec_y [,alpha,beta,op])`
+`call ` [[stdlib_sparse_spmv(module):spmv_kernel_sellc(interface)]] `(nrows,ncols,data,ia,ja,chunk_size,storage,vec_x,vec_y [,alpha,beta,op])`
+
+### Arguments
+
+For the all formats:
```
jalvesz:
```suggestion
Common arguments for all formats
```

jalvesz (changes requested):

doc/specs/stdlib_sparse.md
```diff
@@ -217,6 +217,94 @@ $$y=\alpha*op(M)*x+\beta*y$$
...
+
+### Description
+
+Provide non-object-oriented sparse matrix-vector product kernels for the current supported sparse matrix types.
+
+$$y=\alpha*op(M)*x+\beta*y$$
+
+### Syntax
+
+`call ` [[stdlib_sparse_spmv(module):spmv_kernel_coo(interface)]] `(nrows,ncols,data,index,nnz,storage,vec_x,vec_y [,alpha,beta,op])`
```
jalvesz, line 235: I guess the doc should be updated to reflect that the kernels mandate all arguments to be non-optional.

jvdp1: You are right! I will update the specs

jalvesz (approved):
LGTM @jvdp1 I think this looks ready!

jvdp1:
> LGTM @jvdp1 I think this looks ready!

Thank you @jalvesz . I'll merge it later today.

PR #1186, opened 2026-05-16
feat (linalg_iterative) : add gmres solver to the iterative methods library

### LINKED ISSUE #1182: Adding GMRES  to iterative_solvers Module
### Motivation

Hi,

I’ve been taking a look at the stdlib_linalg_iterative_solvers module and noticed that there isn’t an implementation of GMRES. Would this be a worthwhile addition to the stdlib? If so, I’d be happy to help; I have some implementations that I can adapt.

Motivation: GMRES is one of the most widely used iterative methods: compared to bicgstab, it offers improved robustness, but it has the drawback of consuming significantly more memory (the Krylov basis requires m vectors of size n, where n is the problem size and m is the restart size).

### Prior Art

_No response_

### Additional Information

_No response_

jalvesz:
This PR proposes adding the generalized minimal residual method (GMRES) to the linalg_iterative module.

It was drafted with the help of CoPilot and received the help of @AlexanderGSC to add the Modified Gram Schmidt scheme for increased robustness.

- [x] Public core kernel implementation
- [x] Standard interface for dense and CSR matrix types
- [x] Unit test
- [x] Examples
- [x] Documentation


AlexanderGSC:
Thanks @jalvesz!

If more tests or documentation need to be added, or if the maintainers think it’s necessary to make any changes, I’d be happy to help.

src/linalg_iterative/stdlib_linalg_iterative_solvers.fypp
```diff
@@ -16,8 +16,9 @@ module stdlib_linalg_iterative_solvers
         enumerator :: stdlib_size_wksp_cg = 3
         enumerator :: stdlib_size_wksp_pcg = 4
         enumerator :: stdlib_size_wksp_bicgstab = 8
+        enumerator :: stdlib_size_wksp_gmres = 3
```
jvdp1: is it on purpose the same as stdlib_size_wksp_cg?

src/linalg_iterative/stdlib_linalg_iterative_solvers.fypp
```diff
@@ -219,6 +241,36 @@ module stdlib_linalg_iterative_solvers
...
     public :: stdlib_solve_bicgstab
 
+    !! version: experimental
+    !!
+    !! [Specifications](../page/specs/stdlib_linalg_iterative_solvers.html#stdlib_solve_gmres)
+    interface stdlib_solve_gmres
+        #:for matrix in MATRIX_TYPES
+        #:for k, t, s in R_KINDS_TYPES
+        module subroutine stdlib_solve_gmres_${matrix}$_${s}$(A,b,x,di,rtol,atol,maxiter,restart,kdim,precond,M,workspace)
+            !! linear operator matrix
```
jvdp1, line 251: should it be `!>` as the docstring precedes the declaration?
```suggestion
            !> linear operator matrix
```

test/linalg/test_linalg_solve_iterative.fypp
```diff
@@ -80,6 +82,87 @@ module test_linalg_solve_iterative
...
+        #:for k, t, s in R_KINDS_TYPES
+        block
+        ${t}$, parameter :: tol = 1000*epsilon(0.0_${k}$)
+        ${t}$ :: A(4,4) = reshape([${t}$ :: 5,  1,  2,  0, &
+                                           0,  4, -1,  1, &
+                                           1,  0,  3,  2, &
+                                           2, -1,  0,  6], [4,4])
+        ${t}$ :: x(4), load(4), xref(4)
+
+        xref = [1.0_${k}$, 2.0_${k}$, -1.0_${k}$, 0.5_${k}$]
```
jvdp1:
```suggestion
        ${t}$ :: x(4), load(4)
        ${t}$, parameter :: xref(*) = [1.0_${k}$, 2.0_${k}$, -1.0_${k}$, 0.5_${k}$]
```

test/linalg/test_linalg_solve_iterative.fypp
```diff
@@ -80,6 +82,87 @@ module test_linalg_solve_iterative
...
+        end block
+
+        block
+        ${t}$, parameter :: tol = 1000*epsilon(0.0_${k}$)
+        ${t}$ :: A(3,3) = reshape([${t}$ :: 10,  1,  2, &
+                                            1, 10,  3, &
+                                            2,  3, 10], [3,3])
+        ${t}$ :: x(3), load(3), xref(3)
+
+        xref = [(137._${k}$/218._${k}$), -(9._${k}$/109._${k}$), (87._${k}$/218._${k}$)]
```
jvdp1:
```suggestion
        ${t}$ :: x(3)
        ${t}$, parameter :: xref(:) = [(137._${k}$/218._${k}$), -(9._${k}$/109._${k}$), (87._${k}$/218._${k}$)]
        ${t}$, parameter :: load(*) = [7.0_${k}$, 1.0_${k}$, 5.0_${k}$]
```

test/linalg/test_linalg_solve_iterative.fypp
```diff
@@ -80,6 +82,87 @@ module test_linalg_solve_iterative
...
+
+        call stdlib_solve_gmres(A, load, x, rtol=tol, maxiter=12)
+        
+        call check(error, norm2(x-xref)<tol*norm2(xref), 'error in GMRES solver (restarted dense test)')
+        if (allocated(error)) return
+        end block
+
+        block
+        ${t}$, parameter :: tol = 1000*epsilon(0.0_${k}$)
+        ${t}$ :: A(3,3) = reshape([${t}$ :: 10,  1,  2, &
```
jvdp1:
```suggestion
        ${t}$, parameter :: A(3,3) = reshape([${t}$ :: 10,  1,  2, &
```

test/linalg/test_linalg_solve_iterative.fypp
```diff
@@ -80,6 +82,87 @@ module test_linalg_solve_iterative
         #:endfor
     end subroutine test_stdlib_solve_pcg
 
+    subroutine test_stdlib_solve_gmres(error)
+        type(error_type), allocatable, intent(out) :: error
+
+        #:for k, t, s in R_KINDS_TYPES
+        block
+        ${t}$, parameter :: tol = 1000*epsilon(0.0_${k}$)
+        ${t}$ :: A(4,4) = reshape([${t}$ :: 5,  1,  2,  0, &
```
jvdp1:
```suggestion
        ${t}$, parameter :: A(4,4) = reshape([${t}$ :: 5,  1,  2,  0, &
```

test/linalg/test_linalg_solve_iterative.fypp
```diff
@@ -80,6 +82,87 @@ module test_linalg_solve_iterative
...
+        #:for k, t, s in R_KINDS_TYPES
+        block
+            !Hilbert sp is almost useless. 
+            #:if k == 'sp' 
+            ${t}$, parameter :: tol = 1.e-0_${k}$
+            #:else
+            ${t}$, parameter :: tol = 1.e-3_${k}$
+            #:endif
+            
+            ${t}$ :: A(n, n), b(n), x(n), x_ref(n)
```
jvdp1:
```suggestion
            ${t}$ :: A(n, n), b(n), x(n)
            ${t}$, parameter :: x_ref(n) = 1.0_${k}$
```

jvdp1 (commented):
Thank you for this addition. Here are some suggestions.

src/linalg_iterative/stdlib_linalg_iterative_solvers.fypp
jalvesz: It is not on purpose the as stdlib_size_wksp_cg. The workspace for the GMRES algorithm is bit more tricky than for the other gradient descent methods as not only are there reference buffer vectors with the size of the problem but also there are arrays depending on the `kdim` (number of internal iterations per restart cycle).

The current workspace design :
r(:) → residual (1 vector)
w(:) → Arnoldi work vector (1 vector)
v(:,1:kdim+1) → Krylov basis (kdim+1 vectors)
z(:,1:kdim) → preconditioned basis (kdim vectors)

Actually the last vector `z` could be reduced to z(:) as the preconditionner is applied per each `j` inner iteration.

I'll review this and try to propose a more lean version in terms of internal storage size. This won't change the fact that the workspace depends not only on `n` (problem size) but also on the selected number of internal iterations `kdim`.

Maybe @AlexanderGSC you have some suggestions here?

src/linalg_iterative/stdlib_linalg_iterative_solvers.fypp
jalvesz: Oh yes it could. In that case all the docstring characters in this file should be changed.

src/linalg_iterative/stdlib_linalg_iterative_solvers.fypp
AlexanderGSC: Hi @jvdp1, @jalvesz!

@jalvesz is right; the size of the workspace depends on the size of the restart (kdim), and there isn’t much that can be done about it. Storing only the last vector from the preconditioner saves a lot of memory – well spotted!

I can think of a few ideas for keeping the interface unchanged, each with its own pros and cons:

- move the allocation of v inside the kernel:
pros: keeps the current interface unchanged.
cons: I don’t know if it’s allowed to reserve memory inside the kernel. At the moment there are variables such as the Hessemberg matrix, the rotation vectors, etc. that are allocated inside the kernel, but they are small, depending only on kdim, which by default is 30, meaning very little memory is used.

- Explicit workspace handling: 
Pros: no memory allocation within the kernel.
Cons: we would need to document that the workspace size is 3+(kdim+1). And we would need to add a check in the code to ensure this is the case. Furthermore, in the interface, the value would not reflect the workspace size.

- Implicit kdim handling: the workspace size is checked, and the kdim size is deduced implicitly. If a value of 33 is set in the interface, gmres would run with a default kdim=30. 
Pros: allows running ‘by default’ or with ‘fine-tuning’ without allocating memory in the kernel.
Cons: the way the workspace is managed becomes a little odd and less straightforward. Restart is a parameter that every version of gmres has.

In my opinion, if it isn’t a problem, I would move the v memory allocation into the kernel. If that is a problem, then we could explore other options.

jalvesz: I gave another round to the proposal. I added a `compact` option to allow switching between a memory efficient or a speed oriented implementation versions. Let me know your thoughts on this.

AlexanderGSC: Hello @jalvesz, 

What you’re suggesting is much better; it’s a definitive solution that would allow for much greater flexibility and provide a clean interface for any method requiring a variable number of vectors. 

Is the Hilbert test causing problems again in single precision? If so, I think the best thing is to remove it, or if necessary, I’ll modify it so that the test is skipped entirely in single precision. Adding the interface is much more important. There’ll be time to fix the test later on. 

Great work!

jalvesz: Yes, the Hilbert test in single precision is causing problems with two of the intel jobs, don't know what is best, if just remove it or try with a jacobi preconditioner ?

AlexanderGSC: I don’t think using a preconditioner will make any difference.

The fact that it fails with Intel compilers suggests to me that they may be using aggressive instruction fusion optimisations that cause some numerical drift. The condition number of the Hilbert's matrix must be of the order of $10^{12}$, which is impossible for SP. 

The simplest solution: reduce the size of the Hilbert matrix to something manageable for SP. With $n=4$ it should work.

If you agree, I can push the change to update the test to $n=4$ directly so we can see if the Intel CI jobs finally turn green.

jalvesz: Indeed! I'm thinking that another option would be to check the condition number estimate and pick a size such that the condition is below some critical threshold ? Using
$$\kappa \approx \frac{(1+\sqrt{2})^{4*N}}{\sqrt{N}}$$
a limit size could be precomputed at compile time

... or simply just do the test for kind >sp with fixed size.

src/linalg_iterative/stdlib_linalg_iterative_solvers_gmres.fypp
```diff
@@ -0,0 +1,304 @@
...
+                        v(:,j+1) = zero_${s}$
+                    end if
+
+                    ! Apply the previously accumulated Givens rotations.
+                    do i = 1, j - 1
+                        temp = cs(i) * h(i,j) + sn(i) * h(i+1,j)
+                        h(i+1,j) = -sn(i) * h(i,j) + cs(i) * h(i+1,j)
+                        h(i,j) = temp
+                    end do
+
```
loiseaujc: For `real` matrices, you could use something along the lines of the following code snippet

```fortran
    use stdlib_linalg_lapack, only: lartg, lasr

    pure module subroutine apply_givens_rotation_rdp(h, c, s)
        implicit none(type, external)
        real(dp), target, contiguous, intent(inout) :: h(:)
        !! k-th column of the Hessenberg matrix.
        real(dp), intent(inout) :: c(:)
        !! Cosine components of the Givens rotations.
        real(dp), intent(inout) :: s(:)
        !! Sine components of the Givens rotations.

        integer(ilp) :: k
        real(dp) :: r
        real(dp), pointer :: hmat(:, :)
        !> Size of the column.
        k = size(h) - 1
        !> Apply previous Givens rotations to this new column.
        hmat(1:k, 1:1) => h(:k)
        call lasr("L", "V", "F", k, 1, c(:k-1), s(:k-1), hmat, k)
        !> Compute the sine and cosine compoennts for the next rotation.
        call lartg(h(k), h(k+1), c(k), s(k), r)
        !> Eliminiate H(k+1, k).
        h(k) = r; h(k+1) = 0.0_dp
    end function apply_givens_rotation_rdp
```

It should be easy enough to template for other kinds and `complex` data.

src/linalg_iterative/stdlib_linalg_iterative_solvers_gmres.fypp
```diff
@@ -0,0 +1,304 @@
...
+                call A%matvec(x, r, alpha=-one_${s}$, beta=one_${s}$, op='N')
+                norm_sq = A%inner_product(r, r)
+
+                if (associated(workspace%callback)) call workspace%callback(x, norm_sq, iter)
+            end do
+        end associate
+
+    contains
+
+        subroutine upper_triangular_solve(h, g, y, n)
```
loiseaujc: Any reason why you implemented backsubstitution yourself instead of using `trtrs` from `lapack` for solving this triangular system?

jalvesz:
@loiseaujc thanks for your input! Indeed you are right on both comments, I added the apply_givens_rotation following your suggestion, including a small comment to LightKrylov. I also changed the triangular solve to rely on trtrs, we have to decide how to properly exit in case of a singular triangular matrix. For the moment I left an exit path `if (info/=0) exit`

@AlexanderGSC I pushed a change bypassing the Hilbert test for `sp` following your suggestion.

loiseaujc:
As far as I know, the triangular matrix should never be singular. The cheap residual check `norm_sq = g(j+1) * g(j+1)` is precisely here for that. As soon as `g(j+1)` is small enough (indicating an almost invariant subspace of `A`), the triangular matrix will be of size ` j x j` and invertible. It comes for the on-the-fly QR factorization of the upper Hessenberg matrix.

jalvesz:
Thanks @loiseaujc for the clarification. So in that case the exit path is useless. Would you say it is ok to remove it or what would you advise here?

loiseaujc:
Well, I guess we could have the standard `linalg_state_type` stuff just to be on the safe side but, as far as I know, we should never encounter an error.

jalvesz:
Right, though introducing that would imply changing a bit the API. If needed we could include that in a future PR. So I guess we can leave it as is for the moment.

example/linalg/example_solve_gmres.f90
```diff
@@ -0,0 +1,24 @@
+program example_solve_gmres
+    use stdlib_kinds, only: dp
+    use stdlib_linalg_iterative_solvers, only: stdlib_solve_gmres, pc_jacobi
+    implicit none
+
+    real(dp), parameter :: A(4,4) = reshape([5.0_dp,  1.0_dp,  2.0_dp, 0.0_dp, &
```
jvdp1: Minor comment: `4` could be defined as `integer, parameter :: n=4`

AlexanderGSC:
Hello @jalvesz @loiseaujc , 

@loiseaujc is absolutely right: the cheap residual check (norm_sq < tolsq) should trigger a breakdown exit ($h_{j+1,j} \approx 0$) well before we ever feed a singular system to trtrs.

What I'm not sure will work from a practical standpoint in floating-point computation is using quadratic tolerances (tolsq) and a hard-coded epsilon test within the Arnoldi loop. For example, if $rtol=10^{-6}$ is set in single precision, for an epsilon that will be on the order of $10^{-7}$, it drives the stopping threshold tolsq down to $10^{-12}$ (assuming a normalized system). In single precision, $10^{-12}$ is indistinguishable from pure round-off noise.

Then, the cheap quadratic check (norm_sq < tolsq) might never be mathematically satisfied. The algorithm will keep iterating, the Krylov basis will lose orthogonality, and the Hessenberg matrix will become highly ill-conditioned, and we end up calling trtrs on a nearly-singular matrix where the diagonal elements $H(j,j)$ have collapsed to near-zero. This can lead to division-by-zero, extreme scaling issues, or NaNs.

Keeping the if (info /= 0) exit path after trtrs is actually a very cheap and robust safety net to prevent runtime crashes, or we could avoid the quadratic problem keeping the tolerances and the residual estimate on a linear scale.

These are minor considerations; for now, we can leave it at that and keep refining it step by step. What do you think?

jalvesz:
Thanks for the feedback @AlexanderGSC so I think we agree and the PR is ready!
Any comments @loiseaujc @jvdp1 ?

src/linalg_iterative/stdlib_linalg_iterative_solvers.fypp
```diff
@@ -219,7 +241,47 @@ module stdlib_linalg_iterative_solvers
...
+        #:endfor
+    end interface
+    public :: stdlib_solve_gmres
+
 contains
+    
+    !------------------------------------------------------------------
+    ! helpers
+    !------------------------------------------------------------------
+    integer(int32) function stdlib_size_wksp_gmres(kdim,compact)
```
jvdp1: Could this function be `pure` (or even `elemental`)?

src/linalg_iterative/stdlib_linalg_iterative_solvers_gmres.fypp
```diff
@@ -0,0 +1,309 @@
...
+            character(1), intent(in) :: op
+            #:if matrix == "dense"
+            call gemv(op, m=size(A,1), n=size(A,2), alpha=alpha, a=A, lda=size(A,1), x=x, incx=1, beta=beta, y=y, incy=1)
+            #:else
+            call spmv(A, x, y, alpha, beta, op)
+            #:endif
+            y = merge(zero_${s}$, y, di_)
+        end subroutine
+
+        subroutine precond_none(x,y,alpha,beta,op)
```
jvdp1, line 288: Could it be `pure`?

src/linalg_iterative/stdlib_linalg_iterative_solvers_gmres.fypp
```diff
@@ -0,0 +1,309 @@
...
+            ${t}$, intent(in) :: x(:)
+            ${t}$, intent(inout) :: y(:)
```
jvdp1, line 290: could it be `intent(out)`?

src/linalg_iterative/stdlib_linalg_iterative_solvers_gmres.fypp
```diff
@@ -0,0 +1,309 @@
...
+            ${t}$, intent(in) :: alpha
+            ${t}$, intent(in) :: beta
+            character(1), intent(in) :: op
+            y = merge(zero_${s}$, x, di_)
+        end subroutine
+
+        subroutine precond_jacobi(x,y,alpha,beta,op)
+            ${t}$, intent(in) :: x(:)
+            ${t}$, intent(inout) :: y(:)
```
jvdp1, line 299: Could it be `intent(out)`?

test/linalg/test_linalg_solve_iterative.fypp
```diff
@@ -80,6 +82,98 @@ module test_linalg_solve_iterative
...
+
+        call stdlib_solve_gmres(A, load, x, rtol=1.e-10_${k}$, precond=pc_jacobi)
+
+        call check(error, norm2(x-xref)<tol*norm2(xref), 'error in GMRES solver (Jacobi preconditioned test)')
+        if (allocated(error)) return
+        end block
+
+        block
+        ${t}$, parameter :: tol = 1000*epsilon(0.0_${k}$)
+        ${t}$, parameter :: A(3,3) = reshape([${t}$ :: 10,  1,  2, &
```
jvdp1: `3` could be defined as an `integer, parameter`

jvdp1 (approved):
Nice addition to stdlib. Thank you! I only have some minor suggestions/questions. Feel free to ignore them.

src/linalg_iterative/stdlib_linalg_iterative_solvers_gmres.fypp
jalvesz: The preconditioner procedure here can not be made pure for two related reasons:

These are the actual implementation to which the dummy procedure pointers from
```fortran
type, public :: stdlib_linop_${s}$_type
        procedure(vector_sub_${s}$), nopass, pointer    :: matvec => null()
        procedure(reduction_sub_${s}$), nopass, pointer :: inner_product => default_dot_${s}$
end type
```
Will point to and that are not pure themself. And they were left like that because otherwise one cannot embed MPI calls inside.

src/linalg_iterative/stdlib_linalg_iterative_solvers_gmres.fypp
jalvesz: Not really, for the iterative application of the `matvec` procedure we change wether the data in `y` is recycled or cleaned up with `beta`. if `beta==1` then `y` is fed to itself in an iterative manner.

`intent(out)` could potentially wipeout the contents of `y` when entering the procedure. We don't want that for the iterative solver.

src/linalg_iterative/stdlib_linalg_iterative_solvers_gmres.fypp
jalvesz: same https://github.com/fortran-lang/stdlib/pull/1186#discussion_r3608350636

test/linalg/test_linalg_solve_iterative.fypp
jalvesz: The system defined in the test is not parametric. We could try to replace the system to make it parametric and in that case it would make sense. But the Hilbert matrix case already does that.

jvdp1: The `3` is hard-coded several times. In such a case, I always find a `parameter` cleaner, but I agree it might be a matter of taste.

jalvesz: right! done here https://github.com/fortran-lang/stdlib/pull/1186/commits/c4734a00b8183686116d8223f81030376baeebb6

PR #1196, opened 2026-07-10
fix : style guide violation

jalvesz:
Some previous PR introduced a style guide violation regarding where the `module` attibute should be. This produces spurious builds with CMake < 3.26


jvdp1 (approved):
LGTM. It can be merged. Thank you @jalvesz

PR #1198, opened 2026-07-13
fix (str2num) : handle leading '+' sign in to_num parsers

### LINKED ISSUE #1197: function to_num() does not convert "+1" correctly
### Description

The function to_num() from str2num does not convert correctly if the number has a leading "+". See the following minimal working example:

```
program mwe_to_num
    use stdlib_str2num, only: to_num
    implicit none
    character(len=5) :: tstr
    integer :: k

    tstr = "+1"
    k = to_num(tstr, k)
    print *, k  ! prints "0" (wrong)

    tstr = "1"
    k = to_num(tstr, k)
    print *, k  ! prints "1" (correct)   

    tstr = "-1"
    k = to_num(tstr, k)
    print *, k  ! prints "-1" (correct)
    stop
end program
```

Output:

```
nmznsnch@namizono-desktop:~/Codes/stdlib$ ./mwe_to_num 
           0
           1
          -1
```

### Expected Behaviour

I expect "+1" to be converted to (integer) 1, not 0

### Version of stdlib

Downloaded june 2026

### Platform and Architecture

PC, Ubuntu 24.04.4 LTS, Intel Core i5-6400, gfortran 13.3.0

### Additional Information

_No response_

jalvesz:
Fix #1197 by CoPilot:

`to_num()` silently returned 0 for any numeric string with a leading + (e.g. "+1", "+1.234"). The integer and all four real parsers (to_sp_base, to_dp_base, to_xdp_base, to_qp_base) only consumed a leading -; a + was left unadvanced, the digit loop found a non-digit immediately, and the parser succeeded with value 0.
```Fortran
k = to_num("+1", k)   ! returned 0 (wrong)
k = to_num("-1", k)   ! returned -1 (correct)
```
Changes:

* src/strings/stdlib_str2num.fypp — in all five to_*_base parsers, extended the leading-sign check from an if (minus only) to if / else if that also consumes a leading + and advances p. Updated the comment from "Verify leading negative" to "Verify leading sign".
* test/string/test_string_to_number.fypp — added regression cases for integers ("+1", "+42", "   +99", "+0005") and reals ("+1", "+1.234") across all type variants.

### DIFF
--- a/src/strings/stdlib_str2num.fypp
+++ b/src/strings/stdlib_str2num.fypp
@@ -136,11 +136,13 @@ module stdlib_str2num
         ! Find first non white space
         p = shift_to_nonwhitespace(s)
         !----------------------------------------------
-        ! Verify leading negative
+        ! Verify leading sign
         sign = 1
         if(p<=len(s)) then
             if(iachar(s(p:p)) == minus_sign+digit_0 ) then
                 sign = -1
+                p = p + 1
+            else if(iachar(s(p:p)) == plus_sign+digit_0 ) then
                 p = p + 1
             end if
         else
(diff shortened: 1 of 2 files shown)


jvdp1 (approved):
LGTM. thank you @jalvesz

PR #1200, opened 2026-07-23
Implement unique function to stdlib_sorting

Mahmood-Sinan:
This PR adds a generic `unique` function to `stdlib_sorting` for extracting the distinct elements of rank-1 arrays.

- Support integer and real arrays.
- Support two modes of operation:
  - sorted output (`sorted_output = .true.`)
  - order of first occurrence (`sorted_output = .false.`)
- Add an optional `tolerance` argument for real arrays when `sorted_output = .true.` to support approximate equality comparisons.

### Limitations

- Tolerance-based comparison is currently supported only when `sorted_output = .true.`.
- The order-preserving implementation for `real(xdp)` is currently unavailable because hashing relies on the underlying binary representation.
Solves #940.


doc/specs/stdlib_sorting.md
```diff
@@ -232,6 +236,22 @@ The `RADIX_SORT` needs a buffer that have same size of the input data.
 Your can provide it using `work` argument, if not the subroutine will
 allocate the buffer and deallocate before return.
 
+#### The `UNIQUE` function
+
+`UNIQUE` returns the distinct elements of a rank one array.
+
+When the argument `sorted_output` is `.true.`, the implementation
+sorts a copy of the input array before removing adjacent duplicate
+elements, producing the unique values in non-decreasing order.
```
jalvesz:
```suggestion
elements, producing the unique values in increasing order.
```

jalvesz (changes requested):

src/sorting/stdlib_sorting.fypp
```diff
@@ -477,6 +477,64 @@ module stdlib_sorting
 !!    ! Sort a_data based on the sorting of that component
 !!        a_data(:) = a_data( index(1:size(a_data)) )
 !!    end subroutine sort_a_data
+!!```
+
+    public unique
+!! Version: experimental
+!!
+!! The generic function implementing the `UNIQUE` algorithm to return the
```
jalvesz:
```suggestion
!! The generic procedure implementing the `UNIQUE` algorithm to return the
```

src/sorting/stdlib_sorting.fypp
```diff
@@ -477,6 +477,64 @@ module stdlib_sorting
...
+
+    public unique
+!! Version: experimental
+!!
+!! The generic function implementing the `UNIQUE` algorithm to return the 
+!! distinct elements of a rank-1 array. The result may either preserve the
+!! order of first occurrence or be returned in sorted order.
+!!
+!! Its use has the syntax: 
+!!     output = unique(array, sorted_output[, tolerance] )
```
jalvesz:
```suggestion
!!     call unique(array, output, sorted_output[, tolerance] )
```

src/sorting/stdlib_sorting.fypp
```diff
@@ -477,6 +477,64 @@ module stdlib_sorting
...
+!!
+!!#### Examples
+!!
+!! Remove duplicates while preserving the original order:
+!!
+!!```Fortran
+!!    integer :: a(8) = [4, 2, 4, 1, 2, 3, 1, 5]
+!!    integer, allocatable :: b(:)
+!!
+!!    b = unique(a, .false.)
```
jalvesz:
```suggestion
!!    call unique(a, b, .false.)
```

src/sorting/stdlib_sorting.fypp
```diff
@@ -477,6 +477,64 @@ module stdlib_sorting
...
+!!    ! b = [4, 2, 1, 3, 5]
+!!```
+!!
+!! Return the unique values in sorted order:
+!!
+!!```Fortran
+!!    integer :: a(8) = [4, 2, 4, 1, 2, 3, 1, 5]
+!!    integer, allocatable :: b(:)
+!!
+!!    b = unique(a, .true.)
```
jalvesz:
```suggestion
!!    call unique(a, b, .true.)
```

jalvesz (approved):
LGTM @Mahmood-Sinan thanks for this contribution. Here just a couple of minor corrections.

Mahmood-Sinan:
> LGTM @Mahmood-Sinan thanks for this contribution. Here just a couple of minor corrections.

Thanks for the review, I have made those changes.

doc/specs/stdlib_sorting.md
```diff
@@ -626,6 +645,61 @@ Sorting an array of a derived type based on the data in one component
...
+Experimental
+
+##### Description
+
+Computes the distinct elements of the input array and stores them in
+`output`.
+
+##### Syntax
+
+`call ` [[stdlib_sorting(module):unique(interface)]] `( array, sorted_output, output[, tolerance] )`
```
jvdp1: could `sorted_ouput` be `optional`?

```suggestion
`call ` [[stdlib_sorting(module):unique(interface)]] `( array, output[, sorted_output, tolerance] )`
```

doc/specs/stdlib_sorting.md
```diff
@@ -626,6 +645,61 @@ Sorting an array of a derived type based on the data in one component
...
+
+##### Class
+
+Generic subroutine.
+
+##### Arguments
+
+`array`: shall be a rank one array of any of the types `integer(int8)`, 
+`integer(int16)`, `integer(int32)`,`integer(int64)`, `real(sp)`, 
+`real(dp)`, or `real(qp)`. It is an `intent(in)` argument.
```
jvdp1: `int64` and `qp` are not supported by all compilers. Maybe:

```suggestion
`array`: shall be a rank one array of any supported kind of the types `integer`, 
and `real`. It is an `intent(in)` argument.
```

doc/specs/stdlib_sorting.md
```diff
@@ -626,6 +645,61 @@ Sorting an array of a derived type based on the data in one component
...
+
+`call ` [[stdlib_sorting(module):unique(interface)]] `( array, sorted_output, output[, tolerance] )`
+
+##### Class
+
+Generic subroutine.
+
+##### Arguments
+
+`array`: shall be a rank one array of any of the types `integer(int8)`,
```
jvdp1: Could it be extended to other types than integer and real?

doc/specs/stdlib_sorting.md
```diff
@@ -626,6 +645,61 @@ Sorting an array of a derived type based on the data in one component
...
+`real(dp)`, or `real(qp)`. It is an `intent(in)` argument.
+
+`sorted_output`: shall be a scalar of type default logical.
+It is an `intent(in)` argument.
+If `.true.`, the result contains the unique elements in
+non-decreasing order. Otherwise, the result contains the unique
+elements in the order of their first occurrence.
+
+`output`: shall be an allocatable rank-one array of the same type and
+kind as `array`. It is an `intent(inout)` argument. On return, it
```
jvdp1, line 673: why is it `intent(inout)`? Could it be `intent(out)` only?

example/sorting/example_unique.f90
```diff
@@ -0,0 +1,13 @@
...
+  use stdlib_kinds, only: dp
+  use stdlib_sorting, only: unique
+  implicit none
+
+  real(dp), allocatable :: A(:)
+  real(dp), allocatable :: B(:)
+
+  A = [4.0_dp, 5.0_dp, 4.0_dp, 3.0_dp, -1.0_dp, 4.0_dp, 2.0_dp]
+
+  call unique(A, .true., B)
```
jvdp1: The following feels more natural to me:
```fortran
call unique(A, B, .true.)
```

jvdp1 (commented):
Thank you @Mahmood-Sinan . Overall it sounds good.
I have 2 main comments:
* API: could `sorted_output` be optional?
* Could the implementation be extended to other types (e.g., `character`, `string`, `complex`,...)?

Mahmood-Sinan:
Thanks for reviewing @jvdp1

doc/specs/stdlib_sorting.md
Mahmood-Sinan: One doubt, does `intent(out)` preserve the size of allocatable arrays when function is being called ? Because the intention is to use a preallocated buffer/output array if the user already allocated one.

doc/specs/stdlib_sorting.md
Mahmood-Sinan: working on it, I just extended it to character arrays.

Mahmood-Sinan: I was trying to implement `complex` type version, but got into a problem. Since the macros for `real` and `int` were defined as
```fortran
#:set INT_TYPES_ALT_NAME = list(zip(INT_TYPES, INT_TYPES, INT_KINDS, INT_CPPS))
#:set REAL_TYPES_ALT_NAME = list(zip(REAL_TYPES, REAL_TYPES, REAL_KINDS, REAL_CPPS))
```
I tried to define complex the same way:
```fortran
#:set CMPLX_TYPES_ALT_NAME = list(zip(CMPLX_TYPES, CMPLX_TYPES, CMPLX_KINDS, [""]))
```

And inside `common.fypp`, `REAL_KINDS` are defined as

```fortran
#! Real kinds to be considered during templating
#:set REAL_KINDS = ["sp", "dp"]
#:if WITH_XDP
#:set REAL_KINDS = REAL_KINDS + ["xdp"]
#:endif
#:if WITH_QP
#:set REAL_KINDS = REAL_KINDS + ["qp"]
#:endif
```

And so the `CMPLX_KINDS`:
```fortran
#! Complex kinds to be considered during templating
#:set CMPLX_KINDS = ["sp", "dp"]
#:if WITH_XDP
#:set CMPLX_KINDS = CMPLX_KINDS + ["xdp"]
#:endif
#:if WITH_QP
#:set CMPLX_KINDS = CMPLX_KINDS + ["qp"]
#:endif
```
And the procedures inside `sorting` uses `name1` which corresponds to `*_KINDS` for naming. But `REAL_KINDS` and `CMPLX_KINDS` are the same. So during compiling, I am getting errors like:
```bash
Entity '*_unique’ at (1) is already present in the interface
```
One way to overcome this would be to use `*_SUFFIX` instead of `*_KINDS` for procedure names (I assume purpose of `*_SUFFIX` is so). If we do so, then every function inside sorting would have to follow that. Let me know what you think.

Mahmood-Sinan:
@jvdp1 I have made the changes. Please take a look.

doc/specs/stdlib_sorting.md
jvdp1: @jalvesz what is your opinion about this?

jvdp1:
@Mahmood-Sinan Some tests `sorting_unique` failed. Could you have a look, please?

Mahmood-Sinan:
> @Mahmood-Sinan Some tests `sorting_unique` failed. Could you have a look, please?

thanks for pointing it out, I have fixed them

doc/specs/stdlib_sorting.md
```diff
@@ -626,6 +645,62 @@ Sorting an array of a derived type based on the data in one component
...
 
+#### `unique` - computes the distinct elements of an input array
+
+##### Status
+
+Experimental
+
+##### Description
+
+Computes the distinct elements of the input array and stores them in
```
jvdp1: There is no computations, right? Maybe:
```suggestion
Returns the distinct elements of the input array in
```

doc/specs/stdlib_sorting.md
```diff
@@ -626,6 +645,62 @@ Sorting an array of a derived type based on the data in one component
...
+##### Arguments
+
+`array`: shall be a rank one array of any supported kind of the types `integer`,
+`real`, `character`, `string_type` and `complex`. It is an `intent(in)` argument.
+
+`output`: shall be an allocatable rank-one array of the same type and
+kind as `array`. It is an `intent(inout)` argument. On return, it
+contains one copy of each distinct element of `array`.
+
+`sorted_output` (optional): shall be a scalar of type default logical.
```
jvdp1:
```suggestion
`sorted_output` (optional): shall be a scalar of type default `logical`.
```

doc/specs/stdlib_sorting.md
```diff
@@ -626,6 +645,62 @@ Sorting an array of a derived type based on the data in one component
...
+`SORTED_OUTPUT` is `.false.` by default. This argument is not available
+for complex arrays.
+
+`tolerance` (optional): shall be a scalar of the same real kind as `array`.
+It is an `intent(in)` argument specifying the tolerance used when
+comparing values in sorted order. Two values whose absolute
+difference is less than or equal to `tolerance` are considered
+equal. If omitted, tolerance defaults to zero. This argument is only
+available for real arrays. `tolerance` may only be supplied when `sorted_output`
+is `.true.`.
```
jvdp1, line 688: Why is this the case?

doc/specs/stdlib_sorting.md
```diff
@@ -57,6 +57,9 @@ data:
   that are effectively unordered before the sort;
 * `RADIX_SORT` is intended to sort fixed width intrinsic data 
   types (integers and reals).
+* `UNIQUE` computes the distinct elements of a rank one array,
```
jvdp1:
```suggestion
* `UNIQUE` computes the distinct elements of a rank-one array,
```

jvdp1 (changes requested):
Thank you @Mahmood-Sinan . Overall LGTM. I left a few minor comments. I think it is close to be ready :)

doc/specs/stdlib_sorting.md
Mahmood-Sinan: Because `sorted_output = .false.` uses a hashmap, which only supports exact duplicate detection. With tolerance, we'd need to compare each value against previously accepted values, resulting in O(n²) with a brute-force approach. Since that's inefficient, I restricted `tolerance` to sorted output. Should I implement the brute-force approach for unsorted output as well?

Issue #1187, opened 2026-05-20
add a set_environment_variable procedure to stdlib_system


jalvesz:
### Motivation

While iso_fortran_env provides `get_environment_variable` to query existing env variables, there doesn't seem to be a "set" counterpart availble in the Fortran standard or other standard modules.

A (yet to be) `set_environment_variable` could be added to stdlib by wrapping the C procedures 
* Windows : `_putenv_s`
* Unix: `setenv`

jalvesz:
Something like:

```Fortran
    subroutine set_environment_variable(name, value, overwrite, status)
        character(len=*), intent(in) :: name
        character(len=*), intent(in) :: value
        logical, intent(in), optional :: overwrite
        integer, intent(out), optional  :: status

        logical :: overwrite_
        integer(c_int) :: c_status
        character(kind=c_char, len=:), allocatable :: c_name, c_value
        
        interface
#ifdef _WIN32
            function c_putenv_s(name, value) bind(C, name="_putenv_s")
                import :: c_char, c_int
                integer(c_int) :: c_putenv_s
                character(c_char) :: name(*)
                character(c_char) :: value(*)
            end function
#else
            function c_setenv(name, value, overwrite) bind(C, name="setenv")
                import :: c_char, c_int
                integer(c_int) :: c_setenv
                character(c_char) :: name(*)
                character(c_char) :: value(*)
                integer(c_int), value :: overwrite
            end function
#endif
        end interface

        overwrite_= .true.
        if (present(overwrite)) overwrite_= overwrite

        c_name  = trim(name) // c_null_char
        c_value = trim(value) // c_null_char
        
#ifdef _WIN32
        c_status = c_putenv_s(c_name, c_value)
#else
        c_status = c_setenv(c_name, c_value, merge(1_c_int, 0_c_int, overwrite_) )
#endif
        if (present(status)) status = int(c_status)
    end subroutine set_environment_variable
```

kunalKumar-13:
@jalvesz I would like to take this one, if nobody else is on it.

I have read through `stdlib_system` and your sketch, and I think the shape is clear. Before writing it, three questions where guessing would cost a review round:

[...]

PR #1208, opened 2026-09-07
Add set_environment_variable and delete_environment_variable

kunalKumar-13:
Implements #1187, following the three answers @jalvesz gave on that issue.

**1. C shim, not `#ifdef` in Fortran.** The platform branching lives in `stdlib_system.c` behind a single `bind(c)` interface, the way `is_directory` and `set_cwd` already do it.

**2. One signature on every platform.** `overwrite` is optional, defaults to `.true.`, and selects whether an existing variable is replaced. Windows has no equivalent — `_putenv_s` always replaces — so there the argument has no effect, documented in the spec and in the procedure's own doc block.

**3. A separate `delete_environment_variable`,** rather than removal-by-empty-value, so the intent reads plainly at the call site.

```fortran
call set_environment_variable("STDLIB_EXAMPLE", "hello", err=err)
call set_environment_variable("STDLIB_EXAMPLE", "goodbye", overwrite=.false., err=err)  ! still "hello"
call delete_environment_variable("STDLIB_EXAMPLE", err=err)
```

### Notes on the implementation

Windows has no `unsetenv`; assigning an empty value with `_putenv_s` is how the CRT removes a variable. Deleting a variable that was never set is not an error, matching `unsetenv`.

An empty name, and a name containing `=`, are rejected in Fortran rather than passed down. `setenv` reports both as `EINVAL`, which reaches the user as "Invalid argument" and says nothing about which rule was broken.

The variable is set for the calling process and anything it starts afterwards, not for the parent shell. That is a property of the OS, and the spec says so, since it is the first thing people are surprised by.

### Verification

Ubuntu 26.04 aarch64, gfortran 15.2:

- full `ctest` suite **428/428 passing**
- `test_os` 6/6, including the three new cases
- `example_environment_variable` prints what the spec says it will

The new tests were checked against a deliberately broken implementation rather than only against a working one:

| mutation | result |
|---|---|
| `setenv(name, value, 1)` — ignore `overwrite` | `test_environment_variable_overwrite` FAILED: *overwrite=.false. changed the value to third* |
| `delete` made a no-op | `test_environment_variable` FAILED: *the variable is still set after being deleted* |

The `overwrite` half of that test skips on Windows, where the flag is documented as having no effect.

### Two questions

**On `overwrite` under Windows.** I followed your answer literally — the argument exists everywhere and is ignored on Windows. It could instead be honoured there by checking `getenv` first and returning success without writing, which would make the behaviour identical on both platforms rather than only the signature. That is a three-line change in the shim. I did not do it because you said documenting the difference was enough, but say the word if you would rather have it.

**Unrelated, and I did not touch it:** `master` does not build under its own `-std=f2018` with gfortran 15.2. `stdlib_str2num.f90` trips `GNU Extension: Different type kinds` on `min(p, len(s))`, and `stdlib_system.F90:1058` trips `GNU Extension: LOGICAL dummy argument 'winapi' ... with non-C_Bool kind in BIND(C)`. I built with `-std=gnu` locally to get around it. Worth an issue of its own if it is not already known — happy to open one.


kunalKumar-13:
The intel job is red and I have not been able to clear it. Rather than keep pushing guesses at CI, here is everything I have established, in case someone with ifx to hand sees it immediately.

**The failure**

```
src/system/CMakeFiles/fortran_stdlib_system.dir/stdlib_system.F90-pp.f90:
  error #5633: **Internal compiler error: segmentation violation signal raised**
compilation aborted for ... (code 3)
```

An ICE with a raw stack trace and no source line.

**What is established**

- `Build (ubuntu-22.04, cmake, intel-classic, 2021.10)` — **passes**. Only `ifx 2024.1` fails.
- gfortran 16.2.0 compiles it, and the code is correct there: set, overwrite default, `overwrite=.false.`, delete, delete of an absent variable, a value with a trailing space (preserved), an empty name and a name containing `=` all behave as documented.
- master's intel jobs are green, so it is this branch.
- **All seven commits fail identically, including the first.** Six structural variations changed nothing:

| commit | what was tried |
|---|---|
| 9c628afa9 | original |
| df9fdf38a | interfaces on one line |
| 3b379156c | dummy renamed off `value` |
| 127c96905 | matched `stdlib_set_cwd`'s interface shape |
| 533c86b08 | `to_c_char` results held in locals |
| e5866a8be | automatic arrays instead of allocatables |
| f213d3696 | `c_ptr` by value instead of `c_char` arrays |

**What I ruled out along the way**

- `stdlib_setenv` was the only bind(C) interface in all of `src/` with two `character(kind=c_char)` dummies, which looked like the answer. f213d3696 removed them entirely and it still ICEs.
- The two new subroutines are structurally identical to their neighbours — plain subroutines in the same `contains` section as `set_cwd`, `make_directory` and `remove_directory`, not module procedures.
- `sql.Map`-style marshalling concerns do not apply here; nothing in the added code uses a type with custom behaviour.

The file grows from 1357 to 1487 lines, which is the only other thing that changed.

**Where I am stuck**

There is no ifx for Apple Silicon — Intel dropped macOS support and it is x86-only regardless — so I can only bisect through this CI, and I have spent seven runs doing that already. I did not want to keep burning runs on guesses.

Two questions, if anyone can help:

1. Does anything in the added code obviously trip ifx 2024.1? There is precedent in this repo for working around it — 74b6ebe1 *"change syntax for `ifx` fix"* restructured `module type(process_type) function` declarations in this same file.
2. Would you rather I bisect properly by pushing a commit that keeps only `delete_environment_variable` and drops `set_environment_variable`? That is one run for a definite answer about which subroutine triggers it, but it leaves the PR temporarily incomplete, so I did not want to do it unannounced.

Happy to do whichever is more useful. The functionality itself is finished and tested; this is purely the compiler.

kunalKumar-13:
The `c_ptr` attempt did not work — the last run is on `f213d369`, which is the commit that was supposed to fix it, and ifx 2024.1 still aborts with `error #5633` on `stdlib_system.F90`. So the premise it rested on, that two `character(kind=c_char) :: x(*)` dummies are what ifx chokes on, is not supported.

Looking back, none of the five attempts isolated anything — each one changed several things at once. So I have gone the other way and removed the special-casing entirely.

`set_environment_variable` now uses the shape every other C shim in this module already uses, and that `delete_environment_variable` immediately below it uses unchanged:

```fortran
interface
    integer function stdlib_setenv(cname, cval, overwrite) bind(C, name='stdlib_setenv')
        import c_char
        character(kind=c_char), intent(in) :: cname(*)
        character(kind=c_char), intent(in) :: cval(*)
        integer, intent(in) :: overwrite
    end function stdlib_setenv
end interface
...
code = stdlib_setenv(to_c_char(trim(name)), to_c_char(value), overwrite_)
```

`set_cwd`, `make_directory`, `remove_directory` and `is_directory` are all written like this and all compile under ifx 2024.1 today.

That removes the `target` attribute, both automatic arrays sized from dummy arguments, and both `c_loc` calls. An automatic array carrying `TARGET` whose bound comes from a dummy argument is the one construct in the new code that appears nowhere else in this module, so if anything here is the trigger that is my best guess — but I am guessing, and I would rather the diff be small and idiomatic than clever. It is −15/+6 and the C side is untouched: the arrays still arrive as `const char*` and `overwrite` as `const int*`.

I cannot reproduce this locally — I have no Intel compiler — so CI is the only instrument. If it is still red I will stop guessing and open a separate issue rather than keep pushing at this PR.

Two things worth putting in front of you either way:

- **The pin is old.** `CI.yml` pins `intel` at `2024.1` while gcc runs up to 15. #1142 is a test report against **ifx 2025.3**, so newer Intel is in use by testers but not by CI. If this turns out to be a genuine ifx 2024.1 bug, contorting library source around a compiler two years behind may be the wrong trade — but bumping or excluding that entry is a repo-wide CI decision and does not belong in a feature PR. Happy to open one separately if you want it.
- There is no existing report of `#5633` in this repo; I searched before writing this.

@perazz @jvdp1 — is the shape above what you would want here regardless of the ICE?

kunalKumar-13:
Still red — same `error #5633` on both ifx 2024.1 jobs, so that is eight shapes tried and eight crashes. As I said I would, I have stopped guessing here and moved it to #1214 rather than keep pushing at this PR.

The issue records what is now actually ruled out, including two things I checked against the tree rather than assumed: a dummy named `value` is not special (33 of them in `src/`, all compiling), and a host dummy shadowed by an interface dummy is not either (`set_cwd` does exactly that on master). What survives is that this is the only `bind(C)` interface in the source tree with two `character(kind=c_char) :: x(*)` dummies. There is a two-file reproducer in the issue that should settle it in about a minute for anyone with an Intel compiler — I have none, which is why CI has been my only instrument here.

The code as it now stands is worth keeping regardless of how that resolves: it is the same shape as `delete_environment_variable` and the other four C shims in the module, and it dropped the `target` attribute, two automatic arrays and two `c_loc` calls to get there. Everything except the two ifx jobs is green — 71 checks passing, including `intel-classic 2021.10` and gcc 10 through 15.

PR #1209, opened 2026-09-07
Build under the project's own -std=f2018 on gfortran 15.2

kunalKumar-13:
`config/DefaultFlags.cmake` compiles with `-std=f2018`, but four places rely on GNU extensions that gfortran 15.2 rejects, so a build with the project's own flags stops in `src/strings` and `src/system`:

```
stdlib_str2num.f90:112:    p = min( p , len(s) )
    Error: GNU Extension: Different type kinds

stdlib_system.F90:1058:    type(c_ptr) function strerror(len, winapi) bind(C, name='stdlib_strerror')
    Error: GNU Extension: LOGICAL dummy argument 'winapi' with non-C_Bool kind in BIND(C) procedure

stdlib_system_subprocess.F90:256,259:    call exit(command_state)
    Error: Procedure 'exit' called at (1) is not explicitly declared

stdlib_system_subprocess.F90:374:    wait_time = max(0.0_RTICKS, max_wait_time)
    Error: GNU Extension: Different type kinds
```

### Why CI does not see it

The matrix runs gcc 15 only on `macos-14`; the ubuntu matrix stops at gcc 14. The point release available there still accepts these, so the last full run on master is green. Distro gfortran 15.2 does not.

### The changes

**`stdlib_str2num`** — `min(p, len(s))` compares an `int8` with a default integer. The result can never exceed `p`, so promoting `p` is enough and the value still fits back into `int8`.

**`stdlib_system_subprocess`** — `exit` is a GNU intrinsic, so `implicit none(external)` will not accept it. Bound to C's `exit`, which is what the intrinsic calls, so the reason the comment gives for preferring it to `stop` still holds. And `max(0.0_RTICKS, max_wait_time)` compares `real(RTICKS)` with a default real.

**`stdlib_system`** — this one is worth a closer look, because it is a behaviour fix and not only a build fix.

The interface to `stdlib_strerror` declared `winapi` as a default `logical` passed by reference. The C side is:

```c
char* stdlib_strerror(size_t* len, bool winapi)
```

a `bool` taken by value. So C was reading the *address* of the Fortran logical as the flag — always a non-null pointer, therefore always true, whatever was passed. The interface already imported `c_bool` without using it, which suggests this was the intent and got lost.

On unix nothing changes: the `winapi` branch has no body outside `_WIN32` and falls through to `strerror(errno)` either way. **On Windows it means `FormatMessageA` was always being used, including when `c_get_strerror()` was called with no argument.** That is the intended behaviour restored, but I have no Windows machine and cannot test it — please treat that hunk as the one needing a second opinion.

### Verification

Ubuntu 26.04 aarch64, gfortran 15.2, `-std=f2018` left exactly as the repo sets it:

| | result |
|---|---|
| `master` | build stops, 5 errors across the two files |
| this branch | build completes, **ctest 427/427** |

### What this does not cover

gfortran **16.2** needs more. It additionally rejects the generated BLAS interfaces in `stdlib_linalg_blas.F90` with `Symbol 'lda' is used before it is typed`, roughly 500 of them. That is a separate and much larger change to the generated sources, and I have not touched it. None of the four sites fixed here are among those, so this is a step toward gfortran 16 rather than the whole of it.

Found while building #1208 locally, where I mentioned it and offered to raise it separately.

### DIFF
--- a/src/strings/stdlib_str2num.fypp
+++ b/src/strings/stdlib_str2num.fypp
@@ -109,7 +109,10 @@ module stdlib_str2num
         integer(int8) :: err
         !----------------------------------------------
         call to_num_base(s,v,p,err)
-        p = min( p , len(s) )
+        ! `p` is int8 and `len` returns a default integer, so comparing them
+        ! directly is a GNU extension that -std=f2018 rejects. The result is
+        ! never larger than `p`, so it always fits back into int8.
+        p = min( int(p), len(s) )
         s => s(p:)
         if(present(stat)) stat = err
     end function
--- a/src/strings/stdlib_str2num.fypp
+++ b/src/strings/stdlib_str2num.fypp
@@ -110,8 +110,12 @@ module stdlib_str2num
         !----------------------------------------------
         call to_num_base(s,v,p,err)
         ! `p` is int8 and `len` returns a default integer, so comparing them
-        ! directly is a GNU extension that -std=f2018 rejects. The result is
-        ! never larger than `p`, so it always fits back into int8.
+        ! directly is a GNU extension that -std=f2018 rejects. Promote `p`
+        ! rather than asking for `len(s, kind=int8)`: the string is not bounded
+        ! by 127, and that form wraps silently, with no diagnostic even under
+        ! -std=f2018 -Wall. For a 200 character string it returns -56, so the
+        ! `min` yields -56 and the `s(p:)` below is then a bad substring bound.
+        ! The result here is never larger than `p`, so it fits back into int8.
         p = min( int(p), len(s) )
         s => s(p:)
         if(present(stat)) stat = err
--- a/src/system/stdlib_system.F90
+++ b/src/system/stdlib_system.F90
@@ -1028,7 +1028,10 @@ function c_get_strerror(winapi) result(str)
             import c_size_t, c_ptr, c_bool
             implicit none
             integer(c_size_t), intent(out) :: len
-            logical, intent(in) :: winapi
+            ! The C side takes a `bool` by value. A default `logical` passed by
+            ! reference is neither: -std=f2018 rejects the kind, and the pointer
+            ! that reached C was read as the flag, so it was always true.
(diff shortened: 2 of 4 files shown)


src/strings/stdlib_str2num.fypp
```diff
@@ -109,7 +109,10 @@ module stdlib_str2num
         integer(int8) :: err
         !----------------------------------------------
         call to_num_base(s,v,p,err)
-        p = min( p , len(s) )
+        ! `p` is int8 and `len` returns a default integer, so comparing them
+        ! directly is a GNU extension that -std=f2018 rejects. The result is
+        ! never larger than `p`, so it always fits back into int8.
```
jvdp1: Could you add the reason that the following would not work, please:
```
        p = min( p, len(s, kind=int8) )
```

jvdp1 (approved):
thank you @kunalKumar-13 for this PR. I have only a minor comment.

kunalKumar-13:
Thank you for the review @jvdp1. On `p = min( p, len(s, kind=int8) )` — I put the reasoning in a comment above the line in the last commit, but it is worth stating here too.

`int8` tops out at 127, so `len(s, kind=int8)` wraps for any string longer than that. Compiled and run with gfortran 16.2.0, on a 200-character string:

| expression | value |
|---|---|
| `len(s)` | 200 |
| `len(s, kind=int8)` | **-56** |
| `min(p, len(s, kind=int8))` | **-56** |
| `min(int(p), len(s))` | 5 |

So `min` yields -56, and the `s => s(p:)` on the following line gets a negative substring bound.

The part that decided it for me is that `-std=f2018 -Wall` emits **no diagnostic at all** for the `kind=int8` form. It compiles silently. So that version does not trade a diagnostic for correctness — it trades a program the compiler rejects for one that is quietly wrong on any string over 127 characters, which for a string-to-number routine is not an unusual length.

Promoting `p` instead keeps the comparison in default integer, and since the result is never larger than `p`, it fits back into `int8` without a narrowing concern.

